PR #52, opened 2015-08-04
Add a scrollbar to menus when the screen is too small

cloutier:


cloutier:
You're welcome :smile:

PR #74, opened 2015-08-21
Correct some typos in magics docstrings, etc.

jlutgen:
Just fixing (I hope) a few small typos that I noticed while familiarizing myself with your code base.


haraldschilly:
looks all good to me

williamstein:
These should now be live for people whenever they (re-)start their projects.

PR #126, opened 2015-09-24
Upgrade jQuery to 2.1.4

nikolas:


haraldschilly:
Has this been tested to not break anything?

nikolas:
No, because the [setup instructions](https://github.com/sagemathinc/smc/blob/master/INSTALL.md) are pretty involved, so I haven't been able to set up my own instance yet.

Maybe something like docker / docker-compose could simplify the setup process for development purposes?

But anyways, going from jQuery 2.0.3 to 2.1.4, I've never seen any compatibility issues. But yeah you will have to test this yourself until I figure out how to set up my instance of smc.

williamstein:
I didn't notice any issues with the upgrade to 2.1.4.

williamstein:
now live on the site.

PR #128, opened 2015-09-24
Update three.js to version 72

nikolas:


williamstein:
I merged this.  I then had to fix a number of issues in the SMC library itself involving using threejs version 72, and add in some extra things.  That's done in another commit: https://github.com/sagemathinc/smc/commit/98adcc8059226a2324f816fc68ffdec7677f19d3

I'll make this live soon (when there's a bit less traffic).

williamstein:
Now live on the site.

PR #247, opened 2015-11-10
Add cd to development/install steps

rgaiacs:


haraldschilly:
the installation instructions will change significantly with the next update. please checkout the https://github.com/sagemathinc/smc/tree/smc-in-smc branch

PR #267, opened 2015-11-19
Simplify projects page for new users

johnjeng:
Change projects page to not include Deleted and Hidden buttons if not applicable.

### DIFF
--- a/src/smc-webapp/projects.cjsx
+++ b/src/smc-webapp/projects.cjsx
@@ -583,6 +583,10 @@ NewProjectCreator = rclass
                     description_text : ''
                     error            : ''
 
+    handle_keypress : (e) ->
+        if e.keyCode == 13 and @state.title_text != ''
+            @create_project()
+
     render_input_section : ->
         <Well style={backgroundColor: '#ffffff'}>
             <Row>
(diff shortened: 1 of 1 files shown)


src/smc-webapp/projects.cjsx
```diff
@@ -1164,7 +1205,7 @@ exports.ProjectTitle = ProjectTitle = rclass
 
     render : ->
         if not @props.project_map?
-            return <Loading />
```
williamstein: Why did you make this change?  It got to be wrong.  It means that now if the project_map isn't loaded yet, instead of showing "Loading" it will never show loading and will instead show "(private project)", even if it isn't private.  I would just change it, but I'm puzzled since I don't know why you made the change.

williamstein:
I'm going to just merge this anyways and change back that loading line for now.

johnjeng:
Ah, I believe it was an oversight while fixing the style. Only "return"
should have been deleted.

Merged #267 <https://github.com/sagemathinc/smc/pull/267>.

—
Reply to this email directly or view it on GitHub
<https://github.com/sagemathinc/smc/pull/267#event-469863131>.

PR #280, opened 2015-11-21
Insert space after period on landing page

jasongrout:
There is no space between the "restrict you." and "Upload" in the rendered html.  It seems that if there is not an explicit space (not just a newline) between the period and the `<strong>` tag, the space is deleted in the rendered html.  Reformatting the paragraph ensures that the newline falls in normal text, which seems to not have the problem (evidenced by other paragraphs).

### DIFF
--- a/src/smc-webapp/landing_page.cjsx
+++ b/src/smc-webapp/landing_page.cjsx
@@ -366,8 +366,8 @@ SagePreview = rclass
                     </Col>
                     <Col sm=6>
                         <ExampleBox title="The Sky is the Limit" index={3}>
-                            <SiteName /> does not arbitrarily restrict you.
-                            <strong>Upload</strong> your own files, <strong>generate</strong> data and results online,
+                            <SiteName /> does not arbitrarily restrict you. <strong>Upload</strong> your
+                            own files, <strong>generate</strong> data and results online,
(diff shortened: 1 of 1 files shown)


jasongrout:
Indeed, whitespace is ignored between html tags, and in this case, the text is rendered 
```html
... restrict you.</span>
<strong>Upload...
```
so the whitespace is not rendered.

jasongrout:
(also, I haven't tested this change, since I don't have a working dev setup)

williamstein:
Jason -- here's your development project, setup and working...

https://cloud.sagemath.com/projects/293c1536-461e-44b2-a3b9-b61b5c381f97

To use it, go to https://cloud.sagemath.com/projects/293c1536-461e-44b2-a3b9-b61b5c381f97/files/smc/src/dev/project/project.term and type `./tmux-start-all`.

and your pull request does work fine, so merging now.

williamstein:
Now live

On Saturday, November 21, 2015, Jason Grout <notifications@github.com>
wrote:

> (also, I haven't tested this change, since I don't have a working dev
> setup)
>
> —
> <https://github.com/sagemathinc/smc/pull/280#issuecomment-158689610>.
>


-- 

PR #312, opened 2015-12-02
robustness of smc-env and similar scripts

haraldschilly:
e.g. "`#!/usr/bin/env bash`" in smc-env has no effect, hence the explicit test that it is run in bash. idempotent changes for the path variables are also a good idea (in case this script is called repeatedly), "source" instead of "." for readability, ...


haraldschilly:
That's an oversight/leftover. My linter told me that tempfile isn't used and first I moved it to the right behind a "#"  comment. Then I wasn't sure if I'm breaking something and put it back :wine_glass: :hammer:

Issue #34, opened 2015-06-24
In collaborator/student search -- interpret "Sage Math <sagemath@sagemath.com>" as an exact search for the email address sagemath@sagemath.com


williamstein:
Requested by Michael Forbes

williamstein:
Change a few lines in rethink.coffee, I think, would fix this everywhere...?

williamstein:
Actually, change misc.parse_user_search in smc-util/misc.coffee

williamstein:
Fixed by issues/34

PR #342, opened 2015-12-22
Issues/34

johnjeng:
Added ability to search for emails wrapped in angle brackets.
https://github.com/sagemathinc/smc/issues/34

Undocumented (bug/feature):
```
<correct@email.com>garbageStuff
```
becomes
```
correct@email.com
```
even though having things at the beginning doesn't work ie.
```
garbage<incorrect@email.com>
```
will be understood exactly as is and result in an invalid email by RFC822. This is because emails such as 
```
begin"\<weirdness"email"<>"@odd.provider.com
```
are actually valid.


williamstein:
John,  Can you write some tests for this function and add them here: src/smc-util/test/misc-test.coffee   ?
It should be pretty obvious how to write more tests by looking at what is in that file. 
You can run the tests by typing `npm run test` in the smc-util directory

johnjeng:
Yup, certainly.

williamstein:
Ping.  It would be good to have this in and live before classes start next week.

johnjeng:
Updated the unit test to be on par with current testing style.

Please tell me if more or something different would be preferred.

williamstein:
> Edit: We don't have any mutation testing right?

What is "mutation testing"?

williamstein:
Good job -- I like this; it's now live!

PR #350, opened 2015-12-31
Update graphics.py

fchapoton:
new proposal for conversion of index_face_set

This should do the same job as before, but in a slightly simpler way, not going through the
obj method. Next step would be to introduce face colors in the game.


williamstein:
I merged this and tested it, and it doesn't work (heh -- if not tested, then definitely broken), so I left the code in, but renamed it to _test, and kept the old version.

fchapoton:
Sorry for not testing that before , but I am not able (or patient enough) to install smc on my laptop. I have now tested by just importing graphics.py inside a sage session. As far as I can see, the only difference
between the two methods is that the faces are not given twice in my new version. Could you please say a few words on *how* it failed when you tested it in a running smc ?

PR #352, opened 2016-01-01
change convert_index_face_set

fchapoton:
here is another tentative, that has been "tested", in the sense of "imported and used inside a sage terminal session for different kinds of 3D objects". It has not been tested in a running SMC session.

### DIFF
--- a/src/smc_sagews/smc_sagews/graphics.py
+++ b/src/smc_sagews/smc_sagews/graphics.py
@@ -315,7 +315,7 @@ def graphics3d_to_jsonable(p):
     #####################################
 
     def convert_index_face_set(p, T, extra_kwds):
-        p.triangulate()
+        p.obj()  # just to ensure one call to .triangulate if needed
         if T is not None:
             vertices = [T.transform_point(v) for v in p.vertices()]
         else:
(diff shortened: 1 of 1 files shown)


williamstein:
I tested this in my devel project and it completely breaks 3d graphics. They don't render at all.   This must be incompatible with the frontend renderer for some reason.  Sorry...

williamstein:
I spend some more time trying to make heads or tails of this and failed.

src/smc_sagews/smc_sagews/graphics.py
```diff
@@ -314,13 +314,16 @@ def parse_mtl(p):
...
         if T is not None:
-            p = p.transform(T=T)
-        p.triangulate()
-        face_geometry = [{"material_name": p.texture.id, "faces": [[int(v) for v in f] for f in p.index_faces()]}]
-        vertex_geometry = [json_float(t) for v in p.vertices() for t in v]
+            vertices = [T.transform_point(v) for v in p.vertices()]
+        else:
+            vertices = p.vertices()
+        face_geometry = [{"material_name": p.texture.id,
+                          "faces": [[int(v) for v in f] for f in p.index_faces()]}]
```
williamstein: Replacing ```"faces": [[int(v) for v in f] for f in p.index_faces()]}]``` by
```"faces": [[int(v)+1 for v in f] for f in p.index_faces()]}]``` makes drawing normal 3d plots not break.

HOWEVER, what is the point of this patch?  What is it doing?  Why is it even here?  What's an example to illustrate this working or not working?  I can't even find anything in my email about this, etc.

williamstein:
See https://github.com/sagemathinc/smc/issues/829

PR #427, opened 2016-02-19
Issues/368

johnjeng:
Used existing start_project function.
Start all button is tentatively placed in the middle of the header for the Students tab.

### DIFF
--- a/src/smc-webapp/editor_course.cjsx
+++ b/src/smc-webapp/editor_course.cjsx
@@ -1538,7 +1538,7 @@ Students = rclass
             v.push <option key={key} value={key} label={student_name}>{student_name}</option>
         return v
 
-    start_all_student_projects : ->
+    start_all_student_projects : (e) ->
         @props.redux.getActions(@props.name).start_all_student_projects()
 
     render_add_selector : ->
@@ -1580,7 +1580,7 @@ Students = rclass
                     {<h5>(Omitting {num_omitted} students)</h5> if num_omitted}
                 </Col>
                 <Col md=3>
-                    <Button onClick={@start_all_student_projects()} bsStyle='success'>Start all Projects</Button>
+                    <Button onClick={@start_all_student_projects} bsStyle='success'>Start all Projects</Button>
                 </Col>
                 <Col md=5>
                     <form onSubmit={@do_add_search}>
--- a/src/smc-webapp/editor_course.cjsx
+++ b/src/smc-webapp/editor_course.cjsx
@@ -1576,10 +1576,10 @@ Students = rclass
                         on_change   = {(value)=>@setState(search:value)}
                     />
                 </Col>
-                <Col md=1>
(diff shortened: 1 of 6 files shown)


src/smc-webapp/editor_course.cjsx
```diff
@@ -432,6 +432,12 @@ exports.init_redux = init_redux = (redux, course_project_id, course_filename) ->
             title = "#{store.get_student_name(student_id)} - #{store.getIn(['settings', 'title'])}"
             redux.getActions('projects').set_project_title(student_project_id, title)
 
+        start_all_student_projects: () =>
+            get_store()?.get_students().map (student, student_id) =>
+                student_project_id = student.get('project_id')
+                if student_project_id?
+                    redux.getActions('projects').start_project(student_project_id)
```
haraldschilly: I'm just looking very briefly at this. My only concern is if there are many student projects, like, more than 100. This `map` runs sequentially -- so that could take a lot of time if this `start_project` call is blocking.

OTOH, it would issue such start commands to all of them at once, basically, if it is non-blocking. 

Hmm, but I see, it's written in the same way as everywhere else. So, that should work then, too …

williamstein: On Friday, February 19, 2016, Harald Schilly <notifications@github.com>
wrote:

> In src/smc-webapp/editor_course.cjsx
> <https://github.com/sagemathinc/smc/pull/427#discussion_r53445192>:
>
> > @@ -432,6 +432,12 @@ exports.init_redux = init_redux = (redux, course_project_id, course_filename) ->
> >              title = "#{store.get_student_name(student_id)} - #{store.getIn(['settings', 'title'])}"
> >              redux.getActions('projects').set_project_title(student_project_id, title)
> >
> > +        start_all_student_projects: () =>
> > +            get_store()?.get_students().map (student, student_id) =>
> > +                student_project_id = student.get('project_id')
> > +                if student_project_id?
> > +                    redux.getActions('projects').start_project(student_project_id)
>
> I'm just looking very briefly at this. My only concern is if there are
> many student projects, like, more than 100. This map runs sequentially --
> so that could take a lot of time if this start_project call is blocking.
>

Actually that code will do all 100 at once!  It "sequentially" sets the
"start this project" field in the local table.  All at once that is will
get synced to the backend, which will then do all the starts in parallel.
This isn't Python.





> OTOH, it would issue such start commands to all of them at once,
> basically, if it is non-blocking.
>
> Hmm, but I see, it's written in the same way as everywhere else. So, that
> should work then, too …
>
> —
> <https://github.com/sagemathinc/smc/pull/427/files#r53445192>.
>


-- 

williamstein:
I thought more about the issue of how many get started at once, etc.  I think your code -- exactly as you wrote it -- is the right thing to do.   Rate limitation of when projects get started up *should* be done on the backend no matter what anyways.  So *good*.

PR #468, opened 2016-03-08
File Creation Improvements in Files Tab

johnjeng:
# Highlights 
- "New" button on file listing page changed to "Create" 
- Default extension for created files is sagews
- If no files are found in the search, the create file button that appears is more helpful
- If no text is in the search bar, clicking an option under the "Create" menu gives the user a warning that they should enter a file name.
- Allows creation of nested folders from file search. ie. foo/bar/newfile.ext


johnjeng:
# Test List
List of manual tests for the various functionalities added. Of course the various design choices are subject to change. Preconditions in each header.

## Empty Folder
Or equivocally, nothing found by the search bar.
The small create button should always mimic the functionality of the large create button.
### Empty Search bar
 - Files listing location should have button that navigates to new files tab.
 - Using the create menu should notify the user that they need to enter a file name.
 - Hitting enter should do nothing 

### Non-empty search bar.
 - Hitting enter should behave as clicking "Create"

##### Search bar does not have "/" nor "."
 - Files listing location should have create button with the file name and current extension.
 - Using the create menu should create a new file with the name given in the search bar.
  - Doing so should navigate the user to the new file

##### Search bar includes "/" but not "."
 - If the search bar ends with "/", the large create button should indicate that a folder will be created.
  - If a file extension is clicked, it should behave as if the search bar did not have a trailing "/"
 - If there are multiple "/"s, treat the splitting of "/" as a sequence of subfolders.
  - Consequently, hitting create should create these folders if they don't exist.
  - Characters between two "/"s maybe any legal for a folder name
  - Should be indicated to user in the create button
 - If there is no trailing "/", all characters after the last "/" should constitute the intended file name underneath any given sequence of subfolders.
 - Hitting enter should behave as clicking "Create"

### Search bar includes "." but not "/"
 - Only the last "." should be interpreted as beginning an extension.
  - If a file extension is clicked, it should take the file name literally and then append the extension of the chosen type unless it would result in "filename.foo.foo"
 - The create button should have on it the extension to the file name that will be used if the user were to click on the button or hit enter.

##### Search bar includes both "/" and "."
 - From above, apply "/" rules and then "." rules.
 - "."s between two "/"s should be treated as part of that folder's name.

## File Listing is non-empty
 This should behave as above except hitting enter opens top item of the displayed files. 

Does this last note make sense any longer given what the enter key does above and that a create button is sill available? The problem is, maybe people rely on this functionality. But at the same time, we currently don't give any visual indication that the first item in the file listing is "selected" in any sense.

src/smc-webapp/project_files.cjsx
```diff
@@ -377,24 +377,40 @@ NoFiles = rclass
...
-                    # Only one folder
-                    text = "Creates a folder named #{@props.file_search}"
-            else
-                text = "#{@full_path_text().slice(last_folder_index + 1)} will be created under the folder path #{@props.file_search.slice(0, last_folder_index + 1)}"
             <Alert style={marginTop: '10px'} bsStyle='info'>
-                {text}
+                {@render_help_text(last_folder_index)}
             </Alert>
 
+    render_help_text : (last_folder_index) ->
```
johnjeng: There has to be a better way to render arbitrary styles in the middle of text but google results are only giving me React Native examples.

johnjeng:
There must be a better way to have arbitrary styling on inline text.. [#1cbda9d](https://github.com/sagemathinc/smc/pull/468/commits/1cbda9d04ba0369fe2c174a163060bac8dfacdda#diff-c072a9940d98dbfb78c678ede846fc88R388)

williamstein:
> There must be a better way to have arbitrary styling on inline text under this function

I don't understand the question.

johnjeng:
ie. "Normal text ... *bold text* more normal text"

PR #469, opened 2016-03-09
global font size for editors

haraldschilly:
This changeset introduces
* a "font-size" setting in the DB (hence, hub restart required before webapp rollout!)
* adds an entry in the settings
* all settings, that have units, get a prop for units (hence no more line-breaks, etc.)
* editors, including sagews editor, read this font-size setting as the default – existing local_storage setting is unaffected and if set, overrides it
* jupyter: slight change in how it is instantiated to have access to the local_storage method
* jupyter: two small buttons to change font sizes, and template cleanup
* jupyter: injecting style element into it's iframe dom to override three font sizes – it has essentially the same logic as the editor.


haraldschilly:
will submit a new one later

PR #525, opened 2016-04-30
Add highlighting to file listing page on mouse hover

johnjeng:


williamstein:
My suggestions: do this with two lines of CSS, which should be more efficient than a bunch of js...? Use react className and our sass files.  What do you think?

johnjeng:
I think it's not very React **but** at the moment, things like `:hover` and `media` are a known problem in React so I'm down with your proposed solution with some sort of warning comment in the code.

johnjeng:
Set aside for later.

PR #527, opened 2016-05-02
Zendesk

haraldschilly:


williamstein:
As soon as this has been double checked to work on the backend, I want to test further -- so I can see what a created ticket looks like on the client.

haraldschilly:
I've not really figured out the problem you reported, but maybe it is actually related to the setup to roll it out.

PR #528, opened 2016-05-02
storage/process_update: async-serializing update_db → action → update_db and afterwards clearing tasks[project_id] indicator

haraldschilly:
this might help with getting excessive project open requests under control (it might not solve it, but makes this function more predictable and easier to reason about)


williamstein:
Done a different way with you.

PR #555, opened 2016-05-15
Change python wizard print examples to use python3 function (do NOT merge until Python3 printing is the default in Sage worksheets)

fchapoton:
I am trying to push towards using the python3 print syntax in sage. Let us try to prepare the ground.

### DIFF
--- a/src/wizard/input/python/intro.yaml
+++ b/src/wizard/input/python/intro.yaml
@@ -7,7 +7,7 @@ category: Language / Tutorial
 ---
 title: Hello World
 descr: The "Hello World" program
-code: print "Hello World"
+code: print("Hello World")
 ---
 title: Variables
 descr: >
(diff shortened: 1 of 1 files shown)


haraldschilly:
yes, good idea!

williamstein:
-1

I'm not so sure it's good since

print 1,2

in Python2 is *completely different* (!) than

print(1,2)

This sucks but is what it is.   The first prints two numbers and the second
prints a single tulle.

On Sunday, May 15, 2016, Harald Schilly <notifications@github.com> wrote:

> yes, good idea!
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/sagemathinc/smc/pull/555#issuecomment-219287227>
>


-- 

fchapoton:
That's true. But there is nothing like that in the proposed commit.

williamstein:
On Sun, May 15, 2016 at 8:29 AM, Frédéric Chapoton <notifications@github.com
> wrote:

> That's true. But there is nothing like that in the proposed commit.
>

You can't tell people to use print as a function call in Python 2 when the
meaning is completely different, and many things work wrong...


-- 
William (http://wstein.org)

fchapoton:
Sure. But the plan is to turn print to python 3 behaviour in sage as soon
as possible.

The proposed commit can and should probably wait until this happen.

2016-05-15 17:41 GMT+02:00 William Stein <notifications@github.com>:

haraldschilly:
here are tickets about this, where many are already closed: http://trac.sagemath.org/query?status=closed&status=positive_review&component=python3&order=priority

of course, the best timing will be to merge this ticket, when we update to sage 7.2 (?) as our SMC wide default

williamstein:
Oh heck, I'll just merge this... especially since the wizard is still an easter egg.

PR #595, opened 2016-06-02
pricing.html: link from the top to the FAQ at the bottom. later, mayb…

haraldschilly:
…e just a link to the wiki


haraldschilly:
@GregoryBard this is the PR for adding a link from the top of the pricing page to the FAQ at the bottom. 

at the top of such a PR in github, there is a tab "files changed". when you click there, you see in green what I've added. I'm not sure if that's good english. also, once we figure out what to do with the actual FAQ listing, that link might change and point to the wiki page you're working on.

williamstein:
perfect.

PR #571, opened 2016-06-03
use server clock rather than user clock for timestamps in various places in SMC client code and display a warning if the user's clock is off by > 1 minute.

timothyclemansinsea:
Here's an example commit along these lines: 

https://github.com/sagemathinc/smc/commit/d12bac9edacb5b7a88135707bfa389eb58addedc


williamstein:
Definitely LGTM (looks good to me).  Will merge later today.

PR #460, opened 2016-06-04
When creating project if no subscription then give blurb about subscribing otherwise let me upgrade project

timothyclemansinsea:
Have a checkbox for internet access and auto check it available internet access upgrades and auto checked checkbox for members only hosting


src/smc-webapp/redux_account.coffee
```diff
@@ -171,6 +171,11 @@ class AccountStore extends Store
     # uses the total upgrades information to determine, if this is a paying member
     is_paying_member: =>
         ups = @get_total_upgrades()
+        console.log(JSON.stringify(ups))
```
williamstein: remove this before making live.

williamstein:
Some comments:

Typo in “You can share upgrades with any project you are a collobrator on.”

It seems like “ Adjust your project quota contributions” is too low on the page.


Clicking on “upgrade” in “A project is your own private computational workspace that you can share with others and upgrade.” could maybe scroll the page down to “ Adjust your project quota contributions” (not sure how to do that though…)

Are you rewriting the column labeled “Your contribution”?

I think it would be more consistent for it to be exactly like before, where you give nothing by default, and there is a max button you can click to max out your contributions.

We could also add a max button at the top to max all of the contributions.

haraldschilly:
merged with my changes in 6b2c13c

PR #609, opened 2016-06-09
Make autoscroll happen on submit and not keydown

johnjeng:
No other chat program I know of scrolls to the bottom on entering text. All of them do it on submit. The current behavior makes it impossible to view an old message and type a response at the same time.


haraldschilly:
LGTM

PR #640, opened 2016-06-21
Autofocus email in sign-in form always

johnjeng:
For discussion on whether or not this is a good idea.

I argue that FB does it and our sign up is based on theirs. Also since the sign-up area is so large, a new user is less likely need help finding it. 

On my sample size of 3, autofocusing the email bar makes it immediately obvious where to sign in. Otherwise it is not.

I could be totally wrong. 
- @timothyclemansinsea Wants to merge the two forms. 
- I think that's more confusing.

### DIFF
--- a/src/smc-webapp/landing_page.cjsx
+++ b/src/smc-webapp/landing_page.cjsx
@@ -111,7 +111,7 @@ SignUp = rclass
             <AccountCreationEmailInstructions />
             <form style={marginTop: 20, marginBottom: 20} onSubmit={@make_account}>
                 {@display_error("first_name")}
-                <Input ref='name' type='text' autoFocus={not @props.has_account} placeholder='First and last Name' />
+                <Input ref='name' type='text' autoFocus={false} placeholder='First and last Name' />
                 {@display_error("email_address")}
                 <Input ref='email' type='email' placeholder='Email address' />
                 {@display_error("password")}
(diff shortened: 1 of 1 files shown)


haraldschilly:
+1 to autofocus

williamstein:
Some thoughts:

- If you go to the sagemathcloud website and there is evidence (cookies, localstorage, etc.) that you've signed in before, autofocus the sign-in form.

- If you go to the website and you've never signed in before, and you should be creating an account, it will be confusing that the focus is on sign in.

- With Facebook, almost everybody on the planet already has an account.  Thus if you go to the Facebook page, the probability you want to sign in (instead of create an account) is VERY high. However, with SMC, almost nobody has an account.  Thus if you go to the SMC page, the probability you want to create an account instead of sign in is VERY high.  At least, this is my guess.  We could actually gather data to find out, right?  

It's a fact that in the last month X% of users went to our landing page and signed in and (100-X)% went to the landing page and created an account.  Shouldn't we compute X first to decide which to autofocus?

haraldschilly:
If I understand this correctly, regarding "sessions", this sign-in X% is 85 and the sign-up is 1.5%. the rest are those which just came by and didn't do anything. I can try to filter this down, but I think that's already a good hint.

williamstein:
On Tue, Jun 21, 2016 at 2:24 PM, Harald Schilly <notifications@github.com>
wrote:

> If I understand this correctly, regarding "sessions", this sign-in X% is
> 85 and the sign-up is 1.5%. the rest are those which just came by and
> didn't do anything. I can try to filter this down, but I think that's
> already a good hint.
>

WOW -- if  "this sign-in X% is 85 " is really true.. then let's autofocus
the email address, as in the patch.



-- 
William (http://wstein.org)

PR #656, opened 2016-06-29
Improvements/account deletion

johnjeng:
In staging here: https://cloud.sagemath.com/7552f6e3-0f4f-4476-8954-38b9f14c6069/port/57488/

Just make an account and delete it in account settings. You should get logged out and upon trying to log in, your account should not be found.


williamstein:
Made insanely easy https://github.com/sagemathinc/smc/issues/657 when looking at this.

williamstein:
LGTM -- this works perfectly so far as I can tell.

haraldschilly:
I've read the patch. The only comment I have is, that the confirmation step is entirely checking on the webapp client side, if the intention is really by the user. Maybe, there should be a more sophisticated process, such that there is no way that this can be triggered entirely from the javascript client side?

My first thought was, well, maybe just type in your password! But when you have a federated account, there might not really be one? Another possibility would be, to share some short string between webapp and hub, and only if the correct string is sent to the hub, the deletion is confirmed: e.g. first on the hub, `hash(account_id + shared_secret)[..6]` for 6 unique characters, sent to the webapp, displayed, and then sent back to the hub as confirmation.

Besides this overthinking of mine, it looks all fine.

williamstein:
1. Yes, there very likely might not be a password.

> Besides this overthinking of mine, it looks all fine.

Does your suggestion genuinely increase security/safety in any way?  I'm not seeing it.

PR #658, opened 2016-06-29
Improvements/chat

simonluu:
Improved chat looks, added name to top of message, prevent automatic scroll to bottom


williamstein:
LGTM

PR #659, opened 2016-06-29
Enable search bar to behave like mini terminal.

johnjeng:
Staging here: https://cloud.sagemath.com/7552f6e3-0f4f-4476-8954-38b9f14c6069/port/57488/

No UI changes occur except to show output and to warn user.

To use, begin a search with `/` followed by a miniterm command.

### DIFF
--- a/src/smc-webapp/project_files.cjsx
+++ b/src/smc-webapp/project_files.cjsx
@@ -36,7 +36,11 @@ immutable             = require('immutable')
 underscore            = require('underscore')
 {salvus_client}       = require('./salvus_client')
 
+{salvus_client} = require('./salvus_client')
+
 Combobox = require('react-widgets/lib/Combobox') #TODO: delete this when the combobox is in r_misc
+
+TERM_MODE_CHAR = '/'
 
 exports.file_action_buttons = file_action_buttons =
         compress :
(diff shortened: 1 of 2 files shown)


johnjeng:
@haraldschilly: I mentioned this a little while ago. No different from the current mini-term though so maybe you will not like it so much.

haraldschilly:
well, I don't remember anything like that. Maybe it was @williamstein you were talking to? 

After looking briefly at the code, I could also envision that any space character could trigger this terminal mode. Would be a neat way to educate everyone, that spaces in filenames and paths are a bad thing to do

johnjeng:
>spaces in filenames and paths are a bad thing to do

While I would agree that they can be a huge pain to support, it's extremely natural to do as a user doing basic file and folder stuff.

williamstein:
> Would be a neat way to educate everyone, that spaces in filenames and paths are a bad thing to do

Says who!   What would should do is make it so SMC *does* work fine with spaces in filenames and paths.  It's just a linux hacker perspective to not support spaces (etc.) in filenames.  We should always fully support that as much as possible, rather than creating broken tools.

src/smc-webapp/project_files.cjsx
```diff
@@ -1461,10 +1490,69 @@ ProjectFilesSearch = rclass
         selected_file_index : 0
         num_files_displayed : 0
 
+    getInitialState : ->  # Miniterm functionality
+        stdout : undefined
+        state  : 'edit'   # 'edit' --> 'run' --> 'edit'
+        error  : undefined
+
+    # Miniterm functionality
```
williamstein: I'm confused - why is this big block of coded added, but no corresponding deleting?  Did you just copy the same code again.... seems ugly.

johnjeng: Yup. No idea what will eventually happen to the actual miniterm component. It thought it was potentially getting axed later. Miniterm and its functionality is used no where else. It's certainly not DRY but I don't see a reason for it to be when the parts are well commented and the added functionality itself is experimental. I didn't see the value in abstracting in this case. 

If we want to mimic this in more places, I'm all for moving it out.

haraldschilly:
oh well, I wasn't serious with that last sentence. The intended smiley at the end is missing ...

GregoryBard:
+1. I agree with William. The presence of spaces, periods, dashes, etc..
inside of filenames and folder names has become extremely common among
business people. They seem to like that. I see it all the time. It will
probably spread.
---Greg

Issue #663, opened 2016-06-30
change the new green button in chat to default (so grey)


williamstein:
I really, really don't like the green to bottom button in chatrooms.  Green is usually the "save" button in SMC, and seeing a green button that isn't disabled means "stuff isn't saved", which is kind of disturbing.  Please send PR to change this soon.

johnjeng:
Haha this is a superb reason to hate it. Changing it today.

PR #666, opened 2016-06-30
Change chat scroll down button to default color

johnjeng:
Fixing #663. Also made buttons into a group.


haraldschilly:
I'm reviewing it now

haraldschilly:
ok, it needed an additional merge, but that went fine and the buttons are as advertised …

PR #675, opened 2016-07-04
Hal jupyter

DrXyzzy:
Enable sagews as jupyter client using e.g. `p2 = jupyter("python2")`. Get help and list available kernels with `jupyter?` from sagews.

Requires python package ansi2html. Tested with version 1.1.1.


src/smc_sagews/smc_sagews/sage_salvus.py
```diff
@@ -3571,3 +3571,302 @@ def help(*args, **kwds):
...
+        if scroll:
+            h2 = '<div style="max-height:320px;width:80%;overflow:auto;">' + h2 + '</div>'
+        salvus.html(h2)
+
+    def big_output(f,*args, **kwargs):
+        r"""
+        Temporarily raise ceilings on output limits (for data uris and docs).
+        Arbitrary new upper limit is 2000000 for both.
+        Restore to prior values after calling the function provided.
+        """
```
williamstein: Please wrap this in a try/finally?  Otherwise if anything goes wrong (or there is an interrupt signal), then the limits aren't restored.

src/smc_sagews/smc_sagews/sage_salvus.py
```diff
@@ -3571,3 +3571,302 @@ def help(*args, **kwds):
...
+sage_jupyter.py
+
+Spawn and send commands to jupyter kernels.
+"""
+
+#########################################################################################
+#       Copyright (C) 2016, SageMath, Inc.                                              #
+#                                                                                       #
+#  Distributed under the terms of the GNU General Public License (GPL), version 2+      #
+#                                                                                       #
```
williamstein: Can you add an `AUTHOR: Hal Snyder` thing as well, like we do with Sage.  It's useful to easily know who wrote the first version of something...

src/smc_sagews/smc_sagews/sage_salvus.py
```diff
@@ -3571,3 +3571,302 @@ def help(*args, **kwds):
...
+
+import jupyter_client
+from Queue import Empty
+from ansi2html import Ansi2HTMLConverter
+import tempfile, sys, re
+import base64
+
+def jkmagic(kernel_name, **kwargs):
+    r"""
+    See docs for jupyter
```
williamstein: maybe a URL linking to those docs?

src/smc_sagews/smc_sagews/sage_salvus.py
```diff
@@ -3571,3 +3571,302 @@ def help(*args, **kwds):
...
+            | %auto
+            | a3 = jupyter('anaconda3')
+            | %default_mode a3
+
+        Each magic command connects to its own kernel. So you can have more than
+        one instance of the same kernel type.
+
+            | my_second_python3 = jupyter("python3")
+
+        """)
```
williamstein: Maybe say something about what isn't (or is) supported?

williamstein:
I made some of these changes and separated out the sage_jupyter file.

Issue #67, opened 2015-08-18
Courses --  ignore deleted assignments when adding new assignments


williamstein:
Reported by: Jane Shevtsov

johnjeng:
This is already true.

williamstein:
Do you mean you fixed this in some branch not in master?   It's definitely broken in Master.  I just made  a course and tested it with the simplest example and it failed.

johnjeng:
Interesting. I added an assignment, deleted it in the .course file, and then it doesn't show up in add assignment search.

williamstein:
> Interesting. I added an assignment, deleted it in the .course file, and then it doesn't show up in add assignment search.

Which is the whole point!  That's the problem.   You just confirmed that the bug is still there.

Imagine you are Jayadev Arethraya and you get access to our math 480 course file for a day (like he did).   Then you add the directory `assignments/` as an assignment (like he did).  Then you notice that this is a mistake and delete that assignment (like he did).  Then you're me and you try to add `assignment/assign8`, like I did, but it won't show up!  URGH.  The reason is because assignment search won't show anything that is contained in any existing assignment, both deleted or not.   When this happened last quarter to us, I had to manually edit the .course file with vim to remove `assignments` as an assignment.    Having a way to permanently delete all deleted assignments would also be good (we have that for students, but I never got around to implementing it for assignments).

johnjeng:
Ah. I get it. I thought that was intentional behavior. lol

I'm 100% on board with this. Will fix it tomorrow.

williamstein:
On Tue, Jun 21, 2016 at 8:18 PM, John Jeng <notifications@github.com> wrote:

> Ah. I get it. I thought that was intentional behavior. lol
>
> I'm 100% on board with this. Will fix it tomorrow.
>

Awesome, thanks!  Because I really want this fixed, but I'm trying to
architect an iscsi storage backend thing... and it's really far away from
this.  THANKS!


> —
> You are receiving this because you modified the open/close state.
> Reply to this email directly, view it on GitHub
> <https://github.com/sagemathinc/smc/issues/67#issuecomment-227632555>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/ABN5dqajy0KR6oeCTV4PYEiHcz_qu3k2ks5qOKmSgaJpZM4Ftwvw>
> .
>



-- 
William (http://wstein.org)

johnjeng:
Okay, clarification request on this ticket.
[...]

PR #680, opened 2016-07-05
Course editor handouts

johnjeng:
In staging: https://cloud.sagemath.com/7552f6e3-0f4f-4476-8954-38b9f14c6069/port/57488/

Okay, I claim this is ready.


src/smc-webapp/course/course_funcs.coffee
```diff
@@ -0,0 +1,67 @@
...
+
+# SMC libraries
+misc = require('smc-util/misc')
+{defaults, required} = misc
+
+immutable = require('immutable')
+
+# Transforms Iterable<K, M<i, m>> to [M<i + primary_key, m + K>] where primary_key maps to K
+# Dunno if either of these is readable...
+# Turns Map(Keys -> Objects{...}) into [Objects{primary_key : Key, ...}]
```
johnjeng: I don't even know if this notation makes sense. Either one also reads kind of like code.

src/smc-webapp/course/handouts_panel.cjsx
```diff
@@ -0,0 +1,563 @@
...
+
+# SMC and course components
+course_funcs = require('./course_funcs')
+styles = require('./styles')
+{BigTime, FoldersToolbar} = require('./common')
+{Icon, Tip, SearchInput, MarkdownInput} = require('../r_misc')
+
+# Could be merged with steps system of assignments.
+# Probably not a good idea mixing the two.
+# Could also be coded into the components below but steps could be added in the future?
```
johnjeng: I don't want to mix the two given the express difference in concept between handouts and assignments even though both are just folders. 

In addition to the reason given, structuring the code this way also provides a nice stepping stone for future devs to understand what's happening in the assignments' `STEPS` related code.

src/smc-webapp/course/handouts_panel.cjsx
```diff
@@ -0,0 +1,563 @@
...
+            <Col xs=10>
+                <MarkdownInput
+                    rows          = 6
+                    placeholder   = 'Private notes about this handout (not visible to students)'
+                    default_value = {@props.handout.get('note')}
+                    on_save       = {(value)=>@props.actions.set_handout_note(@props.handout, value)}
+                />
+            </Col>
+        </Row>
+
```
johnjeng: The small render methods after this comment could be moved off since it mostly shares with the same code in assignments. However, it is slightly different enough that I prefer to keep it as is until we have at least a third area where we have the same code use.

src/smc-webapp/course/handouts_panel.cjsx
```diff
@@ -0,0 +1,563 @@
...
+                <Tip title={@props.title} tip={if @props.title=="Handout" then "This column gives the directory name of the handout." else "This column gives the name of the student."}>
+                    <b>{@props.title}</b>
+                </Tip>
+            </Col>
+            <Col md=10 key="rest">
+                {@render_headers()}
+            </Col>
+        </Row>
+
+StudentHandoutInfo = rclass
```
johnjeng: Maintains similar construction to `StudentAssignmentInfo` yet still separate (again).

src/smc-webapp/course/main.cjsx
```diff
@@ -1223,7 +1225,188 @@ exports.init_redux = init_redux = (redux, course_project_id, course_filename) ->
             # Now open it
             redux.getProjectActions(proj).open_directory(path)
 
+        # Handouts
```
johnjeng: Simplified copies of much of what's happening in similarly named assignment methods. Some cases you could literally change `handout` for `assignment` and you'd have the other function. Keeping separate same as previous changes.

src/smc-webapp/course/common.cjsx
```diff
@@ -365,3 +368,180 @@ exports.StudentAssignmentInfo = rclass
...
+            query      : "*#{search.trim()}*"
+            cb         : (err, resp) =>
+                if err
+                    @setState(add_is_searching:false, err:err, add_search_results:undefined)
+                else
+                    filtered_results = @filter_results(resp.directories, search, @props.items)
+                    @setState(add_is_searching:false, add_search_results:filtered_results)
+
+    # Filter directories based on contents of all_items
+    filter_results : (directories, search, all_items) ->
```
johnjeng: Partial fix to #67 here in the following way:

1. Filter out items that are added and *not* deleted. Deleted items will show up in add search.
1. Matches paths exactly for omitting them. ie. if `assignments/` is deleted, `assignments/01` still shows up
2. However, nothing has changed if the deleted item is added..

Okay crap. 3. means that this is probably no good and needs to be changed before merging. My planned solution is to keep 1 and to only undelete deleted items when they get added again. This "fixes" the case where an instructor gets confused when a folder they have isn't active and can't be found (but is actually in the deleted area).

Obviously can be re-implemented to trash/deletion later if desired.

src/smc-webapp/course/common.cjsx
```diff
@@ -365,3 +368,180 @@ exports.StudentAssignmentInfo = rclass
...
+                placeholder   = "Add #{@props.item_name} by folder name (enter to see available folders)..."
+                on_submit     = {@props.do_search}
+                on_escape     = {@clear_and_focus_search_input}
+                buttonAfter   = {@search_button()}
+            />
+            {@render_add_selector() if @state.show_selector}
+         </div>
+
+# Definitely not a good abstraction.
+# Purely for code reuse (bad reason..)
```
johnjeng: At least I think it's a bad reason for the very next comment. This "almost" pure enough result is why I didn't refactor out a lot of the other stuff that is *very* similar between `Assignment` and `Handout`.

src/smc-webapp/course/handouts_panel.cjsx
```diff
@@ -0,0 +1,563 @@
...
+        n = status["not_#{step}"]
+        <Alert bsStyle='warning' key="#{step}_confirm_to_all", style={marginTop:'15px'}>
+            <div style={marginBottom:'15px'}>
+                {misc.capitalize(step_verb(step))} this handout {step_direction(step)} the {n} student{if n>1 then "s" else ""}{step_ready(step, n)}?
+            </div>
+            <ButtonToolbar>
+                <Button key='yes' bsStyle='primary' onClick={=>@copy_handout(step, false)} >Yes</Button>
+                {@render_copy_cancel(step)}
+            </ButtonToolbar>
+        </Alert>
```
johnjeng: These small render methods could be moved off since it mostly shares with the same code in assignments. However, it is slightly different enough that I prefer to keep it as is until we have at least a third area where we have the same code use.

src/smc-webapp/course/handouts_panel.cjsx
```diff
@@ -0,0 +1,563 @@
...
+StudentHandoutInfo = rclass
+    displayName : "CourseEditor-StudentHandoutInfo"
+
+    propTypes :
+        actions    : rtypes.object.isRequired
+        info       : rtypes.object.isRequired
+        title      : rtypes.oneOfType([rtypes.string,rtypes.object]).isRequired
+        student    : rtypes.oneOfType([rtypes.string,rtypes.object]).isRequired # required string (student_id) or student immutable js object
+        handout    : rtypes.oneOfType([rtypes.string,rtypes.object]).isRequired # required string (handout_id) or handout immutable js object
+
```
johnjeng: Maintains similar construction to StudentAssignmentInfo yet still separate (again).

src/smc-webapp/course/main.cjsx
```diff
@@ -1223,7 +1225,188 @@ exports.init_redux = init_redux = (redux, course_project_id, course_filename) ->
...
+        add_handout: (path) =>
```
johnjeng: Simplified copies of much of what's happening in similarly named assignment methods. Some cases you could literally change `handout` for `assignment` and you'd have the other function. 

Here there is much more code to potentially reuse but even more so in this case, I would like to not entwine assignments and handouts into the same code.

williamstein:
Seems perfect.

PR #686, opened 2016-07-07
Change account settings buttons to standard smc behavior

johnjeng:


src/smc-webapp/r_account.cjsx
```diff
@@ -246,49 +242,48 @@ PasswordSetting = rclass
...
+                onChange    = {=>@setState(old_password : @refs.old_password.getValue())}
+            />
+            New password
+            <form onSubmit={(e)=>e.preventDefault();if @is_submittable() then @save_new_password()}>
+                <Input
+                    type        = 'password'
+                    ref         = 'new_password'
+                    value       = {@state.new_password}
+                    placeholder = 'New password'
+                    onChange    = {=>x=@refs.new_password.getValue(); @setState(zxcvbn:password_score(x), new_password:x)}
```
haraldschilly: oh, regarding `zxcvbn`, I see. it assumes to be global. maybe that should be a get_zxcvbn() function and there it does this `require.ensure`.

haraldschilly: well, no, but I hope you get the idea ... :-\

PR #691, opened 2016-07-12
[MAJOR BUG!] create project without upgrades

timothyclemansinsea:
I just created a new project and clicked "Create Project Without Upgrades".  This **still** applied the members only and networking upgrades.  This makes absolutely no sense (i.e., it is a clear bug).  Please fix.


williamstein:
Good job!

Issue #694, opened 2016-07-11
try-except-else raises syntax error in SMC cell


DrXyzzy:
As reported by c. e. larson [in this posting on sage-cloud google group](https://groups.google.com/d/msg/sage-cloud/NWLottYk0CM/-ObQhByFVf8J):

the "else" in the following code raises a "syntax" error when run in a Sage Cloud cell:

```
try:
    print "try now"
except:
    print "some exception"
else:
    print "made it to else"

*but* it runs normally if the code is in a function definition or even a junk for loop, like:

for i in [0]:
    try:
        print "try now"
    except:
        print "some exception"
    else:
        print "made it to else"
```

haraldschilly:
oh, this is very similar to a `for: ... else: ...` which I did report a long long time ago. I bet, you just have to search for where this `else` case is handled in the parser, when it is in a `for` state. the same needs to be done for `try`.

DrXyzzy:
Thanks per @williamstein 

> Probably 2-3 lines added right around here will fix this bug: https://github.com/sagemathinc/smc/blob/master/src/smc_sagews/smc_sagews/sage_parsing.py#L256

DrXyzzy:
You can assign this to me if you like.

haraldschilly:
that's where you'll find a clue: https://github.com/sagemathinc/smc/commit/6fdd71b5ca54bd29d9b2db6ee47d02b51fc6be01

DrXyzzy:
Done.

PR #699, opened 2016-07-12
issue 694 merge else code block with previous try

DrXyzzy:
- one-liner fix, add 'try' to possible parent code blocks of 'else'
- not sure we need to look for 'elif' but I left it there
- decided against any() with list comp because list comp. doesn't short circuit; generator would be overkill


DrXyzzy:
ref https://github.com/sagemathinc/smc/issues/694

williamstein:
Nice!

PR #705, opened 2016-07-15
Fix issue 704

### LINKED ISSUE #704: editing chat does NOT edit the chat you typed
Type a chat with content
```
> foo
```
then edit it and see
```
&gt; test
```

johnjeng:
Fixes #704. Was sanitizing before storage. We just need to sanitize on render which happens.


haraldschilly:
isn't this then storing maybe broken html? but I guess, with that patch the chat would still work, though.

haraldschilly:
ahh, I didn't realize that an issue is referenced. ok yes, let's merge that.

PR #706, opened 2016-07-15
Attach import repl fix

DrXyzzy:
- wrap second import from sage.repl.attach, in sage_salvus.py with try/except

- if either import from sage.repl.attach fails, change attach() function to print warning to worksheet stderr; the import exception messages only go to sage_server.log

### DIFF
--- a/src/smc_sagews/smc_sagews/sage_salvus.py
+++ b/src/smc_sagews/smc_sagews/sage_salvus.py
@@ -3180,42 +3180,52 @@ def load_html_resource(filename):
     elif ext == "js":
         salvus.html('<script src="%s"></script>'%url)
 
-def attach(*args):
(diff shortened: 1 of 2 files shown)


haraldschilly:
I don't know this part enough to review it. E.g. I'm wondering what `attach_available()` is, looks more like a boolean flag in a variable than a function?

DrXyzzy:
That was code from earlier dev that should have been removed. It is now.

It has been tested for failure of either or both sage.repl.attach imports (by temporarily adding misspelled module names to the imports) and for happy path in https://cloud.sagemath.com/projects/ccd4d4a4-29a8-4c39-85c2-a630cb1e9b6c/files/SAGEWS/issue290test.sagews.

PR #707, opened 2016-07-15
Require typing first and last name to delete account

### LINKED ISSUE #688: in account deletion instead of "To proceed, type delete this account below." change to typing something specific about the account
The reason is that typing "delete this account" doesn't in any way force the user to think about **which** account they are deleting.  It would be far better from a UX perspective to actually require typing something related to their account, e.g., the First and Last Name or the email address (if there is one).

johnjeng:
Fixes #688.

### DIFF
--- a/src/smc-webapp/r_account.cjsx
+++ b/src/smc-webapp/r_account.cjsx
@@ -475,9 +475,10 @@ AccountSettings = rclass
                 <Col xs=12>
                     <DeleteAccount
                         style={marginTop:'1ex'}
-                        initial_click={()=>@setState(show_delete_confirmation:true)}
-                        confirm_click={=>@props.redux.getActions('account').delete_account()}
-                        cancel_click={()=>@setState(show_delete_confirmation:false)}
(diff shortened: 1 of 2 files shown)


src/smc-webapp/r_account.cjsx
```diff
@@ -475,9 +475,10 @@ AccountSettings = rclass
                 <Col xs=12>
                     <DeleteAccount
                         style={marginTop:'1ex'}
-                        initial_click={()=>@setState(show_delete_confirmation:true)}
-                        confirm_click={=>@props.redux.getActions('account').delete_account()}
-                        cancel_click={()=>@setState(show_delete_confirmation:false)}
+                        initial_click = {()=>@setState(show_delete_confirmation:true)}
+                        confirm_click = {=>@props.redux.getActions('account').delete_account()}
```
williamstein: There is a `get_fullname` method on the account store, which deals with some special cases (e.g., possibility of first or last name being undefined...)

williamstein:
This looks good to me (except for one comment).

Issue #621, opened 2016-06-12
Embed pay in place in the free server warning banner and in project settings


timothyclemansinsea:

williamstein:
What's the status of this...?

PR #708, opened 2016-07-15
[needs work] Upstream issue 621

timothyclemansinsea:
For #621

### DIFF
--- a/src/smc-webapp/paying_and_upgrading.cjsx
+++ b/src/smc-webapp/paying_and_upgrading.cjsx
@@ -1,3 +1,8 @@
+###
+Put a header here...?
+###
+
+
 {Button, Panel, Row, Col} = require('react-bootstrap')
 {Icon, Space} = require('./r_misc')
 misc = require('smc-util/misc')
@@ -8,6 +13,7 @@ misc = require('smc-util/misc')
 {User} = require('./users')
 {BillingPageSimplifiedRedux} = require('./billing')
 {UpgradeAdjustorForUncreatedProject} = require('./project_settings')
+{ProjectTitle} = require('./projects')
 
 {PROJECT_UPGRADES} = require('smc-util/schema')
 
@@ -15,6 +21,8 @@ redux_name = (project_id) ->
     if project_id
         return "paying-and-upgrading-#{project_id}"
     return "paying-and-upgrading"
+
+exports.upgrade_project_redux_name = redux_name
 
 class PayingAndUpgradingActions extends Actions
     init: (@project_id) =>
@@ -28,6 +36,8 @@ exports.init_redux = init_redux = (redux, project_id) ->
     actions.init(project_id)
 
     redux.createStore(name)
+
+exports.init_project_upgrade_redux = init_redux
 
 UpgradeProject = (name) -> rclass
     reduxProps:
@@ -44,11 +54,12 @@ UpgradeProject = (name) -> rclass
             show_all          : rtypes.bool
         billing :
             customer      : rtypes.object
+        account :
+            stripe_customer : rtypes.immutable
 
     propTypes :
         project_id                           : rtypes.string
-        upgrades_you_can_use                 : rtypes.object
-        upgrades_you_applied_to_all_projects : rtypes.object
+        projects_store                       : rtypes.object
         quota_params                         : rtypes.object.isRequired # from the schema
         actions                              : rtypes.object.isRequired # projects actions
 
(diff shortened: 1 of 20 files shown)


src/smc-webapp/paying_and_upgrading.cjsx
```diff
@@ -0,0 +1,470 @@
...
+                        <b style={fontSize:'12pt'}>Your contribution</b>
+                        <br/>
+                        <Button
+                            bsSize  = 'xsmall'
+                            onClick = {=>@max_upgrades()}
+                            style   = {padding:'0px 5px'}
+                        >
+                            Max all upgrades
+                        </Button>
+                        {' '}
```
johnjeng: We have a <Space/> Component but I guess this is pretty readable too.

johnjeng:
Is this the same thing that's on the Projects listing page when you make a new project?

timothyclemansinsea:
I ran into issues with trying to reuse this new component. Asked help from
William was told doing too much at once so backed out changes to project
settings upgrader.

haraldschilly:
well, for me, the most important issue around project upgrades is, that the UI allows to enter values above the maximum limit. I.e. when someone has >9000mb for memory and enters 9000, it should give an error and not allow that.

I partially fixed this in the upgrade component for the "new project" dialog and hoped, that this part and the adjustment dialog in the project settings will be the same or at least also respects that upper limit. I can see, that there is part of the same code as over there in the new project dialog (because there is this `map_limit`, etc.) -- but it doesn't give any warnings and it also just allowed me to add >8000 mb  of memory upgrades.

In the "new project" component, e.g. this line is showing a warning:
https://github.com/sagemathinc/smc/blob/439fd887997c7f7fb277b0b7351cf0aeaec3bfd4/src/smc-webapp/projects.cjsx#L804
(and related to that, unfortunately, it still lets users create projects with too many upgrades despite that warning error)

So, I'm favor of first fixing those upgrade components regarding entering valid values and respecting upper limits, and then re-using this component from the "new project" upgrade here, such that we can be sure that both are preventing users from upgrading above the limits. (I know, that's just the UI, but we need to show users that there are upper limits and not allow them to still create projects)

src/smc-webapp/projects.cjsx
```diff
@@ -539,12 +539,15 @@ class ProjectsStore extends Store
 
     # Get the total quotas for the given project, including free base values and all user upgrades
     get_total_project_quotas : (project_id) =>
+        console.log('gtpq', project_id)
```
haraldschilly: there is some debugging info left

src/smc-webapp/project.coffee
```diff
@@ -41,7 +41,7 @@ project_store   = require('./project_store')
 {alert_message} = require('./alerts')
 misc_page       = require('./misc_page')
 
-{redux}         = require('./smc-react')
+{redux, ReactDOM} = require('./smc-react')
```
haraldschilly: I think, `ReactDOM` isn't used here. I also think, that inside a .coffee file it shouldn't be.

src/smc-webapp/project.coffee
```diff
@@ -168,8 +173,6 @@ class ProjectPage
                             warning_banner_hidden.push(@project_id)
                             box.hide()
                         box.show()
-                    else
-                        box.hide()
```
haraldschilly: I'm not sure, but IIRC, that last `else box.hide()` has the use, that it hides a maybe visible box, when someone comes back to the project after upgrading. Maybe something has changed, but it had a purpose.

src/smc-webapp/paying_and_upgrading.cjsx
```diff
@@ -0,0 +1,470 @@
...
+    render_addon : (misc, name, display_unit, limit, val, remaining) ->
+        <div style={minWidth:'81px'}>{"#{misc.plural(2,display_unit)}"} {@render_max_button(name, limit) if remaining > 0}{@render_reset_button(name) if val > 0}</div>
+
+    render_get_more_of_this_upgrade : (name) ->
+        # Detrimine if user has subscription that has given upgrade. Like cores requires permium subscription
+        <DropdownButton title='Get more of this upgrade by' bsSize='xsmall'>
+            <MenuItem>Purchasing an additional subscription</MenuItem>
+            <MenuItem>Downgrading other projects</MenuItem>
+            <MenuItem>Resetting this upgrade type for all projects</MenuItem>
+        </DropdownButton>
```
haraldschilly: I still don't know what this DropdownButton really does. it makes this look a bit chaotic, since it repeats so often. maybe for now, we should leave that out?

src/smc-webapp/project.coffee
```diff
@@ -116,11 +116,16 @@ class ProjectPage
         delete @actions
         delete @store
 
+    get_total_project_quotas: (s, project_id) =>
+        quotas = s.get_total_project_quotas(project_id)
+        quotas.project_id = project_id
```
haraldschilly: right here, I got `project.coffee?a24e:121 Uncaught TypeError: Cannot set property 'project_id' of undefined`

so, it looks as if `s.get_total_project_quotas` sometimes returns null or undefined?

haraldschilly:
I also added a few inline comments, e.g. I got an `Uncaught TypeError: Cannot set property 'project_id' of undefined` exception...

timothyclemansinsea:
I really want to keep the dropdowns as they make it possible to rapidly free up quota. It's something that I'll use a lot.

src/smc-webapp/project.coffee
williamstein: > so, it looks as if s.get_total_project_quotas sometimes returns null or undefined?

YES -- absolutely all code **must** be written to allow for cases like this.  If the relevant project/account/etc. information isn't loaded yet from the backend, then you have to fall back to a reasonable default behavior until it is loaded.   Here, doing `quotas?.project_id = project_id` would  I think work.

williamstein:
What's going on with this  PR?

timothyclemansinsea:
I closed it. I lost a lot of my code and need to recover it.

PR #710, opened 2016-07-16
[needs review] smc project: explicitly specify ports

haraldschilly:
for running a local hub in a container, one has to specify the ports. this patch maybe does that, I'm not sure. in any case, it needs review and testing ...


williamstein:
Harald, instead of passing **positional** arguments through the Python code to specify ports, can you instead just pass through arbitrary arguments?    Positional arguments like this are very brittle and error prone.

src/smc_pyutil/smc_pyutil/start_smc.py
```diff
@@ -14,14 +15,21 @@
...
+    # optionally, tcp port and raw port numbers
+    port_args = ''
+    if len(sys.argv) >= 3:
+        port_args += ' %s' % sys.argv[2]
+    if len(sys.argv) >= 4:
+        port_args += ' %s' % sys.argv[3]
+
     # Start local hub server
-    cmd("smc-local-hub start")
+    cmd("smc-local-hub start" + port_args)
```
williamstein: Delete all the above added code and just do `cmd("smc-local-hub start " + ' '.join(port_args))`

src/smc-project/bin/smc-local-hub
```diff
@@ -17,6 +17,14 @@ elif sys.argv[1] == 'start':
...
 target = os.path.abspath(os.path.join(path, 'local_hub.coffee'))
-cmd  = "forever %s -c coffee -l $SMC/local_hub/local_hub.log --pidFile $SMC/local_hub/local_hub.pid -p $SMC/forever/ %s"%(sys.argv[1], target)
+
+# optionally, pass in tcp/raw port numbers for local_hub.coffee
+local_hub_args = ''
+if len(sys.argv) >= 3:
+    local_hub_args += '--tcp_port %s ' % sys.argv[2]
+if len(sys.argv) >= 4:
+    local_hub_args += '--raw_port %s ' % sys.argv[3]
+
```
williamstein: Just do `local_hub_args = ' '.join(sys.argv[2:])`

haraldschilly:
ok, arguments are passed just as concatenated strings now

Issue #2, opened 2015-06-07
Pickling of functions in a file loaded in a Sage Worksheet fails


williamstein:
Somebody (not me) reported this:

Below is a small example demonstrating the issue:
- First create a file called `test.py` with the following content:

``` python
def test_function(): pass
```
- Next open a Sage Worksheet and execute the following code:

``` python
load('test.py')
save(test_function, 'test.sobj')
```
- This crashes with the following traceback:

```
Error in lines 1-1
Traceback (most recent call last):
  File "/projects/8d39729d-4566-4056-84e2-c636147c3fc3/.sagemathcloud/sage_server.py", line 879, in execute
    exec compile(block+'\n', '', 'single') in namespace, locals
  File "", line 1, in <module>
  File "sage/structure/sage_object.pyx", line 989, in sage.structure.sage_object.save (build/cythonized/sage/structure/sage_object.c:11348)
    s = cPickle.dumps(obj, protocol=2)
TypeError: expected string or Unicode object, NoneType found
```
- Switching to `pickle` instead of `cPickle` makes it more obvious what the issue is:

``` python
import pickle
pickle.dumps(test_function, protocol=2)
```

which crashes with the following traceback:

```
Error in lines 2-2
Traceback (most recent call last):
  File "/projects/8d39729d-4566-4056-84e2-c636147c3fc3/.sagemathcloud/sage_server.py", line 879, in execute
    exec compile(block+'\n', '', 'single') in namespace, locals
  File "", line 1, in <module>
  File "/projects/sage/sage-6.7/local/lib/python/pickle.py", line 1374, in dumps
    Pickler(file, protocol).dump(obj)
  File "/projects/sage/sage-6.7/local/lib/python/pickle.py", line 224, in dump
    self.save(obj)
  File "/projects/sage/sage-6.7/local/lib/python/pickle.py", line 286, in save
    f(self, obj) # Call unbound method with explicit self
  File "/projects/sage/sage-6.7/local/lib/python/pickle.py", line 748, in save_global
    (obj, module, name))
PicklingError: Can't pickle <function test_function at 0x7f1af4bdccf8>: it's not found as __main__.test_function
```

If you try the same thing in command-line Sage, then everything works as you would expect.

DrXyzzy:
Problem occurs with functions and classes defined at top level in worksheet as well as loaded from file:

function:
```
import pickle
def f():
    pass
s = pickle.dumps(f, protocol = 2)
==>
[...]

PR #715, opened 2016-07-18
[ready for review] monkey patch save - upstream #2

DrXyzzy:
Upstream issue is [#2](https://github.com/sagemathinc/smc/issues/2)

Please review. Not sure this is the best way to deal with problem:
- Is it ok to insert functions and classes into `__main__`?
- Problem still occurs if user calls pickle or cPickle dump methods directly.

Test worksheet here:
https://cloud.sagemath.com/projects/ccd4d4a4-29a8-4c39-85c2-a630cb1e9b6c/files/SAGEWS/issue0002.sagews

Parent module of worksheet appears to be `__builtin__`, but defined functions and classes at top level do not show up in namespace for `__builtin__`. Problem notes here:
https://cloud.sagemath.com/projects/ccd4d4a4-29a8-4c39-85c2-a630cb1e9b6c/files/SAGEWS/issue0002.md

### DIFF
--- a/src/smc_sagews/smc_sagews/sage_salvus.py
+++ b/src/smc_sagews/smc_sagews/sage_salvus.py
@@ -3225,14 +3225,20 @@ except ImportError:
 # Monkey-patch the save command
 def save(obj, filename=None, compress=True, **kwds):
     """
-    Workaround save() of function or class loaded or defined at top level in worksheet: put object into __main__ module.
+    Save ``obj`` to the file with name ``filename``.
 
-    Limitations:
+    .. NOTE::
 
-    XXX Does not fix the problem if pickle.dumps() is called instead of save()
+        sage_salvus.save() is a workaround to allow saving a function or class
+        loaded or defined at top level
 
-    XXX Docstring is incomplete pending review of PR
+    .. WARNING::
 
+        Not applicable if pickle.dumps() is called instead of save()
+
+    .. SEEALSO::
+
+        :func:`sage_object.save`
     """
     objmod = getattr(obj, '__module__', None)
     if objmod is None or getattr(obj, '__module__', None) == '__builtin__':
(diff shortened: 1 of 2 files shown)


williamstein:
I think this is way better than nothing.  Write the docstring and we'll merge it.

DrXyzzy:
Docstring updated.

Note: there may have been a similar problem with sage
http://ask.sagemath.org/question/8786/how-to-save-a-function-in-sage/

src/smc_sagews/smc_sagews/sage_salvus.py
```diff
@@ -3222,6 +3222,33 @@ def attach(*args):
...
+        sage_salvus.save() is a workaround to allow saving a function or class
+        loaded or defined at top level
+
+    .. WARNING::
+
+        Not applicable if pickle.dumps() is called instead of save()
+
+    .. SEEALSO::
+
+        :func:`sage_object.save`
```
williamstein: Does this work?  I bet it doesn't.     In the worksheet `sage_object.save` isn't even defined.

A serious problem with this docstring is that if 99.99% of people actually just types "save?" to learn how to use save, instead of immediately seeing useful information like they do now, e.g.,

> Save "obj" to the file with name "filename", which will have an
".sobj" extension added if it doesn't have one and if "obj" doesn't
have its own "save()" method, like e.g. Python tuples.  [...] For image objects and the like (which have their own "save()" method), you may have to specify a specific extension, e.g. ".png",
if you don't want the object to be saved as a Sage object (or
likewise, if "filename" could be interpreted as already having some
extension).

they see some useless discussion about saving functions that they could care less about.

A fix is to make the docstring of save just be the original one from sage with your appended to it!
You can do that programmatically (so no copy/paste) -- look for `__doc__` in the same file as save for examples.

Issue #701, opened 2016-07-12
autocompletion in sage worksheet sometimes fails when it shouldn't


DrXyzzy:
Try the following tests of autocompletion in a sage worksheet:
```
8.9 units.length.a<TAB>
8.9 * units.length.a<TAB>
```
The first provides completion alternatives, but after completion the result is not syntactically correct. The second does not offer completions, but if completed manually, for example to
`8.9 * units.length.angstrom`, gives a syntactically correct expression.

DrXyzzy:
Anther example. Leading paren is a workaround.
```
class Xyz:
    numerical_attribute = 42
x1 = Xyz()

--- new cell ---

3 / x1.nu<TAB>  # no completion
3 x1.nu<TAB>    # completion
(3 / x1.nu<TAB> # completion
```

DrXyzzy:
also reported by @haraldschilly:
```
xyz = 23
-- new cell --
xy<TAB>   #completes
[xy<TAB>  #does not complete
[ xy<TAB> #does not complete
(xy<TAB>  #completes
( xy<TAB> #completes
```
adding a fix for this case

Issue #295, opened 2015-11-25
introspection parser doesn't complete properly on "for a in ab[tab]"


williamstein:
Define a variable "abc" in a worksheet, then type -- in another cell -- "for a in ab[tab]".  It should complete, but doesn't, due to the introspection parser in smc_pyutil not being good enough.

DrXyzzy:
This also doesn't introspect, with abc defined: `3 * ab[tab]`.

Issue is similar to #701, similar fix.

DrXyzzy:
Done.

PR #716, opened 2016-07-19
introspection fix for #295

DrXyzzy:
Fixes these two cases from #295 in sage worksheet:
```
abc = [1,2,3]
-- new cell --
for a in ab[tab]
3 * ab[tab]
```
These examples have been added to test file for #701 at https://cloud.sagemath.com/projects/ccd4d4a4-29a8-4c39-85c2-a630cb1e9b6c/files/SAGEWS/issue701.sagews - all tests pass


williamstein:
Agh, how could I have forgot space!

Issue #733, opened 2016-07-23
add a comment at the top of editor_chat.cjsx explaining the format


williamstein:
E.g., a chat line looks like:
```
{"sender_id":"25e2cae4-05c7-4c28-ae22-1e6d3d2e8bb5","event":"chat",
"history":[{"author_id":"25e2cae4-05c7-4c28-ae22-1e6d3d2e8bb5",
"content":"I also talked some with
 CoreOS dev Aaron Levy (https://github.com/aaronlevy) about installing 
k8s (maybe using CoreOS) on bare metal at UW...   ",
"date":"2016-07-22T14:57:07.467Z"}],"date":"2016-07-22T14:57:07.467Z"}
```

It seems clear what sender_id means, though I'm not sure what happens with multiple senders.  The event seems clear... but what events are possible?  Why is the next thing called "history"?  When would it have more than one thing in it?  Why does it have an author and there is also sender_id?  That seems redundant, but probably isn't.   

I'm not suggesting changing the format.  I'm just asking for some docs.

johnjeng:
We'll get it documented.

History gets appended to when someone edits the message. `author_id` and `date` inside each history object refer to who made the change and when. Date on the outside is when it was sent. On  a branch @simonluu has a button that lets you see a single message's edit history.

williamstein:
Re history -- Cool; that seems like a good idea.  You should note that technically the same info is in the syncstring history so it is redundant; nonetheless having it there would make rendering much easier.

williamstein:
seems done.

PR #737, opened 2016-07-23
Add docs to chat message format

johnjeng:
#733 

I could also format it in markdown if we wanted to have automated docs in the future.


williamstein:
Looks good to me.  It doesn't need to be markdown or whatever -- just usable and right.  

Just to be clear, if I edit a 40 line chat and change one letter, and do that say 20 times, will it store 20*40 = 800 lines as a result?     This is exactly the sort of very, very inefficient design decisions that we may want to revisit in the future.

Issue #731, opened 2016-07-22
Chat edit status incorrectly tells you when it's been updated


johnjeng:
Sometimes I'm editing my own chat and suddenly it'll say "You message has been updated". But no one has changed it! I thought it was due to the encoding changing but that got fixed. Something is being checked incorrectly.

johnjeng:
@simonluu has a fix for this incoming.

PR #741, opened 2016-07-24
Chat properly tells you when it's updated

simonluu:
chat edits properly tell you when the message is updated. (No longer says the message is updated when the message is not updated)

issue #731

### DIFF
--- a/src/smc-webapp/editor_chat.cjsx
+++ b/src/smc-webapp/editor_chat.cjsx
@@ -198,6 +198,8 @@ Message = rclass
         new_changes     : false
 
     componentWillReceiveProps: (newProps) ->
+        if @state.history_size != @props.message.get('history').size
+            @setState(history_size:@props.message.get('history').size)
         changes = false
         if @state.edited_message == @newest_content()
             @setState(edited_message : newProps.message.get('history')?.peek().get('content') ? '')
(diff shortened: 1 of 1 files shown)


williamstein:
I'll merge this.  I have to comment that this design kind of goes against a basic core principle of SMC....    SMC is realtime, collaborative, and everybody is together, and the technology I've developed makes it possible to support making everything fully collaborative.   Probably it's not at all clear to you how to do so, which is fine.... for now.

simonluu:
What should I have done? I talked with John and showed him the fix and he told me that I should push out the editing status fix.

williamstein:
As I said above, it's fine **for now**.  However, having mysterious other changes happening and deleting your own changes to see the other's, etc. tends to go against how SMC works.  Perhaps John has a fundamental disagreement with me about this, so I'll have to fix it myself later.    I think that all viewers should *see live* when you edit a past chat, with a cursor for you, etc.  Then if two people edit a past chat at once, there is no issue with conflicts.

simonluu:
Oh, I think I understand. With the chat editing right now, when two people are editing, they have no idea what each other are typing. The only way they find out what the other person sent is to close their own changes. This goes against the basic core principles of SMC. So what chat editing should be is that when more than one person are editing a chat it should change in real time. So everyone editing the chat sees the changes. Kind of like in the sagews.

johnjeng:
I implemented it and realized it was way worse without multiple cursors. I pushed out edits before making real time multiple edits because I think by far the most common use case is making a quick change to something you just wrote.

Not sure how to to multiple cursors.

williamstein:
By multiple cursors maybe you mean "other people's cursors".  If so, we will need to switch to using CodeMirror for input.

Issue #734, opened 2016-07-23
in sagews worksheets support tab completion with jokers 


williamstein:
For example `*prime<TAB>`.

williamstein:
I think Jupyter does this.  The Sage terminal in recent copies of Sage definitely does support this, so adding it to sagews is probably very easy.    Type `sage-develop` in a terminal, then `*prim<tab>` and see it work.

DrXyzzy:
1. I don't think the `*` in `*prime<TAB>` is treated as a glob in sage-develop; I think completion is the same as if there was a space, i.e. `* prime<TAB>`. For example, `*_factors<TAB>' does not offer any completions. Absence of completions in sagews here is probably just another limitation of the simplified parser.
1. On the other hand, `*_factors?` (no TAB) offers matches in sage-develop, while neither `*_factors?<Shift+Enter>` nor `*_factors?<TAB>` yield any result in sagews. Similarly with `*le_prim*?`.

So I think there are two things to look at here. The second case looks more interesting and potentially more valuable.

GregoryBard:
+1 This would really help people like me, who have trouble remembering the
exact names of commands.
---Greg

DrXyzzy:
Implemented in PR753, for tab completion, not help (question mark) mode.

The following now allow tab completion:
- `*_factors[tab]`
- `*le_pr*[tab]`
- `list.re*e[tab]`

PR #753, opened 2016-07-28
[ready for review] allow tab completion on glob expressions upstream #734

DrXyzzy:
Reference #734.

The following now allow tab completion:
- `*_factors[tab]`
- `*le_pr*[tab]`
- `list.re*e[tab]`

There is a test file at https://cloud.sagemath.com/projects/ccd4d4a4-29a8-4c39-85c2-a630cb1e9b6c/files/tabs734.sagews, but misc_page.coffee as well as sage_parsing.py will need to be updated for examples to work.

This follows the precedent in command-line sage in which a get-completions expression is treated as a wildcard expression if it contains one or more embedded asterisks. See [IPython/core/inputtransformer.py](https://github.com/ipython/ipython/blob/8b766e65c15d3f07e05ee6228d60256a5451df0d/IPython/core/inputtransformer.py#L211)

### DIFF
--- a/src/smc-webapp/misc_page.coffee
+++ b/src/smc-webapp/misc_page.coffee
@@ -757,13 +757,21 @@ exports.define_codemirror_extensions = () ->
             that.replaceRange(str, from, to)
 
         if completions.length == 1
-            insert(target + completions[0])
+            # do not include target in appended completion if it has a '*'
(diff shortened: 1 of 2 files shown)


williamstein:
This looks pretty safe in that the code is the same if there is no * in the expression.

PR #767, opened 2016-08-01
[needs work] The cookie set should be a wildcard cookie so if log in via sage.cocalc would be authed if go to cocalc.com or python.cocalc etc

timothyclemansinsea:
According to http://serverfault.com/questions/153409/can-subdomain-example-com-set-a-cookie-that-can-be-read-by-example-com looks like one puts a . in front of domain

### DIFF
--- a/src/smc-hub/auth.coffee
+++ b/src/smc-hub/auth.coffee
@@ -221,7 +221,13 @@ passport_login = (opts) ->
             dbg("set remember_me cookies in client")
             expires = new Date(new Date().getTime() + ttl*1000)
             cookies = new Cookies(opts.req, opts.res)
-            cookies.set(BASE_URL + 'remember_me', remember_me_value, {expires:expires})
+            root_domain = window.location.hostname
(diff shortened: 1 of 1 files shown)


src/smc-hub/auth.coffee
```diff
@@ -221,7 +221,13 @@ passport_login = (opts) ->
             dbg("set remember_me cookies in client")
             expires = new Date(new Date().getTime() + ttl*1000)
             cookies = new Cookies(opts.req, opts.res)
-            cookies.set(BASE_URL + 'remember_me', remember_me_value, {expires:expires})
+            root_domain = window.location.hostname
```
haraldschilly: is there a `window` in node? or is this code used in the client? That confuses me …

williamstein: @haraldschilly -- good point, there is no window in node, so this code doesn't make any sense.

williamstein:
Closing due to nothing happening.

Issue #70, opened 2015-08-21
sagews syntax error after indentions


haraldschilly:
this gives a syntax error:
```
for i in range(1):
    pass
'x'
```
or this
```
if True:
    pass
"x"
```
like
```
Error in lines 1-3
Traceback (most recent call last):
  File "/projects/20e4a191-73ea-4921-80e9-0a5d792fc511/.sagemathcloud/sage_server.py", line 879, in execute
    exec compile(block+'\n', '', 'single') in namespace, locals
  File "<string>", line 3
    'x'
      ^
SyntaxError: invalid syntax
```

while this works:

```
y = 2
"x"
```

DrXyzzy:
Problem is caused by the last line starting with a string literal. Workaround is to put the literal in parens.

Explanation: function [sage_parsing.divide_into_blocks](https://github.com/sagemathinc/smc/blob/master/src/smc_sagews/smc_sagews/sage_parsing.py#L219) merges a line that starts with a literal to the previous block. This causes `exec compile(block+'\n', '', 'single')` to fail in [sage_server.Salvus.execute()](https://github.com/sagemathinc/smc/blob/master/src/smc_sagews/smc_sagews/sage_server.py#L947) because block is not a single statement.

Problem goes away if check `code[i][:2] == '%('` is eliminated in the while statement near [sage_parsing.py:219](https://github.com/sagemathinc/smc/blob/master/src/smc_sagews/smc_sagews/sage_parsing.py#L219). Is there some other case that will fail if this test is removed?

williamstein:
Now live.

PR #786, opened 2016-08-05
[ready for review] Issue70 indent

DrXyzzy:
Fix for: #70 

I tested several python segments with lines beginning with a literal and was not able to cause an error with this code change, but could be missing a case where the comparison `code[i][:2] == '%('` at line 219 in the original is needed.


haraldschilly:
hmm, my little tests look good, but I have no idea how to trigger all those cases in `strip_string_literals`. I did read the code and it is probably fine, but that's kind of hard to understand.

williamstein:
I wish I could remember where `%(` would ever occur.  No idea.

Issue #61, opened 2015-08-13
put a dropzone below/above/near (?) the main file listing


williamstein:
requested by Mesut Karakoç on the sage-cloud list.

haraldschilly:
My idea would be to make the whole area where the files are a dropbox area. I.e. no matter if the thing beneath the cursor is a directory or not ... a dragged file will be uploaded in the current directory. As a visual aid, the entire area is getting grayed out while hovering over it.

that would be just like in google drive and similar services.

timothyclemansinsea:
the whole page can be a dropzone https://github.com/enyo/dropzone/wiki/Make-the-whole-body-a-dropzone

williamstein:
Remember that we are using React or rewriting to user react... so be
careful.  We can't just use things not properly supporting react.

On Sun, Mar 6, 2016 at 10:54 AM, Tim Clemans <notifications@github.com>
wrote:

> the whole page can be a dropzone
> https://github.com/enyo/dropzone/wiki/Make-the-whole-body-a-dropzone
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/sagemathinc/smc/issues/61#issuecomment-192962682>.
>



-- 
William (http://wstein.org)

timothyclemansinsea:
There is a react dropzone and a question about doing full page
https://github.com/okonet/react-dropzone/issues/119

williamstein:
> There is a react dropzone 

Yes, I think that's what we currently use for the +New page, which is written using react...  I do remember it being quite difficult though.

Note that we don't want full page drop -- we won't to make the file listing into a drop zone.   Since there can be multiple projects or pages (like account settings) for which drop doesn't make sense, it shouldn't be full page.  So it's probably pretty easy to do this for the file listing :-)

haraldschilly:
just mentioning, that [drop-onto-files](https://github.com/sagemathinc/smc/tree/drop-onto-files) is my first attempt in doing that

PR #788, opened 2016-08-05
webapp: preventing accidental file drops outside the drop area

haraldschilly:
This is related to #61 and will still be there when the file area is a dropzone -- with a different info text, of course.


williamstein:
Seems like a good idea to me.

PR #790, opened 2016-08-08
Drop onto files

haraldschilly:
- prevent accidental file drops outside a dedicated drop area
- make the file area droppable for uploading files
- explicit upload button
- make room for templates in "new" panel


williamstein:
Issues:

 - The use of an orange *warning* button doesn't follow the semantic meaning of *warning* and is inconsistent with numerous "partially modal button menus" elsewhere in SMC.  Instead, do what's done in the course editor and other places (e.g., account settings) when clicking a menu brings up an "inline modal".  Just try clicking buttons with ... after them in account settings to see. 

 - Closing the upload dialog is inconsistent with the rest of SMC.  It's an orange warning button, but there's no possible danger in closing it.  Also, everywhere else we use either "Cancel" and/or an X.

 - When the upload page is shown, we could mention in small light text (maybe below the box) that one can also simply drag and drop on the file listing to show this dialog.   Otherwise, people won't know that this now works, and longterm users have already been burned by it not working, so won't try.

 - Where it says "use SSH as explained in project settings" it would be nice if clicking "project settings" switches to that tab.

williamstein:
More comments:

 - Clicking on the house (or anywhere in the file path) at the top should probably close the upload mode. 

- It's annoying that I can't see the file listing when the upload page is there, since it would be nice to know what files I have.  Why not show it anyways below the upload panel?

- Create/upload are pretty similar; I wouldn't be opposed to combining them into a button group... not sure though.

- I see absolutely no value whatever in removing the upload box from the +New page and replacing it by "Looking for file upload? Goto "Files" and click on "Upload".   Why do that?     It'll cause trouble for the muscle memory of many of our users, cause friction for tutorials, and does it really make the +New page better?

haraldschilly:
update

* I've fixed the problem, that accidentally dropping anywhere causes the page to reload and show the file. that was no longer working after the top-down rewrite. for that, I've also a little bit cleaned up those two top level pages.

* I've addressed some of the comments from above, but e.g. 

> Clicking on the house (or anywhere in the file path) at the top should probably close the upload mode

ok, that's easy to do, but it makes no sense to me to close it when someone clicks there.

> Why remove the upload box from the +New page ...

my goal is to make room for the template dialog

I'm not sure what to do with this now, though. Are you motivated to review how it is now?

williamstein:
Umh, this is just plain AWESOME!

Issue #399, opened 2016-01-28
preparser(False) is ignored in Sage worksheets right now.


williamstein:

DrXyzzy:
There is a function `preparse()` imported in sagews from sage/repl/preparse.py, but it takes a string of code as first positional parameter. Changing title of this issue to `preparser(False) is ignored...`.

For more info see:
- preparse(False) (still) not working  
http://ask.sagemath.org/question/24696/preparsefalse-still-not-working/
- Last paragraph in selected section of sage docs here:  
http://doc.sagemath.org/html/en/faq/faq-usage.html#i-have-type-issues-using-scipy-cvxopt-or-numpy-from-sage

DrXyzzy:
Sage function `preparser()` sets Sage global `_do_preparse`. This variable is ignored in Sage worksheet processing including preparsing, hence `preparser(False)` is ignored.

DrXyzzy:
Sage FAQ typo `preparse` should also be corrected. I'm opening a trac ticket for that.

DrXyzzy:
Done.

PR #806, opened 2016-08-13
[ready for review] allow sage.repl.interpreter._do_preparse to override preparse argument

DrXyzzy:
Proposed fix for #399 is to override `Salvus.execute()` argument `preparse` if `preparser(False)` has been called.

Test cases are in https://cloud.sagemath.com/projects/ccd4d4a4-29a8-4c39-85c2-a630cb1e9b6c/files/SAGEWS/issue399test.sagews. That project is running the proposed modification.

### DIFF
--- a/src/smc_sagews/smc_sagews/sage_server.py
+++ b/src/smc_sagews/smc_sagews/sage_server.py
@@ -929,10 +929,18 @@ class Salvus(object):
         #code   = sage_parsing.strip_leading_prompts(code)  # broken -- wrong on "def foo(x):\n   print(x)"
         blocks = sage_parsing.divide_into_blocks(code)
 
-        import sage.repl.interpreter
+        try:
+            import sage.repl.interpreter as sage_repl_interpreter
+        except:
+            log("Error - unable to import sage.repl.interpreter")
 
         for start, stop, block in blocks:
-            if preparse and sage.repl.interpreter._do_preparse:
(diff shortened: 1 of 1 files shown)


haraldschilly:
So, I think that's a good fix. The only thing that comes to my mind is, what if the Sage API changes this in a year? Maybe wrap that import in a `try catch` like:

```
try:
    import sage.repl.interpreter as sage_repl_interpreter
except:
    # warning
```
and then
```
getattr(sage_repl_interpreter, '_do_preparse', True)
```
in the if.

DrXyzzy:
Yes, of course, will add try/catch & update PR.

src/smc_sagews/smc_sagews/sage_server.py
```diff
@@ -929,8 +929,18 @@ def execute(self, code, namespace=None, preparse=True, locals=None):
...
+        try:
+            import sage.repl.interpreter as sage_repl_interpreter
+        except:
+            log("Error - unable to import sage.repl.interpreter")
+
         for start, stop, block in blocks:
-            if preparse:
+            # if import sage.repl.interpreter fails, sag_repl_interpreter is unreferenced
+            try:
+                do_pp = getattr(sage_repl_interpreter, '_do_preparse', True)
```
haraldschilly: oh! you don't need to do that try/catch here. `getattr` has **3** arguments here.

DrXyzzy:
try/catch still needed because getattr generates NameError exception if first argument is not defined.

PR #812, opened 2016-08-15
log invites

timothyclemansinsea:


haraldschilly:
ok, this looks good to me. my only slightly related idea is, to also add an `invite_student` to these log activities, i.e. when they are added via the course editor. It might be easy to also record that, and would help us to see who is inviting when, how many, etc. That way, we would also have covered all types of invites, not just these.

src/smc-webapp/project_log.cjsx
```diff
@@ -199,6 +199,12 @@ LogEntry = rclass
         v = if v.length > 0 then r_join(v) else 'nothing'
         <span>set <a onClick={@click_set} style={if @props.cursor then selected_item} href=''>upgrade contributions</a> to: {v}</span>
 
+    render_invite_user : ->
+        <span>invited user <User user_map={@props.user_map} account_id={@props.event.invitee_account_id} /></span>
+        
+    render_invite_nonuser : ->
+        <span>invited nonuser {@props.event.invitee_email}</span>
```
haraldschilly: "nonuser" doesn't sound very english to me. maybe just remove "nonuser" from that string?

Issue #819, opened 2016-08-16
sagews: Comment at indentation 0 taken as start of new block


cben:
In SMC sage notebook:
```
def never_called(a):
    print 'should not execute 1', a
     # comment
    # comment
    print 'should not execute 2', a
   # comment - why red color BTW?
    print 'should not execute 3', a
  # comment - why red color BTW?
    print 'should not execute 4', a
 # comment - why red color BTW?
    print 'should not execute 5', a
# comment at indent 0
    print 'should not execute 6 BUT DOES', a
```
The last line should be part of the function by Python rules, but instead gets executed immediately!
And *if* that causes an exception, the lines numbers are wrong (from start of cell, not notebook).

![screenshot-cloud sagemath com 2016-08-16 11-38-09 comment 0](https://cloud.githubusercontent.com/assets/273688/17693174/30870c4a-63a6-11e6-9f74-dac00b042b79.png)

public share: https://cloud.sagemath.com/projects/328c4733-2f1a-4e40-9d7c-ba21ea605f6f/files/python_0_indent.sagews

cben:
cc @anpc who bumped into this.

haraldschilly:
yes, that seems like a bug in the block parser for the code. I'm assigning this to @DrXyzzy

DrXyzzy:
Thanks for the detailed report. Looking at it now. I suspect the sample program is creating problems with both the Sage worksheet block parser and Codemirror syntax highlighting.

williamstein:
Fixed by PR https://github.com/sagemathinc/smc/pull/823, which is now merged.

I'm closing this.  The syntax highlighting part of this is an issue with the CodeMirror Python mode, as you can confirm here: https://codemirror.net/mode/python/index.html  You might submit a bug report to CodeMirror.  (If so, let us know.)

DrXyzzy:
Reported it here: codemirror/CodeMirror#4276.

PR #823, opened 2016-08-16
[ready for review] issue819 block parser skip comments

DrXyzzy:
Ref #819 

Block parser was substituting comments (as well as strings) with %(Lxx) labels and removing them later. Code that substituted the comments is taken out.

Internal testing at https://cloud.sagemath.com/projects/ccd4d4a4-29a8-4c39-85c2-a630cb1e9b6c/files/SAGEWS/issue819test.sagews - requires code change to be in place.

### DIFF
--- a/src/smc_sagews/smc_sagews/sage_parsing.py
+++ b/src/smc_sagews/smc_sagews/sage_parsing.py
@@ -73,11 +73,12 @@ def strip_string_literals(code, state=None):
             # it's a comment
             newline = code.find('\n', hash_q)
             if newline == -1: newline = len(code)
-            counter += 1
-            label = "L%s" % counter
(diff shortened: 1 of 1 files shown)


haraldschilly:
I tried to understand it, but I'm not much able to follow. I suppose that this fixes this, since it detects a comment and does nothing when it is so. Looks good to me, though.

williamstein:
> Block parser was substituting comments (as well as strings) with %(Lxx) labels and removing them later. Code that substituted the comments is taken out.

I now realize what this was doing.  This completely broke markdown mode, and any other mode where comments aren't hashes...   Sorry that I positively reviewed this.   My fix for https://github.com/sagemathinc/smc/issues/978, which I found via a lot of careful debugging, turned out to be exactly reverting this.  lol.

PR #821, opened 2016-08-17
On hover over avatar for who's viewing file show location of first cursor for that user

timothyclemansinsea:


src/smc-webapp/profile.cjsx
```diff
@@ -147,12 +157,11 @@ UsersViewingDocument = rclass
...
         output = []
         all_users = []
-
+        
         for user_id, events of log
-
             if @props.account_id is user_id
                 continue
-
+            line = @props.get_users_cursors(user_id)?[0]['y'] + 1
```
haraldschilly: Hi, I just read the code and didn't do testing, and the only thing that I saw is, that this line assumes that there is at least one cursor. I've no concrete idea if that's always true, but maybe that array is empty for some reason? So, maybe add a `?` after the `[0]` ?

williamstein:
Live!

PR #830, opened 2016-08-18
about colors in surface plots

fchapoton:
here is the first step: distinguish 2 cases

then one need to know what format should be send to the next step in the toolchain

could you remind me where to find that ?

### DIFF
--- a/src/smc-webapp/3d.coffee
+++ b/src/smc-webapp/3d.coffee
@@ -506,8 +506,8 @@ class SalvusThreeJS
                         when 5
                             push_face4_with_color(v...)
                         else
-                            console.log("WARNING: rendering colored face with #{v.length} vertices not implemented")
-                            push_face4_with_color(v...)
+                            console.log("WARNING: rendering colored face with #{v.length - 1} vertices not implemented")
+                            push_face4_with_color(v[0], v[1], v[2], v[3], v[-1])   # might as well render most of the face...
             else
                 for v in faces
                     switch v.length
--- a/src/smc-webapp/3d.coffee
+++ b/src/smc-webapp/3d.coffee
@@ -428,11 +428,18 @@ class SalvusThreeJS
     add_obj: (myobj)=>
         @show_canvas()
 
-        if myobj.type == "index_face_set"
-            has_local_colors = myobj.has_local_colors
+        if myobj.type == 'index_face_set'
+            if not myobj.has_local_colors?
+                has_local_colors = false
+            elif myobj.has_local_colors == 0
+                has_local_colors = false
+            else
+                has_local_colors = true
+                # then we will assume that every face is a triangle
         else
             has_local_colors = false
 
+            
         vertices = myobj.vertex_geometry
         for objects in [0...myobj.face_geometry.length]
             #console.log("object=", misc.to_json(myobj))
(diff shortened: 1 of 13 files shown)


williamstein:
Thanks for working on this!

fchapoton:
Well, this is not yet done, and may/will require more work. In particular I may need to change also
smc/src/smc-webapp/3d.coffee

And doing this without being able to test will certainly be difficult..

fchapoton:
I think I got a good way to do that, but I cannot test it.

fchapoton:
now this branch needs testing, please.

williamstein:
Thanks!  I will test it tomorrow (too tired now).

src/smc-webapp/3d.coffee
```diff
@@ -428,6 +428,11 @@ class SalvusThreeJS
     add_obj: (myobj)=>
         @show_canvas()
 
+        if myobj.type == "index_face_set"
+            has_local_colors = myobj.has_local_colors
+        else
+            hase_local_colors = false
```
williamstein: typo: hase_ --> has

fchapoton:
typo corrected.

williamstein:
@fchapoton -- unfortunately this is still very broken.  Would you consider learning enough to actually test it?   I can pre-set up an SMC project for you for development...

fchapoton:
damn.. I would really like this to work. Well, I have a SMC account already. Forgot under which email.

williamstein:
I will set everything up for you here: https://cloud.sagemath.com/projects/8208a5a2-ac42-4b75-bf5c-ba7cf1eb89a4/files/, then explain exactly what you need to do to change things and test.

fchapoton:
ok, thanks

williamstein:
That project is now ready -- check out https://cloud.sagemath.com/projects/8208a5a2-ac42-4b75-bf5c-ba7cf1eb89a4/files/chat.sage-chat

fchapoton:
I have some difficulties in acceding to the project right now. Trying to open the chat or to see 3d.coffee just keep loading / do nothing.

haraldschilly:
so, the chat was broken up until a few minutes ago, and for me the project
loads and well, I don't see any problems with 3d.coffee... Just reload the
webpage to get the update for the chat. That should do it!

On Sat, Aug 20, 2016 at 9:38 PM, Frédéric Chapoton <notifications@github.com

fchapoton:
Still the same problem for me: chat loading forever and nothing in 3D.coffee and other files.
Anyway my main problem is the fact that I cannot "pip install" as  William told me to do in the chat,
in order to be able to test my code..

fchapoton:
so, what is happening about that now ?

fchapoton:
this should be ready, once tested..

williamstein:
Wow, it is incredibly hard to get this to work...  E.g., with the version of the code right now, most simple possible example fails: 
```
sphere()
```
results in 
```
...
  File "/projects/45f4aab5-7698-4ac8-9f63-9fd307401ad7/smc/src/data/projects/5b08ef88-9695-4377-91ab-1d09406afc6d/.local/lib/python2.7/site-packages/smc_sagews/graphics.py", line 327, in convert_index_face_set
    face_geometry = [{"material_name": p.texture.id,
AttributeError: 'NoneType' object has no attribute 'id'
```

williamstein:
I also tried running the current live code on the backend and the frontend renderer gives this error `WARNING: rendering colored face with 2 vertices not implemented`

fchapoton:
A suggestion for the p.texture.id problem, but I am not really convinced: could you try to add a line "p.triangulate()" before the line "p.obj()" ?

Faces with 2 vertices ?  Damn..

williamstein:
OK, I tried a combination of things in the code and it seems to work now.

fchapoton:
good. I am curious to see the changes.

williamstein:
This is what I did -- now in master and will be live soon: https://github.com/sagemathinc/smc/commit/d9633ffdd80c219b04fab0d6865baaa17d8d26e1

Issue #825, opened 2016-08-17
sage_eval allows non-string inputs -- support this


williamstein:
`smc_sagews/sage_salvus.py` defines a wrapped `sage_eval` that assumes input is a string.  But tuples are now also supported by sage_eval.  So maybe this will be a better sage_eval:

```
def sage_eval(x, locals=None):
    if isinstance(x, str):
        x = str(x).strip()
        if x.isspace():
            return None
    from sage.all import sage_eval
    return sage_eval(x, locals=locals)
```

williamstein:
Example input:

sage_eval(('f (x) = (x ^ 2-1) / (x-1)', 'f (1.000000000000001)'))

DrXyzzy:
Thanks for detail. Issue related to this [sage-cloud thread](https://groups.google.com/d/msg/sage-cloud/t1AZGdocj10/GJkqVg2vBwAJ)

DrXyzzy:
Also this example from [sage docs](http://doc.sagemath.org/html/en/reference/misc/sage/misc/sage_eval.html) causes an error in sagews:
```
sage_eval('2^3', preparse=False)

==>

TypeError: sage_eval() got an unexpected keyword argument 'preparse'
```

haraldschilly:
I've not time to look into it right now, but your last comment somehow tells me, that we should start writing tests. It doesn't need to be sophisticated, but something simple based on e.g. py.test should be sufficient.

DrXyzzy:
Done.

PR #831, opened 2016-08-18
[ready for review] smc issue #825 sage_eval in sagews

DrXyzzy:
Ref: #825 

Test file https://cloud.sagemath.com/projects/ccd4d4a4-29a8-4c39-85c2-a630cb1e9b6c/files/SAGEWS/issue825_test.sagews
Project is on smc branch for test code at this time.

### DIFF
--- a/src/smc_sagews/smc_sagews/sage_salvus.py
+++ b/src/smc_sagews/smc_sagews/sage_salvus.py
@@ -616,12 +616,13 @@ def interact_control(arg, value):
     c._opts['var'] = arg
     return c
 
-def sage_eval(x, locals=None):
-    x = str(x).strip()
-    if x.isspace():
-        return None
+def sage_eval(x, locals=None, **kwds):
+    if isinstance(x, str):
+        x = str(x).strip()
+        if x.isspace():
+            return None
     from sage.all import sage_eval
-    return sage_eval(x, locals=locals)
+    return sage_eval(x, locals=locals, **kwds)
 
 class ParseValue:
     def __init__(self, type):


src/smc_sagews/smc_sagews/sage_salvus.py
```diff
@@ -616,12 +616,13 @@ def interact_control(arg, value):
...
     return c
 
-def sage_eval(x, locals=None):
-    x = str(x).strip()
-    if x.isspace():
-        return None
+def sage_eval(x, locals=None, **kwds):
+    if isinstance(x, str):
+        x = str(x).strip()
+        if x.isspace():
```
haraldschilly: I'm sorry, I don't get this test. If it is a space, isn't it then the empty string?

`" ".strip().isspace()` → `False`

haraldschilly:
added one question, maybe I don't get it, otherwise ok. the main idea is to pass on the kwds and that's done

DrXyzzy:
@haraldschilly,
Thank you for reviewing.
- The bug initially reported in #825 appears to be fixed by @williamstein suggestion to insert `if isinstance(x, str):`.
- The bug found during testing about ignoring preparse named parameter appears to be fixed, as you noted, by passing through kwds.
- The isspace code was there before. I didn't change it. I wonder if it relates to this: [str.strip() strips ASCII whitespace while isspace() is locale-dependent](https://bugs.python.org/issue25433)
- agree with suggestion about need for testing, have started on that.

haraldschilly:
oh ok, I see. ... and that locale dependency is good to know, could also happen in other places.

PR #148, opened 2016-08-29
Remember all tabs when refreshing

timothyclemansinsea:
Whenever there's a new version of SMC, I am asked to do a 'hard' refresh. Once SMC restarts, it is kind enough to remember the project and file I was working on, but totally forgets all other files and projects. I would like to have it remember everything that is open. Thanks.

### DIFF
--- a/src/smc-webapp/editor.coffee
+++ b/src/smc-webapp/editor.coffee
@@ -1075,13 +1075,13 @@ class exports.Editor
             if @open_file_tabs().length < 1
                 @resize_open_file_tabs()
 
-            if (!(window.localStorage.getItem("opened_files")))
-                window.localStorage.setItem("opened_files", "[]")
-            opened_files = JSON.parse(window.localStorage.getItem("opened_files"))
+            if (!(window.localStorage["opened_files"]))
+                window.localStorage["opened_files"] =  "[]"
+            opened_files = JSON.parse(window.localStorage["opened_files"])
             for value, index in opened_files
                 if value['project_id'] == @project_id and value['path'] == filename
                     opened_files.splice(index, 1)
-            window.localStorage.setItem("opened_files", JSON.stringify(opened_files))
+            window.localStorage["opened_files"] = JSON.stringify(opened_files)
 
             return false
 
--- a/src/smc-webapp/editor.coffee
+++ b/src/smc-webapp/editor.coffee
@@ -1078,9 +1078,7 @@ class exports.Editor
             if (!(window.localStorage["opened_files"]))
                 window.localStorage["opened_files"] =  "[]"
             opened_files = JSON.parse(window.localStorage["opened_files"])
-            for value, index in opened_files
-                if value['project_id'] == @project_id and value['path'] == filename
-                    opened_files.splice(index, 1)
+            delete opened_files[@project_id+'-'+filename]
             window.localStorage["opened_files"] = JSON.stringify(opened_files)
 
             return false
--- a/src/smc-webapp/editor.coffee
+++ b/src/smc-webapp/editor.coffee
@@ -1075,6 +1075,14 @@ class exports.Editor
             if @open_file_tabs().length < 1
                 @resize_open_file_tabs()
 
+            if (!(window.localStorage.getItem("opened_files")))
+                window.localStorage.setItem("opened_files", "[]")
+            opened_files = JSON.parse(window.localStorage.getItem("opened_files"))
+            for value, index in opened_files
+                if value['project_id'] == @project_id and value['path'] == filename
+                    opened_files.splice(index, 1)
+            window.localStorage.setItem("opened_files", JSON.stringify(opened_files))
+
             return false
 
         link.find(".salvus-editor-close-button-x").click(close_tab)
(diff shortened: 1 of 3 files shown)


src/smc-webapp/project_store.coffee
```diff
@@ -194,6 +194,13 @@ class ProjectActions extends Actions
                                 @_project().show_editor_chat_window(opts.path)
                             if opts.foreground
                                 @set_current_path(misc.path_split(opts.path).head, update_file_listing=false)
+        # log file opening to HTML5 storage so we can auto reopen tabs at some point
+        if (!(window.localStorage.getItem("opened_files")))
```
williamstein: why not `window.localStorage['opened_files']`?  Why use getItem and setItem?

src/smc-webapp/project_store.coffee
```diff
@@ -194,6 +194,13 @@ class ProjectActions extends Actions
...
+            window.localStorage.setItem("opened_files", "[]")
+        opened_files = JSON.parse(window.localStorage.getItem("opened_files"))
+        console.log('opened_files:', window.localStorage.getItem("opened_files"))
```
williamstein: I'm not merging code into master with console.log's laying around.

src/smc-webapp/project_store.coffee
```diff
@@ -194,6 +194,13 @@ class ProjectActions extends Actions
...
+        if (!(window.localStorage.getItem("opened_files").indexOf(JSON.stringify({'project_id': @project_id, 'path': opts.path})) > -1))
+            opened_files.push({'project_id': @project_id, 'path': opts.path})
```
williamstein: But then you don't save opened_files...?  So you just pushed a bunch to it, which is thrown away?  I'm confused.

src/smc-webapp/editor.coffee
```diff
@@ -1075,6 +1075,14 @@ class exports.Editor
             if @open_file_tabs().length < 1
                 @resize_open_file_tabs()
```
williamstein: What's the block of code below supposed to do?   I couldn't understand it and there are no comments.

src/smc-webapp/projects.cjsx
```diff
@@ -46,6 +46,17 @@ MAX_DEFAULT_PROJECTS = 50
 
 _create_project_tokens = {}
 
+window.restore_tabs = () ->
```
williamstein: It would be better to make these two methods on `window.smc`, since `window.[blah]` could conflict easily.

src/smc-webapp/editor.coffee
timothyclemansinsea: this not appear to be code I added so I have no clue.

haraldschilly:
so, I tried to understand it, and a couple of thoughts:

* is it in the world of web-browsers already safe to assume that everyone has localStorage? e.g. [here](https://github.com/sagemathinc/smc/blob/master/src/smc-webapp/editor.coffee#L497) we still check if storage exists, and only do this if it is there. I would suggest to safeguard this code in a similar way or wrap get/set of local storage.

* there is a console.log message, it's not critical but once it works we don't need it

* maybe it's a good idea to open all those files in the background? I.e. set `foreground:false`, see [here](https://github.com/sagemathinc/smc/blob/68c23fac85a35d3f167fd277a67c4a197dec5e18/src/smc-webapp/project_store.coffee#L165)

* I'm also not so happy that you do this search for existing entries using a string indexOf operation -- better idea,  change the stored data-structure instead? You're always parsing it first into `opened_files` so you have the actual data right there! You can e.g. save one dictionary of `#{@project_id}-#{path} : true` and hence automatically avoid adding it twice without any need to check for existence. (of course, in the restore step you have to split this string, but that's easy since project_id is fixed length.)

williamstein:
See https://github.com/sagemathinc/smc/issues/470

src/smc-webapp/editor.coffee
williamstein: Um, I think you wrote this code....

williamstein:
This is pretty close!  The editor.coffee thing will have to get redone probably due to @johnjeng's react rewrite (will probably be easier/better...).  

Bug:  if you close a project, then this code doesn't remove all the open files from the `window.localStorage["opened_files"]` dictionary.   In contrast, if you close a file (not project), then it is properly remove.

Perhaps after the top-down react rewrite all open/close of files will go through a common path and the above listed bug will be fixed automatically.  I don't know. 

In short -- I think having this functionality is **really good** -- I can already feel the power playing around with this.  Even having a "restore all open files from last session" button somewhere right when refreshing will be very nice.

src/smc-webapp/editor.coffee
timothyclemansinsea: @williamstein I did not write that code so I have no clue... https://github.com/timothyclemansinsea/smc/blame/454177e2b0d8e55d218b44b6281950d4995dd2e0/src/smc-webapp/editor.coffee#L1075

timothyclemansinsea:
@williamstein am I to wait until @johnjeng 's react rewrite is merged?

williamstein:
You should probably wait until after the react rewrite on this, since some of the relevant code is gone now (e.g., most of editor.coffee)... and how to approach this may be different after that rewrite.

timothyclemansinsea:
@williamstein ah sweet you and I asked question/answer at same time

williamstein:
Re-open if you're going to work on this.

PR #848, opened 2016-08-29
realtime avatars of anyone viewing files in a projects

timothyclemansinsea:


williamstein:
This is really nice to use!

williamstein:
Wow, this is bad.  It breaks opening of any files at all.  Wow did I merge this...?  :-(

williamstein:
No -- it breaks opening chatrooms, not anything at all.   OK now fixed -- issues was the rename of UsersViewingDocument.

PR #849, opened 2016-08-29
Show login form if user is anonymous and directory is private

timothyclemansinsea:


src/smc-hub/hub.coffee
```diff
@@ -1406,7 +1406,7 @@ class Client extends EventEmitter
                         if err
                             cb(err)
                         else if not is_public
-                            cb("project with id '#{mesg.project_id}' is not public")
+                            cb("not_public")
```
williamstein: Since this exact error message is being matched elsewhere, it is CRITICAL to put a comment here saying so and also where it is used.  Otherwise, somebody else could just change the error willy nilly (like you just changed it from what it was).

src/smc-webapp/project_files.cjsx
```diff
@@ -1861,11 +1862,21 @@ ProjectFiles = (name) -> rclass
                 style   = {error_style}
                 onClose = {=>@props.actions.setState(error:'')} />
 
+    render_not_public_error : ->
+        if @props.redux.getStore('account').is_logged_in()
+            <ErrorDisplay title="Directory is not public" error={"You are trying to access a non public project that you are not a collaborator on. You need to ask a collaborator of the project to add you."} />
```
williamstein: Presumably we should make a new issue which is a mechanism for requesting to be added to a project, which would email the current collabs or something else (depending on their prefs). Google groups have something like this.

williamstein:
I can't tell if this works.  It doesn't seem to anymore.   There may be other changes that have broken this?   There is a merge conflict, which is easy to fix. 

I found this new bug while trying to test this: https://github.com/sagemathinc/smc/issues/941

timothyclemansinsea:
No freakin clue why this isn't working. I've looked at both the hub code, made sure hub restarted, and looked at the frontend code.

williamstein:
NOTE to self: when making this life it is critical to restart the hubs with their code updated.

williamstein:
Maybe I'm just being dumb, but I couldn't get this to work for me.

timothyclemansinsea:
@williamstein on my org smc inside smc instance SMC was immediately showing "No Files" When I did a clean install of SMC inside of SMC I got the spinning disc like you likely did. I addressed it by adding two lines in an auto retry function that stops if not_public. Auto retry not stopping for errors that need to be stopped on might be the cause of stuff like #1165

timothyclemansinsea:
Just realized there's some stuff that needs to be removed. Removing now

williamstein:
Nice!

PR #858, opened 2016-09-02
Make realtime avatars for page listing projects

timothyclemansinsea:
On the page listing projects this would show anyone who collaborates on any project the user is a collaborator on that has viewed any file in the last six minutes. Clicking on the icon would take the user to the project page.


williamstein:
This appears to work perfectly.

Issue #860, opened 2016-09-02
terminal autosizing too wide


haraldschilly:
for a while now, my terminal is always about 2 columns too wide (no matter if resized narrow or wide). anyone else noticing that? maybe font related? I don't have any funny zooming going on.

screenshot for illustration, you can see that "16" of 2016 is missing.

![smc-terminal-too-wide](https://cloud.githubusercontent.com/assets/207405/18198718/86b3908e-70fd-11e6-999d-882a19190590.png)

haraldschilly:
The referenced PR #861 fixes this above and also works for me for very very tiny to quite large fonts.

williamstein:
fixed, right?

haraldschilly:
Yes, no more issues for me and I haven't heard anything negative from
others.

PR #861, opened 2016-09-02
webapp/console: fixing col width and more room for line height

haraldschilly:
this addresses #860 and also adds a bit of room for lines to avoid overwriting "_" by the next line.

I also bet that this is browser/OS specific, so maybe it always worked for @williamstein but never for me? My hope is, that my fix here errs on the lower side, such that the terminal is in all cases narrower than before.

### DIFF
--- a/src/smc-webapp/_console.sass
+++ b/src/smc-webapp/_console.sass
@@ -40,8 +40,7 @@
 
 .salvus-console-terminal
   font-family: droid-sans-mono
-  /* really set in console.coffee, overriding this! */
-  padding: 1ex
+  padding: 1px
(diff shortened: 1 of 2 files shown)


williamstein:
I merged this into master since it looks harmless, but it actually completely breaks the terminal on OS X.   And, as you assumed, the previous version (without your patch) does work perfectly for me.

williamstein:
I fixed an issue with this seriously breaking things, due to the scrollbar here:  https://github.com/sagemathinc/smc/commit/0258ad862776eefc0af40b72b8c9cf6e3062692b

Now making live.

PR #866, opened 2016-09-03
fix mobile chat on android (and iphone chrome)

timothyclemansinsea:
You currently can't type messages into the main chatrooms in SMC on mobile chrome (on iOS / Android).  It works fie on mobile safari.


timothyclemansinsea:
I tried to do a more complicated PR where autofocus is set based on status. However onClick does not fire on mobile, see https://github.com/react-bootstrap/react-bootstrap/issues/2193

williamstein:
Thanks!!

williamstein:
This definitely does not fix anything for iPhone Chrome.  No idea if it fixes Android Chrome or not -- i can't test.  But it looks harmless.

williamstein:
Definitely doesn't fix: https://github.com/sagemathinc/smc/issues/867

Issue #470, opened 2016-03-09
clarify the error message when viewing a project that has no public files


williamstein:
Sometimes people report this error:

> I am a TA using SageMathCloud to teach Calculus. When my student tried to share his project folder with me, I saw this message:  Directory listing error  "project with id 'b9e5b094...4b154' is not public"

We get a **a lot** (daily) emails from confused users in such cases.   Let's provide a better message.  E.g., 

> Please ask the person who shared this link with you to visit their project settings at https://cloud.sagemath.com/projects/b9e5b09...b154/settings and add you via your email address (the one you are registered with on SMC) to the list of "collaborators". Then, this project will pop up in your list of projects and you can - of course - also access all files in this project.

This will save a lot of people a lot of time.

LATER -- and this requires more work and thought -- we can replace the above message by a button to _request_ to join a project.  Like a google group, the "ask to join" could have various settings, e.g.,
- a question the user has to answer, and joining must be approved
- any user can join
- only people who know the link can join
- completely private/secret -- don't even give away existence of the project with given id.

etc.

williamstein:
@timothyclemansinsea  -- did you fix this already?

timothyclemansinsea:
@williamstein Yes this is done. Where's the ticket for implementing request to join?

williamstein:
Closing.

> Where's the ticket for implementing request to join?

I don't know -- maybe it got closed by accident.  Please search through closed tickets.

PR #872, opened 2016-09-06
[needs work by wstein] implement request to join project

timothyclemansinsea:


timothyclemansinsea:
things I need help with:

- [ ] removal of request after approve/deny button being pressed
- [ ] determining if user has already submitted a request

src/smc-webapp/project.coffee
```diff
@@ -172,6 +174,16 @@ class ProjectPage
                         box.hide()
 
 
+    requests_to_join_project_alert: (project_id) =>
+        console.log('PID', @project_id, redux.getStore('projects').get_project(project_id).invite_requests.length)
```
williamstein: need to remove this or comment out

src/smc-webapp/project.coffee
```diff
@@ -172,6 +174,14 @@ class ProjectPage
                         box.hide()
 
 
+    requests_to_join_project_alert: (project_id) =>
+        if Object.keys(redux.getStore('projects').get_project(project_id).invite_requests).length > 0
```
williamstein: There is no guarantee that `redux.getStore('projects').get_project(project_id)` isn't undefined when this is called.  Maybe the projects hasn't completely loaded yet?  I'm hitting this.

williamstein: I'll fix this.

src/smc-webapp/project_files.cjsx
```diff
@@ -1825,11 +1829,70 @@ ProjectFiles = (name) -> rclass
...
+        </div>
+
+    render_request_to_join_project_button : ->
+        <Button onClick={=>@request_to_join_project()}>Request to join this project</Button>
+
+    render_not_public_error : ->
+        if @props.redux.getStore('account').is_logged_in()
+            <div>
+                <ErrorDisplay title="Directory is not public" error={"You are trying to access a non public project that you are not a collaborator on. You need to ask a collaborator of the project to add you."} />
+                {if not smc.client.account_id in smc.redux.getStore('projects').get_project(@props.project_id)?.invite_requests then @render_request_to_join_project_button()}
```
williamstein: This line is also wrong.  It'll sort of work if you have been testing for a while, but for a clean test it will fail since invite_requests is undefined.  I'll fix this.    In the future, be more careful in testing.

williamstein: I also realize this cannot and will not work - there's currently no way for the user to know if they have sent an invite -- that data just isn't available to them according to the security model.

williamstein:
1. Please pull from this, which gets it to work again, I think: https://github.com/sagemathinc/smc/tree/timothyclemansinsea-upstream_issue_872
Also, I made it so the request can type a multiline message in the request (very important). 

2. Rewrite this to **not use a modal popup**.  Instead do a state change and make a panel appear, like happens all over in https://cloud.sagemath.com/settings/account, and also in project settings when adding collaborators, etc.       Note the order of buttons and use of ButtonToolbar  as in account settings, e.g., there are two buttons with the "do it" first, then the "cancel" one second.

williamstein:
1. Pull from this https://github.com/sagemathinc/smc/tree/timothyclemansinsea-upstream_issue_872

2. There is at least two hours of work (for me) left to finish this.  There's some subtle issues, and of course I need to implement deleting outstanding requests (on the backend) from the db.

williamstein:
See https://github.com/sagemathinc/smc/issues/470

williamstein:
This has significantly bit rotted -- might be better to start over.

PR #875, opened 2016-09-07
mark chat read if typing in main box or scrolled to bottom

timothyclemansinsea:


timothyclemansinsea:
@williamstein How would I do an action on scroll? How would I mark chat read after scrolled to bottom and new chat comes in? I can't do this in render because I get a can't setstate warning.

johnjeng:
Do it in the function call before rendering from the next state.

timothyclemansinsea:
@johnjeng what "function call" I'm completely lost at what you mean

johnjeng:
Find where the application state changes that triggers a rerender. Wherever that happens, you can mark the file as read.

src/smc-webapp/editor_chat.cjsx
```diff
@@ -785,7 +790,9 @@ ChatRoom = (name) -> rclass
     on_scroll: (e) ->
```
haraldschilly: just as an idea: is `on_scroll` fired all the time? maybe we should wrap it in `throttle`.

src/smc-webapp/editor_chat.cjsx
```diff
@@ -74,7 +74,7 @@ underscore = require('underscore')
 misc = require('smc-util/misc')
 misc_page = require('./misc_page')
 {defaults, required} = misc
-{Markdown, TimeAgo, Tip} = require('./r_misc')
+{Markdown, TimeAgo, Tip, SetIntervalMixin} = require('./r_misc')
```
haraldschilly: is `SetIntervalMixin` still used? I have the impression there is no timer any more …

timothyclemansinsea: it's not used anymore. I'll get rid of it

src/smc-webapp/editor_chat.cjsx
timothyclemansinsea: no clue if it is fired all the time or what throttle is

haraldschilly: So, my thought is, that e.g. right here on GitHub you can open up the development console and type in `document.onscroll = function() { console.log("x"); }`.
Once you scroll even a little bit, you see that function is fired a hundred times. 

[throttle/debounce](http://underscorejs.org/#throttle) would either rate-limit this or defer it until after scrolling is over. I'm thinking of 50ms or so ...

I should check though, if that even applies here. It's just that his is one of those things where scrolling is suddenly getting laggy, etc.

haraldschilly:
@timothyclemansinsea ok, I looked into this and added a change to address this. However, after pushing it created a new branch here -- I obviously missed something with that new github setup, or didn't you mark this branch as being writable by us here? Worked with hal a few days ago, though.

However, please merge this into this branch: https://github.com/sagemathinc/smc/tree/timothyclemansinsea-upstream_issue_875 because I don't want to open up a second PR just for such a one liner and discard this here.

williamstein:
SetIntervalMixin is still required but not used.

timothyclemansinsea:
@simonluu would you be willing to merge this with your chat rewrite like you did for my other issue?

simonluu:
merged with chat rewrite stuff at #1029

src/smc-webapp/smc_chat.cjsx
```diff
@@ -485,7 +486,7 @@ ChatRoom = rclass ({name}) ->
             scroll_to_bottom(@refs.log_container, @props.actions)
 
     mark_as_read: ->
-        @props.redux.getActions('file_use').mark_file(@props.project_id, @props.path, 'read')
+        @props.redux.getActions('file_use').mark_file(@props.project_id, @props.path, 'seen', 0, false)
```
williamstein: What is going on here?  Why does a function called `mark_as_read` actually make the file as **seen**.  This can't possibly be right.  Either change the function name or what is done, or make a big comment explaining why the discrepancy.

src/smc-webapp/smc_chat.cjsx
```diff
@@ -467,6 +467,7 @@ ChatRoom = rclass ({name}) ->
         @set_preview_state = underscore.debounce(@set_preview_state, 500)
         @set_chat_log_state = underscore.debounce(@set_chat_log_state, 10)
         @debounce_bottom = underscore.debounce(@debounce_bottom, 10)
+        @on_scroll = underscore.debounce(@on_scroll, 10)
```
williamstein: Can we make this at least 200, instead of 10?   10 seems really short.

timothyclemansinsea:
What about changing it to mark_as_seen ? Marking as read was not changing
the notification bell marking as seen was.

200 is fine by me

src/smc-webapp/editor_chat.cjsx
```diff
@@ -178,6 +178,11 @@ class ChatActions extends Actions
 
         if messages_before != messages
             @setState(messages: messages)
+            store = @redux.getStore(@name)
+            if store.get('saved_position') + store.get('offset') + 20 > store.get('height') or store.get('height') == store.get('client_height')
```
johnjeng: This condition is hard to understand. What is the 20 for? What's the idea here? Maybe it should be a function in the store?

src/smc-webapp/editor_chat.cjsx
```diff
@@ -178,6 +178,11 @@ class ChatActions extends Actions
...
+                project_id = @name.slice(7,43)
```
johnjeng: This seems like a pretty fragile way to get the project_id and path. I suggest passing it to the action or finding it in the store.

src/smc-webapp/editor_chat.cjsx
```diff
@@ -319,7 +324,8 @@ exports.init_redux = (path, redux, project_id) ->
     store   = redux.createStore(name)
 
     actions._init()
-
+    actions.setState(project_id: project_id)
```
johnjeng: Use the initial state pattern.

timothyclemansinsea:
@johnjeng I have no idea what the initial state pattern is I didn't see any
initial state setting that I could access. I tried accessing syncdb and
couldn't.

haraldschilly:
@timothyclemansinsea I'm not sure what's going on, but my best guess is [getInitialState](https://facebook.github.io/react/docs/react-without-es6.html#setting-the-initial-state).

timothyclemansinsea:
@johnjeng I used init= because there is a middle argument

williamstein:
Sorry, but I just tried this and it completely breaks notifications.  Open a chatroom as two uers, type in one, see no notifications at all in the other.

williamstein:
To be clear -- I think the problem is that the chatroom scrolls when it is opened, hence if you just leave it open it instantly marks everything read.  Not good.

williamstein:
Fixed differently here https://github.com/sagemathinc/smc/commit/bf0a3043e57ff3284bbb135d31efa0954b23fe30 by listening for onMouseMove (and also input field changing).   Listening for onMouseMove is much better than listening for a bunch of clicks in different places and scrolls -- they all typically involve a mouse move anyways.

PR #880, opened 2016-09-07
[review requested] Sage server pytest

DrXyzzy:
This PR adds files under smc/src/smc_sagews/smc_sagews/tests:
- README.md
- conftest.py
- test_sagews.py
- test_sagews_x.py

This is a start at a regression suite to make it easy to add Sage worksheet test cases as issues are fixed.

### DIFF
--- a/src/smc_sagews/smc_sagews/tests/README.md
+++ b/src/smc_sagews/smc_sagews/tests/README.md
@@ -11,7 +11,8 @@
 ### Prerequisites
 
 - pytest must be installed
-- make sure ~/.smc/sage_server/sage_server.log exists and has the
+- sage_server must be running
+- file ~/.smc/sage_server/sage_server.log must exist and have the
    current port number around line 3, like this:
    ```
    9240 (2016-09-07 04:03:39.455): Sage server 127.0.0.1:43359
@@ -20,6 +21,6 @@
 ### Running the Tests
 
 ```
-cd .../src/smc_sagews/smc_sagews
-pytest tests [-s]
+cd smc/src/smc_sagews/smc_sagews
+python -m pytest tests [-s]
 ```
--- a/src/smc_sagews/smc_sagews/tests/README.md
+++ b/src/smc_sagews/smc_sagews/tests/README.md
@@ -1,0 +1,25 @@
+# Pytest suite for sage_server
+
+## Purpose of These Tests
+
+- verify that sage worksheets return correct results for given cell inputs
+- back-end only test, uses message protocol to run tests against running sage_server
+- NOT to provide unit tests of functions internal to sagews modules
+
+## How to Use These Tests
+
+### Prerequisites
+
+- pytest must be installed
+- make sure ~/.smc/sage_server/sage_server.log exists and has the
+   current port number around line 3, like this:
+   ```
+   9240 (2016-09-07 04:03:39.455): Sage server 127.0.0.1:43359
+   ```
+
+### Running the Tests
+
+```
+cd .../src/smc_sagews/smc_sagews
+pytest tests [-s]
+```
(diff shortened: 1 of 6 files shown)


src/smc_sagews/smc_sagews/tests/README.md
```diff
@@ -0,0 +1,25 @@
...
+- make sure ~/.smc/sage_server/sage_server.log exists and has the
+   current port number around line 3, like this:
+   ```
+   9240 (2016-09-07 04:03:39.455): Sage server 127.0.0.1:43359
+   ```
+
+### Running the Tests
+
+```
+cd .../src/smc_sagews/smc_sagews
```
haraldschilly: tripple dot?

src/smc_sagews/smc_sagews/tests/README.md
```diff
@@ -0,0 +1,25 @@
...
+pytest tests [-s]
```
haraldschilly: in my setup, not sure why, it wanted to run this with python 3. What did work for me is to be more specific about the interpreter, i.e. `python -m pytest tests/` (where python is the system-wide python 2.7)

src/smc_sagews/smc_sagews/tests/conftest.py
```diff
@@ -0,0 +1,394 @@
...
+                line = inf.readline().strip()
+                m = re.search("Sage server (?P<host>[\w.]+):(?P<port>\d+)$", line)
+                if m:
+                    host = m.group('host')
+                    port = int(m.group('port'))
+                    break
+            else:
+                raise ValueError('Server info not found in log_file',log_file)
+    except IOError:
+        pytest.fail("Unable to open log file %s\nYou may need to open a sage worksheet"%log_file)
```
haraldschilly: this and the previous error message are maybe a little bit too vague? Isn't in both cases the problem, that the sage server isn't running? I would rephrase that by something like: "there is probably no sage server running. you either have to open a sage worksheet or run `smc-sage-server start`" … at least, that's what I think needs to be done.

src/smc_sagews/smc_sagews/tests/README.md
```diff
@@ -0,0 +1,25 @@
...
+## How to Use These Tests
+
+### Prerequisites
+
+- pytest must be installed
+- make sure ~/.smc/sage_server/sage_server.log exists and has the
+   current port number around line 3, like this:
+   ```
+   9240 (2016-09-07 04:03:39.455): Sage server 127.0.0.1:43359
+   ```
```
haraldschilly: tied to my comment in the fixture about error messages, maybe it's better to explain here, that the sage server needs to be running and that there is this logfile? i.e. just the fact that the logfile exists is not enough, technically :-)

haraldschilly:
made a few cosmetic comments … running the tests in my own dev project worked fine, though, so this is good to go in.

DrXyzzy:
Thanks for comments. Ready for further review. Glad to know you got 6 tests to pass. 

__triple dot__. 
```
cd .../src/smc_sagews/smc_sagews
```
It was meant  to say, wherever the root of the smc git repo is. Changed to dots to smc.
```
cd smc/src/smc_sagews/smc_sagews
```

__pytest wanting python3__. I didn't see that problem, but changed README.md to say use `python -m pytest`. Here's are the versions of pytest and python I'm using.
```
~$ pytest --version
This is pytest version 2.7.2, imported from /usr/lib/python2.7/dist-packages/pytest.pyc
~$ which python
/usr/bin/python
~$ python --version
Python 2.7.10
```

__sage\_server__ must be running. Failure message in conftest.py and prerequisites in README.md updated as suggested.

haraldschilly:
> Here's are the versions of pytest and python I'm using

yes, that looks good, but for example (I don't remember what I did, though) in my smc dev project there is no "pytest" ... just some "py.test" which is python 3. So, my guess is, for some reasons both of our wrapper scripts differ a little bit.

PR #883, opened 2016-09-07
Video Chat and Chat Room style side chats [updated]

simonluu:
The chat button in sagews, md, and other files with side chats opens a chat that has the same style as the sage-chat chat rooms. The video button opens up a new window with an appear.in video chat room. Each file has a specific room id that is displayed in the chat room.

Also added a video chat button in the sage-chat chat rooms that does the same thing. It opens up a new window with an appear.in chat room. It has its own room id displayed in the chat log.

Also minor detail, while testing the video chat room ids, I created a bunch of sage-chat files and noticed that the create drop down menu doesn't have chat rooms. So I added it to the drop down menu.

---------------------------------------------------------------------------------------------------------------------

Split editor_chat.cjsx into side_chat.cjsx and smc_chat.cjsx. editor_chat.cjsx has all the common methods between side chat and smc chat.


src/smc-webapp/editor_chat.cjsx
```diff
@@ -799,10 +896,16 @@ ChatRoom = (name) -> rclass
...
+    send_video_id: ->
+        if not @has_video_id()[0] and not @props.is_side_chat
+            @_video_chat_id = misc.uuid()
+            @props.actions.send_video_chat("Video Chat Room ID is: #{@_video_chat_id}")
+
     componentWillMount: ->
         @set_preview_state = underscore.debounce(@set_preview_state, 500)
         @set_chat_log_state = underscore.debounce(@set_chat_log_state, 10)
         @debounce_bottom = underscore.debounce(@debounce_bottom, 10)
+        @send_video_id = underscore.debounce(@send_video_id, 10)
```
haraldschilly: I'm just reading the patch to see what's going on. Here, I see it is debouching to 10ms only? My feeling is, that this is still too fast.

src/smc-webapp/syncdoc.coffee
```diff
@@ -79,42 +83,43 @@ class SynchronizedDocument extends AbstractSynchronizedDoc
...
             payload    : required  # event-dependent dictionary
             cb         : undefined # callback
 
         new_message = misc.to_json
-            sender_id : redux.getStore('account').get_account_id()
-            date      : new Date()
+            sender_id : "Unknown"
             event     : opts.event_type
             payload   : opts.payload
+            date      : new Date()
```
haraldschilly: `salvus_client.server_time()` ? 

Just asking, I have no real idea. If it should be the client date, then maybe a comment about the why would be nice.

haraldschilly:
I just read the patch, and my only very general comment is that all those `@props.is_side_chat` make it very much spaghetti code. I don't see an easy solution for that, though.

on an unrelated meta level, I'm also questioning if those files should still be called "sage-chat". "smc-chat" comes to my mind first, but yet I think those endings should not contain the product name. maybe just "chat" ? anyways, discussing that is something for another day …

src/smc-webapp/editor_chat.cjsx
simonluu: I have it set to 10ms so that @send_video_id calls a little after mounting so that every time a new chat room is created, there will be a message that is sent that is the video chat id. If the message was sent when the chat was mounted, the chat message would be created but it would not show up, and then once you refresh the page, the message would show up.

src/smc-webapp/syncdoc.coffee
simonluu: Not sure, I think Vivek wrote that when he implemented the side chats years ago. The new Date() was there when I was working on it, I just moved it around.

simonluu:
I can see why all those @props.is_side_chat make it spaghetti code. At first I made a new file for the side chats but we wanted to make it look like the chat rooms, so we are using a lot of similar things from the chat rooms. So I decided to just to keep it in the same file.

johnjeng:
I suggest breaking it up into 2 different higher level components. If you need to reuse code, reuse them as child components.

simonluu:
Not sure how I would do this because I'm reusing code in the Message component and ChatRoom component. Should I just split all the reused code from both components into two different components?

simonluu:
Okay, split up editor_chat to fix some spaghetti code.

johnjeng:
Will review tomorrow if necessary.

williamstein:
Can somebody fix the merge conflicts on this.

simonluu:
ok fixed merge conflicts.

williamstein:
There are merge conflicts with master.  I attempted to merge but was so overwhelmed I couldn't.

williamstein:
Some cosmetic issues:


1. This immediately shows an actual visible chat that says "Video Chat Room ID is: 9053f75f-7bd2-4f18-97a7-28c9f4aa6933".

screen shot 2016-09-26 at 8 27 32 pm

The input box at the bottom of the side chat doesn't go to the bottom for me (Chrome, OS X):

screen shot 2016-09-26 at 8 29 04 pm

Less important -- the tooltips shouldn't literally say "Button" in them:

screen shot 2016-09-26 at 8 30 58 pm

williamstein:
Oh, and thanks and besides the cosmetic issues I mentioned, this seems to work fine.

simonluu:
1. At first I had it so that when the first person clicks on the video chat button, it would send a chat message with the video chat id. That would just show up at the bottom like all the other messages and as more messages are added, it gets lost in the middle. I wasn't sure how to make it so that when the first person clicks the video chat button it sends a chat and it shows up at the top, so I made it so when you first open a new chat room, it will just have a message with the chat room id.

2. CSS is dumb. The height of the chatlog is a fixed number. So if I change it to a different fixed number to make it go to the bottom, it will be at the bottom for some people but it will go off on other screens for others. (This was happening at SD81 with Vivek. I made it so it go to bottom for me but on Vivek's laptop it was going off screen). So I tried making it height dependent on the size of the screen but there was a problem.

3. Okay I can remove the button.

williamstein:
1. Clearly whatever shares the id should be hidden from the user's view.  If you don't know how to implement that, just say so.

2. Regardless of whether or not CSS is dumb, this problem has to get solved.

3. ok

simonluu:
Oh, I think I misunderstood what you told me before. I will fix these problems.

williamstein:
Heh, this totally works perfectly, as far as I can tell.  I still don't like the vertical height of the input box, but it's better, and probably after switching to the top down react rewrite, you can fix it right.  This is https://github.com/sagemathinc/smc/issues/1001

PR #896, opened 2016-09-12
webapp/quotas: enforcing quota limit from the upgrade specs in the UI…

haraldschilly:
… (also being a bit more explicit about this limit to avoid confusion)

### DIFF
--- a/src/smc-webapp/project_settings.cjsx
+++ b/src/smc-webapp/project_settings.cjsx
@@ -174,6 +174,9 @@ UpgradeAdjustor = rclass
             Max
         </Button>
 
+    render_addon : (misc, name, display_unit, limit) ->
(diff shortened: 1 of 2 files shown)


haraldschilly:
how to test:

1. get a nice package of upgrades for testing
1. in the project settings, increase for example the memory to the max. it should be either at the overall limit of the quotas of the user or at 8000 (which is the limit defined in the upgrade-spec file, which is also publicly communicated on the pricing page)
1. when creating a new project, the quota adjustments in the "new project dialog" should behave the same like with the project settings.

PS: I hate code duplications ... increases my dev time by at least 2x

williamstein:
I created a follow-up issue: https://github.com/sagemathinc/smc/issues/906

PR #897, opened 2016-09-12
webapp/files: attempt to retry file listings

haraldschilly:
This is my attempt to automatically retry getting that file listing for newly started projects. I think I'm on the right track, but it causes warnings: "ERROR getting username for account with id null" somewhere down the line.


williamstein:
I happened to do this on the plane home not realizing it was here.  But this is broken as is anyways, with it trying to do something that can take up to 45s... but with a max_time of 30s.

PR #907, opened 2016-09-16
webapp: fixing directory listing sort and hidden arguments

haraldschilly:
I think that's the problem with the directory listings, but I can't test it since everything fails in my dev environment.


haraldschilly:
deployed this in the test and prod namespaces as the image `fix-file-listing-sort-d3eda9`

PR #868, opened 2016-09-20
mobile chat issues: it is impossible to close the chat (there should be a "Files" button like with all other editors).

timothyclemansinsea:


timothyclemansinsea:
I got rid of the two columns so that the buttons would be on the same row. If there are avatars they'll show up on the same row as the buttons or below it.

simonluu:
I think right now the container for the chat is completely blocking out the bar that has the files, create new files, project activity log, search and settings buttons because the position of the container is set to position: fixed with a top: 0px.

williamstein:
(i made a few additional changes... but don't like this at all; still kind of ugly/mess, as discussed above.  But at least usable now.)

PR #827, opened 2016-09-21
Minor typo in course handout distribution

timothyclemansinsea:
When distributing a handout in a course, if some students have already received it and some have not, a button appears which says something like, "to the n students not already distributeed to"

There is an extra "e" in distributed.


williamstein:
Still a problem.  Why is there a PR and no issue?  I've made an issue: https://github.com/sagemathinc/smc/issues/1094

johnjeng:
#1094 has since been closed so I presume this should be too.

PR #850, opened 2016-09-21
don't show line number if line = 1 in users viewing document

timothyclemansinsea:
This would address editors like chat where line numbers make no sense

### DIFF
--- a/src/smc-webapp/profile.cjsx
+++ b/src/smc-webapp/profile.cjsx
@@ -94,8 +94,7 @@ Avatar = rclass
         return merge(style, @props.style)
 
     render_line: ->
-        if @props.line
-            <span> (Line {@props.line})</span>
+        <span> (Line {@props.line})</span>
 
     tooltip: ->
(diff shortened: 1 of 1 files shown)


williamstein:
(sorry if my review sounds harsh -- your actual code here was nice code; it's just not quite the right approach, imho).

timothyclemansinsea:
@williamstein is there already a way to know if a given editor is a codemirror editor? If not is this something that could be easily added after react rewrite? I'd like to do something like `if @props.is_codemirror then @render_line_number`

williamstein:
I think maybe the cursor position for that user would be undefined instead of an actual number (?).  Not sure...

timothyclemansinsea:
@williamstein prior there was `if @props.line`  and yet there was still a line number showing for chat.

williamstein:
It's necessary to understand things more deeply than just looking at `props.line`, e.g., in syncstring.coffee...

Issue #669, opened 2016-07-01
On an iPhone using Google Chrome with SMC chatrooms, it is not possible to type input.


williamstein:
This is for chatrooms. You can focus the input field, but then the keyboard starts fickering like crazy and you can't type.  It's probably something involving scrolling or rendering on click, which makes an assumption that is wrong (or a bug in mobile chrome that we should maybe work around).

Note: safari on iPhone works fine.

johnjeng:
Also on Android using mobile Chrome. Seems to continually redraw and refocus the text box. Sometimes you get time to fit in a letter or there.

williamstein:
Works now.

Issue #670, opened 2016-07-01
On mobile, do not autofocus most forms.


johnjeng:
Automatically pulling up the keyboard is awful.

timothyclemansinsea:
For now don't change this for chat as without autofocus=true for chat it doesn't work at all in Chrome on Mobile.

williamstein:
Make a specific list or I'll close.  This seems too vague.  I also frequently use SMC on mobile, and am not actually that annoyed by the current autofocus behavior.  Anyway, this too vague.

johnjeng:
Fine with me.

PR #930, opened 2016-09-23
Iphone and android chrome input fixes

simonluu:
Fixes issue #669 
Changed auto focus to false on ios since we don't want to auto focus most forms on mobile. Changing auto focus to false on android messes up the input, so auto focus is set to true and it works.

### DIFF
--- a/src/smc-webapp/editor_chat.cjsx
+++ b/src/smc-webapp/editor_chat.cjsx
@@ -66,7 +66,7 @@ editing   : immutable.Map
 
 # standard non-SMC libraries
 immutable = require('immutable')
-{IS_MOBILE} = require('./feature')
+{IS_MOBILE, isMobile} = require('./feature')
 underscore = require('underscore')
 
 # SMC libraries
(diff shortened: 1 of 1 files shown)


williamstein:
@timothyclemansinsea or @haraldschilly -- can you review this now?   I don't have an android phone....

src/smc-webapp/editor_chat.cjsx
```diff
@@ -1022,7 +1022,7 @@ ChatRoom = (name) -> rclass
                 <Row>
                     <Col xs={10} style={padding:'0px 2px 0px 2px'}>
                         <Input
-                            autoFocus   = {true}
+                            autoFocus   = {if isMobile.Android() then true else false}
```
haraldschilly: `isMobile.Android()` already returns a boolean value, so just write `isMobile.Android()` there

haraldschilly: I fixed this, don't worry about it

haraldschilly:
my only trouble with reviewing this is, that my dev project for the webapp is still broken. that's so ridiculous :expressionless:

src/smc-webapp/editor_chat.cjsx
haraldschilly: oh, and more general, maybe you both actually want it the other way around ? I'm trying to test it.

haraldschilly:
Well, by experimenting around, I was able to more or less test and verify it. My observation is, that the input box is always focused on android. What I don't know is, what happens on mobile non-android systems? Is it still possible to enter anything? What if that `autoFocus=true` is only set if on Android and otherwise it isn't set at all, would that work on iOS mobile devices?

Besides that I am not able to really follow what's going on here any more. The associated ticket is set to be closed and working 9h ago, 4h later william asks to review this on android. Wouldn't it be less confusing, if the ticket is closed after the fix is deployed and verified to work?

My assumption is, that setting `autoFocus=false` on iOS mobile devices does work and that's why the original ticket is closed.

simonluu:
Setting autoFocus=false on iOS is fine but setting autoFocus=false on android breaks the input and won't let you type.

williamstein:
> Wouldn't it be less confusing, if the ticket is closed after the fix is deployed and verified to work?

The original ticket title is "On an iPhone using Google Chrome with SMC chatrooms, it is not possible to type input." and it is definitely the case (as verified by my iPhone) that on live SMC right now things work fine. 

From what you guys say above, it seems safe to merge this.

williamstein:
WAIT: You say above "Setting autoFocus=false on iOS is fine"... except that right now in the live site the running code is `autoFocus   = {true}` and everything seems to work fine on iPhone.  The whole point of this PR seems to be to break iPhone support while maintaining android support?  Huh?

I think I should just close this PR.

williamstein:
Hold on -- I didn't realize this was already merged.  Let me just say: whatever is live NOW works perfectly on iphone so I'm a happy camper.

simonluu:
it doesn't matter if autoFocus is set to true or false on iOS, it works either way. Setting it to false is for issue #670 . I tried setting it to false for android but that just breaks the input for android.

haraldschilly:
So, that's what did confuse me. The very first commit here says it solves the ticket #669 … and it got closed after that. Everything else is just code cleanup. The autoFocus=true is definitely necessary for android.

Issue #189, opened 2015-10-26
add a "check your spam folder" message (in bold/red) to the "reset your password" dialog, and fix some other clumsy bugginess in that modal


williamstein:
In the very rare case a password reset message doesn't work, when we get to the bottom of it, the user says:

> Thank you very much for your help. indeed the password reset emails had gone to the spam.

williamstein:
We still need to to do this.  The message now only says not this.  

<img width="597" alt="screen shot 2016-09-23 at 12 22 58 pm" src="https://cloud.githubusercontent.com/assets/1276278/18798688/8b23a1fe-8188-11e6-8a84-66c7806befd6.png">

Also, the forgot password modal popup doesn't even autofocus the input.

Clicking on the modal at all (e.g., to copy paste) just closes the dialog, which is disconcerting.

This whole thing should be rock solid, not a mess, given how important it is...

haraldschilly:
for the sake of mentioning it, my old branch for a completely different way how passwords are done is here: https://github.com/sagemathinc/smc/tree/password-reset-plain

williamstein:
(it is a major stretch that this is a bug... but it is really important.)

williamstein:
fixed and live now.

PR #931, opened 2016-09-23
added a "check your spam folder" and fixed other things in the modal.

simonluu:
Fixes issue #189 
- Added a bold red "check your spam folder"
- moved the dialog to below the input box. I think that makes more sense than being above it.
- made the input box auto focus when you click "forgot password?".
- noticed a bug while fixing this when you click reset password with an empty input box it just gives you the same dialog with an empty email. I think that it should just say "Please enter an email address". So I made it so it says that if you try to reset password with an empty email address. It also focuses the input again if that happens.

### DIFF
--- a/src/smc-webapp/_account.sass
+++ b/src/smc-webapp/_account.sass
@@ -16,7 +16,7 @@
 
 .smc-sign-in-form
   line-height: 1em
-  font-size: 12px !important
+  font-size: 11pt !important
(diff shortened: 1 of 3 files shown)


williamstein:
Dude, you're killing it!

williamstein:
I made some more changes and merged this in master.

PR #397, opened 2016-09-25
sometimes a project's home directory is owned by root (so wrong permissions)

timothyclemansinsea:
Sometimes /projects/[project_id] doesn't have group/owner equal to that of the project.  Add a check or something in src/smc_pyutil/smc_pyutil/smc_compute.py to ensure it does whenever starting the project.

(Don't just chown the whole tree though, since that can take a very long time and is usually wasteful.)


williamstein:
Looks good actually!

williamstein:
Wow, I was over-confident this would work that I made it live, and it instantly killed all of SMC project starting, of course.  Last time I make that mistake.  Will test fix properly now.

williamstein:
OK, I fixed this in a follow up PR #d4962c90569cd7c963a657ffcc917feba72e25d3.   The code was clearly -- in retrospect -- wrong and was never tested.     It was trying to do "chmod" on the file with empty name in case of not recursive.

PR #238, opened 2016-09-25
make it possible to draw plots in %octave mode on Sage worksheets

timothyclemansinsea:
Even the most basic thing is completely utterly broken.  This

```
%octave
x = -10:0.1:10;
plot (x, sin (x));
```

hangs forever, and similarly, in the command line, this hangs forever:

```
sage: octave.eval("x = -10:0.1:10;\nplot (x, sin (x));")
```

Help from Nils/Andrey:

```
On Friday, November 6, 2015 at 2:59:56 PM UTC-8, Andrey Novoseltsev wrote:
My understanding is that (unlike for R) we do not try to do anything for Octave plotting, 
so everything is "default". Plotting commands in Octave use either gnuplot or OpenGL,
 but it seems that in our setup only gnuplot is available. How extractly it is used I don't 
know, but my guess is that Octave creates some input file and then passes it to "gnuplot" 
and you have an option to change the name of the called binary and supply some
 options to it. For whatever reason, default options result in just ASCII plot in 
SageMathCell and SageNB. When I try to add explicit commands for saving in a 
particular format I am just getting the same ASCII plot one more time and some 
warnings. Starting Octave from a local terminal allows me to open plots in a new 
window, while saving files works but still shows warning about missing things. 
It would be nice if someone who knows Octave and gnuplot figured out what do 
we need to do to get some plots and no warnings - I am stopping for now.

I don't know either very much, but I have succeeded on my local octave (on 
Fedora) to produce some plots in files via essentially this code:

https://sagecell.sagemath.org/?z=eJxLy0wvLUrVMNRRKssszkzKSVXSUcpPS1PS5OUqLi1K0yhITcwuBnESy1ITi0HqwCJ6BXnpIDUAiBwTdA==&lang=octave 

figure(1,"visible","off")
surf(peaks)
saveas(1,"peaks.png")

(saving to peak.pdf works too).

I did install "transfig" (a part of xfig), as well as "pstoedit". 

Turning the "visible" off ensures that you don't get the silly ascii art.

Other remarks:

I do need to run octave with "DISPLAY" unset. Otherwise it seems to want to use the 
graphics windows things, and from there any attempt to saving the graphics leads to a
 segfault [which has significantly diminished my trust in octave, or at least its fedora
 packaging]. Even with "visible" off, I wasn't getting saved graphics (but no segfaults either).

It would be great if we could get basic octave graphics support running on sagecell. I 
think plenty of matlab-centric courses could make excellent use of it.
```

### DIFF
--- a/src/smc_sagews/smc_sagews/sage_jupyter.py
+++ b/src/smc_sagews/smc_sagews/sage_jupyter.py
@@ -84,6 +84,7 @@ class JUPYTER(object):
     __doc__ = property(_get_doc)
 
 jupyter = JUPYTER()
+octave = jupyter('octave')
 
 import jupyter_client
 from Queue import Empty


timothyclemansinsea:
I need help on this one. Thought this would be easy to just put `octave = jupyter('octave')` after `jupyter = JUPYTER()` but after upgrading via pip and restarting I don't get any sagews output

src/smc_sagews/smc_sagews/sage_jupyter.py
```diff
@@ -84,6 +84,7 @@ def _get_doc(self):
     __doc__ = property(_get_doc)
 
 jupyter = JUPYTER()
+octave = jupyter('octave')
```
haraldschilly: I'm a bit puzzled about what you try to do here. Trying to understand your intentions, this line means that each time a sagews starts, octave starts up too? That doesn't sound like a good idea to me.

DrXyzzy:
1. See %sh from after [PR911](https://github.com/sagemathinc/smc/pull/911) for code to create a sagews pre-defined mode - need to a) start the jupyter client b) create the mode function c) add `jupyter_kernel` attribute to the mode function for autocomplete.https://github.com/sagemathinc/smc/blob/master/src/smc_sagews/smc_sagews/sage_salvus.py#L1997
1. Merely adding a pre-defined mode for octave jupyter client by doing the equivalent of `octave = jupyter('octave')` will not be backward compatible with the Sage interpreter interface to Octave: http://doc.sagemath.org/html/en/reference/interfaces/sage/interfaces/octave.html __This is a requirements issue.__ Do we
 - replace the old octave now and give up some of its functionality
 - add a second octave mode with plotting but not sage interpreter interface
 - take time to add support to octave interpreter sage interface with jupyter client before replacing
1. I'm adding pytest cases for new sagews features & fixes. I'll submit a PR with pytest refactoring for sagews modes within an hour or so, based on experiences with %sh jupyter bridge.

williamstein:
> will not be backward compatible with the Sage interpreter interface to Octave

This is true.  However, how much do we loose?  The sage interpreter interface to Octave (which written by yours truly) is -- honestly -- so broken as to be useless.

haraldschilly:
Question, can we simply close this? William's initial example works in a sagews and it always worked in jupyter notebooks anyways. Well, I'm closing it. If there is something wrong, we should probably start with an actually broken test case in the first place.

williamstein:
Yes -- +1 to close.  Explicit testable fixable well defined issues are the goal.

Issue #920, opened 2016-09-21
%sh default_mode autocomplete and capture not working


DrXyzzy:
New problems since PR #911 

- [x] fix autocomplete
- [x] fix capture

DrXyzzy:
Autocomplete fixed. Still working on %capture.

haraldschilly:
made you two checkboxes, such that we know that this is half done

DrXyzzy:
Capture might be broken with default_mode in general, not just with the new sh:
```
%default_mode r
... new cell ...
x <- c(9,7,44)
... new cell ...
%capture(stdout='foo')
mean(x)
... new cell ...
%sage
print(foo)
NaN
```

DrXyzzy:
Done, verified in production.

PR #939, opened 2016-09-25
sagews %sh autocomplete

DrXyzzy:
Ref. #920 

This PR fixes autocomplete for direct and default %sh mode and adds pytest for those.

Test with
```
python -m pytest test_sagews_modes.py::TestShMode
```

Not fixed in this PR: %capture with %default_mode sh.


williamstein:
Now making live by typing this on all compute nodes: `salvus@compute1-us:~$ ./update`

PR #943, opened 2016-09-26
fix capture for default modes for sh and r, add pytest cases

DrXyzzy:
Second part of #920.
- fix %capture with %default_mode sh
- fix %capture with %default_mode r
- add pytest cases for above and simple computation in r

To run pytest cases for this issue:
```
python -m pytest test_sagews_modes.py
```

### DIFF
--- a/src/smc_sagews/smc_sagews/tests/test_sagews_modes.py
+++ b/src/smc_sagews/smc_sagews/tests/test_sagews_modes.py
@@ -34,6 +34,7 @@ class TestShMode:
 
     def test_sh_display(self, execblob, image_file):
         execblob("%sh display < " + str(image_file))
+
     def test_sh_autocomplete_01(self, exec2):
         exec2("%sh TESTVAR29=xyz")
     def test_sh_autocomplete_02(self, test_id, sagews):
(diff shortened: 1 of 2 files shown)


haraldschilly:
So, I started to test this. The automatic tests pass, good, but I found one problem, which is maybe not related to this, but I'm mentioning it anyways. When running a command that doesn't exist in `%sh` it crashes completely.

```
%sh
xyz
```
```
Fatal Python error: Py_Initialize: Unable to get the locale encoding
  File "/projects/sage/sage-7.3/local/lib/python/encodings/__init__.py", line 123
    raise CodecRegistryError,\
                            ^
SyntaxError: invalid syntax

Current thread 0x00007fd0d4a7a700 (most recent call first):
Aborted (core dumped)
```

I'll play around with it anyways, and also check the code.

src/smc_sagews/smc_sagews/tests/test_sagews_modes.py
```diff
@@ -46,3 +47,64 @@ def test_sh_autocomplete_02(self, test_id, sagews):
...
+        exec2("pwd\n", html_pattern=">/")
+
+    def test_multiline(self, exec2):
+        exec2("FOO=hello\necho $FOO", html_pattern="hello")
+
+    def test_date(self, exec2):
+        exec2("date +%Y-%m-%d", html_pattern = '\d{4}-\d{2}-\d{2}')
+
+    def test_capture_sh_01(self, exec2):
+        exec2("%capture(stdout='output')\nuptime")
```
haraldschilly: I'm not 100% sure about this test. I tried running these examples directly in a sagews, and what did work for me is:

```
%capture(stdout='output')
%sh
uptime
```
and then
```
%sage
output
```

Just like in the R case below, there is an additional `%sh` like there is an additional `%r`  in the `%capture`. 

Or is there a default set, which I'm not seeing here? In any case, the test did work, but in my sagews the `output` was empty.

haraldschilly: ok, :sparkler:, my fault. the first test sets the default. so that's fine.

Issue #945, opened 2016-09-26
jupyter client core dump with %sh command not found


DrXyzzy:
As reported in [PR943 comment](https://github.com/sagemathinc/smc/pull/943#issuecomment-249523900), calling an undefined shell command does not fail gracefully. 
```
%sh
xyz
```
```
Fatal Python error: Py_Initialize: Unable to get the locale encoding
  File "/projects/sage/sage-7.3/local/lib/python/encodings/__init__.py", line 123
    raise CodecRegistryError,\
                            ^
SyntaxError: invalid syntax

Current thread 0x00007fd0d4a7a700 (most recent call first):
Aborted (core dumped)
```

Problem also appears with jupyter notebook running bash kernel.

DrXyzzy:
Done, verified in production.

PR #946, opened 2016-09-26
add workaround for %sh command not found and pytest case

DrXyzzy:
Workaround for issue/bug #945, with pytest case added.


DrXyzzy:
to test:
```
python -m pytest test_sagews_modes.py::TestShMode
```

Issue #355, opened 2016-01-03
%octave mode in sage worksheets is flaky simply because Sage's own octave.eval is flaky


williamstein:
Unfortunately the expect-based Sage interface to octave isn't sufficiently robust at this time. 

You'll have to find another solution until either we figure out how to install an Octave kernel for Jupyter (which might be more robust and you could use instead), or write a new interface between Octave and Sage that is better than the current one. 

Email from Daniel Krenn:
```
On Sun, Jan 3, 2016 at 8:08 AM,  wrote:
I've used SMC when teaching SageMath and Octave. With Octave, the notebook (as the interface to octave) seems to be not as stable as it should be. Here some observations and a bug (all done on the SMC-Notebook):

1) In our first lecture, we (my students) had to do a lot of restarts of the worksheet (and even the project server). Of course, when using a software for the first time, you have to get used to it. They were typing in some simple commands and for sure a lot of syntax errors, but there should not be any need to restart that often.
To be a bit more precise: During 90 minutes of class all of the 15 persons did restart the worksheet at least once; about one half of them frequently. Common reason: evaluation did not work at all anymore; weird results appeared as output (were correct after the restart).

Also when using SageMath in the SMC-notebook, it is more unstable than the sagenb-notebook was (I've really had fewer complaints in recent years when using sagenb for teaching).

Unfortunately I cannot make this all more precise...

2) In the first cell:

%default_mode octave
a = [1 2 3]

does not work (produces a Python syntax error). Only all other cells seem to be switched to octave-mode, but not this first cell. This was/is unexpected. If intended: Why?

3) %default_mode octave

randn(10)

The second cell with randn(10) hangs. randn(9) works.


Best,

Daniel
```

haraldschilly:
about (2): One important thing to know (let's not argue if this is a bug or feature), that each cell that starts with `%auto` or `%default_mode` is in itself always in "Sage" mode. So, please try to evaluate the cell with only `%default_mode octave` in it _first_, and then the other cells.

(3) is definitely a bug with the wrapper in sagemath and should be reported upstream.

evelynmitchell:
I found https://github.com/calysto/octave_kernel 
which is a Jupyter kernel for Octave. Looks to be under active development.

There's also octave magic from Oct2py
https://github.com/blink1073/oct2py

billpage:
I think there are good reasons to want to continue to use Octave from a
Sage worksheet - the primary one being interoperability with other parts of
[...]

PR #952, opened 2016-09-27
Issue355octave

DrXyzzy:
Issue/bug #355.
- %octave mode uses jupyter kernel (this will hide sage octave())
- add pytest for calculation, plot, capture in default mode octave
- omit output count numbers "[1] xxx" etc. from jupyter kernel outputs
- more pytest refactoring

```
python -m pytest test_sagews_modes.py::TestOctaveMode
python -m pytest test_sagews_modes.py::TestOctaveDefaultMode
python -m pytest test_graphics.py::TestOctavePlot
```

### DIFF
--- a/src/smc_sagews/smc_sagews/sage_jupyter.py
+++ b/src/smc_sagews/smc_sagews/sage_jupyter.py
@@ -156,7 +156,7 @@ def _jkmagic(kernel_name, **kwargs):
         stdinj = kc.stdin_channel
 
         # buffering for %capture because we don't know whether output is stdout or stderr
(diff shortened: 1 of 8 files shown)


haraldschilly:
Is this ready to review? I'll have a look

haraldschilly:
I've tested the octave mode quite extensively. Went through many examples, some plots, added a test for "for loops", etc. Sometimes it just got stuck for no obvious reason and some plots look ugly. I blame that on octave. It's definitely way better than before, i.e. especially that lost sync with for loops is solved.

Issue #112, opened 2015-09-14
$ sign shouldn't work in <pre> in md files


AndreWin:
For example I need to write the following code:
```
$ cd ~
$ mkdir apps
```
This code block contains $ signs and is converted to \<pre\> in preview side. This code in \<pre\> should be shown as it was typed, but Mathjax tries to parse content in this code block.

You can see more information about it [here](https://groups.google.com/forum/#!topic/sage-cloud/JAwrzv7XS3k)

Best regards, Andrey.

williamstein:
Not parsing $'s in pre environments would save us a lot of pain in a lot of places!!

haraldschilly:
so, uhm, in my eyes this ticket is a hard one. however.

1. I can't reproduce the originally reported problem. I've tested single or double dollars in chats inside a pre, in a markdown file, and in `%md` in a sagews. everything fine. Hence, even interpreting this more broadly, this bug is ok. 

2. Despite that, I assume @williamstein  has tested this to be still the case 4 days ago, so I'm definitely missing some corner cases.

3. a case where I know something is off is ```x `$abc` y``` an  a chat. I boiled this down to how markdown processing is done in `markdown.coffee:markdown_to_html`. I've submitted PR #954 to address this detail.

4. the only real undecidable case in markdown is this: 

        $a^b
        $xxx

   where there is no newline between the two and they're also not in `pre` or ```. It's somehow automatically converted to being a  one-liner and hence the heuristic in `misc.parse_mathjax` no longer detects it.

williamstein:
fixed

PR #954, opened 2016-09-27
webapp: rendering mathjax by first running marked and then analyzing mathjax code. this catches corner cases like `\$999$` in sagews' %md and `foo $999` in chat

haraldschilly:
This is motivated by #112 but actually solves:

* ```foo `\$999$` bar``` in `%md` in sagews
* `foo $999` in chat

here is some markdown for testing this:

    $a^b$

    `Payment of $99`

    abc `text $ 123` abc

    abc `$a^b$` test

    <pre>
    $a^b$
    </pre>

    ```
    asdf $a^b$ asdf
    asdf
    ```

    <pre>
    $ cd ~
    $ mkdir apps
    </pre>

    ```
    $ nosetests pandas ...
    ..........
    OK (SKIP=2)
    $ test
    ```

    $xxx

    x `$abc` y

    payment of \$ab99

    payment of `\$ab $`

    payment of `\$999$`

    `foo $999`

I've also explicitly added the tags that mathjax should ignore -- there were none specified and these are the defaults.

### DIFF
--- a/src/smc-webapp/editor_chat.cjsx
+++ b/src/smc-webapp/editor_chat.cjsx
@@ -443,7 +443,9 @@ Message = rclass
         value = misc.smiley
             s: value
             wrap: ['<span class="smc-editor-chat-smiley">', '</span>']
+        #value_old = value
         value = misc_page.sanitize_html(value)
+        #console.log "sanitize: '#{value_old}' -> '#{value}'"
 
         font_size = "#{@props.font_size}px"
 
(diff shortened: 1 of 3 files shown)


haraldschilly:
and well, despite looking simple, I fear there are some cases where this breaks something – or well – maybe it's ok to break some corner cases for the greater good :-)

williamstein:
Looks better to me.

Issue #585, opened 2016-05-27
bug in markdown/html output font size in sage worksheets


williamstein:
Create a cell with some text and headers/indents using markdown in a Sage worksheet:

```
%md
# Foo
## bar
This is a test
- one
- two
```

Now use the font-size icons to increase the font.  The text increases, but the headers don't.

REPORTED BY: Joel Laakso

williamstein:
Same problem for html, e.g., 
```
%html
<h1>
    Foo
</h1>

bar
```

williamstein:
See https://cloud.sagemath.com/projects/4a5f0542-5873-4eed-a85c-a18c706e8bcd/files/support/2016-09-22-153951-issue585.sagews

simonluu:
There was an issue, that I can't find anymore, similar to this with chats where changing the font size in account settings doesn't change the font sizes of the markdown chat messages.

williamstein:
NOTE: So I think the right solution for pr 955 (issue #585) is to get rid of the Bootstrap CSS rule that hard codes the h1, h2, etc., font size in pixels. Just deleting that rule in chrome tools immediately fixes everything uniformly without any of that hackish code. Thoughts?

simonluu:
That was my first solution, but I'm not sure how to do that.

williamstein:
>  I'm not sure how to do that.

Figure out how... or if that is too hard, Harald or I will hopefully do so.

williamstein:
In a sage worksheet, I put these lines in a file `a.css`
```
h1 { font-size: 250% !important }
h2 { font-size: 200% !important }
h3 { font-size: 170% !important }
```
Then I typed `%load a.css`. 

In the rest of the worksheet the problem is solved...  Maybe need more than just h1,h2,h3, and the choice of percents was sort of arbitrary.

PR #955, opened 2016-09-27
fixes md/html output font size not changing in sagews

simonluu:
issue #585


williamstein:
I realize this wasn't marked needs review yet, but looked anyways.  As it, this code doesn't work at all.  It makes things worse.  Try with the input example 
```
%md
# Foo
## bar
This is a test
- one
- two
```

and zoom in or out *a lot*.   Try refreshing...

src/smc-webapp/editor.coffee
```diff
@@ -1932,13 +1932,33 @@ class CodeMirrorEditor extends FileEditor
...
             else if @default_font_size?
-                @set_font_size(cm, @default_font_size)
+                @set_font_size(cm, @default_font_size, 0)
 
-    set_font_size: (cm, size) =>
+    set_font_size: (cm, size, delta) =>
         if size > 1
             elt = $(cm.getWrapperElement())
+            if delta != 0
+                for html, i in elt.find('.sagews-output-html').children()
```
haraldschilly: so, my first instinct here is, that since this is jquery, you do not have to iterate. just specify in the find-query that you mean all the children and then run an operation once on all of them.

src/smc-webapp/editor.coffee
```diff
@@ -1932,13 +1932,33 @@ class CodeMirrorEditor extends FileEditor
...
+                    original_font_size = parseInt($(html).css('font-size'))
+                    $(html).css('font-size', original_font_size + delta + 'px')
+                    @local_storage("html_font#{i + html}", original_font_size + delta)
```
haraldschilly: uhm, and all that into local storage? does it even support keys of arbitrary length? or do I miss what `html` is? I think I should finish reading this first ...

williamstein:
NO way -- this is hackish code, especially the `setTimeout(..., 1000)`.   

So I think the right solution for pr 955 (issue #585) is to get rid of the Bootstrap CSS rule that hard codes the h1, h2, etc., font size in pixels. Just deleting that rule in chrome tools immediately fixes everything uniformly without any of that hackish code. Thoughts?

williamstein:
> I would never ever pollute local storage that way just for font sizes. 

Yes, this is really terrifyingly bad.  Do you know that on many browsers localStorage is restricted to maybe 5MB?    And you really need to clean it up (we don't do this enough) -- you shouldn't just write to it and never clean it up later, unless you want the app to just start mysteriously breaking.

simonluu:
My first thought was that I had to remove the bootstrap css from the h1, h2, ..etc but I'm not sure how to do this, so I went to overriding the css of the heading tags.

williamstein:
I'm closing this and will fix this properly with a css style.

Issue #476, opened 2016-03-17
in a sage worksheet: in the docstring for "show?" it says "import graphics"...


williamstein:
... but it should say `import smc_sagews.graphics; smc_sagews.graphics.graph_to_d3_jsonable?`

DrXyzzy:
Done. Verified change in production manually and with pytest.

PR #957, opened 2016-09-27
issues 476 and 594 with pytest case

DrXyzzy:
- Made change to docstring per #476
- Added pytest case `tests/test_sagews.py::TestBasic::test_show_doc`, first test case for `code` response from front end.

### DIFF
--- a/src/smc_sagews/smc_sagews/sage_salvus.py
+++ b/src/smc_sagews/smc_sagews/sage_salvus.py
@@ -2530,12 +2530,16 @@ def show(*objs, **kwds):
                 return "$\\displaystyle %s$"%s
             else:
                 return "$%s$"%s
+    sys.stdout.flush()
+    sys.stderr.flush()
     s = show0(objs, combine_all=True)
     if s is not None:
         if display:
             salvus.html("<div align='center'>%s</div>"%cgi.escape(s))
         else:
             salvus.html("<div>%s</div>"%cgi.escape(s))
+        sys.stdout.flush()
+        sys.stderr.flush()
 
 # Make it so plots plot themselves correctly when they call their repr.
 Graphics.show = show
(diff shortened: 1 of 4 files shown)


DrXyzzy:
Rather than deal with multiple colliding branches off SMC master, I added work for #594 to this PR as well. Flush stdout and stderr before and after plot, to preserve order of mixed output of text & plot. Added pytest case as well:
```
python -m pytest test_graphics.py::TestShowPlot
```

src/smc_sagews/smc_sagews/tests/test_sagews.py
```diff
@@ -50,6 +50,20 @@ def test_search_doc(self, exec2):
...
+        code = "show?"
+        patn = "import smc_sagews.graphics; smc_sagews.graphics.graph_to_d3_jsonable?"
+        m = conftest.message.execute_code(code = code, id = test_id)
+        sagews.send_json(m)
+        typ, mesg = sagews.recv()
+        assert typ == 'json'
+        assert mesg['id'] == test_id
+        assert 'code' in mesg
+        assert 'source' in mesg['code']
+        assert re.sub('\s+','',patn) in re.sub('\s+','',mesg['code']['source'])
```
haraldschilly: hmm, in an ideal world, we would have a simple tokenizer and a fuzzy text matcher at hand. I had to do a similar trick testing some multiline output for `%octave` yesterday. but in this case, that's definitely ok.

haraldschilly: throwing ideas into the room: assert [SequenceMatcher(a= .., b= ..).ratio()](https://docs.python.org/3/library/difflib.html#difflib.SequenceMatcher.ratio) > .7 ?

src/smc_sagews/smc_sagews/sage_salvus.py
```diff
@@ -2412,7 +2412,7 @@ def show(*objs, **kwds):
        - svg: (default: True); if True, show 2d plots using svg (otherwise use png)
 
        - d3: (default: True); if True, show graphs (vertices and edges) using an interactive D3 viewer
-           for the many options for this viewer, type 'import graphics; graphics.graph_to_d3_jsonable?'
+           for the many options for this viewer, type 'import smc_sagews.graphics; smc_sagews.graphics.graph_to_d3_jsonable?'
```
haraldschilly: what I'll be trying to fix here is, that this line-wraps and isn't copy/paste friendly (adds spaces and of course, the line-break ...)

Issue #589, opened 2016-05-28
Fix/improve the checkbox that enables users to obtain and edit the generated latex when printing worksheets


williamstein:
Here are the manual instructions:

Not easily.  However, if you make a terminal (+New --> Terminal) next to your worksheet, then type

  smc-sagews2pdf --remove_tmpdir=false Appendix.sagews

then look at the directory at the end of the output, which will look like this:

  /tmp/tmp4hf9YO

you can then type (in the terminal)

 cp -r /tmp/tmp4hf9YO .; open tmp4hf9YO/tmp.tex

You'll then see the latex version of your worksheet, where you can fine tune things.   E.g., you can remove the heading completely by putting % at the beginning of the line \maketitle. 

Whenever you save, it will recompile.   You can click the download button on the far right to download the pdf to your computer.

williamstein:
The "Keep generated files in a sub-directory. This is useful for debugging printing issues or additional editing." checkbox when printing sagews should do exactly this...

BUT that doesn't work at all right now!

haraldschilly:
Hmm, I just tried it, and it works :confused:

williamstein:
I just tried with any random file.... but **NOT in the root directory**.  

I now see that (1) the generates files are in the root directory, even if the sagews is in a subdirectory.  (2) After doing the build, there is no link to the generated files.

haraldschilly:
ok, I understand. so, it's at least an issue with the paths and I never worked on showing a link to it at all, so that aspect is also not solved either.

PR #960, opened 2016-09-28
webapp/pyutil: printing sagews and keeping it correctly creates a subdirectory and adds a link to open the generated tex file

haraldschilly:
This addresses issue #589. I also took the liberty to remove the easy flag, because getting this work involved more than I expected :smiley:


src/smc-webapp/editor.coffee
```diff
@@ -2067,8 +2070,28 @@ class CodeMirrorEditor extends FileEditor
                                 cb(err)
                             else
                                 url = mesg.url + "?nocache=#{Math.random()}"
-                                window.open(url,'_blank')
                                 dialog.find(".salvus-file-printing-link").attr('href', url).text(pdf).show()
+                                if is_subdir
+                                    subdir_texfile = p.head + "/#{base}-sagews2pdf/tmp.tex"
```
williamstein: I tried this code a few times in my dev project, and it failed in every case in various ways.  It never worked.   E.g., when the sagews is in a subdirectory the generated file wasn't even in that subdir.  When the sagews was in the home directory, this subdir_texfile ended up with a / at the front.

haraldschilly:
regarding your two comments:

* the first one sounds like you didn't update smc_pyutil/sagews2pdf
* I've dealt with the homedirectory case

still, I've found another problem, so I'll set it to needs review when it is ready

williamstein:
This works beautifully.

PR #962, opened 2016-09-28
use nbviewer.jupyter.org to preview publicly accessible ipynb files

haraldschilly:
this is essentially using it in an iframe.

for testing with smc-in-smc, the nested project url doesn't work. there is a commented src link in the code to try it with a real path.


williamstein:
works; worrisome to depend on that -- would be better to run on our server or something in a container... but this is massively better than nothing!

Issue #862, opened 2016-09-02
jupyter bridge: wrong newlines with print vs. display of last object


haraldschilly:
When running this in  a sagews, I get a funny output. Probably just a wrongly timed or missing `\n`:

```
%sage
a3 = jupyter("anaconda3")
---
%a3
x=1
print("x = %s" % x)
x
```

![smc-jupyter-bridge-output-bug](https://cloud.githubusercontent.com/assets/207405/18202191/cc9a24f8-710e-11e6-9cb2-fc80011e8b7a.png)

DrXyzzy:
Problem is that, in order to get ANSI color to show, most text output from jupyter is wrapped in html `<pre>...</pre>`, and that automatically inserts either newline or space. Either choice will cause a problem sometime.

Proposed workaround for this bug is to bypass the html-ansi wrapper if output does not contain the ansi CSI (Control Sequence Introducer) "\1b[". There will still be cases that don't line up properly, but lots fewer of them.

DrXyzzy:
Done. Verified in production manually and passes sagews pytest suite:
```
collected 58 items
...
58 passed in 80.79 seconds
```

PR #969, opened 2016-09-30
fix newlines with jupyter bridge #862

DrXyzzy:
Reference: #862.

Added pytest cases for
- #862 
- anaconda3 jupyter bridge
- julia jupyter bridge
- sagemath jupyter bridge

Put sage & sage_server timing tests at start of default test run because they are most robust about initial state of sage_server.

To run just the new tests:
```
python -m pytest test_sagews_modes.py::TestJupyterModes
```


DrXyzzy:
Thanks for catching change to smc-project-sagews-dev.sh. Done.

Issue #972, opened 2016-09-30
the "Auto close brackets: automatically close brackets" in haskell mode is buggy


williamstein:
When enabled (which it is by default), typing
```
    "hello"
```
inserts a " suddenly at the very end (and only then).  Probably the best solution for this issue for now is just never enable Auto close brackets: automatically close brackets" for `.hl` haskell files.

This could also get reported to Codemirror upstream...

DrXyzzy:
Reported it here: codemirror/CodeMirror#4277.

PR #975, opened 2016-09-30
force auto close brackets false for .hs and .lhs files; #972

DrXyzzy:


src/smc-webapp/editor.coffee
```diff
@@ -1624,6 +1624,12 @@ class CodeMirrorEditor extends FileEditor
                 evaluate_key = "Shift-Enter"
             extraKeys[evaluate_key] = (editor)   => @action_key(execute: true, advance:true, split:false)
 
+        # turn off autoCloseBrackets for haskell files SMC #972
+        if misc.filename_extension_notilde(filename) == "hs" || misc.filename_extension_notilde(filename) == "lhs"
```
williamstein: It's a little ugly to call `misc.filename_extension_notilde(filename)` twice (also `or` is same as `||` in coffeescript).  Does `misc.filename_extension_notilde(filename) in ['hs', 'lhs']` work?  Could you just make the line `autoCloseBrackets       : opts.auto_close_brackets` be 
```
autoCloseBrackets       : opts.auto_close_brackets and (misc.filename_extension_notilde(filename) not in ['hs', 'lhs'])
```
?

PR #983, opened 2016-10-02
webapp+raw_server: tell browsers to always download a file if it should be downloaded

### LINKED ISSUE #982: fix the download button to always download a file
If you try to download a file there is a button and a link. The button may or may not work depending on browser, file type, etc.,  Using buttons to initiate a download robustly is really only possible using (crappy) Flash, I think.

haraldschilly:
fixes #982

testing:
* try it with a pdf or svg file
* webapp recompilation
* restart the project to pick up the modified `smc-project/raw_server.coffee` file

### DIFF
--- a/src/smc-project/raw_server.coffee
+++ b/src/smc-project/raw_server.coffee
@@ -56,6 +56,12 @@ exports.start_raw_server = (opts) ->
             base = "#{base_url}/#{project_id}/raw/"
             winston.info("raw server: port=#{port}, host='#{host}', base='#{base}'")
 
+            raw_server.use base, (req, res, next) ->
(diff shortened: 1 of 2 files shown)


haraldschilly:
possible addition: avoid forcing downloads for html pages, by not adding `?download` to the url for their URLs. (came up in #984)

williamstein:
Tested.  This does work for png files now (etc.), though it didn't work (with [this sort of exception](https://github.com/gorhill/uBlock/issues/2029)) when my ad blocker (ublock) extension was enabled.  But we obviously can't workaround extensions.  So seems good to me.

Issue #328, opened 2015-12-13
Need easy way to get the raw url of a file


timothyclemansinsea:
I'm trying to view HTML page I'm editing and can't remember how to get the raw url. It would be nice if there was a button.

williamstein:
Click the checkbox then Download. It has a link.

On Sunday, December 13, 2015, timothyclemansinsea <notifications@github.com>
wrote:

> I'm trying to view HTML page I'm editing and can't remember how to get the
> raw url. It would be nice if there was a button.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/sagemathinc/smc/issues/328>.
>


-- 
Sent from my massive iPhone 6 plus.

williamstein:
I got it again in support this week

haraldschilly:
@timothyclemansinsea what do you think about my #984 idea?

williamstein:
Fixed!

PR #984, opened 2016-10-02
webapp/file listing: add download buttons that also support right-click → copy url

haraldschilly:
This is a simple solution for all issues related to #328 

Note, that it also support right-click and "copy url", since it is in the href of an anchor.

What I haven't done is to think much about the styling (we could add a hover text), nor if this also shows up for public file listing (would it work there?). Still, I like it, it's very minimal and it doesn't add much clutter.

It also definitely conflicts with my recent PR #983 since it moves that raw URL creation into a separate function ... merging is trivial, though.

### DIFF
--- a/src/smc-webapp/project.coffee
+++ b/src/smc-webapp/project.coffee
@@ -595,11 +595,14 @@ class ProjectPage
             # unfortunately, download_file doesn't work for pdf these days...
             opts.auto = false
 
-        url = "#{window.smc_base_url}/#{@project_id}/raw/#{misc.encode_path(opts.path)}"
+        url = @download_href(opts.path)
         if opts.auto
             download_file(url)
         else
             window.open(url)
+
+    download_href: (path) =>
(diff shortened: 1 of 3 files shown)


timothyclemansinsea:
@haraldschilly I like the simplicity.

timothyclemansinsea:
@haraldschilly one issue I have with this is if say html documents are being forced to download even when go to url in a separate tab

haraldschilly:
Hmm, maybe we should make paths ending in htm and html a special case and not force them to be downloaded? (I.e. just leave the content type as it is or undefined and then it will show up)

haraldschilly:
Ok yes, just checked, handling html pages as not being downloaded makes sense, but this is actually something for #982 and not here -- i.e. right now, without #982, the html page shows up normal in the new tab with this patch.

williamstein:
screen shot 2016-10-04 at 12 10 53 pm

PR #987, opened 2016-10-04
fixed blue info file dropdown button

simonluu:
issue #284 The mobile version was using the same heights as the desktop version, so increased the size.


williamstein:
Rewrite this to not duplicate code. Eg., define an object style outside and set style.height='45px' only in the mobile case.

williamstein:
1. Though I described the problem as "MOBILE" in the ticket, the problem isn't really mobile, but when responsive mode changes the button sizes when changing the screen to be narrow.   Thus looking at IS_MOBILE is wrong, and doesn't even fix the problem on desktop, and even introduces problems  on a mobile device with high resolution or zoomed out.

2. Hardcoding 45px is lazy and error prone.  Instead you should be adding the appropriate class to the button to match what is done to make the other buttons larger in the first place.

src/smc-webapp/editor.coffee
```diff
@@ -1147,8 +1147,11 @@ class exports.Editor
             # responsive mode
             @project_page.destroy_sortable_file_list()
             width = "50%"
+            $("#file_info_button").addClass("btn-lg")
```
williamstein: 1. The selector `$("#file_info_button")` changes file_info_button EVERYWHERE, not just in this particular document.  You might want `@element.find("#file_info_button")`, except that...

2. The whole idea of CSS responsive mode is that the size of the button is determined by CSS entirely rather than javascript.  What you're doing here is just directly fighting against how CSS is meant to work.   Please see http://getbootstrap.com/css/#responsive-utilities

src/smc-webapp/r_misc.cjsx
```diff
@@ -1159,7 +1160,10 @@ EditorFileInfoDropdown = rclass
             @render_menu_item(name, icon)
 
     render : ->
-        <DropdownButton style={marginRight:'2px'} id='file_info_button' bsStyle='info' title={<Icon name='info-circle' />} className='pull-left'>
+        if is_responsive_mode()
+            bs_style = "large"
+
```
williamstein: This again (as above) goes against the idea of responsive design, which lets css do the work...

(Also, you are using the variable name `bs_style` but you should call it `bs_size`... but of course don't even do any of this.)

haraldschilly:
@johnjeng in `feature.coffee` is a simple `is_responsive_mode` check which analyzes an empty div in the dom (see bottom of `index.jade`). however, no idea if this is maybe dead code …

src/smc-webapp/r_misc.cjsx
```diff
@@ -35,6 +35,7 @@ Combobox    = require('react-widgets/lib/Combobox')
 misc        = require('smc-util/misc')
 immutable   = require('immutable')
 underscore  = require('underscore')
+{IS_MOBILE} = require('./feature')
```
williamstein: This shouldn't be here -- it's not used anymore.

src/smc-webapp/r_misc.cjsx
```diff
@@ -1159,9 +1160,14 @@ EditorFileInfoDropdown = rclass
             @render_menu_item(name, icon)
 
     render : ->
-        <DropdownButton style={marginRight:'2px'} id='file_info_button' bsStyle='info' title={<Icon name='info-circle' />} className='pull-left'>
-            {@render_menu_items()}
-        </DropdownButton>
+        <div>
```
williamstein: This code duplication is clearly bad and in no way necessary.  Just make a function `render_dropdown_button`, which takes as input bsSize and className, and call it twice in succession. Then there is no *code* duplication, but we still get two things in the DOM eventually, and CSS works fine... and we can still maintain this code.

I'm serious about bouncing this, because it's important to get basic easy things like this right, so it's possible to do more complicated programming later.

williamstein:
Nice!

PR #988, opened 2016-10-05
fixed cancel and x buttons in mobile new cell sagews modal

simonluu:
issue #986 cm is undefined when clicking cancel or x, changed it to @codemirror and works.

### DIFF
--- a/src/smc-webapp/sagews.coffee
+++ b/src/smc-webapp/sagews.coffee
@@ -802,8 +802,8 @@ class SynchronizedWorksheet extends SynchronizedDocument2
                 if result
                     @_handle_input_cell_click0(e, mark)
                 else # what the user really wants...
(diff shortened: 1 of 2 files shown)


src/smc-webapp/sagews.coffee
```diff
@@ -802,8 +802,8 @@ class SynchronizedWorksheet extends SynchronizedDocument2
                 if result
                     @_handle_input_cell_click0(e, mark)
                 else # what the user really wants...
-                    cm.focus()
-                    cm.setCursor({line:mark.find().from.line+1, ch:0})
+                    @codemirror.focus()
```
williamstein: Instead of what you've done here, you should do `cm = @focused_codemirror()`

williamstein:
This looks good...

Issue #483, opened 2016-03-23
settings→file sort is not respected in default file sorting


haraldschilly:
user report: "In SMC, under preferences, i have chosen the "default file sort" to be "by name", but it doesn't work. Files are always listed by date."

johnjeng:
Looks like this used to work and then for some reason there had to be a temporary solution that allowed sorting the files but never read any data from user's settings. Additionally, it seems that these settings aren't event being saved correctly to the database.
Though I could be completely wrong on the last bit, I'm still not entirely clear on how interfacing with the database works.

PR #996, opened 2016-10-06
fixed file sort to work with default file sort

simonluu:
issue #483 
The default file sort under user settings now sorts the file listing by the selected option.


williamstein:
Nice work!

Issue #63, opened 2015-08-17
make show_identifiers work correctly in Sage Worksheets in SMC.


williamstein:
See 

  https://cloud.sagemath.com/projects/4a5f0542-5873-4eed-a85c-a18c706e8bcd/files/support/2015-08-16-162239-show-identifiers.sagews

Doing

     sage.misc.session.state_at_init = dict(globals())

mostly fixes things.

Reported by saad khalid on sage-support.

PR #999, opened 2016-10-07
show_identifiers() #63

DrXyzzy:
Reference: #63 

Fix was tricky because
- I was only able to set sage.misc.session.state_at_init after first code block has executed
- sage.misc.session.init() is not called until first invocation of show_identifiers()
- need to set sage.misc.session.state_at_init before first user cell executes or risk missing variable assignments in that cell
- worksheet server adds 'smc', 'salvus', 'require', and sometimes other identifiers to session namespace - I made the requirements choice of removing these before returning show_identifiers() because user can't shadow them anyway.

Pytest cases have been added for show_identifiers(), reset(), save_session(), load_session(). To run just the new tests:
```
python -m pytest test_sagews.py::TestIdentifiers
```

### DIFF
--- a/src/smc_sagews/smc_sagews/sage_salvus.py
+++ b/src/smc_sagews/smc_sagews/sage_salvus.py
@@ -2969,6 +2969,8 @@ def reset(vars=None, attached=False):
     sage.misc.reset.reset_interfaces()
     if attached:
         sage.misc.reset.reset_attached()
+    # reset() adds 'pretty_print' and 'view' to show_identifiers()
+    exec('sage.misc.session.state_at_init = dict(globals())',salvus.namespace)
 
 reset.__doc__ += sage.misc.reset.reset.__doc__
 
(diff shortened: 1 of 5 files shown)


williamstein:
I wanted to just give this a positive review (for your birthday) but quickly found a bug. Try doing
```
sage = 1
show_identifiers()
```
This fails because `show_identifiers()` uses eval...

I think it would be better to just copy over the source of show_identifiers from `sage.misc.session.show_identifiers` rather than doing `eval('sage.misc.session.show_identifiers()',salvus.namespace)`, due to these subtle issues.

Also, I wrote `sage.misc.session.show_identifiers` in the first place, and if you look you'll see it's about 3 lines of code total, including functions it calls... 

Here's the original function it was based on: https://magma.maths.usyd.edu.au/magma/handbook/text/51#429

DrXyzzy:
Setting `sage`. I didn't see that coming. Yet not using `eval()` somehow pulled in the wrong context for `show_identifiers()`, even with salvus namespace. I puzzled over that for more hours than I want to admit.

Reimplement it is.

DrXyzzy:
This workaround - b59b339 - came to me, uninvited. I added a test case for assigning to `sage`, too. Still want the rewrite?

williamstein:
> Still want the rewrite?

I'll merge this, but let's consider it technical debt, since I don't like these eval, which make me nervous.  But this is a solid workaround, and there are many other important BUGS to squash!

PR #997, opened 2016-10-08
chatrooms: on mobile have a submit button below message when message is being edited

timothyclemansinsea:
I tried to edit a message on mobile only to discover I couldn't do so because mobile doesn't have shift-enter.

### DIFF
--- a/src/smc-webapp/editor_chat.cjsx
+++ b/src/smc-webapp/editor_chat.cjsx
@@ -404,7 +404,6 @@ Message = rclass
                 @props.actions.set_editing(@props.message, false)
 
     save_edit: ->
-        console.log('saving edit')
         mesg = @refs.editedMessage.getValue()
         if mesg != @newest_content()
             @props.actions.send_edit(@props.message, mesg)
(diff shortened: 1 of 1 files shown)


src/smc-webapp/editor_chat.cjsx
```diff
@@ -431,6 +443,17 @@ Message = rclass
             </div>
         </Col>
 
+    is_user_editing: ->
+        other_editors = @props.message.get('editing').remove(@props.account_id).keySeq()
+        current_user = @props.user_map.get(@props.account_id).get('first_name') + ' ' + @props.user_map.get(@props.account_id).get('last_name')
```
williamstein: This is a potential weird definition of current_user... which seems to not even be used at all in this function?

src/smc-webapp/editor_chat.cjsx
```diff
@@ -431,6 +443,17 @@ Message = rclass
             </div>
         </Col>
 
+    is_user_editing: ->
```
williamstein: In fact, this is_user_editing function isn't called anywhere.  Just delete it.

williamstein:
I made a few line comments.

williamstein:
Please resolve the merge conflict.

williamstein:
This is impossible to merge.  I would recommend just trying to redo what you did in the new version of the code.

williamstein:
Or maybe @simonluu could move this code into the new chat rewrite.

simonluu:
see #1018

PR #992, opened 2016-10-08
the "Upgrades you have applied to projects" in https://cloud.sagemath.com/settings/upgrades includes entries with no upgrades applied to them

timothyclemansinsea:
I think the problem is it includes probably everything that has ever been upgraded and then downgraded too.  Fixing this is probably nearly trivial -- just recognizing 0 versus undefined...

### DIFF
--- a/src/smc-webapp/projects.cjsx
+++ b/src/smc-webapp/projects.cjsx
@@ -576,9 +576,10 @@ class ProjectsStore extends Store
         v = {}
         @get('project_map').map (project, project_id) =>
             upgrades = @getIn(['project_map', project_id, 'users', account_id, 'upgrades'])?.toJS()
-            for upgrade of upgrades
-                if upgrades[upgrade] > 0
+            for upgrade,val of upgrades
+                if val > 0
                     v[project_id] = upgrades
+                    break
         return v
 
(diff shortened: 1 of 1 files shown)


src/smc-webapp/projects.cjsx
```diff
@@ -576,8 +576,9 @@ class ProjectsStore extends Store
         v = {}
         @get('project_map').map (project, project_id) =>
             upgrades = @getIn(['project_map', project_id, 'users', account_id, 'upgrades'])?.toJS()
-            if misc.len(upgrades)
-                v[project_id] = upgrades
+            for upgrade of upgrades
+                if upgrades[upgrade] > 0
+                    v[project_id] = upgrades
```
williamstein: Shouldn't you add a `break` line right directly below `v[project_id] = upgrades`.  Otherwise, it'll waste time doing `v[project_id] = upgrades` a whole bunch of times for no reason.

You could also do it like this (so use the val part of `of`):
```
for upgrade,val of upgrades
    if val > 0
        v[project_id] = upgrades
        break
```

williamstein:
@timothyclemansinsea can you make the trivial change I suggest?

Issue #291, opened 2015-11-24
In +New page, don't allow creating a file with no extension at all without first asking for user confirmation


williamstein:
If a user clicks +New, then hits return, they get a basically useless file.  People *often* do this, expecting a sage worksheet, and get very confused that SMC is broken.  "Code won't run!"

I would bet that this might account for a significant amount of lost users.

In addition we could also make it so that if a user presses shift+enter when editing a file with empty extension, they get a message that they should instead create a sage worksheet.

johnjeng:
Do people still often do this? Does anyone type a name expecting a file with that exact name?

williamstein:
I can't see how this could not still be a problem.

johnjeng:
I dunno, people used to type sage into the quick terminal and I would have expected that to also still happen but when we checked, it didn't.

williamstein:
This is I think now **the most popular support request**.  People are constantly creating plain text files (no extension) and getting confused that they aren't Sage worksheets.   I'm promoting this to high priority.   E.g., the latest support request a few minutes ago: 

> There isn't even an option to run this code, and the top of the screen is telling me that I can no longer do anything in Sage without paying a subscription fee. What's up with this?

PR #1003, opened 2016-10-08
Added a warning message when creating a file with no extension

simonluu:
issue #291 
When the user creates a file with no extension, a warning message shows up asking if they are sure they want to create a file with no extension.

There were many different ways that the alert message could be implemented, I tried a modal, and this warning message below. Not sure which one is preferred. So I went with this one.


williamstein:
1. Change the sentence "Warning: You tried to create a file with no extensions. Are you sure you want to create a file with no extensions?" to "Warning: Are you sure you want to create a file with no extension?"

2. Change the `[Yes]  [No]` buttons to `[Create file]  [Cancel]` with Create file green (do it like in change password in account settings).

src/smc-webapp/project_new.cjsx
```diff
@@ -272,6 +273,33 @@ ProjectNew = (name) -> rclass
             on_error     : on_error
             switch_over  : true
 
+    keydown : (e) ->
+        if e.keyCode==13 and @refs.project_new_filename.getValue().split(".").length < 2
+            @setState(warning : true)
+            ReactDOM.findDOMNode(@refs.project_new_filename.refs.input).disabled = true
+
+    accept_file : ->
```
williamstein: What?  Why not just call create_file directly?  This is bizarre.

src/smc-webapp/project_new.cjsx
```diff
@@ -272,6 +273,33 @@ ProjectNew = (name) -> rclass
...
+        @create_file()
+
+    decline_file : ->
+        @setState(warning : false)
+        ReactDOM.findDOMNode(@refs.project_new_filename.refs.input).disabled = false
```
williamstein: This makes me very uncomfortable -- smells like a hack?

simonluu: The ReactDOM.findDOMNode(@refs.project_new_filename.refs.input)? I'm trying to get the input so I can disable the input while the warning message is open because I think it would confuse users if you can change the file name while the warning message is open. ReactDOM.findDOMNode(@refs.project_new_filename) only gets the form-group div, and not the actual input. 

There is probably another way to get the input but I'm not sure how.

src/smc-webapp/project_new.cjsx
simonluu: Ooops. Changing.

src/smc-webapp/project_new.cjsx
simonluu: oh..I'm dumb

src/smc-webapp/project_new.cjsx
```diff
@@ -239,8 +240,11 @@ ProjectNew = (name) -> rclass
         e.preventDefault()
         if @state.filename[@state.filename.length - 1] == '/'
             @create_folder()
-        else
+        else if @refs.project_new_filename.getValue().includes('.')
```
johnjeng: Needs changing to new way of getting input values. Search code for `<FormControl`. It's what's replaced `<Input` and no longer has a `getValue()` function.

src/smc-webapp/project_new.cjsx
```diff
@@ -239,8 +240,11 @@ ProjectNew = (name) -> rclass
...
             @create_file()
+        else
+            @setState(warning : true)
+            @refs.project_new_filename.getInputDOMNode().disabled = true
```
johnjeng: Should this just always be disabled when `@state.warning == true` ?

src/smc-webapp/project_new.cjsx
```diff
@@ -272,6 +276,25 @@ ProjectNew = (name) -> rclass
             on_error     : on_error
             switch_over  : true
 
+    decline_file : ->
+        @setState(warning : false)
+        @refs.project_new_filename.getInputDOMNode().disabled = false
```
johnjeng: This leads me to believe my hunch from before is true. Just change the ref component to render based on `@state.warning` instead of disabling it via the DOM.

johnjeng: There are examples in the code of `clear_and_focus_input()`

src/smc-webapp/project_new.cjsx
```diff
@@ -271,6 +274,25 @@ ProjectNewForm = rclass ({name}) ->
             on_error     : on_error
             switch_over  : true
 
+    decline_file : ->
+        @setState(warning : false)
+        ReactDOM.findDOMNode(@refs.project_new_filename).disabled = false
```
haraldschilly: Hmm, I think you can delete that line. One core idea of react is, that such attribute state data is wrapped in components. I.e. that "project new filename" already has a `disabled    = @state.warning` property. Setting it right here directly isn't good. I.e. the setState from above should be enough, right?

src/smc-webapp/project_new.cjsx
```diff
@@ -271,6 +274,25 @@ ProjectNewForm = rclass ({name}) ->
...
+        ReactDOM.findDOMNode(@refs.project_new_filename).focus()
```
haraldschilly: that's ok again, at least I don't know a better way how to focus it.

src/smc-webapp/project_new.cjsx
```diff
@@ -238,8 +239,10 @@ ProjectNewForm = rclass ({name}) ->
         e.preventDefault()
         if @state.filename[@state.filename.length - 1] == '/'
             @create_folder()
-        else
+        else if ReactDOM.findDOMNode(@refs.project_new_filename).value.includes('.')
```
haraldschilly: hmm, shouldn't this be just `@state.filename.includes('.')`? every change of the value in the box is automatically stored in the state, and it's also the state that's used for creating the filename. i.e. look into line 230:

```
create_file : (ext) ->
        @props.actions.create_file
            name         : @state.filename
```

src/smc-webapp/project_new.cjsx
```diff
@@ -238,8 +243,10 @@ ProjectNewForm = rclass ({name}) ->
         e.preventDefault()
         if @state.filename[@state.filename.length - 1] == '/'
             @create_folder()
-        else
+        else if @state.filename.includes(".")
```
simonluu: One thing I thought about doing was checking if the @state.filename has one of the new_file_button_types extensions. If it does have it, then it would create the file. If it doesn't it would ask if the user wants to create a file with an invalid extension.

williamstein: no -- There are tons of extensions that aren't explicitly listed anywhere (or known to me), but are not "invalid extensions".  

Maybe a warning would be good though, just saying we have no special support for editing that file type.  Maybe.

williamstein:
I'm going to merge this, but not until rewriting it yet again, since it's bad.

- don't just call the state variable "warning" -- that is not descriptive.  Call it "extension_warning".
- Determine if there is an extension using the function for that in misc.
- Remove english mistake by actually putting "Warning: Are you sure you want to create a file with no extension?"
- Fixed bad spacing between buttons and sentence.

Issue #1012, opened 2016-10-10
saving sagews: unable to read json message in worksheet: [empty string?]


haraldschilly:
There are several warnings when saving a sagews, appearing in sagews.coffe: `SynchronizedWorksheetCell.output`, where "x" seems to be an empty string. Could be a symptom of a more severe issue ...

haraldschilly:
Via debugging I can see, that this warning always happens when there is a cell without any output.

williamstein:
fixed by pr 1013

PR #1013, opened 2016-10-10
webapp/sagews: discarding empty strings when deserializing cell output -- related to #1012

haraldschilly:
see #1012 ... not sure if this is a fix at all, it does at least handle all the corner cases and finishes fixing [this change from a while ago](https://github.com/sagemathinc/smc/commit/5215886c8352504f199e00cc402960dff7269518#diff-9d796ba4eb777e83c8f17a2169b073dbL2416)

So, the point is, this PR should go in asap, but I'm not sure if the story behind why #1012 came up in the first place is fixed by that.

### DIFF
--- a/src/smc-webapp/sagews.coffee
+++ b/src/smc-webapp/sagews.coffee
@@ -2415,11 +2415,15 @@ class SynchronizedWorksheetCell
 
     output: =>
         v = []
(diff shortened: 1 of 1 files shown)


williamstein:
Looks perfect to me. Will merge, test again, and make live asap.

PR #1023, opened 2016-10-11
Course: Fix self closing assignment/folder adder

### LINKED ISSUE #835: In the course editor, the add searchbox (for assignments and handouts) often instantly disappears which drives users crazy
For some reason after some time, the results box closes itself after a little bit.

My guess is that something is rerendering it? Or data is getting updated somewhere incorrectly.

johnjeng:
Fixes #835 

Tested in Chrome, Firefox, and Edge on Windows 10

### DIFF
--- a/src/smc-webapp/course/common.cjsx
+++ b/src/smc-webapp/course/common.cjsx
@@ -10,6 +10,8 @@ misc = require('smc-util/misc')
 
 {ErrorDisplay, Icon, Space, TimeAgo, Tip, SearchInput} = require('../r_misc')
 
+immutable = require('immutable')
+
 # Move these to funcs file
 exports.STEPS = (peer) ->
     if peer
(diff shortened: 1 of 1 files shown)


haraldschilly:
Hmm, when testing this after merging with master, I'm getting `DOMChildrenOperations.js: Uncaught TypeError: Failed to execute 'insertBefore' on 'Node': parameter 1 is not of type 'Node'.` errors. I'm checking if this is caused by this or if this is actually in master...

src/smc-webapp/course/common.cjsx
```diff
@@ -497,7 +490,11 @@ exports.FoldersToolbar = rclass
                     @setState(add_is_searching:false, err:err, add_search_results:undefined)
                 else
                     filtered_results = @filter_results(resp.directories, search, @props.items)
-                    @setState(add_is_searching:false, add_search_results:filtered_results)
+                    if filtered_results.length == @state.add_search_results.size
+                        merged = @state.add_search_results.merge(filtered_results)
+                    else
+                        merged = immutable.List(filtered_results)
+                    @setState(add_is_searching:false, add_search_results:merged)
```
haraldschilly: I don't understand what's going on here, probably missing the big picture.

haraldschilly:
ok, interestingly, I had to compile just with "webpack" without watch or map -- now it works. Hence :+1: for this from my side...

Issue #1006, opened 2016-10-08
the chat message format is replicated in 3 different files, massively increasing the chances of documentation becoming out of sync with reality


williamstein:
There's a (nice) documentation block beginning with `Chat message JSON format:` in `editor_chat.cjsx`, `side_chat.csjx`, and `smc_chat.cjsx`.  Fix this.  The docs should be in one file only.

PR #1035, opened 2016-10-12
fixed chat documentation to appear in only editor_chat

simonluu:
#1006


johnjeng:
Solve these conflicts first. I think this PR is moot now.

PR #1043, opened 2016-10-15
Do something toward react/chat refactoring to clean up code a little and avoid too much rendering

williamstein:
This is based on top-down-react, which is why the diff with master is huge.  I'll pick this up and finish it later.  It was just some work on the plane.


williamstein:
I'll return to this later...

Issue #195, opened 2015-10-29
HTML table output


rbeezer:
Not a bug, but it would be nice if `html.table(list)` was implemented

```python
File: /projects/..../.sagemathcloud/sage_salvus.py 
Source: 
def table(self):
  raise NotImplementedError, "html.table not implemented in SageMathCloud yet"
```

A good first step would be for the function to have another argument, so the not-implemented error would be raised in the situation where it would normally be used.

williamstein:
Still true: 

https://cloud.sagemath.com/projects/4a5f0542-5873-4eed-a85c-a18c706e8bcd/files/support/2016-09-23-120437%20issue%20195.sagews

williamstein:
Also there is a normal table command, which doesn't work in typeset mode: https://github.com/sagemathinc/smc/issues/89

haraldschilly:
Rolled out and verified, that the examples from https://github.com/sagemathinc/smc/pull/1055 work.

PR #1055, opened 2016-10-17
sagews: implementing html.table (issue #195)

haraldschilly:
food for testing, issue #195 

```
rows = [['a', 'bäßbb', 'cö'],[100,2,3], [4,5,60]]
html.table(rows, header=True)

import numpy as np
html.table(np.random.rand(5,5))

m = matrix([[1,2.2,3], [2,3,1]])
html.table(m)
```


haraldschilly:
merging this is low risk, since that method didn't exist before (was a NYI error)

Issue #1002, opened 2016-10-08
chat: there is no way to know somebody is trying to video chat unless they explicitly write that in the chatroom


williamstein:
Imagine you click the video chat button.  Nobody else in the room has a clue you're trying to initiate video chat. 

I'm counting this as a bug, since it's a regression compared to what we had before and kind of makes chat unlikely to work...

A fix could be to edit the line of the chatroom that has the chat id and also put info about who is trying to use the chatroom.  Another approach would be to add a new chatline saying when a person tries to open the video chat...  A third approach would be if there is some sort of API for getting info about how many people have the video chat room open and just display that next to the video chat button.

williamstein:
This is now fixed in the new sidechatright branch.

PR #1058, opened 2016-10-17
added users to video chat to show how many users are in video chat

simonluu:
issue #1002 
Added a list of users to video event that shows the users currently in the video chat.


src/smc-webapp/editor_chat.cjsx
```diff
@@ -291,6 +301,15 @@ class ChatActions extends Actions
         w.document.write('</body></html>')
 
         w.addEventListener "unload", () =>
+            video = (@store.get('video')?.toJS()) ? {}
```
johnjeng: I was concerned about this conversion from Immutable but looks like `@save_shared_video_info` expects an object, so in this context it's fine I guess.

haraldschilly:
I fear this is too outdated. I just tried to test it, and I get this:

```
Uncaught TypeError: Cannot read property 'size' of undefined
    at Constructor.render_video_chat_off_button 
```

The point is, that I've merged with master prior to testing. Please also do this when fixing this, i.e. `git fetch --all` and then after checking out this branch, `git merge master`.

src/smc-webapp/editor_chat.cjsx
```diff
@@ -287,6 +297,15 @@ class ChatActions extends Actions
...
+            users = video.users
+            # deletes the user from the user list
+            index = users.indexOf(account_id)
+            if index != -1
+                users.splice(index, 1)
+                video.users = users
```
williamstein: This line `video.users = users` doesn't do anything, since `video.users` and `users` are just references to the same object.

simonluu:
I made some changes, not sure if this is something along the lines of what you were suggesting I do. If it is are there things I can improve on?

williamstein:
Already done in the sidechatright branch

Issue #1039, opened 2016-10-13
clean up rendering of side chat


williamstein:
Similar to #1038, but for the side chat.  Make similar clean up of the rendering code for messages.

PR #1059, opened 2016-10-17
cleaned up rendering of side chat

simonluu:
issue #1039 
Did similar thing to the chat rooms and cleaned up rendering of side chat.

### DIFF
--- a/src/smc-build/build.py
+++ b/src/smc-build/build.py
@@ -470,6 +470,7 @@ class BuildSage(object):
             "patch_banner",
             "patch_sage_env",
             "user_site",
+            "install_sagemanifolds",
             "install_sloane",
             "install_projlib",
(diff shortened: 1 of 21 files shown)


haraldschilly:
@simonluu well, this is also ok, I guess, but maybe I can convince you to tidy this up little bit more before I merge? E.g. I don't like that there is this code duplication with `show_history` and `hide_history`. Rather, I would make just *one* method `toggle_history` and make its content depending on `@state.show_history`. Next up, just have one method `toggle_history_side_chat(bool)` instead of two, and call it like `toggle_history_side_chat(@state.show_history)`.

simonluu:
I fixed the code duplication you suggested, also went into the chat version of the show_history/hide_history and changed that too.

Issue #959, opened 2016-09-28
sagews2pdf should hide html css styles for jupyter kernel outputs


haraldschilly:
printing

    sh = jupyter('bash')
    %sh
    date

or something like that shows all the html css styling in the pdf. 

note: interestingly, the inserted css happens with the jupyter command...

haraldschilly:
that's solved, since the PR #1063 removed an initialization line which was inserting all that html for no good reason.

besides that, I bet it is not possible to decide between "wanted" html output and "unwanted" one. If there is ever another example coming up, we should open a new issue with a concrete case where this does happen.

PR #1063, opened 2016-10-18
sagews/jupyter: removing conv.convert from jupyter kernel initialization.

haraldschilly:
makes no sense (?) and if at all, it should be called with full=False. issue #959

test with the example given in the #959 ticket

### DIFF
--- a/src/smc_pyutil/smc_pyutil/sagews2pdf.py
+++ b/src/smc_pyutil/smc_pyutil/sagews2pdf.py
@@ -198,6 +198,7 @@ site = 'https://cloud.sagemath.com'
 
 import argparse, base64, cPickle, json, os, shutil, sys, textwrap, HTMLParser, tempfile, urllib
 from uuid import uuid4
+from pprint import pprint
 
 def escape_path(s):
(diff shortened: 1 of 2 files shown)


haraldschilly:
@DrXyzzy please take a look at this! that one line I removed is probably just some leftover and since there is "full=False" for each cell, that issue doesn't happen in reality.

DrXyzzy:
The change causes many tests in `src/smc_sagews/smc_sagews/tests/test_sagews_modes.py` to fail. I think it relates to initial DOCTYPE. Doing a few more tests.

haraldschilly:
Hmm, good argument regarding the tests. The point for me is, that inserting this whole html document just for an empty string makes no sense to me. It has no impact on the other html and ansi-ascii snippets inserted later.

DrXyzzy:
Looks good in all my automated and (modified) tests. I will update the pytest suite so all tests pass as soon as this change propagates.

haraldschilly:
ok, great! thank you!

DrXyzzy:
The one-line code change here is included in my PR #1082

haraldschilly:
closing, is part of #1082

Issue #378, opened 2016-01-18
Suggestion: new button, called "Duplicate", in "File" operations


jpedroan:
I see buttons: Download, Delete, Rename, Move, Copy, Share.

I was used to do "copy/paste" of a file, using the keyboard and producing "File (copy).tex", for example.  Then rename it and this is fast.

In SMC, one has to do "Copy" but to another directory,  then rename the file,  and finally move it back.
It is not possible to do "Save As"   also.

Currently, as far as I know, to solve this
* I have to open a terminal and do "cp File.tex to OtherFile,tex".
* get inside the directory of the file and use the top-right command line to write the "cp File.tex to OtherFile,tex"
* all of them need one step more that "copy/paste" ou "Save As".

williamstein:
@rolles -- you might want to do this.

williamstein:
Making bug since this is a basic function whose absence constantly confuses and frustrates people.

williamstein:
It's been done.

PR #1072, opened 2016-10-18
webapp/files: adding duplicate functionality ...

haraldschilly:
addresses issue #378 by extending the rename feature and adding a suggestion function for new names

testing: check files named like `file-11.sagews` or `test.md` and click on the duplicate button.

### DIFF
--- a/src/smc-util/misc.coffee
+++ b/src/smc-util/misc.coffee
@@ -2,7 +2,7 @@
 #
 # SageMathCloud: A collaborative web-based interface to Sage, IPython, LaTeX and the Terminal.
 #
-#    Copyright (C) 2014, William Stein
+#    Copyright (C) 2014 -- 2016, SageMath, Inc.
 #
 #    This program is free software: you can redistribute it and/or modify
 #    it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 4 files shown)


src/smc-util/misc.coffee
```diff
@@ -2,7 +2,7 @@
 #
 # SageMathCloud: A collaborative web-based interface to Sage, IPython, LaTeX and the Terminal.
 #
-#    Copyright (C) 2014, William Stein
```
williamstein: yep -- we should do this.  I made a ticket for doing this everywhere: https://github.com/sagemathinc/smc/issues/1095

williamstein:
Code looks good.  It works.  Obvious missing thing -- what about directories?   I'll try now to see if that was an oversite or is hard...?

williamstein:
Adding directory functionality was trivial. DONE.  (1 line) 
 
Next obvious thing missing or wrong, which is done right by OS X (say), is that it should definitely NOT auto-suggest a filename if there is already a file in that directory with that name.   And given the suggested name, it's quite likely this will happen.  Also, we know the directory listing, so it should be easy to deal with this.  I'll try.

williamstein:
NOT auto-suggest a filename -- ok, that was easy and done.

Issue #1074, opened 2016-10-18
clean up rendering of history and preview


simonluu:
History and Preview rendering got messed up when #1038 updated the messages.

haraldschilly:
I've merged #1075 since it is clearly better to have that in sooner than later. I don't know about this ticket  ... sounds very vague to me. Should we close it and you open up one with more specific goals? Just comment when you read this, no hurry ...

simonluu:
Ok, that sounds good. I was just trying to mention this because of the change of the messages.

PR #1075, opened 2016-10-18
fix history and preview rendering

simonluu:
cleans up history and preview rendering #1074

### DIFF
--- a/src/smc-webapp/editor_chat.cjsx
+++ b/src/smc-webapp/editor_chat.cjsx
@@ -64,10 +64,6 @@ editing   : immutable.Map
 
 ###
 
-###
-This file is all the parts that are similar between side_chat.cjsx (The side chat rooms of all the files) and smc-chat.cjsx (The sagemathcloud chat room)
-###
-
 # standard non-SMC libraries
 immutable = require('immutable')
 {IS_MOBILE, isMobile} = require('./feature')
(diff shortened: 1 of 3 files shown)


src/smc-webapp/editor_chat.cjsx
```diff
@@ -388,35 +384,36 @@ exports.render_markdown = render_markdown = (value, project_id, file_path) ->
...
             s: value
             wrap: ['<span class="smc-editor-chat-smiley">', '</span>']
         value = misc_page.sanitize_html(value)
-        author = user_map.get(history_author[date]).get('first_name') + ' ' + user_map.get(history_author[date]).get('last_name')
-        if history[date].trim() == ''
+        author = user_map.get(objects.author_id)?.get('first_name') + ' ' + user_map.get(objects.author_id)?.get('last_name')
+        if value.trim() == ''
             text = "Message deleted "
         else
             text = "Last edit "
```
haraldschilly: I might sound pedantic, but maybe text should just be "Edited " ?

haraldschilly:
hmm, all this works, although in general, I find it a bit odd that the history is below (and outside) of the chat message itself. maybe somehow nested inside would be better, but anyways, that's not the goal here. I'm also not sure what preview changes it addresses, but this can be merged.

PR #1077, opened 2016-10-19
Remove user's upgrades when they delete a project

### LINKED ISSUE #282: when deleting a project, remove all upgrades automatically

johnjeng:
Fixes #282. Only removes the current user's contributions. A collaborator may not have even wanted the project deleted! 

Also add the ability for map_limit to take a number as the second argument. Seems like a natural extension to me.


src/smc-util/misc.coffee
```diff
@@ -1458,10 +1458,15 @@ exports.map_diff = (a, b) ->
     return c
 
 # limit the values in a by the values of b
+# or just by b if b is a number
 exports.map_limit = (a, b) ->
     c = {}
-    for k, v of a
-        c[k] = Math.min(v, (b[k] ? Number.MAX_VALUE))
+    if typeof b == 'number'
```
williamstein: Please change to `typeof(b)` -- I don't like to use the no parens coffee-script syntax except for callbacks.

src/smc-webapp/project_settings.cjsx
```diff
@@ -827,6 +827,9 @@ HideDeletePanel = rclass
             <Row>
                 <Col sm=8>
                     {@delete_message()}
+                    {<Alert bsStyle="info" style={padding:'8px', marginTop:'4px'} >
+                        <b>New -- </b>This will also clear all upgrades from this project. Undeleting the project will not automatically restore them.
```
williamstein: Why does it say `New --`?    Think of the psychology.  Most SMC users are new or don't remember how things worked before anyways, and nowhere else in SMC do we ever say "heh this is new!".    Also, if we did (which we don't) it should only happen once or for a limited time.  There's nothing worse than explicitly saying something is new and seeing it 10 years later in your codebase.  I bet there is stuff like that in Sage.

johnjeng:
Opened #1101 for (3).

In this PR, help menu looks like:
delete menu

Also changed `Adjust Quotas` button to stick around (consistent with rest of smc `...` style buttons).
adjust quotas

src/smc-webapp/project_settings.cjsx
```diff
@@ -518,6 +528,37 @@ HideDeletePanel = rclass
...
+            onClick = @show_delete_conf
+            disabled = is_expanded
+
+        <Button bsStyle='danger' style={float: 'right'} onClick={onClick} disabled={disabled}>
+            <Icon name='trash' /> {text}
+        </Button>
+
+    render_expanded_delete_info : ->
+        <Well>
+            <Alert bsStyle="info" style={padding:'8px'} >
```
williamstein: Sorry to nitpick: can you change this to only get shown if the users has **actually applied** upgrades to this project?  If they haven't (usually the case?), it would be safe to just not show the alert, I think.

Issue #964, opened 2016-09-28
the avatar counter is still wrong


williamstein:
<img width="546" alt="screen shot 2016-09-28 at 3 03 54 pm" src="https://cloud.githubusercontent.com/assets/1276278/18934106/d6e7eec6-858c-11e6-92bc-db97a6adcede.png">

Above it says +2 but should say +1...

haraldschilly:
I merged #1081 since I think it fixes this. We should keep an eye on that, though, maybe some corner case comes up.

PR #1081, opened 2016-10-19
fixed avatar count

simonluu:
issue #964 
fixed the avatar count to show the right number of people


src/smc-webapp/profile.cjsx
```diff
@@ -307,7 +307,7 @@ UsersViewing = rclass
         if all_users.length > num_users_to_display
             rest =
                 <span style={fontSize:"small", cursor:"pointer", marginBottom:"4px", marginRight:"10px"}>
-                    {"+ #{all_users.length-4}"}
+                    {"+ #{all_users.length-5}"}
```
haraldschilly: hmm, maybe I'm blind because of all the pasta, but shouldn't the `5` better be `num_users_to_display` ?

Issue #976, opened 2016-10-01
have sagews %r mode use jupyter ir kernel


DrXyzzy:

DrXyzzy:
Done

Issue #240, opened 2015-11-09
Set R's working directory path to the one of sagews


haraldschilly:
When running `%r` in a sagews which is in a subdirectory, the R interpreter is still running in the home directory. Call R's `setwd()` appropriately when starting it up.

williamstein:
Still broken
```
%r
getwd()
```
https://cloud.sagemath.com/projects/4a5f0542-5873-4eed-a85c-a18c706e8bcd/files/support/2016-09-23-102034-issue240.sagews

Seems pretty damned serious too.

williamstein:
Switching to the Jupyter R kernel would just fix this:
```
ir = jupyter('ir')
```
---
```
%ir
getwd()
```

and it works fine.

williamstein:
@DrXyzzy  any thoughts about just switching to the R jupyter kernel...?

DrXyzzy:
+1.

haraldschilly:
this works, ticket is merged, closing bug

PR #1082, opened 2016-10-19
%r with jupyter kernel

DrXyzzy:
Re: #240 and #976.

Much of old %r supporting code is commented out in sage_salvus.py starting around line 2035. I wasn't sure we want to get rid of all of it.

I don't think it's worth fighting with R and Jupyter to make plots come out in .svg. Suggest referring users to the usual R workaround of saving plot to a file if non-png format is desired.

### DIFF
--- a/src/smc_sagews/smc_sagews/sage_jupyter.py
+++ b/src/smc_sagews/smc_sagews/sage_jupyter.py
@@ -170,6 +170,7 @@ def _jkmagic(kernel_name, **kwargs):
         # buffering for %capture because we don't know whether output is stdout or stderr
         # until shell execute_reply message is received with status 'ok' or 'error'
         capture_out = ""
+        capture_mode = not hasattr(sys.stdout._f, 'im_func')
 
         # handle iopub messages
         while True:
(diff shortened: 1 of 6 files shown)


DrXyzzy:
This PR includes the code change for #1063 as well as pytest changes to make all tests in src/smc_sagews/smc_sagews/tests work.

haraldschilly:
Hmm, despite trying to kill the sage processes and so on, two graphics tests fail everytime: `tests/test_graphics.py F.F`

first

```
            else:
                assert typ == 'json'
                if 'html' in mesg:
                    assert want_html
                    want_html = False
                    print('got html')
                else:
>                   assert want_name
E                   assert False
```


second: 

```
        typ, mesg = sagews.recv()
        assert typ == 'json'
        assert mesg['id'] == test_id
>       assert 'html' in mesg
E       assert 'html' in {'done': False, 'event': 'output', 'id': 5, 'stdout': 'AFTER PLOT 0'}
```

DrXyzzy:
Thank you for the review.
## 1. Tests
> Hmm, despite trying to kill the sage processes and so on, two graphics tests fail everytime: tests/test_graphics.py F.F

All tests pass repeatedly in my project at this commit  https://github.com/DrXyzzy/smc/commit/3c2c41fbd2f39104987737297c12e69b1f8a7ca6 and git status says all changes are committed and pushed to this PR. Does that help?
```
~/smc/src/smc_sagews/smc_sagews/tests$ python -m pytest
============================= test session starts ==============================
platform linux2 -- Python 2.7.10, pytest-3.0.2, py-1.4.30, pluggy-0.3.1
rootdir: /projects/ccd4d4a4-29a8-4c39-85c2-a630cb1e9b6c/smc/src/smc_sagews, inifile:
collected 67 items 
 
test_00_timing.py ....
test_graphics.py ...
test_sagews.py ................
test_sagews_modes.py ............................................
 
========================= 67 passed in 109.18 seconds ==========================
```
## 2. svg
> Regarding plotting, does it use svg now or not?

The ir jupyter kernel always sends .png so there is no chance to select .svg option; the `default_image_format` setting is ignored. I can improve comments & docstring to indicate that. Default text and image formats were a start at generic config framework for all jupyter kernels.

I think limitation to png (unless plot is written to a file) is part of the R display ecosystem, not only jupyter. That's why it may be reasonable to leave it to the user if R image output other than .png is desired.

PR #1083, opened 2016-10-19
Fix free project warning to update automatically.

### LINKED ISSUE #1031: the message "WARNING: This project runs on a free server without internet access" (or its variation) MUST auto-update on store changes
It turns out you have to close and open the project for "WARNING: This project runs on a free server without internet access" to update.  Watching users, this causes **enormous confusion** precisely at the point when we don't want it to.  It's especially confusing because everything else in SMC automatically updates in response to changes.

johnjeng:
Fixes #1031 Again another problem with showing stale data since the func pointer itself won't update. 

Issue made for the general problem #1084.

### DIFF
--- a/src/smc-webapp/project_page.cjsx
+++ b/src/smc-webapp/project_page.cjsx
@@ -137,6 +137,8 @@ FreeProjectWarning = rclass ({name}) ->
 
     reduxProps :
         projects :
+            # get_total_project_quotas relys on this data
+            # Will be removed by #1084
             project_map : rtypes.immutable.Map
             get_total_project_quotas : rtypes.func
         "#{name}" :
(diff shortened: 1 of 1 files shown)


haraldschilly:
Well, I think that's fine, but I have no clue how to test it. I finally found a way how to enable this customization to reply `{"commercial":true}` but there are still no additional UI elements.

johnjeng:
As an admin account, go to the admin settings and change site settings. 
screenshot_2016-10-20_10-32-52

haraldschilly:
oh wow! thank you! who knew that that's there ... my value was "true" by the way, which is obviously not ok. "yes" works.

williamstein:
I added a trivial commit that explicitly says the options are 'yes' or 'no':  53626897dffe25ac0e7059437b92aeef4478270e

williamstein:
1. Your solution to this problem seems to be to make it so any time anything about **any project** changes, then every project has their page updated.  For someone like me with 600+ projects, that's **constantly**.     

2. Just adding a dependency via reduxProps on something, with zero comment or explanation at all, seems to be begging for trouble later.  Somebody will see that and just delete it.

So: (1) need a comment, and (2) needs a shouldComponentUpdate to check whether the part of the project_map that matters actually changed.

PR #1087, opened 2016-10-20
Fix assignment expansion state

johnjeng:
Persist the expansion state of assignments in the course management. 

Implemented with an `immutable.Set`. Does not propagate to other clients.

Without this change, expansion state of assignments is loss between unmounts which makes things like grading a lot slower.


haraldschilly:
code works for me, so that's good, and it's also done in a nice way. However, for the sake of playing devils advocate when reviewing: that same functionality should also be done for the handouts panel ... I know that it is not related to assigning grades, but it's a silly asymmetry to not have that there, too.

johnjeng:
Updated to do the same thing for Students and Handouts.

PR #1089, opened 2016-10-20
Change all found instances of load_target to correct call

### LINKED ISSUE #1086: load_target called when it shouldn't be
I just hit this code in projects.cjsx:

```
# Should not be necessary/here for React/Redux
exports.load_target = load_target = (target, switch_to) ->
    throw "Fix this! #{arguments}"
```

Called from misc_page line 143: `require('./projects').load_target(target, not(e.which==2 or (e.ctrlKey or e.metaKey)))`

**Obviously** we need to do a full text search through the source for load_target (and the other similar functions) and fix all this.   This clearly just got forgot due to not keeping a good enough todo list?

johnjeng:
Fixes #1086

### DIFF
--- a/src/smc-webapp/misc_page.coffee
+++ b/src/smc-webapp/misc_page.coffee
@@ -140,7 +140,7 @@ $.fn.process_smc_links = (opts={}) ->
                         else if opts.project_id and opts.file_path?
                             # realtive to current path
                             target = "#{opts.project_id}/files/#{opts.file_path}/#{decodeURI(target)}"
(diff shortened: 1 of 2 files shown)


williamstein:
Also, should delete this from projects.cjsx -- I'll do it:
```
# Should not be necessary/here for React/Redux
exports.open_project = open_project = (opts) ->
    throw "Fix this! #{arguments}"

# Should not be necessary/here for React/Redux
exports.load_target = load_target = (target, switch_to) ->
    throw "Fix this! #{arguments}"
```

Issue #1019, opened 2016-10-10
typing `open *.foobar` in a terminal creates a file literally named `*.foobar`


williamstein:
ug.

williamstein:
Fix will just involve changing this file: https://github.com/sagemathinc/smc/blob/master/src/smc_pyutil/smc_pyutil/smc_open.py

williamstein:
So I think `open *.foobar` should just do NOTHING if *.foobar doesn't match anything, by the way.

PR #1102, opened 2016-10-22
Issue1019termopen

DrXyzzy:
Fix for #1019.

Also has fixes for pytest files smc_sagews/test_graphics.py and smc_sagews/test_sagews_modes.py

### DIFF
--- a/src/smc_pyutil/smc_pyutil/smc_open.py
+++ b/src/smc_pyutil/smc_pyutil/smc_open.py
@@ -50,6 +50,10 @@ def main():
         print "Opens each file (or directory) in the Sagemath Cloud web-based editor from the shell."
         print "If the named file doesn't exist, it is created."
     else:
+        # exception if any args have ? or * after glob expansion
+        ff = filter(lambda x: any(wc in x for wc in ['?', '*']), sys.argv[1:])
+        if len(ff) > 0:
+            raise ValueError('No matching files for wildcard argument(s)', ff)
         process(sys.argv[1:])
 
 if __name__ == "__main__":
--- a/src/smc_pyutil/smc_pyutil/smc_open.py
+++ b/src/smc_pyutil/smc_pyutil/smc_open.py
@@ -1,6 +1,6 @@
 #!/usr/bin/python
 
-import json, os, sys, re
+import json, os, sys
 
 home = os.environ['HOME']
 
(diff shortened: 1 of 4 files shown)


DrXyzzy:
Oops just saw #1019 comment about doing nothing. If there are multiple command line args, it could skip any that are unmatched globs and open the rest. I'm changing the PR now to do that.

src/smc_pyutil/smc_pyutil/smc_open.py
```diff
@@ -50,7 +50,8 @@ def main():
         print "Opens each file (or directory) in the Sagemath Cloud web-based editor from the shell."
         print "If the named file doesn't exist, it is created."
     else:
-        process(sys.argv[1:])
+        # ignore args that have ? or * after glob expansion
+        process([p for p in sys.argv[1:] if re.search('[\?\*]', p) is None])
```
haraldschilly: I think at that point, it's better to keep the `sys.argv[1:]` and do that filtering up there in the `process` function itself. Then, all the logic parts are kept in one place. Despite that, I'm wondering about more explicitly checking if such a file exists. There are also other shell-expansions, e.g. `{2...3}` or something like that, which aren't covered.

williamstein: Harald has a good point about other shell expansions, so that needs to be looked up.

Also, using a regular expression not only makes I think the code harder to read and work on, it turns out it is also much, **much** slower:  https://cloud.sagemath.com/projects/4a5f0542-5873-4eed-a85c-a18c706e8bcd/files/support/2016-10-22-issue1019.sagews

Of course the speed doesn't matter -- in both cases (with or without re) it is fast.  However, regular expressions are often notoriously hard to read and lead to weird bugs, whereas doing, e.g., `'?' in p` is very clear and easy to read code... and there is no loss in speed (in fact, it is over 10x faster).

williamstein: Yep, "what if my filename really has a question mark or star?" -- that was my "IMPORTANT" point 3 below too.

DrXyzzy:
Changed code per comments above. Please review again.

williamstein:
The PR branch had a bunch of other stuff in it related to jupyter kernels.  I copy pasted the two relevant lines out into some new code, added some docs, and also limited the total number of files to open at once: https://github.com/sagemathinc/smc/commit/e677c2bff0d9d36e26f55480b9de8cb0ae722ece

Closing.

Issue #1078, opened 2016-10-19
trying to open a file in the log that no longer exists results in a confusing and non-professional error message


williamstein:
<img width="1415" alt="screen shot 2016-10-18 at 9 35 41 pm" src="https://cloud.githubusercontent.com/assets/1276278/19505835/ed61fb12-957a-11e6-9440-d2e77620e879.png">

williamstein:
The message above is "You should not be here! editor-smc/src/smc-webapp/client_browser.coffee".

johnjeng:
In this case do we want to create a new file or give a more professional/helpful error message?

williamstein:
> In this case do we want to create a new file or give a more professional/helpful error message?

You could put a professional message, then have a button to create the file, and another to cancel/close.

haraldschilly:
I've pushed 284c71f595c5d986ee245c to fix the strings, but the problem is definitely deeper. I.e. I have seen that string before, but I wasn't able to replicate it. If I'm reading the code correctly, maybe the actual problem is that `open_files` is sometimes not defined? So maybe, this is actually some race condition?

Also, more general, I think files opened from the log shouldn't be created. There should be exactly this message that this file no longer exists there. (I'm always rolling with my eyes, if I accidentally create an empty file by clicking in the log)

williamstein:
The problem is gone, but now it creates the file.  @haraldschilly maybe you want to create a new issue to handle that, though I don't consider this a flat out bug anymore (it was the behavior before).  Anyways, now there is no "nonprofessional error message".

PR #1114, opened 2016-10-24
webapp: better error message when file does not exist or `open_files` is not defined

haraldschilly:
issue #1078

This improves the error message a little bit, but the underlying problem is definitely deeper


williamstein:
Different approach here: https://github.com/sagemathinc/smc/commit/0969631e442d89b5fff4a37004c593b71f772308

Issue #445, opened 2016-02-29
Sage worksheet underscore for last output does not work when typeset_mode is true


williamstein:
Reported by: Kelly McCoun

Example: https://cloud.sagemath.com/projects/4a5f0542-5873-4eed-a85c-a18c706e8bcd/files/support/2016-02-29-underscore_and_typeset_mode.sagews

Basically _ stops getting updated after doing 

    %typeset_mode true

haraldschilly:
Ok, I know why this happens. With typeset true, the usual displayhook isn't run. I.e. inside the normal one, also the `_system_sys_displayhook` is called, which is (I guess) [the one from ipython](https://github.com/ipython/ipython/blob/master/IPython/core/displayhook.py) ... at least, that contains code to update and initialize the underscores. I also noticed today for the first time, that double and triple underscores aren't supported, so that's yet another issue.

However, I don't really know how to fix this. I don't think it's wise to fix it by simply replicating parts of the functionality. I'm looking into a way how to call that displayhook directly without creating any output.

haraldschilly:
ok, forget my last comment ... I thought there is ipython involved. pushing PR for this now...

haraldschilly:
and as an additional note: the referenced fix just adds feature parity between typeset mode on and off. what isn't updated in both cases are those filtered output types (i.e. plots, etc.).
we could easily also update `_` for plots ... although I don't think that makes much sense in practice.

williamstein:
@hsy fixed this today!

PR #1115, opened 2016-10-24
sagews: directly update the builtin "_" just like the normal system wide displayhook does

haraldschilly:
issue #445

testing is explained in the ticket, basically in a sagews: 

```
123
---
%typeset_mode true
---
321
---
_
[here should be 321, not 123]
```


williamstein:
nice -- works perfectly.

Issue #1117, opened 2016-10-24
adding collaborator or student when there is just one in the list doesn't work any more


haraldschilly:
tracked this down to this commit: https://github.com/sagemathinc/smc/commit/995f96e0#diff-1d4654a4776f0825cd0ac2f439629410R1035

likely, the same holds true for adding students in a course

PR #1119, opened 2016-10-24
one-click adding for single search result when adding a collaborator or student

haraldschilly:
issue #1117 

how to test:

1. search for a new collaborator and if there is just one name as a result, clicking on "add user" is enough (not clicking on the name to actually select it)

2. the same in the student panel of a course, i.e. search for a name and if there is just one result, clicking on the button is also enough


williamstein:
make sense; works perfectly

PR #1125, opened 2016-10-24
Update url on path change

### LINKED ISSUE #1105: more react rewrite fallout: the url on the top doesn't match the current directory
Go to files for any project with a folder `foo`, and type `foo` then enter in the upper left search box.   You'll then be in the folder `foo`, but the url won't have foo in it at the top.

johnjeng:
Whenever the path changes, update the URL

Fixes #1105


haraldschilly:
follow up on this is here: #1134

williamstein:
The fix caused a major bug.  Please try again and fix this right :-)

PR #1136, opened 2016-10-25
compute server: limit the number of parallel state updates

haraldschilly:
I think it's a sane move to limit this map a little bit, e.g. after restart or during high load. of course, if it takes longer than two minutes, we're in in any case in a gameover situation.


williamstein:
Looks good to me.  Usually number is 1 (I looked in logs) but on startup can be higher. This is safe.

Issue #1008, opened 2016-10-09
Side chat exhibits odd redraw behavior on window resize


johnjeng:
When you resize the window and side chat is open, shrinking vertically can cover the input area and shrinking horizontally is very slow. Opening and closing the side chat redraws as expected with any new window size.

williamstein:
raising priority -- this is actually really bad and easy to reproduce.

flex model!

williamstein:
I'm working on a complete rewrite of side chat and editor placement/sizing.   Some screenshots:

<img width="1440" alt="screen shot 2016-11-06 at 11 44 32 am" src="https://cloud.githubusercontent.com/assets/1276278/20040987/88629352-a416-11e6-9e0a-82df47b8c038.png">
<img width="1438" alt="screen shot 2016-11-06 at 11 44 12 am" src="https://cloud.githubusercontent.com/assets/1276278/20040988/88641826-a416-11e6-997c-cbc2040da55c.png">
<img width="1440" alt="screen shot 2016-11-06 at 11 44 07 am" src="https://cloud.githubusercontent.com/assets/1276278/20040989/88643996-a416-11e6-9e90-9b0142fc57d1.png">

williamstein:
Fixed right in the sidechatright branch.

PR #1150, opened 2016-10-27
Side chat exhibits odd redraw behavior on window resize fix

simonluu:
This fixes issue #1008 
Shrinking horizontally changes the width of the code mirror which fixes the horizontal redraw behavior.

Shrinking vertical sets the height of the chat log so that it fits the window height.


src/smc-webapp/editor.coffee
```diff
@@ -1545,6 +1545,12 @@ class CodeMirrorEditor extends FileEditor
         # show the codemirror editors, resizing as needed
         @_show_codemirror_editors(cm_height, width)
 
+        # changes the width on resize change
+        width_resize = () =>
+            width = @element.find(".salvus-editor-codemirror-chat-column").offset().left
+            @_show_codemirror_editors(cm_height, width)
+        $(window).resize(width_resize)
```
williamstein: Wow. Just wow.  Every single time the editor is shown you're creating a new function and assigning a new event handler.   It's never freed.  This is some seriously leaky code.

src/smc-webapp/syncdoc.coffee
```diff
@@ -130,15 +130,16 @@ class SynchronizedDocument extends AbstractSynchronizedDoc
...
 
         @_chat_redux_name = side_chat.render(@editor.project_id, @editor.chat_filename, @editor.chat_elt[0], redux, chat_height)
 
         height_resize = () =>
-            chat_height = $($(".salvus-editor-codemirror")[1]).css("height").split("px")[0] - 140
-            side_chat.render(@editor.project_id, @editor.chat_filename, @editor.chat_elt[0], redux, chat_height)
-        $(window).on("resize", height_resize)
+            chat_height = $(document).height() - $(".navbar").height() - $(".smc-file-tabs").height() - @element.find(".salvus-editor-codemirror-button-row").height() - 63
+            @_chat_redux_name = side_chat.render(@editor.project_id, @editor.chat_filename, @editor.chat_elt[0], redux, chat_height)
+        $(window).resize(height_resize)
```
williamstein: Same remark as above about leaks.  I see here that every time the chat window is shown it creates a new leak too -- that's some old code I didn't write, which needs to be excised.

If there is **any possible way** to do this resizing via CSS rather than Javascript, let's figure out how to do so and implement it.   If it really can only be done with javascript, don't leak.

simonluu:
I tried to use css vh units to make the height of the chat change based on the viewport but that didnt work because of the button bar.

williamstein:
I'll work on this.  It needs to get done properly, which may require some more nontrivial changes higher up.

PR #1151, opened 2016-10-27
Improve start/stop all student projects UI feedback

johnjeng:
Edit:
Removed `wait until` madness.  
Now tries for 5 minutes to keep all projects running or closed. 
Clicking the other option while this happens removes the timeout and interval of the previous action.

### After clicking `Start...`.

start

### After clicking `Start All`.

starting

### After clicking `Stop All` similar to above.
stopping

### DIFF
--- a/src/smc-webapp/users.cjsx
+++ b/src/smc-webapp/users.cjsx
@@ -32,20 +32,21 @@ immutable = require('immutable')
 # Register the actions
 class UsersActions extends Actions
     fetch_non_collaborator: (account_id) =>
-        salvus_client.get_usernames
-            account_ids : [account_id]
-            use_cache   : false
-            cb          : (err, x) =>
-                if err
-                    console.warn("ERROR getting username for account with id '#{account_id}'")
-                else
-                    obj = x[account_id]
-                    if obj?
-                        obj.account_id = account_id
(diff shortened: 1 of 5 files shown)


src/smc-webapp/users.cjsx
```diff
@@ -32,20 +32,21 @@ immutable = require('immutable')
...
-                    console.warn("ERROR getting username for account with id '#{account_id}'")
-                else
-                    obj = x[account_id]
-                    if obj?
-                        obj.account_id = account_id
-                        user_map = store.get('user_map')
-                        if user_map? and not user_map.get(account_id)?
-                            user_map = user_map.set(account_id, immutable.fromJS(obj))
-                            @setState(user_map : user_map)
+        if account_id?
```
williamstein: could you just do `return if not account?` or 
```
if not account_id?
    return
```
at the top?  Then the rest of the code doesn't have to be indented another level, and it is easier to read (instead of searching from  an else on the next page...)

williamstein:
referee discussion at https://cloud.sagemath.com/projects/369491f1-9b8a-431c-8cd0-150dd15f7b11/files/work/2016-10-27.sage-chat

williamstein:
OK, this is **really good**!!

Issue #501, opened 2016-04-08
button cutoff latex editor


haraldschilly:
The issues button is cut off, while the PDF and Build buttons are not visible.

I am currently using the latest version of Firefox.

![image](https://cloud.githubusercontent.com/assets/207405/14395058/caec6b32-fdcf-11e5-9ad7-ab8629016349.png)

williamstein:
Rewriting in react is would fix this...

williamstein:
Hi priority since this is really embarrassing, and we really don't want our latex editor to scare off users.

williamstein:
Question: does anybody have any clue how to actually replicate this?  I've seen it before, but I just tried for 6 minutes with chrome and firefox and can't replicate it.

haraldschilly:
I've noticed this yesterday several times, it's also in the "normal" md editor. So, that's not just latex... maybe there is just a reason that it is more easily happening there. I'll read that code for the preview iframe, maybe I see something...

haraldschilly:
just FYI, I know the problem ... fixing this also means to clean up some parts.

haraldschilly:
This commit is work in progress, I need a break since all this is a minefield.

williamstein:
Fixed!

PR #1157, opened 2016-10-28
webapp: md and latex editor render tweaks: preview no longer does not overlaps with buttons, fixing margin and offset computations, triggering recalc on latex building, some style cleanup.

haraldschilly:
issue #501

The core problem with #501 is, that some layout calculations are done too early, done when not needed (editor not shown), or not at all (when latex compiles, the spinner in the build button might cause a linebreak).  It also deals with issues like adding some negative top margins in the css, while adding positive top space via element styling. Despite that, it removes a non-existing button bar from the embedded pdf preview, makes the divider line more visible, and uses underscore's debounce instead some custom throttling.

I've tested it in chrome and firefox, with 100% and 150% zoom, also narrowing the window to trigger the responsive view for mobile. For me it looks all good, so I hope it's fine everywhere.

### DIFF
--- a/src/smc-webapp/_editor.sass
+++ b/src/smc-webapp/_editor.sass
@@ -142,7 +142,10 @@
   background-color : #eee
   overflow-x       : auto
   overflow-y       : scroll
-  border-left      : 2px solid #ddd
+
+.salvus-editor-pdf-preview-page-single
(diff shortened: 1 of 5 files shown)


williamstein:
Looks more than good enough for now to me

Issue #591, opened 2016-05-30
opening files not in the project's directory tree (home) no longer works (using a symlink also doesn't work)


williamstein:
Try using open on a file in /tmp or make a symlink -- either way it no longer works.

(It used to -- this must have broke with the sync rewrite.)

Reported by: Ben Salisbury

DrXyzzy:
Bug can be reproduced from .term command line with bash. I don't think it's in smc_open.py.

williamstein:
Raising priority, since I keep hitting this...

haraldschilly:
I tried to figure out what's going on, but I failed.

The error disappears when I add this

        if opts.path[0] == '/'
            # uses symlink to '/', which is created by start_smc
            opts.path = '.smc/root' + opts.path

in the constructor of `smc-util/syncstring.SyncDoc`. Then, the follow up error is in the hub about a missing key in the syncstrings table. I think it's just a matter of also fixing that derived key, or well, symlinks are maybe also dereferenced and hence those hashed keys no longer match?

error is this

```
Client(Ny40dSAViLHvAefoAAAC).user_query: user_query error: {"name":"ReqlNonExistenceError","msg":"Cannot perform pluck on a non-object non-sequence `null`.","frames":[0],"message":"
Cannot perform pluck on a non-object non-sequence `null` in:\nr.db(\"smc\").table(\"syncstrings\", {\"readMode\": \"outdated\"}).get(\"fd7e307e0d29879c8c050e48e9352816ea2174d4\").pluck(\"project_id\")\n^^^^^^^^^^^^^
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                    "}
```

haraldschilly:
ah, got it!!!

PR #1158, opened 2016-10-28
sync: dealing with file paths not inside the project

haraldschilly:
issue #591

this patch indirectly prefixes the filename for SyncDoc (and derived) objects with the symlink to the root directory in .smc

I think that's not the whole deal, since that sha1 sum comes up at more places, but at least `open /tmp/text.md` works.


williamstein:
This seems exactly right and was how I think things used to work.

Issue #852, opened 2016-08-30
Print Button Problems


johnjeng:
List of problems
- [x] Does nothing in Codemirror editor. ie. `file_name`, `thing.txt`, 
- [x] Entirely useless in html editor
- [x] Conversion from markdown to pdf is not great
  - lists don't work

williamstein:
Lists do work, but the markdown "standard" is different, e.g., it requires the nested lists to be done using asterisks.   It uses pandoc I think; I don't know if there is a better alternative in existence today -- search around!     In 2013 pandoc was probably the best choice.  Today?  I doubt it.

johnjeng:
This website seems to convert markdown pretty well. https://gitprint.com/

Looking in the [git repo](https://github.com/adamburmister/gitprint.com/blob/master/routes/index.js), it looks like he uses [markdown-pdf](https://github.com/alanshaw/markdown-pdf)

williamstein:
See also this feature request: https://github.com/sagemathinc/smc/issues/82

williamstein:
See https://github.com/sagemathinc/smc/pull/1160

haraldschilly:
I'll fix it, it's something many use and it should work or not work in a
well defined way.

williamstein:
Plan:
1. Make a list of editors that have a print button and the status of the implementation.
2. Make it non-buggy, either by explicitly eliminating the print button, or using a better approach to printing.
3. For those editors that should have printing, but don't, create enhancement tickets to implement printing, and do that after all bugs are fixed.

haraldschilly:
my status on this is, that I've collected "print" utility classes in a new "printing" module. That's necessary, because there are cross-cutting concerns between the editors (e.g. txt can be printed via pandoc like md or html, but it is not suitable for the specialized md-html editor).

based on that information, the UI only shows and arms the print button, if there is a known class for printing the file.

the status I've collected so far is the following:

* sagews → SagewsPrinter (calls this print_to_pdf on the backend, which is actually only for sagews ... could be refactored)
* tex → LatexPrinter (checks if the pdf exists, then opens it in a new tab)
* ['md', 'html', 'htm', 'rst', 'wiki', 'mediawiki', 'txt'] → PandocPrinter (calls pandoc, then opens pdf)

fixing the actual printing of specific document types (or even adding new printers) should be further more specific tickets.

haraldschilly:
[...]

PR #1160, opened 2016-10-29
webapp/latex: retrofitting a simple pdf-to-print functionality and fixing overlapping hints for the buttons on the right

haraldschilly:
I've no idea if there is a ticket for this, but these are two things that bugged me yesterday fixing the show function:
1. the tooltips on the right hand side overlap, so I move the to the top or to the bottom (also, consitent with the other small tooltips which are also at the bottom already!)
2. the print button doesn't disappear, because some "magic" somewhere resurrects it. Despite finding out what this is about, I've added a "print" feature to the "open new tab" functionality in the store.


src/smc-webapp/editor.coffee
```diff
@@ -1085,8 +1085,11 @@ class CodeMirrorEditor extends FileEditor
                 that.click_edit_button($(@).data('name'))
                 return false
 
+        # FIXME: removing the button here is pointless, since it will be added later again (somewhere)
         # FUTURE: implement printing for other file types
-        if @filename.slice(@filename.length-7) != '.sagews'
+        is_sagews = @filename.slice(@filename.length-7) != '.sagews'
```
williamstein: What's going on here?  

1. The name `is_sagews` suggess it should be `@filename.slice(@filename.length-7) == '.sagews'` instead of `@filename.slice(@filename.length-7) != '.sagews'`?

2. Can you use use filename_extension from misc, rather than the hardcoded length of 7, which is a little more dangerous code?

haraldschilly: oh well, I wasn't thinking much. before that, it was this `if @filename.slice(@filename.length-7) != '.sagews'` and that's how it is right now. I just wanted to extend this to .tex files without having a too long line. It's a bit odd, yes. Let me fix this nicely.

williamstein: I know -- I lazily wrote that crappy code -- just encouraging you to fix it...

That said, the variable name has to be `is_not_sagews`, right?

haraldschilly: yes, i wrote this using muscle memory. PR is updated ...

williamstein:
1. Printing is a mess still -- that's https://github.com/sagemathinc/smc/issues/852

2. This patch makes things worse in some ways.  E.g., on the live site without this if you open a "markdown" .md file and click print it just works fine.  With this patch, it still works, but it also pops up a notification saying "printing only works for sage worksheets". 

For now, the right fix should be to hide the print icon unless print is implemented. This patch doesn't seem to actually do that - it leaves the button and one sees that message about only sage worksheets having printing implemented (which is itself completely false).   Of course, this patch doesn't add that message, which is already there, but it suddenly make it a lot more prevalent.

haraldschilly:
well, as I feared, there are conflicts now that the other one is merged. I'll fix it.

src/smc-webapp/editor.coffee
```diff
@@ -574,9 +572,10 @@ templates = $("#salvus-editor-templates")
 #
 
 class FileEditor extends EventEmitter
-    constructor: (@project_id, @filename, content, opts) ->
+    # ATTN it is crucial to call this constructor in subclasses via super(@project_id, @filename)
```
williamstein: I would be more comfortable if this was enforced some way.  E.g.,  in the code that does `new ThingThatDervicesFromFileEditor`, have it check that constructor was called, or at least the reason for it being called holds, e.g., that @ext is defined.   That would be 2 lines of code and would make it impossible for anybody to ever have an editor work at all that doesn't call super properly.

src/smc-webapp/editor.coffee
```diff
@@ -342,10 +343,7 @@ PUBLIC_ACCESS_UNSUPPORTED = ['terminal','latex','history','tasks','course', 'cha
 exports.public_access_supported = (filename) ->
     ext = filename_extension_notilde(filename)
     x = file_associations[ext]
-    if x?.editor in PUBLIC_ACCESS_UNSUPPORTED
-        return false
-    else
-        return true
+    return x?.editor in PUBLIC_ACCESS_UNSUPPORTED
```
williamstein: For what it is worth, I think PUBLIC_ACCESS_UNSUPPORTED is no longer used or valid -- what is supported is determined by which editors are *registered* to support public access.  See the directory `public/`.

src/smc-webapp/editor.coffee
```diff
@@ -933,7 +932,7 @@ class CodeMirrorEditor extends FileEditor
         if opts.read_only
             @set_readonly_ui()
 
-        if @filename.slice(@filename.length-7) == '.sagews'
+        if misc.filename_extension(@filename)?.toLowerCase() == 'sagews'
```
williamstein: I have to wonder if filename_extension and filename_extension_notilde should just always do `toLowerCase()` ?

src/smc-webapp/editor.coffee
```diff
@@ -1089,17 +1088,21 @@ class CodeMirrorEditor extends FileEditor
...
-        for name in ['search', 'next', 'prev', 'replace', 'undo', 'redo', 'autoindent',
-                     'shift-left', 'shift-right', 'split-view','increase-font', 'decrease-font', 'goto-line', 'print' ]
+        button_names = ['search', 'next', 'prev', 'replace', 'undo', 'redo', 'autoindent',
+                        'shift-left', 'shift-right', 'split-view','increase-font', 'decrease-font', 'goto-line' ]
+
+        # if the file extension indicates that we know how to print it, show and enable the print button
+        if printing.can_print(@ext)
+            button_names.push('print')
+        else
+            @element.find("a[href=\"#print\"]").remove()
```
williamstein: Why not `'a[href="#print"]'` instead of the backslashes?

src/smc-webapp/editor.coffee
```diff
@@ -1282,42 +1302,45 @@ class CodeMirrorEditor extends FileEditor
...
                                 cb(err)
                             else
                                 pdf = _pdf
                                 cb()
                 (cb) =>
-                    salvus_client.read_file_from_project
-                        project_id : @project_id
-                        path       : pdf
-                        cb         : (err, mesg) =>
+                    redux.getProjectActions(@project_id).file_nonzero
```
williamstein: What does `file_nonzero` mean?  I've never heard of it, and the name is not very descriptive.  I really have no clue what it might be.  Can you change the name?

williamstein: Actions are not supposed to return anything.   They're supposed to set something in motion, whose impact is only via changes to the store.  There are a few critical cases (e.g., private internal methods) with callbacks.   Is there any way you can do whatever file_nonzero does but without a callback or move it somewhere that isn't an action?

ACTIONS:  take input and modify the store.  Do not return values. 

STORE: Functions/props compute stuff about the store, but never change the store. 

The Actions vs store business follows this design pattern:  http://martinfowler.com/bliki/CQRS.html

src/smc-webapp/editor.coffee
```diff
@@ -1282,42 +1302,45 @@ class CodeMirrorEditor extends FileEditor
...
                                 dialog.find(".salvus-file-printing-link").attr('href', url).text(pdf).show()
                                 if is_subdir
                                     {join} = require('path')
                                     subdir_texfile = join(p.head, "#{base}-sagews2pdf", "tmp.tex")
                                     # if not reading it, tmp.tex is blank (?)
-                                    salvus_client.read_file_from_project
-                                        project_id : @project_id
-                                        path       : subdir_texfile
-                                        cb         : (err, mesg) =>
+                                    redux.getProjectActions(@project_id).file_nonzero
```
williamstein: This is callback hell with ridiculous nesting.  Rewrite using async.series.

src/smc-webapp/editor.coffee
```diff
@@ -3126,6 +3157,10 @@ class JupyterNBViewerEmbedded extends FileEditor
         @iframe.maxheight()
 
 {HTML_MD_Editor} = require('./editor-html-md/editor-html-md')
+html_md_exts = []
+for ext, opts of file_associations
```
williamstein: What's this doing and why?

williamstein: Oh, this is the same as the (more readable) one-liner list comprehension:
```
html_md_exts = (ext for ext, opts of file_assocations when opts.editor == 'html-md')
```
This is like Python, but with paren instead of square brackets.  

That said, maybe exactly this snippet of code (suitably optimized) should actually be done automatically when registering editors.... something to worry about later.

src/smc-webapp/editor_latex.coffee
```diff
@@ -51,6 +53,11 @@ class exports.LatexEditor extends editor.FileEditor
             @update_preview()
             @spell_check()
 
+        @latex_editor.print = () =>
+            outfn = @filename.slice(0, @filename.length - 3) + "pdf"
```
williamstein: Change to use `misc.split...` ?

src/smc-webapp/editor_latex.coffee
```diff
@@ -657,10 +664,10 @@ class exports.LatexEditor extends editor.FileEditor
             elt.find(".salvus-latex-mesg-content").show().text(mesg.content)
         return elt
 
-
-    download_pdf: () =>
+    download_pdf: (print = false) =>
         redux.getProjectActions(@project_id).download_file
-            path : @filename.slice(0,@filename.length-3) + "pdf"
+            path : @filename.slice(0, @filename.length - 3) + "pdf"
```
williamstein: Use `misc.split...`

src/smc-webapp/editor.coffee
haraldschilly: I see, probably best to just remove that function.

src/smc-webapp/editor.coffee
haraldschilly: ok

src/smc-webapp/project_store.coffee
```diff
@@ -1258,6 +1127,159 @@ class ProjectStore extends Store
...
+                    cb         : (err, output) =>
+                        cb(err)
+        ], (err) =>
+            if err
+                cb(err)
+            else
+                cb(undefined, filename.slice(0,filename.length-3) + 'sagews')
+        )
+
+    convert_docx_file: (filename, cb) =>
```
williamstein: ACTIONS: 
  - they change state -- functions that do things and modify the store.  Do NOT return values.

STORE: 
  - stores state and has get'ers that return values.  These functions do NOT change state.

You just moved a bunch of functioniality to the store, but these are definitely not methods that belong in the store.  They could be actions if they do something, then store a result in the store.    However, these seem to just be traditional async callback functions.  There is no rule that **everything** needs to be an action or in the store.    In fact, there are rules on how the actions and stores work and what sort of functions they have, so we can easily reason about them.

williamstein: CQRS = command query responsibility segregation.

It means that you have to different things you communicate with (e.g., "Actions" and "Store").  One is used *only* for commands (=actions).  The other is used only to get out data (=store).  Segregation means they are completely separate.  It's the design pattern we're using here.  

If something doesn't fit into that at all, and you don't want to change the store or UI to reflect that, then that something should be somewhere else -- not in actions or store.

williamstein:
I just carefully read the whole PR, and everything looks fine except for putting a bunch of state changing code in the store.  Can you please change that.   I haven't actually tested yet.

src/smc-webapp/project_store.coffee
williamstein: > * not project related, maybe callback, maybe return value, doesn't change store → misc

It depends.  misc is pure javascript and generally just stuff that could be used on both the frontend and backend and doesn't have anything to do with sage_client communication.  It's utility functions.

> * project related, maybe callback, no return value, doesn't change store → store ? (e.g. `ensure_directory_exists` ?)

No.  The methods of the store should all be "pure" functions of the immutable js state.  There should be no callbacks ever, and nothing that should have any impact on any state anywhere.    In fact, @johnjeng is currently rewriting the store for the last week to ensure this (that's what we've been talking about in chat), motivated by his experience doing the top down rewrite.   The store is a container of state and the interface is "ways to get at that state".   (Exception: there is a method calld "wait", which calls a callback when a certain condition on the store holds.)

* project related, maybe callback, has return value, doesn't change store → somewhere else ?

Somewhere else, e.g,. a module scope function or class or function in client.coffee.   We want to minimize these as much as possible, as they are harder to reason about, but obviously sometimes they are necessary.    Example: synctables involve tons of such methods.  

* project related, no callback, no return value, changes store → action

Yes, that's the ideal case.    These can of course be asynchronous functions -- e.g., copying a file -- but rather than expressing what happens as it progresses via callback(s), the instead update the store as the run.   Then the UI can display what happens (or not). 

* project related, has callback, no return value, maybe changes store → action, but only for "internal" methods

Yes, to write clean code the non-public api for Actions can have all kinds of such "traditional" methods.

* project related, no callback, has return value, changes store → shouldn't exist at all

Yes, exactly.

Of course, nothing is perfect -- the above is our design pattern, what we should rewrite everything to do, etc., but mistakes have been made.

haraldschilly:
I've moved everything back to where it was, except for those two which check the existence of a path or directory. all other changes and fixes I made are still there, though. they're now in their own file and I named it "project_tasks". not sure if that's a good name, but at least they're separate and only have trivial dependencies. this should make it easy to refactor them later on.

williamstein:
Now looks great; I will test code soon and make live if good.

williamstein:
Interestingly, this PR breaks opening .tex files.   I get a back console traceback in the browser when attempting to open a tex file.

williamstein:
Wow, the problem is caused by `@local_storage("layout")` in the code, and that in the NEW branch I've been working on that can now have the value 0!  Amusing.  I'll fix this now.

haraldschilly:
oh ok, I just tried to see what's going on after merging with master, but there are no actual changes with the editors. I see, forward-breaking code, dangerous^2

williamstein:
I definitely made changes to editor: https://github.com/sagemathinc/smc/commit/d825795546245584896bb30230ee5e942529da3b

Issue #1168, opened 2016-10-31
sagews %javascript: make robust by catching exceptions


haraldschilly:
when evaluating arbitrary javascript in a sagews, exceptions corrupt the sagews rendering, because the evaluation stops.

in particular, the try block should be in sagews'.s `process_output_mesg` / `if mesg.javascript? and @allow_javascript_eval()` around those `eval( ... )`s.

Steps to reproduce:

1. new sagews file, and enter a few cells like that:

        %javascript
        print(window.foobar())
        ---
        1
        ---
        2
        ---
2. eval all of them (console will show `Error rendering worksheet TypeError: window.foobar is not a function`)
3. close/open sagews

I'm labeling it a bug, because we want to avoid broken sagews files.

PR #1174, opened 2016-10-31
webapp/sagews: catching javascript exceptions

haraldschilly:
and some cleanup (there is also a try-catch in there!) for _recieve_broadcast

issue #1168

testing:

```
%javascript
print(foo)
---
%coffeescript
print(bar)
---
123
```

before: evaluating the cells throws in javascript and closing/opening the worksheet shows corrupted cells. after: no problems, and one can see the error messages.

what I don't know, is how to test this `_recieve_broadcast`, I fixed the indentation which was all red, etc.

I'm also not sure, why e.g. `print(window.bar)` doesn't show any errors -- however, it also doesn't corrupt anything (maybe it's just undefined?). ah well, maybe one should also test `print(window.bar())` as a function call ...


williamstein:
undefined prints as nothing.  I'm fixing now. 

I ended up incorporating your code into another thing I'm about to push...

PR #1175, opened 2016-10-31
Add empty values to add payment FormControl

### LINKED ISSUE #1170: webapp/billing/credit card: uncontrolled vs. controlled
when I enter a credit card number in my dev project, I get this warning:

```
Warning: FormControl is changing an uncontrolled input of type text to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components
```

johnjeng:
Fixes #1170.

### DIFF
--- a/src/smc-webapp/billing.cjsx
+++ b/src/smc-webapp/billing.cjsx
@@ -180,10 +180,14 @@ AddPaymentMethod = rclass
         on_close : rtypes.func.isRequired  # called when this should be closed
 
     getInitialState : ->
-        new_payment_info : {name : @props.redux.getStore('account').get_fullname()}
-        submitting       : false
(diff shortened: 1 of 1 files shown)


src/smc-webapp/billing.cjsx
```diff
@@ -272,7 +276,7 @@ AddPaymentMethod = rclass
             return true
 
         x = info[name]
-        if not x?
+        if not x? or x == ""
```
williamstein: I think this is the same as `if not x`

williamstein: i changed this...

Issue #1184, opened 2016-11-02
salvus.html() won't open image file with absolute path


DrXyzzy:
Start with valid image test.png and copy to /tmp. This fails with broken image icon:
```
salvus.html('<img src="/tmp/test.png">')
```
but this succeeds:
```
salvus.file('/tmp/test.png')
```

DrXyzzy:
This works:
```
os.chdir('tmp')
salvus.html('<img src="test.png">')
```

And for image in home directory:
This fails
```
salvus.html('<img src="{0}">'.format(os.path.join(os.getcwd(),'test.png')))
```
but this suceeds
```
salvus.file(os.path.join(os.getcwd(),'test.png'))
```

haraldschilly:
ok. I made a patch that covers all this in a sagews: Are these all the cases to worry about?

```
plot(sin, figsize=(1,1)).save("test.png")
os.system("cp test.png /tmp/")
---
salvus.file("test.png")
---
salvus.html('<img src="{0}">'.format("test.png"))
---
fn_blob = salvus.file("test.png", show=False)
fn_blob
---
salvus.html('<img src="{0}">'.format(fn_blob))
---
salvus.html('<img src="{0}">'.format('/tmp/test.png'))
---
salvus.html('<img src="{0}">'.format(os.path.join(os.getcwd(),'test.png')))
```

haraldschilly:
since #1185 is merged, I assume we can close this.

PR #1185, opened 2016-11-02
webapp/sagews: when fixing the src of an image in some html, make sure to deal with absolute paths and images from the blobstore

haraldschilly:
issue #1184 -- @DrXyzzy please check if this is really fixing what you meant (and not breaking other cases)

### DIFF
--- a/src/smc-webapp/sagews.coffee
+++ b/src/smc-webapp/sagews.coffee
@@ -1,3 +1,24 @@
+##############################################################################
+#
+# SageMathCloud: A collaborative web-based interface to Sage, IPython, LaTeX and the Terminal.
+#
+#    Copyright (C) 2015 -- 2016, SageMath, Inc.
+#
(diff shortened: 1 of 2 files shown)


DrXyzzy:
Passes test for viewing the following with `salvus.html()` with img link:
- path in /tmp: `/tmp/image.png`
- full path in project: `'/projects/ccd4d4a4.../smc/src/data/projects/1e76ad83.../image.png'
- blob file u'/ccd4d4a4.../port/58594/blobs//tmp/image.png?uuid=d558...' created with `salvus.file(..show=False)`
- same blob file after unlinking original image.png

I'll run a few more tests but looks good so far.

DrXyzzy:
Image display checks out ok with:
- sage plot(cos(x),x,0,pi)
- %sh display < foo.png
- %octave x = -10:0.1:10;plot (x, sin (x));
- %ir kernel (not started via sage_salvus built-in mode - different defaults)

Issue #1095, opened 2016-10-20
(automatically) re-assign copyright to SageMath, Inc. for all SMC code.


williamstein:
So there will be a big patch with lots of things like this:

<img width="846" alt="screen shot 2016-10-20 at 2 54 05 pm" src="https://cloud.githubusercontent.com/assets/1276278/19579374/263d7ece-96d5-11e6-8b40-4bc8a204c143.png">

Everybody has agreed to this, and it will help avoid future IP issues.

PR #1189, opened 2016-11-03
mass fix copyright notices

timothyclemansinsea:
#1095

### DIFF
--- a/src/admin.py
+++ b/src/admin.py
@@ -3,7 +3,7 @@
 #
 # SageMathCloud: A collaborative web-based interface to Sage, IPython, LaTeX and the Terminal.
 #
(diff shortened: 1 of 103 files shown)


src/webapp-lib/policies/copyright.html
```diff
@@ -60,7 +60,7 @@
 
                     Deliver this Notice, with all items completed, to SageMath’s Designated Copyright Agent:<br><br>
 
-                    William Stein (Copyright Agent)<br>
+                    William Stein (Copyright (C) 2016, Sagemath Inc.
```
williamstein: This is the only one that would actually be visible to a user.  The parenthesis are wrong (no close).

Issue #1192, opened 2016-11-03
jupyter back-end: scale high res plots accordingly


haraldschilly:
follow up of #1178

DrXyzzy:
Done, with #1193.

PR #1193, opened 2016-11-03
jupyter/sagews: using image scaling factor to resize it in sagews

haraldschilly:
issue #1192

This is actually really weird. When the output renders, the image is drawn (at least) two times. The first time around it is all fine, the naturalWidth is like it should be, scaling works. It should just stop there!

But immediately afterwards, probably triggered by some sync mechanism, the output is rendered again, the naturalWidth is unfortunately 0 (I assume, the image is removed from the DOM but not loaded) and the calculated width is zero. Therefore, here in this code, there is this addition to do the calculation after it has been loaded the second time around.

### DIFF
--- a/src/smc-webapp/sagews.coffee
+++ b/src/smc-webapp/sagews.coffee
@@ -1316,6 +1316,16 @@ class SynchronizedWorksheet extends SynchronizedDocument2
         for x in a
             y           = $(x)
             src         = y.attr('src')
+            # see https://github.com/sagemathinc/smc/issues/1192
+            img_scaling = y.attr('smc-image-scaling')
+            if img_scaling?
+                y.get(0).onload = ->
+                    width     = this.naturalWidth
+                    factor    = parseFloat(img_scaling)
+                    if not isNaN(factor)
+                        new_width = width * factor
+                        $(this).css('width', "#{new_width}px")
+            # checking, if we need to fix the src path
             is_fullurl  = src.indexOf('://') != -1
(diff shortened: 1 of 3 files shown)


DrXyzzy:
my eval:

- I did not see the double-render problem at all with Firefox, and only saw it with Chrome with the e762fee commit. So it looks as if it is fixed.
- I found this which may be related, which suggests doing this.onload = null;after scaling: http://stackoverflow.com/questions/7699621/display-image-at-50-of-its-native-size/12607106#12607106
- I think what was bothering me about switch from style 50% to onload is that the images rescaled with onload appear about 20% larger than those rescaled with style. Setting `r.jupyter_kernel.smc_image_scaling=0.35` in the worksheet creates images about the same size as with `style="width: 50%; height: 50%;"`.
- latest commit be9b8f9 passes all 71 pytests and manual tests of sage, %sh, %octave

__LGTM__

williamstein:
Please support %octave plots also -- you're only fixing %r plots...

haraldschilly:
Well, what about making that a follow up issue? I don't know how to set the default resolution for octave, and how that is integrated into the octave kernel. I can see if I find it out, or if it requires patching the kernel wrapper.

williamstein:
> Well, what about making that a follow up issue? 

Absolutely yes -- in fact, I've merged this and made the code live.  It is perfect as is.  I just wanted to record adding octave support.

williamstein:
octave issue: https://github.com/sagemathinc/smc/issues/1202

Issue #1202, opened 2016-11-04
improve png sizing for sagews octave mode, like what was done for %r.


williamstein:

haraldschilly:
I've looked into the code itself to figure out what to do. The goal is to change the `plot_settings` dictionary before the very first evaluation. To do that, one either has to get an instance of the kernel (I've no idea how, I think that's rather impossible since they just talk via TCP with each other) or know how to set this when it is instantiated (which is a mystery to me, since maybe there is a way that meta kernels do get parameters via the kernel.json file, but I don't see where the octave kernel would respect that)

Relevant line to understand what's going on: https://github.com/Calysto/octave_kernel/blob/master/octave_kernel/kernel.py#L149
and later, that `res = self.plot_settings['resolution']` when it is rendered to png.

Hence, the reason why this is harder than `%r` is, that inside octave there are no print related settings. It's in the state of the jupyter kernel itself and that's wrapped nicely and tucked away...

DrXyzzy:
Above noted, thanks.

williamstein:
For what it is worth, the default plot resolution right now is already pretty high. If I make a random plot it fills up my entire screen.  The solution to this problem might simply be to take whatever the current output png's are and simply scale then down by 50% -- and done.

DrXyzzy:
Something changed very recently according to pytests. Octave plot is now sending png from jupyter kernel. A couple days ago it was sending svg.

DrXyzzy:
So it might be as simple as enabling onload scaling for octave at this point.

DrXyzzy:
FWIW text returned by the octave kernel changed, too. Used to collect multiple lines of output into a single message to client. Now they come back in separate messages.

haraldschilly:
@DrXyzzy  wow, interesting that so much has changed. this is all a side effect of #953 where I had to update this kernel and the metakernel library in the background.

PR #1203, opened 2016-11-04
1202octavepng

DrXyzzy:
Ref: #1202 

Recent config changes for octave in jupyter kernel already changed plot file format from svg to png. All that was needed is scale factor; 0.66 looks good.

Server-side pytests updated. `71 passed in 114.32 seconds`


williamstein:
Seems impossible that this could break anything -- tested on Hal's project.

PR #1209, opened 2016-11-05
webapp/sagews: display output spans as blocks, span-wrap images and improve multiline-selections of text output

haraldschilly:
this little change makes it easier to select multiline output. my test-case is to copy paste the whole output of `random_matrix(GF(2), 10, 10)`. works much better than it does right now.

I also checked it in firefox, and well, all other output types like images and md aren't touched by that. I also tested to switch to other codemirror themes...

### DIFF
--- a/src/smc-webapp/_editor.sass
+++ b/src/smc-webapp/_editor.sass
@@ -329,10 +329,11 @@
 //
 // this messes up codemirror themes
 
-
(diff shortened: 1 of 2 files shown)


williamstein:
screen shot 2016-11-05 at 2 11 53 pm

PR #1214, opened 2016-11-07
smc_pyutil: using python2-nosite to avoid interferences with ~/.local

### LINKED ISSUE #1213: make it so python code in ~/.local can't possibly mess up starting projects
It is silly that *python* code in ~/.local that conflicts with Sage (?) or anything else would have any impact on starting projects... since the project local hub server itself is written in Node.js...  It would be good to have a workaround to stop this problem.

haraldschilly:
this reverts 2f6bf61a9f5f and fixes #1213 

for reviewing: my test was to temporarily tell the setup.py to always change the executable (i.e. `if True: # ...`)  and then to see what it produces in my `~/.local` and that my local `smc-status` works.

`/usr/loca/bin/python2-nosite` is [this globally installed wrapper](https://github.com/sagemathinc/smc/blob/master/src/smc-build/smc-ansible/files/python2-nosite)

### DIFF
--- a/src/smc-build/smc-ansible/all-install.yaml
+++ b/src/smc-build/smc-ansible/all-install.yaml
@@ -36,8 +36,8 @@
           dest="{{ item[1] }}"
           owner=root group=root mode="u=rx,g=rx,o=rx"
     with_together:
-      - [ "files/python2-nosite",         "/usr/local/bin/python2-nosite"]
-      - [ "files/python3-nosite",         "/usr/local/bin/python3-nosite"]
+      - [ "files/python2-nosite",            "files/python3-nosite"]
+      - [ "/usr/local/bin/python2-nosite",   "/usr/local/bin/python3-nosite"]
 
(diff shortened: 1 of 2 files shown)


williamstein:
Quick question -- won't this break: 
  - the docker image
  - anybody who tries to setup SMC themselves and doesn't happen to also create python*-nosite?

Shouldn't it only use those scripts if they exist -- otherwise just do the normal thing?

haraldschilly:
Hmm, I was only thinking about the compute environments.  Yes, testing that they exist is a good idea!

haraldschilly:
added a check for file existence and being executable

williamstein:
Nice!

williamstein:
Crap -- I merged this and made it live and it **completely broke everything**!   No project would start.   I'm changing it with an `if False` for now.  Please try to fix **and test** (by installing into a project on the live system with `--user` but with the thing that avoids user commented out).

Issue #740, opened 2016-07-24
add sagews builtin %anaconda3 mode using jupyter anaconda3 kernel


DrXyzzy:

DrXyzzy:
Closing because issue is the now the same as #739, since deciding to implement a single built-in mode `%python3` using the `anaconda3` jupyter kernel.

Issue #739, opened 2016-07-24
have sagews %python3 mode use jupyter python3 kernel


DrXyzzy:

DrXyzzy:
Plan:

1. Replace existing %python3 with jupyter bridge running anaconda3 kernel. Work on this enhancement will also take care of #740.
1. Fix so error output with ANSI codes is displayed in color:

    ```
    %a3
    xyz*
    # output below here --
    [0;36m  File [0;32m"<ipython-input-1-1e273f11e5d1>"[0;36m, line [0;32m1[0m
    [0;31m    xyz*[0m
    [0m        ^[0m
    [0;31mSyntaxError[0m[0;31m:[0m invalid syntax
    ```
1. Add support for python3 `input()`which causes cell execution to hang at present.

haraldschilly:
I just saw this ticket when clicking through this pile here. I'm in general more happy, if `%python3` uses the system-wide `/usr/bin/python3` and `%anaconda3` uses `$ANACONDA3/bin/python` (the "3" just means that it is python 3 based).

I'm saying this, because I did some work towards those new compute images for the k8s switch, which actually have 4 different python environments. This would also imply, that a `%python2` is using `/usr/bin/python2`.

DrXyzzy:
Just saw comment from @haraldschilly above, but this is not about /usr/bin environments, but rather about which jupyter kernel to use with `%python3` mode. Recommendation in 2016-07-23 chat was to run the `anaconda3` kernel with the built-in `%python3` mode.

DrXyzzy:
Done.

Issue #358, opened 2016-01-05
raw_input in sage worksheets does not work sufficiently robustly


kcrisman:
See [this SO question](http://stackoverflow.com/questions/34621146/double-input-and-extra-letter-in-list-in-sagemath) where the doubling nature of `raw_input` becomes clear.  I assume this is not desired?

Related: #78

williamstein:
Simple example:
```
age = int(raw_input('Enter your age: '))
if age < 18:
    print('You may vote in {difference} years.'.format(difference = 18-age))
else:
    print('You are allowed to vote.')
```

williamstein:
Marking hard since the coordination between frontend and backend to get this robust is really tricky.

williamstein:
See also https://github.com/sagemathinc/smc/issues/78

williamstein:
heh in that comment I said this is NOT fixed by that commit!

williamstein:
process: got angry/curious about this while merging PR https://github.com/sagemathinc/smc/pull/1799

PR #1217, opened 2016-11-09
Issue739python3

DrXyzzy:
Ref: #739 and #740.

- Invoke `anaconda3` jupyter kernel when `%python3` mode is used
- fix ansi highlighting of errors as reported in #739 
- add support for input() for %python3 mode (actually for any mode using either python3 or anaconda3 kernel)
- fix latex output generated by %python3
- `debug` is no longer an optional argument to the `jupyter()` kernel launch function. Instead, set e.g. `python3.jupyter_kernel.debug = True` after starting a kernel and jupyter messages will be sent to sage_server log
- add pytest cases for %python3 - this PR passes all 79 tests


haraldschilly:
ok, read the code, looks good. the part about detecting plain text vs. latex is odd, but probably the only way to make this work well. what I'm also a bit puzzled about are all those additional `continue` statements. I assume that has something to do with how the protocol actually works. At some point, this should obviously be cleaned up, but let's focus on getting it work well first.

haraldschilly:
testing it in my own dev project reveals, that it works fine and all tests pass.

what's not working is that new input box. initially it looks fine, but then some sort of "hiccup" happens and it duplicates. of course, that could be entirely related to how the output is rendered, but it's not working well. Sometimes, I get it to work initially for a split second, but then it's broken.

smc-python3-jupyter-input-broken

DrXyzzy:
Thank you for the review. The input hiccup is a known problem with sagews `raw_input()`, #358 which is on track to be fixed soon per yesterday's chat.

haraldschilly:
ok, I see, I've merged it

Issue #1212, opened 2016-11-07
sagews: %prun broken


haraldschilly:
`%prun factor(12)` is broken.

haraldschilly:
`%prun` was just a special case, this is a more general one:

```
@interact
def f(sort=("foo", selector([('bar', 'baz')], width='100%'))):
    print sort
```

without that `width = ...` it works

haraldschilly:
ok, got it, that's a really weird one

haraldschilly:
just checked, and also works in prod.

PR #1227, opened 2016-11-14
sagews: quick fix for %prun -- #1212 tracks the root cause

haraldschilly:
for testing:

```
%prun
1+1
```


haraldschilly:
don't merge

PR #1228, opened 2016-11-14
webapp/interact: parse_width spelling error -- #1212

haraldschilly:
This also fixes

```
%prun
1+1
```

It's really strange, since that spelling error has been there since 2013? how is that possible?


williamstein:
>  how is that possible?

Nobody has complained because nobody was sufficiently annoyed and tried it.  We need more users and need to encourage them strongly to complain more :-)

Issue #1225, opened 2016-11-13
autocomplete with * [space] variable


DrXyzzy:
Case 1. This does not autocomplete in sagews:
```
z = 45.9 * units.len[tab]
```
but this does
```
z = 45.9 + units.len[tab]
```

Case 2. This does not autocomplete:
```
class TC:
    def __init__(self, xval):
        self.x = xval
y = TC(49)
-- new cell --
z = 12 * y.[tab]
```
but this does:
```
z = 12 + y.[tab]
```

Autocompletion also succeeds if the operator is omitted.

haraldschilly:
I think that's solved, closing it -- reopen if I'm wrong.

PR #1239, opened 2016-11-16
1225autocomplete

DrXyzzy:
- Fix for #1225 plus regressions on #295 and #701.
- Add test cases for #295, #701, #734 and #1225 (all autocomplete)
- Add introspection test fixture.

### DIFF
--- a/src/smc_sagews/smc_sagews/sage_parsing.py
+++ b/src/smc_sagews/smc_sagews/sage_parsing.py
@@ -371,10 +371,11 @@ def introspect(code, namespace, preparse=True):
                 get_help = False; get_completions = True; get_source = False
                 i      = expr.rfind('.')
                 target = expr[i+1:]
-                if target == '' or is_valid_identifier(target) or '*' in expr:
-                    # this case includes list.*end<tab>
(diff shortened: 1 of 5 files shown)


DrXyzzy:
Just added fix for %python3 autocomplete to this PR.

src/smc_sagews/smc_sagews/sage_parsing.py
```diff
@@ -386,7 +387,8 @@ def introspect(code, namespace, preparse=True):
         if get_completions and target == expr:
             j      = len(expr)
             if '*' in expr:
-                # this case includes *_factors<TAB>
+                #if '*' in expr and '* ' not in expr:
```
haraldschilly: nitpicking, this commented if line shouldn't be there at all, right?

haraldschilly:
I'm giving this a positive review, and I went through all the referenced tickets which can all be closed. However, after wrapping my head a bit around what the code does, I can easily come up with a new issue -- i.e. what about `2*sq[TAB]` where I aim that to be `2*sqrt(2)`? that's now excluded.

PR #1246, opened 2016-11-17
Fix a space

johnjeng:


haraldschilly:
Hmm, not sure. `Space` could change it's actual meaning. What I usually do in a situation like that, is `{' '}`. That makes the space character not dissapear.

williamstein:
> Space could change it's actual meaning.

But I defined space for this purpose, so I think we control its meaning.

PR #1249, opened 2016-11-18
Allow copying file and folder names from the files listing

### LINKED ISSUE #1247: make it so it is possible to copy filenames from the file listing

johnjeng:
Fixes #1247 
- Allows highlighting text without opening the file
 - Applied to projects in the projects listing as well.
- Opens the file if you click on it
- Nothing is highlighted when you check multiple rows with `shift + click`


williamstein:
Wow, this is pretty cool!

Issue #1251, opened 2016-11-18
smc_sagews tests depend on current working directory


haraldschilly:
when running those tests, they fail if I'm in a parent directory (i.e. `python -m pytest -x smc_sagews/tests/`) ... uhm, they even fail, if I run them exactly as explained in the tests readme file.

```
assert 'stdout' in {'done': False, 'event': 'output', 'id': 35, 'stderr': 'Error in lines 1-1\nTraceback (most recent call last):\n  File "...ttach\n    raise IOError(\'did not find file %r to attach\' % fname)\nIOError: did not find file u\'a.sage\' to attach\n'}
```

i propose, that there should be a change of the working directory, at least for those tests where the file is being attached.

DrXyzzy:
Done.

Issue #1252, opened 2016-11-18
sagews autocomplete 2*sq[TAB]


haraldschilly:
<strike>After #1239 is in</strike> Ok, it was broken before, too.

The autocompletion of `2*sq[TAB]` for e.g. `2*sqrt(2)` is not working.

DrXyzzy:
Also `2+sq[TAB]` is not working, and it goes through different code.

DrXyzzy:
Done.

PR #1258, opened 2016-11-19
1251testdir

DrXyzzy:
Fix for #1251 so smc_sagews pytests can be run from any directory. Added fixture to set directory for read-only data files.

I updated the README.md to reflect where I usually run the tests. But tests now also work if you invoke them per the previous instructions, i.e.
```
cd smc/src/smc_sagews/smc_sagews
python -m pytest tests
```


DrXyzzy:
Added fixes & test cases for #1252.

Issue #1271, opened 2016-11-21
fix projects-level ghost tabs -- something wrong seemingly caused by computed values


williamstein:
See commented out code in init_app.coffee in the close_project_tab function.

williamstein:
To work on this or test it, you have to do `npm run webpack-watch-production`; each build on change will take a LONG TIME.

williamstein:
I'm assigning you @johnjeng since this problem isn't visible before I merged computed values, and is visible afterwards.  Moreover, you implemented ghost tabs, I think.

PR #1283, opened 2016-11-21
Do nothing on closing a project if it's not open

johnjeng:
Fixes all known replication cases of #1271. 

App was trying to close tabs which didn't exist on every store change.

### DIFF
--- a/src/smc-webapp/init_app.coffee
+++ b/src/smc-webapp/init_app.coffee
@@ -59,6 +59,9 @@ class PageActions extends Actions
         active_top_tab = page_store.get('active_top_tab')
 
         index = open_projects.indexOf(project_id)
+        if index == -1
+            return
+
         size = open_projects.size
         if project_id == active_top_tab
             if index == -1 or size <= 1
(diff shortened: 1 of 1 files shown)


williamstein:
This works.  I have to wonder -- I put in console.log and see that ...
```
        index = open_projects.indexOf(project_id)
        if index == -1
            return   # this return code path very frequently gets called, even when opening a project...
```

why?  That seems odd.

johnjeng:
It's a listener on the projects store. https://github.com/sagemathinc/smc/blob/600b6ee7f61ad2f5872d0694e4fb719db7669c99/src/smc-webapp/project_store.coffee#L1210

It's old code that got moved into the new store definition. Really it should use the Table since long polling vs. pushed notifications..

williamstein:
>  Really it should use the Table since long polling vs. pushed notifications..

What do you mean by "long polling vs. pushed notifications" in this context?

johnjeng:
I was thinking about listening on all store changes vs. something telling the projects controller that the project is no longer available. 

But now I think the real solution is to let the project stay open when the user gets removed from the collaborators list. Rendering would be identical to public viewing of the project.

PR #1298, opened 2016-11-24
Prevent moving folders into themselves

### LINKED ISSUE #1295: don't allow trying to move a directory into itself
To replicate:

1. create a directory
2. try to move it into itself using the UI

Issues:
1. The error message has some warning about BASH shells, which is totally off.
2. You shouldn't be able to do this in the first place -- the directory selector box should exclude the directory you're copying/moving.

Images showing this:

screen shot 2016-11-23 at 10 08 18 am

screen shot 2016-11-23 at 10 08 27 am

johnjeng:
Fixes #1295 

Doesn't show any folders for when a selected folder is its ancestor.

### Full folder tree
full-tree

### Just top2 selected
top2-highlighted

### All contents of top2 selected
top2-contents


johnjeng:
Need to check if my assumptions about string sorting is true.

johnjeng:
I think the correct way is change how `find` is being called using some options to `fetch_directories`. Not sure I understand entirely how `@exec` works in all the args. Failing to use approaches found [here](http://stackoverflow.com/questions/4210042/exclude-directory-from-find-command).

src/smc-util/client.coffee
```diff
@@ -1124,15 +1125,30 @@ class exports.Connection extends EventEmitter
...
+        if opts.exclusions?
+            exclusion_args = _.map opts.exclusions, (excluded_path, index) =>
+                "-a -not \\( -path '#{opts.path}/#{excluded_path}' -prune \\)"
+            args = args.concat(exclusion_args)
+
+        args = args.concat(tail_args)
+        # The exclusion args don't work if not in the command for some reason
+        if opts.exclusions
+            command = "find #{args.join(' ')}"
+            args = undefined
```
johnjeng: test

src/smc-util/client.coffee
```diff
@@ -1122,17 +1123,31 @@ class exports.Connection extends EventEmitter
...
```
johnjeng: I gave up on a fruitless search to find out why it wouldn't work.

src/smc-util/client.coffee
```diff
@@ -1122,17 +1123,31 @@ class exports.Connection extends EventEmitter
...
 
         @exec
             project_id : opts.project_id
-            command    : "find"
+            command    : command ? "find"
+            args       : args
             timeout    : 15
-            args       : [opts.path, '-xdev', '!', '-readable', '-prune', '-o', '-type', 'd', '-iname', opts.query, '-readable', '-print']
-            bash       : false
```
johnjeng: I don't know what this does, but it's the default in `@exec`.

williamstein:
Your code seems to makes some false assumptions about state that result in tracebacks.  Does that make sense?  I could never get this to work (it always failed for me).  See:

screen shot 2016-11-30 at 8 12 08 am

screen shot 2016-11-30 at 8 12 19 am

williamstein:
Now it never shows any options in the target directories dropdown, no matter how long I wait or how many times I try.

What about starting over from scratch and just filtering out the directory on the client?  Shouldn't that be like 2 lines of code and easy, and can't possibly go wrong like this...?

williamstein:
screen shot 2016-11-30 at 3 10 22 pm

johnjeng:
Ahhh, I was always testing Move and then testing Copy.

>Shouldn't that be like 2 lines of code and easy, and can't possibly go wrong like this...?

We also have to filter out subdirectories and since you're just given a list of all the directories not necessarily in any order, it's O(||all directories|| * ||selected directories||). 

Also, adding it to the find command means that moving large folders with many sub folders becomes much faster. eg. if I wanted to move an installed SMC repo.

williamstein:
If/when you can get things to work that's way better than client side filtering (I guess)...

johnjeng:
Okay, this should work since I'm testing it correctly now.

src/smc-webapp/projects.cjsx
```diff
@@ -242,7 +242,7 @@ class ProjectsActions extends Actions
         opts = defaults opts,
             exclusions : undefined # Array<String> of sub-trees' root paths to omit
         # WARNING: Do not change the store except in a callback below.
-        block = "_fetch_directory_tree_#{project_id}"
+        block = "_fetch_directory_tree_#{project_id}_#{opts.exclusions?.toString()}"
```
williamstein: @johnjeng  -- this makes me nervous...

However, I don't care, since the whole directory tree will be sync'd via the db at some point in the future, which will remove all this.

williamstein:
Works finally, so far as I can tell, within the constraints of this approach!

PR #1307, opened 2016-11-25
show login form or nicely formatted error for non public files

timothyclemansinsea:


timothyclemansinsea:
@williamstein @haraldschilly @johnjeng could one of you please tell me how to how to display <AccountPage /> instead of an editor?

williamstein:
> ... how to how to display instead of an editor?

How to display what?  Is there a word missing from your sentence?

timothyclemansinsea:
@williamstein sorry I didn't realize `<AccountPage />` would get stripped out

johnjeng:
I'm not sure I understand what you're requesting. Looking at the code, it seems like you're trying to put login forms when the user tries to access a private file. 

You'll want to understand [how public editors are made](https://github.com/sagemathinc/smc/blob/master/src/smc-webapp/project_file.cjsx) and how files are [opened](https://github.com/sagemathinc/smc/blob/master/src/smc-webapp/project_store.coffee#L269). My guess is that you'll want to make a fallback public editor which is a login form and once they log in, it will stay open on that file.

johnjeng:
For clarification, changing the public code mirror editor is not the right way to do this.

haraldschilly:
so, I think it's much better to add a link or button to show the account page with the login form. it's complex to get that form right and I'm not sure there is much gain in spending much time in figuring out how to properly extract it and use it in two locations.

johnjeng:
I think from a user perspective it's definitely nicer to have it in place and open your file immediately. But if you're arguing that it isn't worth the developer hours, my response is "maybe...". 

I certainly want the log in stuff not attached to the account page anymore but it hasn't been a priority.

Issue #1016, opened 2016-10-10
improve projects/new project UI


haraldschilly:
The enhanced dialog for creating a new project in [projects.cjsx](https://github.com/sagemathinc/smc/blob/master/src/smc-webapp/projects.cjsx) has a couple of issues.
- [ ] deduplicate the code from [project_settings](https://github.com/sagemathinc/smc/blob/master/src/smc-webapp/project_settings.cjsx) since [this patch](https://github.com/sagemathinc/smc/commit/c29b274f91e8172ce062786c8c8d99dd4d11b33e#diff-1d4654a4776f0825cd0ac2f439629410) fixes a couple of issues that aren't there yet. The best strategy would be to use one single component that's used in both places!
- [ ] It is confusing to show so much information to new users. Instead, condense this to a clear explanation and only reveal the upgrade dialog with billing information, when a button is pressed.
- [ ] If a user is already a member, show the upgrade dialog directly. Additionally, in such a case when there are unused quotas for internet and member hosting left, automatically apply them for a newly created project (it's sensible to assume, that with unused upgrades a user wants to use them -- and explicitly unchecks the box if not)

haraldschilly:
see also https://github.com/sagemathinc/smc/issues/1015

timothyclemansinsea:
Code deduplication needs to be its an own issue because it's such an involved task. I tried to do it and failed. 

My proposal:
Change explanation text to "A project is your own private computational workspace that you can share with others. Each project is free. What's not free is members only hosting and direct use of internet resources. Project resources like ram, cpu, and disk space can be upgraded. If you have any questions, please email help@sagemath.com immediately."
Have three buttons:
"Create project without any upgrades"
"Create project with members only hosting and direct internet access"
"Create project with customized upgrades"

Subscription/billing only show up if the 2nd button is clicked and upgrade is not possible without changing subscription/upgrade allocation or the 3rd button is clicked.

timothyclemansinsea:
If I end up doing this I would like to break this down into a few E-easy/E-medium issues rather than one big E-hard issue.

timothyclemansinsea:
This either needs to be closed or updated given the recent improvements.

PR #1309, opened 2016-11-26
Simplify the create project component

timothyclemansinsea:
Note: make sure you set commercial to yes in settings and set up stripe

For #1016

### DIFF
--- a/src/smc-webapp/projects.cjsx
+++ b/src/smc-webapp/projects.cjsx
@@ -702,7 +702,7 @@ NewProjectCreator = rclass
             title_text        : ''
             description_text  : ''
             error             : ''
-            create_button_hit : '' # Options are 'with_members_and_internet' and 'with_custom_upgrades'
+            create_button_hit : '' # Options are 'with_members_and_network' and 'with_custom_upgrades'
 
     toggle_editing: ->
         if @state.state == 'view'
@@ -765,12 +765,14 @@ NewProjectCreator = rclass
         else if @state.state == 'saving'
             <Alert bsStyle='info'>Working hard to build your project... <Icon name='circle-o-notch' spin /></Alert>
 
-    create_project_with_members_and_internet: ->
+    create_project_with_members_and_network: ->
         remaining_upgrades = misc.map_diff(@props.upgrades_you_can_use, @props.upgrades_you_applied_to_all_projects)
         if remaining_upgrades.member_host > 0 and remaining_upgrades.network > 0
-            @create_project({member_host: 1, network: 1})
+            @setState(create_button_hit: 'with_members_and_network')
+            @scroll_to_billing()
         else
-            @setState(create_button_hit: 'with_members_and_internet')
+            @setState(create_button_hit: 'custom_upgrades')
+            @scroll_to_billing()
 
     scroll_to_billing: ->
         setTimeout ( ->
@@ -781,11 +783,10 @@ NewProjectCreator = rclass
         <ButtonToolbar>
             <label>Create this project with:</label><br/>
             <Button
-                disabled = {@state.title_text == '' or @state.state == 'saving' or @state.create_button_hit == 'with_members_and_internet'}
+                disabled = {@state.title_text == '' or @state.state == 'saving' or @state.create_button_hit == 'with_members_and_network'}
                 bsStyle  = 'success'
-                title    = 'Upgrades you to members only hosting which is more reliable and network access so you can use internet resources directly from your code'
-                onClick  = {=>@create_project_with_members_and_internet()} >
-                <Icon name="arrow-circle-up" /> Hosting/network upgrades
+                onClick  = {=>@create_project_with_members_and_network()} >
+                <Icon name="arrow-circle-up" /> Hosting and network upgrades…
             </Button>
             <Button
                 disabled = {@state.title_text == '' or @state.state == 'saving' or @state.create_button_hit == 'with_custom_upgrades'}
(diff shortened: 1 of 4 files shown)


williamstein:
Is this ready for review?  Has not label.

williamstein:
I'm labeling it needs work, where the work my simply be to set the label to needs review.

src/smc-webapp/projects.cjsx
```diff
@@ -774,6 +765,67 @@ NewProjectCreator = rclass
...
+            <Button
+                disabled = {@state.title_text == '' or @state.state == 'saving'}
+                bsStyle  = 'success'
+                onClick  = {=>@create_project(false)} >
+                Create project without upgrades
+            </Button>
+            <Button
+                disabled = {@state.title_text == '' or @state.state == 'saving' or @state.create_button_hit == 'with_members_and_internet'}
+                bsStyle  = 'success'
+                onClick  = {=>@create_project_with_members_and_internet()} >
```
johnjeng: This unexpectedly creates a project even if I don't have members hosting spots available. This is because the function only checks if I have some upgrades in members and internet but not if they're available.

src/smc-webapp/projects.cjsx
```diff
@@ -697,10 +698,11 @@ NewProjectCreator = rclass
...
         @setState
-            state            : 'view'
-            title_text       : ''
-            description_text : ''
-            error            : ''
+            state             : 'view'
+            title_text        : ''
+            description_text  : ''
+            error             : ''
+            create_button_hit : ''
```
johnjeng: Comment what the possible states are. Not a huge fan of this name but it makes okay sense knowing what the possible states are.

src/smc-webapp/projects.cjsx
```diff
@@ -774,6 +765,67 @@ NewProjectCreator = rclass
...
+
+    render_create_button: ->
+        <ButtonToolbar>
+            <Button
+                disabled = {@state.title_text == '' or @state.state == 'saving'}
+                bsStyle  = 'success'
+                onClick  = {=>@create_project(false)} >
+                Create project
+            </Button>
+            <Button
```
johnjeng: This doesn't render well. See screenshot.

timothyclemansinsea:
@johnjeng I fixed 2 of the 3 issues. I don't have a good idea for how to make this look better. What's your ideas?

johnjeng:
I'm not sure. One of the problems is that the button text feels too long. One option to fix this using names for the types of projects. eg. The buttons could say `Create a free project`, `Create an elevated project`, `Create a project with custom upgrades`. (Caveats, "free project" and "elevated project" could be confusing")

Other ways one could describe a project with members hosting and internet:
- Deluxe Project
- Plus Project
- Project+
- Power Project
- Privileged Project
- First-class Project
- Advantage Project
- Vantage Project

Okay, some of these are just silly, but you get the idea.

Ways to describe a project without any upgrades:
- Basic Project
- Barren Project
- Default Project
- Destitute Project

Explanation of project upgrades and stuff can come after the buttons. The text is a little clunky. "Each project is free. What's not free is...". It could be something like "All projects are free but upgrading ... costs resources which you may purchase through subscriptions."

timothyclemansinsea:
@johnjeng @williamstein what do you think of:
image

johnjeng:
I'd say change the text "Projects are free ... and disk space." to "Creating basic projects without upgrades is free while upgrades require a subscription. Core upgrades are members only hosting and network access. You may also upgrade the CPU, RAM, and disk space."

- Change "Create a project with:" to "Create this project with:"
- Move the button bar below that text.
- Add a helper alert between the text and the buttons if there's no title.
  - Probably `warning` definitely not `danger`
- Change the button text "Basic upgrades" to "Core upgrades".
- Put the "No upgrades" button last after "Custom upgrades"
- Make the background of "No upgrades" gray but not an "inactive-looking" gray.
- Make "Core upgrades" primary
  - imo, we want to suggest users do this. However, I might also decide I hate how it looks in the end.

johnjeng:
Also, since it's just a text change, could you add to this PR changing all relevant instances of `internet access` to `network access`?

haraldschilly:
I remember from some support requests, that users didn't understand that blocked network access means that it's not possible to download files from the internet from within a project. also, blocking network access would render projects completely useless because they can't interact with the other components of smc. maybe there is some other good phrase to replace it with? like "firewalled" vs "not firewalled"? a link or hover-textbox then explains what this means. it's also a more complicated term, which will cause those who don't understand it to look it up. "network" and "internet" are much broader and if someone reads that with the wrong assumptions doesn't start re-evaluating ones assumptions.

timothyclemansinsea:
@haraldschilly I'm actively working on this at the moment. "No direct internet" is my proposal. Right now I have my code set to Network access

timothyclemansinsea:
I'm finding it a bit hard to make the core/custom upgrade buttons stand out from the no upgrades button. Maybe put the no upgrades button on the right side?

image

johnjeng:
>stand out from the no upgrades button

Looks good enough for me. Maybe add the [arrow circle up](http://fontawesome.io/icon/arrow-circle-up/) icon in front of `Core upgrades` and [cog](http://fontawesome.io/icon/cog/) or [wrench](http://fontawesome.io/icon/wrench/) in front of `Custom upgrades`.

Thinking about this more, I would move the button stuff above all the text and put the "You can easily change ..."  directly under the description section like it is on master.

williamstein:
This seems really close:

- [ ] the "Custom upgrades" button needs ... after it, since it opens up a further "dialog".
- [ ] mine had a "Core upgrades" button.  I don't know what it does.  It should maybe have ... and an explanation that appears, then a confirm/cancel button.  Maybe just get rid of this button for now.
- [ ] The "No upgrades" button is grey: we don't use a grey dialog anywhere else.  Hmm.  Maybe just use default (so white)?
- [ ] When you click "Custom upgrades..." the dialog that appears below is so far down you'll never even see it unless you have a big screen.  The new details of upgrades should appear immediately below the button.

timothyclemansinsea:
@williamstein Maybe core -> members hosting & network access?

williamstein:
Definitely can't be cores since one of the quotas is "CPU cores".

williamstein:
1. Can you fix the spacing here? -- it just looks uneven:



2.... Actually I can't think of anything in particular to complain about.   It does all seem to work, to me.

PR #1317, opened 2016-11-28
Fix updating current directory url

### LINKED ISSUE #1312: go to the file listing in a project and click "Backups" -- the URL at the top doesn't change to indicate the current path (.../.snapshots)
This is likely a bug relating to the top down rewrite...

johnjeng:
Fixes #1312.

### DIFF
--- a/src/smc-webapp/project_store.coffee
+++ b/src/smc-webapp/project_store.coffee
@@ -156,7 +156,7 @@ class ProjectActions extends Actions
     # Updates the URL
     set_active_tab: (key) =>
         store = @get_store()
-        if store.get('active_project_tab') == key
+        if store.active_project_tab == key
             # nothing to do
             return
         @setState(active_project_tab : key)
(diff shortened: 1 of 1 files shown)


williamstein:
This breaks things badly.  The most basic test is:

1. open project
2. make directory
3. click directory and observe the URL at the top.  It should end in the directory name.  Nope.

screen shot 2016-11-28 at 5 58 46 pm

johnjeng:
I can't replicate your broken test. 

Maybe it didn't build this PR yet? Because it's actually broken on master.

I think ProjectActions might need to get refactored because [this patch](https://github.com/sagemathinc/smc/commit/9f9db60f2eccdfce630e148f1826b10e5153793f#diff-e6453b90cf71dc76e152191a2985bafdR156) is what broke the url changing when opening folders. Except it's totally reasonable code to have so there's some bad coupling going on.

williamstein:
> Maybe it didn't build this PR yet? Because it's actually broken on master (and live).

Oh, wow, that's true and really terrible.  Raising bug to blocker status.

> I think ProjectActions might need to get refactored because this patch is what broke the url changing when opening folders. Except it's totally reasonable code to have so there's some bad coupling going on.

I'm worried that there may be some fundamental misunderstanding about what the store needs to do, which is being exposed here. 

I'll try testing this branch again...

williamstein:
I rebuilt (stopping/starting the watcher), and in fact you're right  -- your branch **works perfectly**.

I'll be making this live in a few minutes.

PR #1324, opened 2016-12-02
Issue/1234

### LINKED ISSUE #1234: toolbar md: wrapping selected text for a list splits cursor
This is not really wrong, but it is doing odd things:

1. open md file (or `%md` in sagews)
2. write "foo" in a new line, select the whole word "foo"
3. click on the "insert unordered list" button

Result is, that there are now *two* cursors, one before and one after the insertion.

My dream behavior would be, that I can select more than one line (e.g. `foo\nbar\nbaz`) and clicking on ordered or unordered list inserts a `1. ` or ` -` on each line. That's probably similar to commenting each line by prefixing a `# `.

haraldschilly:
this fixes #1234 

tests in a md file:

1. select multiple lines, click on list or enumeration ... each line does get a marker
2. no second cursor appears when multiple line changes happen, i.e. select some text and change the font.
3. similar to the above, it also does no longer happen when text is selected and a horizontal line is inserted

### DIFF
--- a/src/smc-webapp/buttonbar.coffee
+++ b/src/smc-webapp/buttonbar.coffee
@@ -355,14 +355,18 @@ exports.commands =
             wrap :
                 left  : "\n#### "
                 right : ""
-        format_code :  # FUTURE: I think indentation is probably nicer?  on single line ` is nicer.
+        format_code :
             wrap :
-                left  : '\n```'
-                right : '\n```\n'
+                left  : '    '
+                right : ''
+                multi : true
+                space : false
         indent :
             wrap :
-                left  : "\n> "
-                right : ""
+                left  : '>'
+                right : ''
+                multi : true
+                space : true
         horizontalRule:
             wrap:
                 left  : "\n------------------\n"
--- a/src/smc-webapp/buttonbar.coffee
+++ b/src/smc-webapp/buttonbar.coffee
@@ -325,12 +325,16 @@ exports.commands =
                 right : '~~'
         insertunorderedlist :
             wrap :
-                left  : "\n - "
-                right : "\n"
+                left  : " -"
+                right : ''
+                multi : true
+                space : true
         insertorderedlist :
             wrap :
-                left  : "\n 1. "
-                right : "\n"
+                left  : "1."
+                right : ''
+                multi : true
+                space : true
         format_heading_1 :  # FUTURE -- define via for loop below
             strip : ['format_heading_2','format_heading_3','format_heading_4']
             wrap :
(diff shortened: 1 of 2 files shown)


johnjeng:
A nice side effect of this is that `ctrl+clicking` multiple lines also works with bullets and lists as expected now.

johnjeng:
Thoughts on what this patch is trying to do: LGTM

Thoughts in general:
Because of how `highlight` + `tab` works, I would expect highlighting part of a line to make the whole line into a list as well. To me, making lists and bullets is a transformation on the line and not the highlighted text specifically.

What happens:

(before)
highlight

(after)
after

What I expect the result to be:
alternative
or
expected

However, inserting around the ends of the highlighted text matches things like bold or italics so I could be wrong.

src/smc-webapp/buttonbar.coffee
```diff
@@ -351,14 +355,18 @@ exports.commands =
...
+                multi : true
+                space : false
         indent :
             wrap :
-                left  : "\n> "
-                right : ""
+                left  : '>'
+                right : ''
+                multi : true
+                space : true
```
johnjeng: I'm not sure what's going on, but quoting seems slightly off.
highlight
quoted

haraldschilly:
well, maybe your comment about partial selections for bullets/lists should be a ticket. compare it to how it behaves now, where it inserts  a new line and the marker, ignores the remaining lines, and introduces a second cursor. to do what you want, we have to come up a new parameter (e.g. `full_line: true (default false)`) for the button bar definition and handle it in the code.

horizontal lines aren't defined to work on each line, they're essentially wrapped around the selection (with linebreaks)

and no idea what's going on with the additional space for quotes... I'll try to replicate it and see if there is something obvious to fix.

johnjeng:
Yeah, I don't expect this PR to address any of those concerns. It looks great! The quoting thing seems like something to fix though.

haraldschilly:
ok, I found out the part about quoting. The trick to replicate yours is to start each line with one or more spaces. I thought all lines are starting right at the left hand side!

So, looks like for the first line there is some trimming going on. Let's see if I can find it.

haraldschilly:
@johnjeng I've added some additional code to deal with some special cases. So, e.g. your single space indentation list works, it also deals with adding a new line at the end when the selection is inside a line (otherwise md doesn't parse it properly), etc. It's not changing the whole line outside the selection, that's definitely something for another ticket and maybe needs some additional discussions. I was also careful to not change the existing behavior with empty, words and single line selections -- although it starts  to feel like a whack a mole game.

williamstein:
Works for me.  I agree with @johnjeng  that making partial select work on a line basis makes sense, but it definitely doesn't have to be in this PR.

PR #1328, opened 2016-12-05
webapp: remove sagews2html easter egg

haraldschilly:
Attention, merging this branch removes the sagews2html beta testing easter egg and adds a new button for the current sagew2pdf printing.


williamstein:
Works for me and makes sense.

Issue #1330, opened 2016-12-05
Codebase has inconsistencies about directory path assumptions


johnjeng:
In some places we assume that paths are directories iff it ends with `/`. Other places, we assume that a path not ending in `/` could still be a directory and we check against the `isdir` property in the displayed directory listing.

This issue proposes that the front end should assume everywhere that folders end in `/` and files don't. Some parts of the front end receives folder paths without the trailing `/` but context lets us convert these cases. This convention is enforced by linux since folder and file names cannot include `/`.

williamstein:
> This issue proposes that the front end should assume everywhere that folders end in / and files don't.

What's your argument for doing this, instead of **the opposite** -- namely NEVER putting / at the end of a name? I think the arguments for never ever putting / at the end of paths is much more compelling:

1. Consistency with UNIX:   In command line, scripts (i.e., linux) folders never end in / (e.g., if you are doing something with a pipe or ls, etc.).
2.  Encoding **metadata** about the path (whether it is a path to a file or folder) in the path name itself is a bad idea, and isdir is just metadata.  The metadata should go somewhere else.   For example the [smc-ls script](https://github.com/sagemathinc/smc/blob/master/src/smc_pyutil/smc_pyutil/git_ls.py) has several pieces of metadata about paths: filename, size, last change, and whether or not the path is a folder.  The latter is just metadata, just like the size and last modification time.     You could argue that all the metadata should be encoded in the filename, e.g, "foo/2016-10-11/456/", but that way clearly madness lies.

williamstein:
@haraldschilly -- very interested in your opinion...

johnjeng:
1. Under [POSIX](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_13) [specified](http://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xbd_chap04.html#tag_21_04_13) utilities, using a trailing slash is a kind of type checking requiring that it's a folder.
   - This is nice in certain cases. [Example](http://unix.stackexchange.com/a/50533)
2. We have to ask the server if something is a directory.
    - For example when linking to directories in other projects.

After some thought, the opposite of what I'm suggesting is also rather nice:
- Various APIs would become agnostic to folder/file.
  - eg. `open_path` instead of `open_file` and `open_directory`
- Project tree information as a necessity will be easy to access

johnjeng:
Some more arguments one way or the other. 
[...]

PR #1331, opened 2016-12-05
Fix opening copied directories from log

### LINKED ISSUE #1326: Opening a Directory from Project Log Fails
1. Copy a folder to anywhere
2. Note the record in `Log`
3. Click on the link to the folder
4. See failure

opening folder from log

The Project Log is probably just falsely assuming all entries are related to files.

johnjeng:
Fixes #1326. 

In my opinion, this isn't the prettiest solution, but a better solution (I would rather all folder paths end with `/`) is *much* more involved with a *lot* more testing. #1330 proposes this change.

### DIFF
--- a/src/smc-webapp/project_files.cjsx
+++ b/src/smc-webapp/project_files.cjsx
@@ -2064,7 +2064,7 @@ exports.ProjectFiles = rclass ({name}) ->
             </div>
 
     update_current_listing: ->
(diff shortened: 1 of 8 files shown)


williamstein:
Can you look at my comments/arguments on https://github.com/sagemathinc/smc/issues/1330 and let me know if you still think doing this is a good idea?

johnjeng:
This solution has flaws that would require hacking around the problem specified by #1330. I'll implement a proper solution to this problem when a consensus is reached on that issue.

PR #1333, opened 2016-12-06
webapp/sagews/interact: html-process descriptions to fix #1332

### LINKED ISSUE #1332: broken images in %exercise questions
This shows a broken image, because the url isn't fixed up:

```
%exercise
plot(sin(33*x)).save('plot2.svg')
question = "<p>what is this <img src='plot2.svg'></p>"
```

it works for interacts and `html( ... )`, so there must be something missing regarding how `question = ...` is processed.

haraldschilly:
this is an attempt to fix #1332 

I ended up doing it this way after a lot of trial and error. Feels like the wrong way to do it and I hope there is a more elegant way. Also, those additional processing steps in between inside interact.coffee are probably not necessary.

### DIFF
--- a/src/smc-webapp/interact.coffee
+++ b/src/smc-webapp/interact.coffee
@@ -40,6 +40,7 @@ $.fn.extend
             desc                : required
             execute_code        : required
             process_output_mesg : required
+            process_html_output : required
             start               : undefined
             stop                : undefined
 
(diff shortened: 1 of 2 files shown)


williamstein:
Though ugly, this works.   Rendering will get reactified at some point, which will change everything anyways...

Issue #1349, opened 2016-12-08
maybe we should somehow display or warn on invisible unicode characters in the input of a sage worksheet...?


williamstein:
This drove me crazy for a bit!

<img width="780" alt="screen shot 2016-12-08 at 1 31 06 pm" src="https://cloud.githubusercontent.com/assets/1276278/21028577/9b0e11aa-bd4a-11e6-9663-b6417df74232.png">

See https://cloud.sagemath.com/projects/4a5f0542-5873-4eed-a85c-a18c706e8bcd/files/support/2016-12-08-131537.sagews

This is exactly the sort of funny thing people are likely to do, since they copy/paste out of PDF or other funny files, or work with other editors (Word?) and paste into SMC...  Yes, this came from a support request.

haraldschilly:
I've another idea. I bet there is only a handful of such invisible characters and we should search for a list of them. Then, whenever a user pastes something into codemirror, all those characters are replaced by a space.

What won't work are those ligature characters, e.g. that "ft" is a one-character letter. In theory, we could also detect them and replace them with two characters.

williamstein:
That wikipedia table looks great.  

It's probably best to do two things:

 - one scanning for these characters on the backend  **WHEN THERE IS AN ERROR (and only then!)** and adding that message in the output.  That's all @DrXyzzy did, I think.   So your funny unicode examples would be fine if there is no error.

- also do something on the frontend...   THOUGH, that's actually less clear to me.  E.g., what if the user is pasting these into a string, where everything would work fine.

williamstein:
I merged https://github.com/sagemathinc/smc/pull/1350 which does something useful.  However, probably something much better client side involving paste and non-visible characters should get done too.

DrXyzzy:
Merged solution is not perfect but there haven't been specific suggestions for awhile. Let's close this.

PR #1350, opened 2016-12-09
warn about non-ascii characters in code if there is an error #1349

DrXyzzy:
Ref: #1349 

How does this look? I didn't add pytest cases yet because I'm not sure the test is specific enough.
Example:
```
# code with non-printing characters
def f():
    a = [0]
    a[0]
---
Error in lines 3-3
  File "/projects/ccd4d4a4-29a8-4c39-85c2-a630cb1e9b6c/.local/lib/python2.7/site-packages/smc_sagews/sage_server.py", line 977, in execute
    exec compile(block+'\n', '', 'single') in namespace, locals
  File "", line 1, in <module>
NameError: name 'a' is not defined
*** WARNING: Code contains non-ascii characters ***
```

### DIFF
--- a/src/smc_sagews/smc_sagews/sage_server.py
+++ b/src/smc_sagews/smc_sagews/sage_server.py
@@ -919,6 +919,14 @@ class Salvus(object):
 
     def execute(self, code, namespace=None, preparse=True, locals=None):
 
+        ascii_warn = False
+        code_error = False
+        if sys.getdefaultencoding() == 'ascii':
+            for c in code:
+                if ord(c) >= 128:
+                    ascii_warn = True
+                    break
+
         if namespace is None:
             namespace = self.namespace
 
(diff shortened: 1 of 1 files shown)


williamstein:
Thanks for working on this, but I have some very serious concerns:

- for most anybody programming outside the US, it's very reasonable to have tons of non-asci all over, e.g., in comments, strings,... and with python3 even variable names. 

- the problem isn't non-ascii but **invisible** non-ascii

- it's very important to also give an indication of where the worrisome characters are -- here, showing the repr(...) of the input can be very helpful.

williamstein:
All that said, I don't know what "invisible non-ascii" even means, and as this is just an extra little warning, it's way better than nothing, without being annoying.  So I would be OK with this going in as is, unless you can think of ways to address the above.

DrXyzzy:
Ok I will think about it a bit more. Block parser could help ignore comments and strings, for example.

williamstein:
> Block parser could help ignore comments and strings, for example.

But in python3 variables names can be unicode... 

Big question: which unicode are invisible (in codemirror?).   

or in Python `"\xa0"`.  This is evidently a "NO BREAK SPACE" in unicode according to http://stackoverflow.com/questions/26068832/how-to-remove-this-xa0-from-a-string-in-python, so maybe instead of trying to solve the general case, we could just change your patch to warn about NO BREAK SPACE in the input?   

I've never had a support request about anything else, and I bet pasting in "NO BREAK SPACE" and not realizing it is common.

haraldschilly:
So, -1 to this idea, that will break with my "#äüßö" immediately ...

williamstein:
> So, -1 to this idea, that will break with my "#äüßö" immediately ...


I think you are wrong.  It only ever shows a warning if the code fails to run.  Code that works, continues to work.

Issue #1352, opened 2016-12-09
Add Conley's quote to landing page


timothyclemansinsea:
> How about just do ONE THING:
> 
> Put an abbreviated version of will Conley’s quote under the video on the landing page, making the whole thing a link to the quotes page.
> 
> That’s it!!!!!!
> 
> Abbreviated version:
> 
> “SageMathCloud provides a user friendly interface. Students don’t need to install any software at all. They just open up a web browser and go to cloud.sagemath.com and that’s it. They just type code directly in, hit shift+enter and it runs and they can see if it works. It provides immediate feedback. The course management features work really well.”
> Also, put a picture of him to the right of the quote, so when you click on the picture you get his video.
> 
> The video: https://www.youtube.com/watch?v=ZcxUNemJfZw

PR #1357, opened 2016-12-10
Add Conley's quote

timothyclemansinsea:
For #1352


williamstein:
Can you also link to the other quotes?

timothyclemansinsea:
@williamstein done

Issue #1347, opened 2016-12-08
broken mathjax for published ipynb


haraldschilly:
1. enter some latex into an ipynb, e.g. a markdown cell with `# $U_1$`
2. publish it
3. open the html file in incognito mode

expected result: mathjax runs and the formulas are converted.

haraldschilly:
also noteworthy, opening the html file directly (e.g. [this one](https://cloud.sagemath.com/14eed217-2d3c-4975-a381-b69edcb40e0e/raw/scratch/latex.html) ) shows the formulas. So it must be something with the iframe embedding into smc.

williamstein:
For completeness:
 - This should work but doesn't: https://cloud.sagemath.com/projects/14eed217-2d3c-4975-a381-b69edcb40e0e/files/scratch/latex.html
 - This does work and should be the same as the above, but isn't: https://cloud.sagemath.com/14eed217-2d3c-4975-a381-b69edcb40e0e/raw/scratch/latex.html
 - This also **does** work via a third party nbviewer iframe embed: https://cloud.sagemath.com/projects/14eed217-2d3c-4975-a381-b69edcb40e0e/files/scratch/latex.ipynb

williamstein:
now fixed!

PR #1361, opened 2016-12-12
webapp: set public html iframe srcdoc when supported -- #1347

haraldschilly:
This is the best I came up with for #1347. The root problem is, that setting the innerHTML in an iframe doesn't trigger any javascript to run. Maybe that's a security measure, I don't know (even a `console.log` in a script tag doesn't work) and therefore no mathjax.

However, setting the srcdoc works in almost all browsers and this falls back to jquery's `html(...)` for the other ones.

Besides that: It removes a margin which make no sense.

### DIFF
--- a/src/smc-webapp/_editor.sass
+++ b/src/smc-webapp/_editor.sass
@@ -356,7 +356,7 @@
 
 .salvus-editor-static-html-content
   overflow-y : auto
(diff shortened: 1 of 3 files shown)


williamstein:
Very nice!

Issue #1359, opened 2016-12-12
"Failed to open link"


williamstein:
I have seen this new -- horribly mysterious and useless -- error message several times after merging some recent PR from @johnjeng:

<img width="432" alt="screen shot 2016-12-11 at 8 45 11 pm" src="https://cloud.githubusercontent.com/assets/1276278/21088321/e1b118ba-bfe2-11e6-9986-0df8f5553f84.png">

This message is useless, as it doesn't say what the file is.   

This might only be something that admins see when they try to visit a link for a project they are not a collab on.  Not sure.  Anyway, I *AM* able to open said file if I just click on it.

haraldschilly:
I'm closing this ,since after merging #1367 the error message is definitely different

PR #1367, opened 2016-12-12
Fix invisible error when opening hidden files

### LINKED ISSUE #1365: traceback in console log when opening history
Consider this code in project_store.coffee:
```
                    @get_store().wait
                        until   : (s) => s.directory_listings.get(parent_path)
                        timeout : 3
                        cb      : (err, listing) =>
                            if err
                                alert_message(type:'error', message:'Failed to open link')
                            else
                                item = listing.find (val) => val.get('name') == last
                                if item.get('isdir')
                                    @open_directory(full_path)
```

Where it says `val.get('name')` above, I'm getting a traceback frequently when opening a file.  For me it's this file in my dev project: 
  
  https://cloud.sagemath.com/45f4aab5-7698-4ac8-9f63-9fd307401ad7/port/54249/projects/9f2e5869-54b8-4890-8828-9aeba9a64af4/files/.x.md.sage-history

Here's the traceback:

screen shot 2016-12-12 at 9 44 24 am

johnjeng:
Fixes #1365. Includes a debug log upon error for #1359 which I have been failing to replicate on test.sagemath.com and on smc-in-smc. Replication on production servers also seems inconsistent.

### DIFF
--- a/src/smc-webapp/project_store.coffee
+++ b/src/smc-webapp/project_store.coffee
@@ -1205,10 +1205,10 @@ class ProjectActions extends Actions
                         until   : (s) =>
                             listing = s.directory_listings.get(parent_path)
                             return listing?.find (val) => val.get('name') == last
-                        timeout : 3
+                        timeout : 30
                         cb      : (err, item) =>
                             if err
-                                console.log err
+                                alert_message(type:'error', message:"There was an error related to opening the link: #{err}")
                                 @open_directory(parent_path)
                             else
                                 if item.get('isdir')
--- a/src/smc-webapp/project_store.coffee
+++ b/src/smc-webapp/project_store.coffee
@@ -1200,15 +1200,17 @@ class ProjectActions extends Actions
                 if target[target.length-1] == '/' or full_path == ''
                     @open_directory(parent_path)
                 else
-                    @fetch_directory_listing(parent_path)
+                    @fetch_directory_listing(parent_path, true, true)
                     @get_store().wait
-                        until   : (s) => s.directory_listings.get(parent_path)
+                        until   : (s) =>
+                            listing = s.directory_listings.get(parent_path)
+                            return listing?.find (val) => val.get('name') == last
                         timeout : 3
-                        cb      : (err, listing) =>
+                        cb      : (err, item) =>
                             if err
-                                alert_message(type:'error', message:'Failed to open link')
+                                console.log err
+                                @open_directory(parent_path)
                             else
-                                item = listing.find (val) => val.get('name') == last
                                 if item.get('isdir')
                                     @open_directory(full_path)
                                 else


src/smc-webapp/project_store.coffee
```diff
@@ -1200,15 +1200,17 @@ class ProjectActions extends Actions
                 if target[target.length-1] == '/' or full_path == ''
                     @open_directory(parent_path)
                 else
-                    @fetch_directory_listing(parent_path)
+                    @fetch_directory_listing(parent_path, true, true)
                     @get_store().wait
-                        until   : (s) => s.directory_listings.get(parent_path)
+                        until   : (s) =>
+                            listing = s.directory_listings.get(parent_path)
+                            return listing?.find (val) => val.get('name') == last
```
williamstein: Just a comment -- no need to do anything now:   It seems like this line is doing a linear search through a potentially large data structure, which is clearly not the optimal way to structure our data -- instead shouldn't there be a map, so it's a O(log(N)) search.

src/smc-webapp/project_store.coffee
```diff
@@ -1200,15 +1200,17 @@ class ProjectActions extends Actions
...
                         timeout : 3
-                        cb      : (err, listing) =>
+                        cb      : (err, item) =>
                             if err
-                                alert_message(type:'error', message:'Failed to open link')
+                                console.log err
```
williamstein: This console.log probably shouldn't be here.  If you really have an error that the user should know about, it shouldn't be hidden in the console.   Instead give an alert message that is helpful, e.g., include the path in it.

It seems like you changed the code to basically silently ignore an error and do something different (show a directory).

src/smc-webapp/project_store.coffee
```diff
@@ -1200,15 +1200,17 @@ class ProjectActions extends Actions
...
                     @open_directory(parent_path)
                 else
-                    @fetch_directory_listing(parent_path)
+                    @fetch_directory_listing(parent_path, true, true)
                     @get_store().wait
-                        until   : (s) => s.directory_listings.get(parent_path)
+                        until   : (s) =>
+                            listing = s.directory_listings.get(parent_path)
+                            return listing?.find (val) => val.get('name') == last
                         timeout : 3
```
williamstein: This timeout of 3s makes me very nervous!  Can't all this easily take much longer, especially if project isn't opened or even connected to.

src/smc-webapp/project_store.coffee
johnjeng: Yes, this is true. I wonder what most number of children any folder has right now though.

src/smc-webapp/project_store.coffee
johnjeng: This is also true. What about 30 seconds? Or even longer but with an intermediary loading notification?

src/smc-webapp/project_store.coffee
johnjeng: Yes, because right now I can't seem to reproduce #1359 anywhere except on live and even then only rarely. In this case, I want to find out what the error is first because the error might not even matter to the user. 

I think opening the parent directory is fine though.

src/smc-webapp/project_store.coffee
williamstein: 30s is probably reasonable for now -- better than 3s.

src/smc-webapp/project_store.coffee
williamstein: Yes, doing that is fine if the error isn't silent. 

Any displayed red error any user sees (which appears in the upper right) gets logged to the database, btw... in case you were to want to search them.

src/smc-webapp/project_store.coffee
johnjeng: I think right now I should just write a query for a checking for directory or file. I'll look into it. I can probably just adjust `fetch_directory_listing`.

williamstein: Agreed.  As I said, I don't consider this a show stopper either.   Just something to keep in mind.

PR #1371, opened 2016-12-13
Always render \$ as $.

### LINKED ISSUE #1329: \$ gets rendered as \$ in markdown
E.g., try this in a chat message
```
Get some $ money \$
```
and you'll see `\$` rather than just $.

johnjeng:
Fixes #1329 in all markdown renderers which use the `r_misc.Markdown` component.

### DIFF
--- a/src/smc-webapp/r_misc.cjsx
+++ b/src/smc-webapp/r_misc.cjsx
@@ -713,7 +713,7 @@ exports.Markdown = rclass
 
     update_escaped_chars: ->
         node = $(ReactDOM.findDOMNode(@))
-        node.text(node[0].innerText.replace(/\\\$/g, '$'))
+        node.html(node[0].innerHTML.replace(/\\\$/g, '$'))
 
     update_mathjax: ->
         if @_x?.has_mathjax?
--- a/src/smc-webapp/r_misc.cjsx
+++ b/src/smc-webapp/r_misc.cjsx
@@ -711,6 +711,10 @@ exports.Markdown = rclass
     shouldComponentUpdate: (newProps) ->
         return @props.value != newProps.value or not underscore.isEqual(@props.style, newProps.style)
 
+    update_escaped_chars: ->
+        node = $(ReactDOM.findDOMNode(@))
+        node.text(node[0].innerText.replace(/\\\$/g, '$'))
+
     update_mathjax: ->
         if @_x?.has_mathjax?
             $(ReactDOM.findDOMNode(@)).mathjax()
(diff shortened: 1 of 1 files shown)


williamstein:
Don't make the \$ --> $ change after rendering everything to the DOM.  Instead, do the \$ --> $ transform in the function to_html, which does the markdown to html computation in the first place (and all kinds of sanitization).   In particular, investigate why one of these
```
     @_x = markdown.markdown_to_html(v)
     v = require('./misc_page').sanitize_html(v)
```
doesn't already do the \$ --> $ transform.

johnjeng:
The problem with that is mathjax gets applied afterwards. If we transform `\$` to `$` before mathjax, mathjax will parse it as `$ ... $` which is probably not intended at all. 

Oh, a question though: What do you think about behavior inside ` ``` `? What should
```
'''
\\\$$$
'''
```
render to?
I'm thinking
```
\\\$$$
```

(This PR needs fixing in that regard too then)

williamstein:
> The problem with that is mathjax gets applied afterwards.

Mathjax runs asynchronously, so the solution you propose will work on small examples and fail on large ones. 

HEH, check out http://docs.mathjax.org/en/latest/tex.html#tex-and-latex-math-delimiters  !  Mathjax itself has an option `processEscapes: true` to deal with all this itself.   @haraldschilly -- can you look into this tomorrow...?

johnjeng:
>Mathjax runs asynchronously, so the solution you propose will work on small examples and fail on large ones.

Oh fun. Well, I'll let @haraldschilly look at this and see what he comes up with.

haraldschilly:
We have `processEscapes` enabled since ever: [setup is here](https://github.com/sagemathinc/smc/blob/master/src/smc-webapp/last.coffee#L67)

what you want is to remove a character which you've actually typed and has no additional meaning. @johnjeng's patch is simple, but does that. I wouldn't expect it to do this in an blockquote code environment, yes.

what I don't understand so far is, why you really want to do that? Maybe we should just improve the heuristic for formula parsing and forget about dealing with a backslash? My first thought is, that I'm always writing formulas in dollar signs, when they're right next to another character. I.e. `$1+1$` and not `$ 1 + 1 $`. We could detect that by looking for no space character inside of them. Not sure how, though.

williamstein:
"johnjeng <https://github.com/johnjeng>'s patch is simple, but does that"

Since mathjax processes the DOM asynchronously, I'm not convinced this will
always work.

williamstein:
> mathjax... async: 

Of course, that async issue is easily solved by only running the substitution code when mathjax says it is done: https://github.com/mathjax/mathjax-docs/wiki/Event-when-typesetting-is-done%3F-Rescaling-after-rendering...

williamstein:
OK, I'm just going to merge this less-than-perfect approach for now.

PR #1373, opened 2016-12-13
Fix bug where project log search would not show

johnjeng:
Fix a bug that by luck hasn't shown up before.


williamstein:
I noticed this (already there) and fixed it when refereeing (in cjsx):
```
value        = @props.search
```

Issue #1374, opened 2016-12-13
project creation user interface polish


williamstein:
What happens now:

1. Click to create a new project and then click the "Hosting and network upgrades..." button.
2. Receive a horrible terrifying jarring experience as the screen glitches leaving one looking at this:
<img width="853" alt="screen shot 2016-12-12 at 10 02 34 pm" src="https://cloud.githubusercontent.com/assets/1276278/21129516/cee3cea6-c0b6-11e6-9350-e0327d063677.png">

What should happen:

1. Click to create a new project and then click the "Hosting and network upgrades..." button.
2. Absolutely **NO AUTOMATIC SCROLLING**, and instead a small info box appears before the button with the "Create this project on a members-only host with full network access" content.  Exactly like happens all over in account settings, billing, and course management.  E.g.,

timothyclemansinsea:
@williamstein OK so are you saying you want "Create this project on" directly below the buttons toolbar for different ways to create project? And what about when clicking the custom upgrades button?

williamstein:
> OK so are you saying you want "Create this project on" directly below the buttons toolbar for different ways to create project? 

**YES**

> And what about when clicking the custom upgrades button?

I'm not saying anything about that.

PR #1378, opened 2016-12-14
polish create project with hosting and network upgrades

timothyclemansinsea:
For #1374


williamstein:
I just tested this and it seems that you (1) fixed the position, but (2) missed my other request about scrolling, so I've made it very clear here:

## **Absolutely NO AUTOMATIC SCROLLING**!

williamstein:
Further complaint: there is a typo you introduced in your code -- "memebers".  It's "members".

PR #1380, opened 2016-12-15
Re-generate Editors if Client relationship to them changes

### LINKED ISSUE #1263: opening a private file after being signed out doesn't refresh on sign in
1. goto a private file (not public!), e.g. `test.sagews` or `test.ipynb` (with ipynb, there isn't even a visible error message
1. copy the exact URL
1. goto account, sign out, let it reload
1. paste the copied URL and load the page ... the point is, that you tell the webapp that you want to go exactly to that file while you're not signed in
1. the file opens with the error message `Error opening file -- path 'insert.sagews' of project with id '955988e5-0998-49de-9833-e66b683c1c05' is not public`

so far, so good

1. now goto account's "sign in" and sign in.
1. you're diverted back to the file, but it hasn't reloaded. it shows the same error message. this failed update is even more pronounced with ipynb and probably extends to the whole project.
1. going to files for a listing and re-opening the file solves it, so it's definitely a problem of state staleness.

johnjeng:
Fixes #1263.

### DIFF
--- a/src/smc-webapp/project_store.coffee
+++ b/src/smc-webapp/project_store.coffee
@@ -354,6 +354,11 @@ class ProjectActions extends Actions
                         # Only generate the editor component if we don't have it already
                         # Also regenerate if view type (public/not-public) changes
                         if not open_files.has(opts.path) or open_files.getIn([opts.path, 'component']).is_public != is_public
+                            was_public = open_files.getIn([opts.path, 'component'])?.is_public
+                            if was_public? and was_public != is_public
+                                @setState(open_files : store.open_files.delete(opts.path))
+                                project_file.remove(opts.path, @redux, @project_id, was_public)
+
                             open_files_order = store.open_files_order
 
                             # Initialize the file's store and actions
--- a/src/smc-webapp/project_store.coffee
+++ b/src/smc-webapp/project_store.coffee
@@ -181,6 +181,9 @@ class ProjectActions extends Actions
                 @redux.getActions('file_use')?.mark_file(@project_id, path, 'open')
                 @push_state('files/' + path)
                 @set_current_path(misc.path_split(path).head)
+                if redux.getStore('projects').get_my_group(@project_id) != store.open_files.getIn([path, 'component']).is_public
+                    @open_file(path : path)
+                    # Reopen the file if relationship has changed
 
     add_a_ghost_file_tab: () =>
         current_num = @get_store().num_ghost_file_tabs
@@ -1214,7 +1217,8 @@ class ProjectActions extends Actions
                                 alert_message(type:'error', message:"There was an error related to opening the link: #{err}")
                                 @open_directory(parent_path)
                             else
-                                if item.get('isdir')
+                                item = listing?.find (val) => val.get('name') == last
+                                if item?.get('isdir')
                                     @open_directory(full_path)
                                 else
                                     @open_file
--- a/src/smc-webapp/project_store.coffee
+++ b/src/smc-webapp/project_store.coffee
@@ -622,7 +622,6 @@ class ProjectActions extends Actions
             map = store.directory_listings.set(path, if err then misc.to_json(err) else immutable.fromJS(listing.files))
             @setState(directory_listings : map)
             delete @_set_directory_files_lock[_key] # done!
-            opts.cb(err, listing)
         )
 
     # Increases the selected file index by 1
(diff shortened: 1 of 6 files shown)


johnjeng:
Conflicts need resolving and testing redone.

src/smc-webapp/project_store.coffee
```diff
@@ -181,6 +181,9 @@ class ProjectActions extends Actions
                 @redux.getActions('file_use')?.mark_file(@project_id, path, 'open')
                 @push_state('files/' + path)
                 @set_current_path(misc.path_split(path).head)
+                if redux.getStore('projects').get_my_group(@project_id) != store.open_files.getIn([path, 'component']).is_public
```
williamstein: This line of code **can't possibly be correct** because the left hand side is a string such as 'owner', 'collaborator', 'public' , and the right hand side is true or false (or undefined).

williamstein:
Looks good except for one line that seems wrong to me.

williamstein:
By the way, I'm **really glad** you are fixing this!  The analogue of this is horribly broken at https://ask.sagemath.org/questions/, which is one reason I don't like that site at all.

src/smc-webapp/project_store.coffee
```diff
@@ -1200,18 +1210,29 @@ class ProjectActions extends Actions
...
+                                foreground : foreground
+                                foreground_project : foreground
+                    else
+                        @fetch_directory_listing
+                            path         : parent_path
+                            show_hidden  : true
+                            finish_cb    : =>
+                                store = @get_store()
+                                listing = store.directory_listings.get(parent_path)
+                                item = listing?.find (val) => val.get('name') == last
```
johnjeng: Want to try exactly once after `fetch_directory_listing` so `store.wait` isn't any good.

williamstein:
How do you test this code?  What is it supposed to do?  I tried the following (after full rebuild repeatedly and clearing browser cache):

1. Sign out from an smc-in-smc project.
2. Paste in the URL of a project with some public files.
3. Open one of them (html say)
4. Click sign in and sign in (as the owner of those files).
5. Look at the html file I opened -- it's still opened as a public editor.  Same with the jupyter notebook and sagews I tested.

Another bug I noticed -- If I have chat opened for a file when signed in, when I later try to open that same file publicly it consults local storage and decides to open chat too... which makes no sense in public mode (or maybe chat should be available?).

Anyway, as far as I can tell, this code doesn't work at all.  What am I doing wrong?  What were you testing?

johnjeng:
I tested with markdown files. Testing with HTML files gives your results. Time to look more deeply.

johnjeng:
> Another bug I noticed

This is true on the live servers. I'll look into fixing this bug afterwards #1384

johnjeng:
Okay, this should work now. Tested on test.sagemath.com

src/smc-webapp/projects.cjsx
```diff
@@ -157,13 +157,10 @@ class ProjectsActions extends Actions
...
         store = redux.getProjectStore(opts.project_id)
         actions = redux.getProjectActions(opts.project_id)
-        sort_by_time = store.sort_by_time ? true
-        show_hidden = store.show_hidden ? false
-
         relation = redux.getStore('projects').get_my_group(opts.project_id)
         if not relation? or relation in ['public', 'admin']
             @fetch_public_project_title(opts.project_id)
-        actions.fetch_directory_listing(store.current_path, sort_by_time, show_hidden)
+        actions.fetch_directory_listing()
```
williamstein: What a great change!

williamstein:
> Okay, this should work now. Tested on test.sagemath.com

It does not work at all for me, either on test.sagemath.com or when I install this myself.  I won't look at this again until you give step-by-step instructions that you promise you already carefully tested.  I will then start by verifying that those work, before testing other related things.  As it is now, I can't seem to get anything to work -- but maybe that is me just not understanding what this PR is supposed to accomplish.  If you give me step-by-step instructions (with expectations), at least I know what you think works.

johnjeng:
Video of what I've tested. https://youtu.be/cLKVd3TmeSo

1. Be logged out
2. Go to url for a public file
3. Click Sign in.
4. Sign in.
5. SMC should load into the file you were on and it should no longer be public view.

johnjeng:
Merge conflicts.

johnjeng:
Manually tested with the following steps:

1. Open a public file as the owner while logged in.
2. Copy the URL
3. Log out
5. Paste the public file URL into the address bar
6. Wait until the file loads
7. Sign in
8. See the the file should have refreshed
9. Do step 5 again but open more public files

The last step should result in the first file being opened after logging in. In any case, opening files after logging in should refresh the view to editable.

johnjeng:
This should also fix opening TimeTravel from the project log. 

Tested with the following steps:
1. Log in
2. Open any file with TimeTravel (I created and opened a `.sagews`)
3. Open TimeTravel for the file
4. Close both files
5. Go to the Project Log
6. Open the `.sage-history` log entry.

williamstein:
What you implemented does what you claim, and I'll merge this asap.  Thanks!

PR #1383, opened 2016-12-16
sagews2html: embedding videos

haraldschilly:
error handling is missing. besides that a first working idea... I'll finish this another day.

### DIFF
--- a/src/smc-webapp/printing.coffee
+++ b/src/smc-webapp/printing.coffee
@@ -360,6 +360,7 @@ class SagewsPrinter extends Printer
                         out += "<div class='output sage3d'><img src='#{data_url}'></div>"
                 else if ext == 'webm'
                     out ?= ''
+                    # 'raw' url. later, embed_videos will be replace by the data-uri if there is no error
                     out += "<video src='#{mesg.file.url}' class='sagews-output-video' controls></video>"
                 else
                     # console.log 'msg.file', mark, mesg
@@ -384,9 +385,7 @@ class SagewsPrinter extends Printer
             # ignored
         else
             console.warn "ignored mesg", mesg
-        html = @html_post_process(out)
-        if html?
-            @_html.push(html)
+        return @html_post_process(out)
 
     html_post_process: (html) ->
         # embedding images and detecting a title
@@ -432,7 +431,8 @@ class SagewsPrinter extends Printer
         # the following fits mentally into sagews.SynchronizedWorksheet
         # progress takes two arguments: a float between 0 and 1 [%] and optionally a message
         {MARKERS}    = require('smc-util/sagews')
-        @_html       = [] # list of elements
+        _html        = [] # list of elements
+        full_html    = '' # end result of html content
         @_title      = null # for saving the detected title
         @_output_ids = {} # identifies text marker elements, to avoid printing show-plots them more than once!
         cm           = @editor.codemirror
@@ -480,8 +480,8 @@ class SagewsPrinter extends Printer
                     break
             if input_lines.length > 0
                 input_lines = input_lines.map(process_line).join('') # no \n linebreaks!
-                #@_html.push("<div class='mode'>#{input_lines_mode ? input_lines_default_mode} mode")
-                @_html.push("<pre class='input'>#{input_lines}</pre>")
+                #_html.push("<div class='mode'>#{input_lines_mode ? input_lines_default_mode} mode")
+                _html.push("<pre class='input'>#{input_lines}</pre>")
             input_lines      = []
             input_lines_mode = null
 
@@ -490,10 +490,11 @@ class SagewsPrinter extends Printer
         process_collected_mesg_stdout = =>
             # processing leftover stdout mesgs from previous iteration
             if mesg_stdout.stdout.length > 0
-                @html_process_output_mesg(mesg_stdout)
+                # it's ok to leave `mark` undefined
+                om = @html_process_output_mesg(mesg_stdout)
+                _html.push(om) if om?
                 mesg_stdout = {stdout : ''}
 
-        _html = ''
         process_lines = (cb) =>
             # process lines in an async loop to avoid blocking on large documents
             line = 0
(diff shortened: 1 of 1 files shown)


src/smc-webapp/printing.coffee
```diff
@@ -356,6 +358,9 @@ class SagewsPrinter extends Printer
                         data_url = scene.static_image
                         out ?= ''
                         out += "<div class='output sage3d'><img src='#{data_url}'></div>"
+                else if ext == 'webm'
+                    out ?= ''
+                    out += "<video src='#{mesg.file.url}' class='sagews-output-video' controls></video>"
```
williamstein: I'm sure you're aware, and I agree with this -- but unlike with images, this will require internet access to work...

src/smc-webapp/printing.coffee
```diff
@@ -406,6 +411,7 @@ class SagewsPrinter extends Printer
             img.height = scaling * img.height
             ext = misc.filename_extension(img.src).toLowerCase()
             ext = ext.split('?')[0]
+            ext = if ext == 'jpg' then 'jpeg' else ext
```
williamstein: `if ext == 'jpg' then ext = 'jpeg'` is shorter and clearer, imho.

src/smc-webapp/printing.coffee
```diff
@@ -525,6 +532,9 @@ class SagewsPrinter extends Printer
                 process_collected_mesg_stdout()
                 line++
             input_lines_process(final = true)
+            # combine all html snippets to one html block
+            @_html = (h for h in @_html).join('\n')
```
williamstein: This looks suspicious to me, since for the right hand side to work, `@_html` must be an array, and after this code runs, then `@_html` is a string.  However, this is done inside of a function which is clearly called many times.   So `@_html` must be getting changed back and forth between array and string, which is just a little nerve racking to me (who grew up in statically typed C++), and feels like it might make surrounding code a little confusing to work on.  I'm sure it works fine... but I wonder if you could change it to not change the data type of `@_html`?

src/smc-webapp/printing.coffee
```diff
@@ -547,6 +557,36 @@ class SagewsPrinter extends Printer
                 cb        : (err) ->
                     cb(err, data_base64)
 
+        $html = null
```
williamstein: For the record, I think we've never once used a variable name starting with $ in all the SMC source code, since it's not valid in Python, and I try to have a somewhat pythonic style.

src/smc-webapp/printing.coffee
```diff
@@ -547,6 +557,36 @@ class SagewsPrinter extends Printer
...
+        embed_videos = (cb) =>
+            # downloading and embedding all video files (especially for animations)
+            # _html is a string and we have to wrap this into a div
+            $html = $('<div>' + _html + '</div>')
+            vids = (vid for vid in $html.find('video'))
+            vids_num = 0
+            vids_tot = vids.length
+            vembed = (vid, cb) ->
+                console.log "embedding #{vids_num}/#{vids_tot}", vid
```
williamstein: Comment out this console.log?

src/smc-webapp/printing.coffee
```diff
@@ -547,6 +557,36 @@ class SagewsPrinter extends Printer
                 cb        : (err) ->
                     cb(err, data_base64)
 
+        $html = null
+        embed_videos = (cb) =>
+            # downloading and embedding all video files (especially for animations)
+            # _html is a string and we have to wrap this into a div
+            $html = $('<div>' + _html + '</div>')
+            vids = (vid for vid in $html.find('video'))
```
williamstein: I think `vids = $html.find('video')` would work just as well, since isn't a jquery wrapped object iterable?

src/smc-webapp/printing.coffee
```diff
@@ -547,6 +557,36 @@ class SagewsPrinter extends Printer
...
+                progress(.4 + (5 / 10) * (vids_num / vids_tot), "video #{vids_num}/#{vids_tot}")
+                xhr = new XMLHttpRequest()
+                xhr.open('GET', vid.src)
+                xhr.responseType = 'blob'
+                xhr.onreadystatechange = ->
+                    if this.readyState == 4 and this.status == 200
+                        blob = this.response
+                        reader = new FileReader()
+                        reader.addEventListener "load", ->
+                            console.log(reader.result[..100])
```
williamstein: Comment out console.log?  (Or use a function log with a DEBUG flag?)

src/smc-webapp/printing.coffee
```diff
@@ -577,7 +617,7 @@ class SagewsPrinter extends Printer
                     console.debug("write_text_file_to_project.resp: '#{resp}'")
                     cb?(err)
 
-        async.parallel([sagews_data, process_lines], finalize)
+        async.parallel([sagews_data, process_lines, embed_videos], finalize)
```
williamstein: Stupid annoying overall question: why is this all written using jquery and direct construction of HTML in the DOM?  Couldn't it be done in a cleaner, more easily debugable and maintained way by using React instead?   I know React is harder to write, but I think it is easier to debug, improve, and reason about, especially for non-authors...  Is there a good reason for not using react?

src/smc-webapp/printing.coffee
haraldschilly: no, it won't work if _html is `<video> ... </video>`

src/smc-webapp/printing.coffee
haraldschilly: yes, that's one of the things I had to change in order to find a way to do this with the video tag. the `process_lines` is only called once and I think it's best to drop `@_html` entirely.

src/smc-webapp/printing.coffee
haraldschilly: It's being replaced in `embed_videos` with such a data uri, but only if there isn't a problem getting the file. I wished there is a way to do the same canvas trick as with images, though.

Issue #1310, opened 2016-11-26
Make it easy to steal quota from other projects when upgrading


timothyclemansinsea:

timothyclemansinsea:
@williamstein I'm changing this to M-high because it would lay the foundation for adding upgrade in place to #1583

haraldschilly:
licenses move automatically between projects, depending on their run state. no need for this any more.

PR #1399, opened 2016-12-19
add ability to steal quota

timothyclemansinsea:
For #1310

### DIFF
--- a/src/smc-webapp/r_misc.cjsx
+++ b/src/smc-webapp/r_misc.cjsx
@@ -1329,6 +1329,45 @@ exports.UpgradeAdjustor = rclass
     render_addon: (misc, name, display_unit, limit) ->
         <div style={minWidth:'81px'}>{"#{misc.plural(2,display_unit)}"} {@render_max_button(name, limit)}</div>
 
+    remove_upgrade_from_all: (projects_with_this_upgrade, name) ->
+        for project_id in projects_with_this_upgrade
+            @remove_upgrade_from_project(project_id, name)
+
+    render_remove_upgrade_from_all: (projects_with_this_upgrade, name) ->
+        <MenuItem
+            eventKey="0"
+            onClick={=>@remove_upgrade_from_all(projects_with_this_upgrade, name)}>
+            All
+        </MenuItem>
+
+    remove_upgrade_from_project: (project_id, name) ->
+        quotas_to_apply = redux.getStore('projects').get_projects_upgraded_by()[project_id]
+        console.log(JSON.stringify(quotas_to_apply))
+        console.log('name', name)
+        delete quotas_to_apply[name]
+        console.log(JSON.stringify(quotas_to_apply))
+        redux.getActions('projects').apply_upgrades_to_project(project_id, quotas_to_apply)
+
+    render_remove_upgrade_from_menuitems: (projects_with_this_upgrade, name) ->
+        for project_id, i in projects_with_this_upgrade
+            <MenuItem 
+                eventKey={i}
+                onClick={=>@remove_upgrade_from_project(project_id, name)}>
+                {redux.getStore('projects')?.get_title(project_id)}
+            </MenuItem>
+
+    render_remove_upgrade_from: (name) ->
+        upgrades = redux.getStore('projects').get_projects_upgraded_by()
+        projects_with_this_upgrade = []
+        for project_id, project_upgrades of upgrades
+            if project_upgrades[name]
+                projects_with_this_upgrade.push(project_id)
+        if projects_with_this_upgrade.length > 0
+            <DropdownButton className="small" title="Remove upgrade from" key={name} id={name}>
+                {@render_remove_upgrade_from_all(projects_with_this_upgrade, name) if projects_with_this_upgrade.length > 1}
+                {@render_remove_upgrade_from_menuitems(projects_with_this_upgrade, name)}
+            </DropdownButton>
+
     render_upgrade_row: (name, data, remaining=0, current=0, limit=0) ->
         if not data?
             return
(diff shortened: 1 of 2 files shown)


src/smc-webapp/r_misc.cjsx
```diff
@@ -1329,6 +1329,45 @@ exports.UpgradeAdjustor = rclass
...
+            onClick={=>@remove_upgrade_from_all(projects_with_this_upgrade, name)}>
+            All
+        </MenuItem>
+
+    remove_upgrade_from_project: (project_id, name) ->
+        quotas_to_apply = redux.getStore('projects').get_projects_upgraded_by()[project_id]
+        console.log(JSON.stringify(quotas_to_apply))
+        console.log('name', name)
+        delete quotas_to_apply[name]
+        console.log(JSON.stringify(quotas_to_apply))
```
johnjeng: You still have some console logs in here :P

johnjeng: You still have some console logs in here :P

timothyclemansinsea:
@johnjeng fixed the console logs

src/smc-webapp/r_misc.cjsx
```diff
@@ -1329,6 +1329,42 @@ exports.UpgradeAdjustor = rclass
...
+        quotas_to_apply = redux.getStore('projects').get_projects_upgraded_by()[project_id]
+        delete quotas_to_apply[name]
+        redux.getActions('projects').apply_upgrades_to_project(project_id, quotas_to_apply)
+
+    render_remove_upgrade_from_menuitems: (projects_with_this_upgrade, name) ->
+        for project_id, i in projects_with_this_upgrade
+            <MenuItem 
+                eventKey={i}
+                onClick={=>@remove_upgrade_from_project(project_id, name)}>
+                {redux.getStore('projects')?.get_title(project_id)}
```
johnjeng: Formatting. I think we need a linter.

johnjeng:
```
<MenuItem 
    eventKey={i}
    onClick={=>@remove_upgrade_from_project(project_id, name)}
>
    {redux.getStore('projects')?.get_title(project_id)}
```

src/smc-webapp/r_misc.cjsx
```diff
@@ -1329,6 +1329,42 @@ exports.UpgradeAdjustor = rclass
...
+        <MenuItem
+            eventKey="0"
+            onClick={=>@remove_upgrade_from_all(projects_with_this_upgrade, name)}>
+            All
+        </MenuItem>
+
+    remove_upgrade_from_project: (project_id, name) ->
+        quotas_to_apply = redux.getStore('projects').get_projects_upgraded_by()[project_id]
+        delete quotas_to_apply[name]
+        redux.getActions('projects').apply_upgrades_to_project(project_id, quotas_to_apply)
```
johnjeng: This looks like this removes the upgrade when it gets selected. I think expected behavior is for all the removals to happen when the user saves the changes.

timothyclemansinsea:
*@johnjeng* I think the behavior should be remove immediately. How else is
the user supposed to get alerts about the quota they want being too high?
Seems like doing it as you suggest would complicate things a lot.

johnjeng:
Complicate things a lot of us or for the user? For the user it's not any more complicated. They still get warnings about choosing quotas that are too high. We just have to know if it's too high based on the quotas they will have available after the removal operations. 

Why only actually remove them after saving? Because I think a common pattern would be "Let me adjust these quotas... Remove this one... and this one... Oh, I don't have enough that I want to free from other projects (because they're more important). Never mind then. *Cancel*. " In that scenario, I would be very surprised to find that the quotas got removed from the first two projects. I certainly don't want to manually have to reassign them.

timothyclemansinsea:
How about we just make it easy to undo stealing?

johnjeng:
That's what I'm saying. There's nothing as easy as hitting cancel.

timothyclemansinsea:
@johnjeng I list out the removals with undo buttons

johnjeng:
Design concerns:
1. What happens when I have a ton of projects? This list is going to be gigantic
2. I can only remove the total for any one upgrade. It would be common that I only want to remove some of something from a different project
3. I still think undoing is too complicated. I want to think of this as one operation: Move upgrades from my stash or from other projects to this project. *Not* "Move upgrades from my stash or move upgrades from other projects to the stash then move those upgrades from my stash to this project". This way it's a single operation I can revert from doing with a single

timothyclemansinsea:
@johnjeng it's going to take me a few hours to get this coded up to be one operation 
I've started with the checkboxes. Only show the dropdown if the particular upgrade is maxed out. The dropdown now is ten projects in order of last edited ascending. The theory being that if I haven't used a project in ten months that I'm more likely to steal from it then I am for one I used yesterday. For the non checkbox upgrades I'll make a more complex thing rather a drop down. 
image

williamstein:
> The dropdown now is ten projects in order of last edited ascending. The theory being that if I haven't used a project in ten months that I'm more likely to steal from it then I am for one I used yesterday. 

Quick remark:  if you're a teacher or TA for a course (so the most likely person to have upgrades to apply by far), then you can easily have 100s of projects you own, and they could all be being modified very recently.    We really need a better component to select from all projects, and *any* such selector has to at least provide all projects (not just 10).   This complexity is also a problem for copying directories between projects.    

I just want you to be aware of the above.  it doesn't mean that the first version -- just show last 10 **then the rest** isn't OK for now.  Also, note that as an instructor, most of those projects are hidden, so maybe aren't in your list at all.

timothyclemansinsea:
@williamstein thank you for the heads up.

timothyclemansinsea:
@williamstein @johnjeng what I'm thinking is to stick with "ten projects in order of last edited ascending." for this pull request and make handling the hundreds of projects one might have another issue.

williamstein:
> what I'm thinking is to stick with "ten projects in order of last edited ascending." for this pull request 

I think that's fine; somebody can just ignore this functionality entirely if they have a class.

timothyclemansinsea:
Closing because scope, trying to make it look pretty and add stealing "some", is too much for me to handle.

Issue #1316, opened 2016-11-28
set "reply-to" field in invitation emails to user's email address


haraldschilly:
in transactional invitation emails, set the "reply-to" email address to the user's email address. don't set the "from" email field, otherwise it's spam and breaks dkim!

the outcome should be, that replies to invitations no longer end up in our email servers.

(and well, if there is no email address set, don't set the reply-to field)

haraldschilly:
working on this one, which will include a full api update to version 3 (our version 2 is deprecated)

williamstein:
you did this already.

PR #1406, opened 2016-12-20
webapp/hub: setting reply-to in emails such that recipients automatically contact the correct one

haraldschilly:
(unless email_address is undefined, then it falls back) and updating the sendgrid API to v3. -- issue #1316

for testing: update the sendgrid npm package, then send an invitation email and a course invite. in both cases, the normal "reply" in the email will be the actual user, not help@sagemath.com. also, in gmail, when clicking on "show original" in the email's drop down, SPF and DKIM both pass.


williamstein:
Looks (very) good to me.  I didn't actually test, but I trust you did via test.sagemath.com or something.

haraldschilly:
I've invited myself on the prod site and sure indeed, the email arrived just like it should.

Issue #1408, opened 2016-12-21
missing error handling for directory listings


haraldschilly:
This is something I've discovered in `smc-webapp/project_store.coffee` and either needs fixing or I'm misunderstanding it.

Assume, there is an error getting the file listing via `fetch_directory_listing` in line around 617:

```
map = store.directory_listings.set(path, if err then misc.to_json(err) else immutable.fromJS(listing.files))
@setState(directory_listings : map)
```

Then, a bit later, in line 1207

```
@get_store().wait
  until   : (s) =>
    listing = s.directory_listings.get(parent_path)
    return listing?.find (val) => val.get('name') == last
```

`find( ...)` doesn't work because it's not an immutable object, but a simple string.

Somewhere else in that file is a `if typeof(listing) == 'string'` check, maybe that needs to be consolidated.

haraldschilly:
Isn't that already fixed? I'm closing it.

PR #1409, opened 2016-12-21
webapp: dealing with file listing problems -- issue #1408

haraldschilly:
the way I did test it, is to reload the whole page while the url contains a full path to a file

### DIFF
--- a/src/smc-webapp/project_store.coffee
+++ b/src/smc-webapp/project_store.coffee
@@ -1538,7 +1538,8 @@ get_directory_listing = (opts) ->
         method = salvus_client.project_directory_listing
     else
         method = salvus_client.public_project_directory_listing
-    listing = undefined
+    listing     = undefined
+    listing_err = undefined
     f = (cb) ->
         method
             project_id : opts.project_id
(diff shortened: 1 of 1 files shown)


williamstein:
> the way I did test it, is to reload the whole page while the url contains a full path to a file

And...?  What is the expected behavior?  How does this actually test anything?  How do you simulate each of the errors.  

Please give me step by step instructions to test this PR.

williamstein:
I tried:
 - just refresh on a valid directory: works as expected, which I guess tests we didn't break things horribly again
 - refresh on a directory that doesn't exist -- Loading spins "forever" (several minutes), and I give up.  There *should be* an almost immediate error in this case.

williamstein:
I did read all this code and in certainly looks fine.   I'm OK with it being merged.   I just don't know what it is supposed to accomplish exactly or how to test it.

haraldschilly:
hmm, the proper way would be some mocking to trigger a file listing error. and yes, I meant to do this reloading, because that's where the codepath has those changes in it. reloading a non-existing directory should fail, yes, but I have no idea what's going on. I'll have a look.

haraldschilly:
I've pushed 05731ae which (hopefully) addresses that problem of opening a non-existing directory. The tricky detail is, that calling smc-ls works fine even though it returns an error. This "good" error triggers calling the function again (until the timeout of 4 minutes).

williamstein:
Fantastic!

PR #1416, opened 2016-12-27
safari mobile: make idle notification have a pointer cursor #1415

haraldschilly:
based on http://stackoverflow.com/a/17490775/54236 -- and I also added a second event listener and also check for touch movements.

@timothyclemansinsea please test if this helps with safari and chrome/WKWebView ... and well, it should preferably not break the existing behavior


williamstein:
100% works; completely fixes problem and doesn't break anything.

PR #1418, opened 2016-12-27
Issue/1412 -- making mathjax in chat great again

haraldschilly:
This is a fix for the failing mathjax processing. It also cleans up odd things with html sanitizing and only calls mathjax if the markdown processing told it that there are formulas (right now, it would do it for all messages or never)

Things to test
* lines like `xyz \$1+2+3+4\$ asdf $4^2$ $100` ← `\$` should show up as `$` while 4^2 and the dollar amount works.
* still sanitizing html (but not `inside code blocks`)

I also sneaked in a max-width of 100% for images ... was always bugging me

### DIFF
--- a/src/smc-webapp/_editor.sass
+++ b/src/smc-webapp/_editor.sass
@@ -589,3 +589,6 @@ h6 { font-size: 110% }
 
 .smc-message-from-viewer a:link, .smc-message-from-viewer a:visited
    color : #FFEB3B
+
+.smc-chat-message img
+  max-width: 100%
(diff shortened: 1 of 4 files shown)


haraldschilly:
> sanitizing html - I tested embedded html, it looks ok. Is that what needs testing?

actually no, but it's also no longer clear to me what exactly to test. the main thing to look for is if malformed html (e.g. just an open `<div>` but no closing div) do not break the html dom. That would lead to a broken webapp, and symptoms are that the page no longer updates correctly. what should not happen though, is that such a `<div>` in a code block (as it is here in my text) is corrected to have a closing tag.

Issue #977, opened 2016-10-01
Move sign out to top of account page


timothyclemansinsea:
Some users find it hard to find the sign out button. Let's move it to the very top of the account page.

haraldschilly:
well, it's already quite at the top. I think the issue is they don't know that this is in the account page.

haraldschilly:
closing this. at the moment it is at the top and always visible (i.e. sign out isn't hidden when you open another tab than the prefs)

![Screenshot from 2020-02-18 20-31-03](https://user-images.githubusercontent.com/207405/74771008-b217ec00-528d-11ea-82b6-c2d071ccce5e.png)

PR #1423, opened 2016-12-27
move sign out buttons to top

timothyclemansinsea:
For #977 

image


williamstein:
Put a screenshot...

williamstein:
Thanks for the screen shot.  Your buttons on the upper right push the entire account settings tabs and page down a full button height, which looks weird.    What about not pushing down?  

Also, screenshot when the screen is skinny...

timothyclemansinsea:
@williamstein I was not able to get the buttons on the same row as the tabs. So I put them in the right column at the top. Now I got to figure out how to make them appear at top when in small screen mode.

williamstein:
> Now I got to figure out how to make them appear at top when in small screen mode.

@timothyclemansinsea  -- You left this for five days...?

timothyclemansinsea:
I dont have a clue how to know if in single column mode. Best i know of is
is_mobile

johnjeng:
This breaks the design pattern for the page so I won't okay it. An alternative is to make a sign out button in the top nav which gives you an option to sign out everywhere or not when you click on it.

timothyclemansinsea:
@williamstein any thoughts on putting sign out in the topbar?

williamstein:
> any thoughts on putting sign out in the topbar?

No, because most people never have to click that button, and it'll just be sitting there wasting space all the time.  There should never be anything on the screen all the time that most people won't derive regular value from.   Look at your top bar right now.  Nothing there is redundant, and I bet you look at and use everything there.

williamstein:
Also, no major websites put sign out buttons at the top visible level, that I know of.

timothyclemansinsea:
OK then I guess I should close this PR then.

williamstein:
Where are the sign out buttons for Google, Facebook, and a few other sites?  The problem still stands -- people sometimes can't find the buttons.

timothyclemansinsea:
Google's sign out is on a dropdown. Facebook's sign out is pretty hidden in
a dropdown. Github's sign out is in a dropdown. Twitter's sign out is in a
dropdown.

williamstein:
> ... is in a dropdown...

Maybe we should have a dropdown?

johnjeng:
Maybe we should have a dropdown on our `Account` button?

williamstein:
@timothyclemansinsea -- thanks for investigating; @johnjeng  -- sounds good to me!

timothyclemansinsea:
Closing because it would take me about an hour or more to fix and this is not a M-high issue.

PR #1434, opened 2016-12-31
issue/1424 Rmd edit mode 

DrXyzzy:
Ref: #1424

Initial version, allows you to open a .Rmd file and see it recompiled in real time.

Adds `rmd2html.py` (committed in `smc_pyutil`) which must be in command path.

Features possibly to be added after this:
- view pdf
- view md
- clear knitr cache
- buttonbar
    - add `r` code chunk
    - menu of code chunk options

### DIFF
--- a/src/smc-webapp/editor-html-md/editor-html-md.coffee
+++ b/src/smc-webapp/editor-html-md/editor-html-md.coffee
@@ -54,6 +54,8 @@ class exports.HTML_MD_Editor extends editor.FileEditor
             @opts.mode = 'htmlmixed'
         else if @ext == 'md'
             @opts.mode = 'gfm'
+        else if @ext == 'rmd'
+            @opts.mode = 'gfm'
         else if @ext == 'rst'
             @opts.mode = 'rst'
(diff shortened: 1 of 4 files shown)


DrXyzzy:
Here's a sample .Rmd file. https://raw.githubusercontent.com/yihui/knitr-examples/master/001-minimal.Rmd

haraldschilly:
Well, long-term, I think that this won't be enough, but short term it's ok to have some sort of starting point.

However, this isn't enough to make it work. The missing link is to make that python script "installable". There is a little bit of python setup tools magic, which creates global commands: [look here into setup.py](https://github.com/sagemathinc/smc/blob/master/src/smc_pyutil/setup.py#L61) and add a line for `smc-rmd2html` and then fix the code to call `smc-rmd2html` and not the `*.py` file directly.

haraldschilly:
good enough for a first stealth release, but it's definitely not enough for a proper release.

also, I did ba746fb2ebc153a5609436656ae68f8b2240 to properly fix the image and object paths

DrXyzzy:
Thank you for the merge. It will save time and allow others to test more conveniently.

PR #1440, opened 2017-01-02
add %java

timothyclemansinsea:


src/smc_sagews/smc_sagews/sage_salvus.py
```diff
@@ -3731,7 +3731,64 @@ def go(s):
...
+        }
+
+    You can set the whole worksheet to be in java mode by typing
+
+        %default_mode java
 
+    NOTE:
+
+    - There is no relation between one cell and the next.  Each is a separate
+      self-contained go program, which gets compiled and run, with the only
```
williamstein: go --> java

src/smc_sagews/smc_sagews/sage_salvus.py
```diff
@@ -3731,7 +3731,64 @@ def go(s):
...
+# Java mode
+########################################################
+def java(s):
+    """
+    Run a Java program.  For example,
+
+        %java
+        public class YourName {
+            public static void main(String[] args) {
+
```
williamstein: maybe too much whitespace?

src/smc_sagews/smc_sagews/sage_salvus.py
```diff
@@ -3731,7 +3731,64 @@ def go(s):
...
+        sys.stdout.flush()
+        sys.stderr.flush()
+        if not os.path.exists(name+'.class'): # failed to produce executable
+            return
+        (child_stdin, child_stdout, child_stderr) = os.popen3('java %s'%name)
+        sys.stdout.write(child_stdout.read())
+        sys.stderr.write('\n'+child_stderr.read())
+        sys.stdout.flush()
+        sys.stderr.flush()
+    finally:
```
williamstein: When I tried this there was also some red in the output -- that's kind of weird, right?

timothyclemansinsea: @williamstein I'm not sure what the right thing to do about that is. I assume it is stderr

williamstein: It is stderr.  I'm not sure either.   We can leave it as is for now, but it is annoying, right?

timothyclemansinsea: @williamstein yep it is annoying. Maybe ignore those particular lines?

PR #1441, opened 2017-01-02
add a java file editor that shows output on the right

timothyclemansinsea:
I didn't want to write a bunch of code for this so I'm piggybacking off of the md file editor.

image

### DIFF
--- a/src/smc_pyutil/smc_pyutil/java2html.py
+++ b/src/smc_pyutil/smc_pyutil/java2html.py
@@ -1,0 +1,55 @@
+#!/usr/bin/python
+
+# rmd2html.py - used by rmd edit mode
+
+import os, sys, subprocess, errno, re
+
+def rmd2html(path):
+    if not os.path.exists(path):
+        raise IOError(errno.ENOENT, os.strerror(errno.ENOENT), path)
+
+    (root, ext) = os.path.splitext(path)
+    if ext.lower() != ".java":
+        raise ValueError('Rmd input file required, got {}'.format(path))
+    
+    with open(path, 'r') as f:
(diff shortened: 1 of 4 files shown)


src/smc_pyutil/smc_pyutil/java2html.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/python
+
+# rmd2html.py - used by rmd edit mode
+
+import os, sys, subprocess, errno, re
+
+def rmd2html(path):
```
haraldschilly: the documentation string above and this function name is definitely wrong

timothyclemansinsea: @haraldschilly fixed

williamstein:
This does seem to work.  It would be nice to somewhere show all output of compilation...

williamstein:
This makes me very nervous.  Right now, any time you make any change -- even 2 or 3 characters -- and pause, it will run javac, then run the program, and display the output, almost in real time.   There's not even a delay to wait for the user to click "Save", like with latex.  This design makes sense for the .rst mode, where compilation is very fast and predictable, but is kind of insane for compiling and running a java program. 

So modify the editor to not autosave on change for this java mode.  Probably the same should be done for the Rmd mode!

williamstein:
I just pushed the change I mentioned, namely only update on save.

williamstein:
The change  pushed is here: https://github.com/sagemathinc/smc/commit/6225d4a36bce79be55db9d6333a1b169c2a0def4

I couldn't push straight to this PR for some reason.

PR #1442, opened 2017-01-02
rmd2html: knitr::knit writes output file to current directory

DrXyzzy:
@haraldschilly please see if this fixes the problem of preview not being shown.

- fix for problem where `.md` file was written to different directory than `.Rmd` file
- save intermediate `.md` file

There may still be a problem with save button not being enabled. Workaround is that system autosaves within seconds after a change in edited text.

Ref: #1424 #1434

### DIFF
--- a/src/smc_pyutil/smc_pyutil/rmd2html.py
+++ b/src/smc_pyutil/smc_pyutil/rmd2html.py
@@ -12,10 +12,15 @@ def rmd2html(path):
     if ext.lower() != ".rmd":
         raise ValueError('Rmd input file required, got {}'.format(path))
 
-    cmd = '''Rscript -e "library(knitr); knit('{}')" >/dev/null 2>/dev/null'''.format(path)
+    (head,tail) = os.path.split(path)
+    if head == '':
+        head = "./"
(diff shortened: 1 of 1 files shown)


haraldschilly:
the PR https://github.com/sagemathinc/smc/pull/1443 would break what's going on here. I don't know if besides some path-changes there is anything else in this patch. If not, we should close it without merging.

DrXyzzy:
Appreciate the review.
- Besides path, a second change is that this PR saves the `knitr` .md output to a file as well as piping it to the front end. The md output file is sometimes needed for submitting assignments. Writing it at this point is a workaround for not having a UI option to create markdown.
- To use `warnings` variable, the script needs to be changed to delete `2>/dev/null` in the two `subprocess.call()` lines.
- Will the fix in PR #1443 work if smc-rmd2html is invoked when the user has current directory anywhere and the .Rmd file is at some possibly different path?
- To debug the previous problem I instrumented a local version of smc-rmd2html so that the internal shell commands were written to a log file. I didn't see an option in the smc_pyutils framework for debug mode.

williamstein:
@DrXyzzy please fix the conflicts with master.

DrXyzzy:
This PR is superseded by #1443 and should not be applied.

PR #1443, opened 2017-01-02
webapp/rmd: showing errors when back-end fails to run and switching path for subdirectories

haraldschilly:
@DrXyzzy  this patch does two things

1. it switches to the directory where the files is -- before `smc-rmd2html` is even called. hence, all relative files will be relative to it, etc.

2. there is a variable `warnings` collecting the `stderr` output. The only problem is, in `@to_html` this warnings variable from the callback is not used at all. Therefore, I wasn't able to see any issues but now I could.

what I don't understand is what is going on with saving the file.

### DIFF
--- a/src/smc-webapp/editor-html-md/editor-html-md.coffee
+++ b/src/smc-webapp/editor-html-md/editor-html-md.coffee
@@ -430,9 +430,11 @@ class exports.HTML_MD_Editor extends editor.FileEditor
         cb(undefined, m.s)
 
     rmd_to_html: (cb) =>
+        split_path = misc.path_split(@filename)
         @to_html_via_exec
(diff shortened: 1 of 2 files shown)


haraldschilly:
oh, and before I forget, your other `rmd` branch would break this.

DrXyzzy:
I added the changes we discussed this morning to `rmd2html.py`. plus one more:
1. use `os.chdir()` to set working directory before running subprocesses
1. remove `2>/dev/null` so `warning` will be returned
1. don't output to `stderr` if knitr completes without error (knitr writes to stderr on normal runs)

Tested in my smc-in-smc.

haraldschilly:
safe to merge and the changes in the .py script work

PR #1455, opened 2017-01-04
add note for educators on projects page

timothyclemansinsea:


williamstein:
Professors are our primary target market for marketing purposes.  They are not, however, likely to be more than 1-2% of of our **actual users**.  Thus I don't think it makes sense to clutter the UI with help text that is only of interest to them.   

Let's approach this differently.

A. We will discuss and add a new profile setting in account settings, with some entries like this:

   - I teach using SMC: 0-5 (or unset)
   - I use SMC in a class I'm taking: 0-5 (or unset)
   - I do research using SMC: 0-5 (or unset)
   - I write papers using SMC: 0-5 (or unset)

Users can fully set them however they want, and these scores will impact additional help advise like you have here.  

B. We take some actions to adjust the above scores, including:
  (a) politely requesting that people adjust them (maybe a quick pop-up survey, and then mention they can easily change their answers later).
  (b) we might iteratively adjust them based on things we observe, at least for new users.  E.g., if you create a course, then your "I teach using SMC" score could go up.  If you are in a course, then "I use SMC in a class I'm taking" goes up.

Thoughts?

timothyclemansinsea:
@williamstein what about a select defaulted to no on sign up "Are you an instructor?" or on projects page?

PR #1457, opened 2017-01-06
add smc-m2sagews utility

DrXyzzy:
Patterned after `smc-ipynb2sagews`, this PR adds a utility to take one or more `.m` files on the command line and make a `.sagews` file for each one, with a header cell setting the default mode and a single body cell containing the file contents.

Input file foo.m:
```
x = eye(5)
```

After running `smc-m2sagews foo.m`, get output file `foo.sagews`:
```
%auto
# This cell automatically evaluates on startup -- or run it manually if it didn't evaluate.
# Here, it starts the Jupyter octave kernel and sets it as the default mode for this worksheet.
%default_mode octave
--- 
# foo.m
x = eye(5)
```

### DIFF
--- a/src/smc_pyutil/setup.py
+++ b/src/smc_pyutil/setup.py
@@ -14,7 +14,7 @@ def readme():
     with open('README.md') as f:
         return f.read()
 
-from setuptools import setup
+from setuptools import setup, find_packages
 
 # This checks, if setup.py is run with 'install --user'
 # in that case we assume it is installed for development and do NOT change the python executable.
@@ -47,7 +47,7 @@ setup(
     author           = 'SageMath, Inc.',
     author_email     = 'office@sagemath.com',
     license          = 'GPLv3+',
-    packages         = ['smc_pyutil'],
+    packages         = find_packages(),
     install_requires = ['markdown2', 'psutil', 'PyYAML', 'ansi2html'],
     zip_safe         = False,
     classifiers      = [
--- a/src/smc_pyutil/setup.py
+++ b/src/smc_pyutil/setup.py
@@ -78,6 +78,7 @@ setup(
             'smc-git              = smc_pyutil.smc_git:main',
             'smc-html2sagews      = smc_pyutil.html2sagews:main',
             'smc-rmd2html         = smc_pyutil.rmd2html:main',
+            'smc-m2sagews         = smc_pyutil.m2sagews:main',
         ]
     },
     include_package_data = True
--- a/src/smc_pyutil/smc_pyutil/ipynb2sagews.py
+++ b/src/smc_pyutil/smc_pyutil/ipynb2sagews.py
@@ -35,149 +35,10 @@ import sys
 import os
 import codecs
 import textwrap
-import json
 # reading the ipynb via http://nbformat.readthedocs.io/en/latest/api.html
 import nbformat
-from ansi2html import Ansi2HTMLConverter
 
-from sws2sagews import MARKERS, uuid
-
-
-class IpynbCell(object):
-
-    '''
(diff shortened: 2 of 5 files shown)


haraldschilly:
I haven't looked at this for a long time and right now I'm pretty tired ... but I'm wondering if this isn't a massive code duplication? What about importing that "cell"-class from a common python file? (I.e. I'm thinking about something like `from smc_pyutil.lib import Cell` or so...

DrXyzzy:
Thanks for the comment. The utility is useful to me at the moment, so I thought to suggest it. I'm not sure it belongs in the release as is, either. I'll have a look at `from smc_pyutil.lib import Cell`.

Motivation for this PR is courses like the now-classic Machine Learning [MOOC](https://www.coursera.org/learn/machine-learning) from Stanford and Coursera, which assign homework in the form of Octave `.m` template files to be completed by the student.

Agree, there is duplication with `smc-ipynb2sagews`. We might extend to other types of files for similar teaching use cases, e.g. python2, python3, R. Should `Convert to (kernel type) Sage Worksheet` be an option in the UI for any 1-up programming language editor tab?

williamstein:
> Thanks for the comment. The utility is useful to me at the moment

@haraldschilly (and I) want this to exist!  It's just that it shouldn't duplicate code.  He's just saying: refactor this first so we can include it.

DrXyzzy:
Refactored to extract common class SagewsCell.
```
smc_pyutil/
  setup.py
  smc_pyutil/
    __init__.py
    ipynb2sagews.py
    m2sagews.py
    ...
    lib/
      __init__.py
      sagews_cell.py
```

williamstein:
@DrXyzzy: 

This looks fine to me, and I would merge it if you reminded me of how to run the test suite to verify that it doesn't break everything.  That said:

1. Wouldn't basically the **exact same** code work to do basically the same thing for a bunch of file types?    You could try to autodetect the input file type, but have an option to explicitly set it.  The script could just be called `smc-file2sagews` and could be used like this `smc-file2sagews options`.  I think pandoc's system for specifying file type behavior might be inspiring (I would look at it to design this).
2. Why not just create a worksheet and use some sort of load command to bring in the code?  What's the use case?  Not saying there isn't one!
3. This script (esp 1) could power a little converter that is run via an export button...

DrXyzzy:
0. To test the two modified/added utility commands, run them:

    - smc-inpynb2sagews filename.ipynb ...
    - smc-m2sagews filename.m ...

    AFAIK there is no test suite for smc_pyutil (yet?). And this PR makes no changes to the back end. The only slightly risky piece might be replacing literal `[smc_pyutil]` with `find_packages()` in setup.py. I could not test this as admin but it appears to work doing a user pip install of smc_pyutil.

1. A generic script could work although extension `.m` is used for Mercurial and Objective-C as well as Octave, and extension `.py` could be used to generate sage worksheets with any of several python2 and python3 kernels.
1. My use case is a bundle of .m files for each week's assignments in a class. I like running a shell command to create several sage worksheets at once. It looks as if there was a `cell.input()` command in [sage_salvus.py](https://github.com/sagemathinc/smc/blob/master/src/smc_sagews/smc_sagews/sage_salvus.py#L1151) at one time that might be used to load a cell from an object, but it's commented out, and would require reading in the file and creating the auto + default_mode header cell as well as the body cell. I played with it briefly, but personally am happier with shell utility at the moment.
1. I agree that once script(s) are built, it should be easy to add convert button(s) to the program file editor GUI for users who prefer point and click. These buttons could take care of kernel selection in case multiple kernels are available.

williamstein:
Got it. Merging.

PR #1461, opened 2017-01-07
webapp/latex: support for *.Rnw knitr files (like sagetex, but for R)

haraldschilly:
This is a minimal implementation for supporting `*.Rnw` files.

### DIFF
--- a/src/smc-webapp/editor.coffee
+++ b/src/smc-webapp/editor.coffee
@@ -117,6 +117,7 @@ codemirror_associations =
     pyx    : 'python'
     r      : 'r'
     rmd    : 'gfm2'
+    rnw    : 'stex2'
     rst    : 'rst'
     rb     : 'text/x-ruby'
     ru     : 'text/x-ruby'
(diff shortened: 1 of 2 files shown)


williamstein:
I tried this as my first example:

   http://kbroman.org/knitr_knutshell/pages/latex.html

(This file: http://kbroman.org/knitr_knutshell/assets/knitr_example.Rnw)

It has lots of plots, etc., and really does take a long time (30s?) of with running running at 100% cpu, every time you save your file.    sagetex instead caches things that don't need to be recomputed.

Other than this, I think this looks fantastic and works fine, and I strongly support merging this ASAP!   Positive review.

williamstein:
Want me to make it live, or do you plan to push more first?

haraldschilly:
caching is "explicit" instead of "implicit". inside those blocks in the first configuration line you can add `, cache=TRUE` and it saves some stuff into a `./cache` subdirectory.

last hour I spent figuring out how to fix forward/inverse search ... that requires to patch the synctex file and it works (surprisingly). what's not working are those issues/errors buttons, but we can make this live without this detail. seems to be mainly about me figuring out what this really does.

haraldschilly:
Ah, even more interesting, there are [global options and automatic dependency management](https://github.com/yihui/knitr-examples/blob/master/017-auto-dependson.Rnw)

Issue #1448, opened 2017-01-03
java file editor preview 


williamstein:
I've disabled (in editor.coffee) this because:
(1) it leaves a visible .html file sitting around, which seems cluttered and confusing
(2) it doesn't work at all when the .java file is in a subdirectory -- this is a huge bug.

It would also be good if there were some indication that compilation is happening, and also compilation should happen when the document is first loaded.

Resolution of this issue is to either: (a) fix the above bugs (should be easy), or (b) -- worse -- revert the code of https://github.com/sagemathinc/smc/pull/1441.

PR #1466, opened 2017-01-08
Make java preview work with subfolders and get rid of html file creation

timothyclemansinsea:
See #1448


williamstein:
1. Doesn't work if there is a space in the path, e.g., `a b/HelloWorld.java`.

2. No syntax highlighting.

screen shot 2017-01-07 at 10 43 13 pm

3. It needs to run java once on startup (not just after change and save).

Issue #1469, opened 2017-01-08
when editing java, do not show the text editor input button row.


williamstein:
<img width="1417" alt="screen shot 2017-01-07 at 11 16 03 pm" src="https://cloud.githubusercontent.com/assets/1276278/21747268/45407496-d52f-11e6-92fe-9de1747e330d.png">

PR #1470, opened 2017-01-08
hide text input buttons for java files

timothyclemansinsea:
For #1469

### DIFF
--- a/src/smc-webapp/editor-html-md/editor-html-md.coffee
+++ b/src/smc-webapp/editor-html-md/editor-html-md.coffee
@@ -49,7 +49,7 @@ class exports.HTML_MD_Editor extends editor.FileEditor
         #     * source editor -- a CodeMirror editor
         #     * preview/contenteditable -- rendered view
         # console.log("HTML_MD_editor", @)
-
+        $('.salvus-editor-textedit-buttonbar').show()
         if @ext == 'html'
             @opts.mode = 'htmlmixed'
         else if @ext == 'md'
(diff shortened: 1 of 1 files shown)


src/smc-webapp/editor-html-md/editor-html-md.coffee
```diff
@@ -49,7 +49,7 @@ class exports.HTML_MD_Editor extends editor.FileEditor
         #     * source editor -- a CodeMirror editor
         #     * preview/contenteditable -- rendered view
         # console.log("HTML_MD_editor", @)
-
+        $('.salvus-editor-textedit-buttonbar').show()
```
williamstein: NO!!

src/smc-webapp/editor-html-md/editor-html-md.coffee
```diff
@@ -62,7 +62,9 @@ class exports.HTML_MD_Editor extends editor.FileEditor
             # canonicalize .wiki and .mediawiki (as used on github!) to "mediawiki"
             @ext = "mediawiki"
             @opts.mode = 'mediawiki'
-        else if @ext != 'java'
+        else if @ext == 'java'
+            $('.salvus-editor-textedit-buttonbar').hide()
```
williamstein: Hell no!   Do `@element.find('.salvus-editor-textedit-buttonbar').hide()`...

The above then this is scary, since you have two bugs that sort of cancel out, but if you understand what you just did, you'll see that this approach very badly breaks things, e.g,. if you have another .md file open in a tab...

timothyclemansinsea:
@williamstein I fixed this

PR #1482, opened 2017-01-10
Make file creation errors show again

### LINKED ISSUE #1481: files: selecting something in the "create" dropdown does nothing
1. don't enter a filename
2. click the dropdown to create a file
3. e.g. select markdown
4. wonder why nothing happens

I would either expect a warning or a timestamp filename.

johnjeng:
Fixes a bug where `file_creation_error` wasn't hooked up to the store.
**Edit**: Fixes #1481

Should look like this.
error

Test:
1. Go to file page
2. Make sure the search bar is *empty*
2. Click the down arrow of `Create`
3. Click on one of the file types.
4. You should see a notice.

This is currently broken on live. Who knows how long it's been broken for.

### DIFF
--- a/src/smc-webapp/project_files.cjsx
+++ b/src/smc-webapp/project_files.cjsx
@@ -1828,6 +1828,7 @@ exports.ProjectFiles = rclass ({name}) ->
             error               : rtypes.string
             checked_files       : rtypes.immutable
             selected_file_index : rtypes.number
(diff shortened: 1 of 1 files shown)


williamstein:
I'm also making a few small changes -- this is an error not a warning, so use style error; also, the wording is not so clean.

Issue #1473, opened 2017-01-08
brand new account (first time) should show projects page instead of account settings


williamstein:

williamstein:
I think I basically closed this (or a closely related) issue, so I'm re-openning since it is still an issue.  Easy to fix, probably?

timothyclemansinsea:
Please post the steps for not getting projects page right after sign up.

williamstein:
1. Create new account

<img width="636" alt="screen shot 2017-01-09 at 12 24 06 am" src="https://cloud.githubusercontent.com/assets/1276278/21757799/f9368044-d601-11e6-8331-3c603356f231.png">

2. Click "Sign up!" and get the accounts page.  

<img width="855" alt="screen shot 2017-01-09 at 12 24 11 am" src="https://cloud.githubusercontent.com/assets/1276278/21757800/ff54a87a-d601-11e6-95ef-e2798bc67d96.png">

williamstein:
OK.  Maybe read the code and understand it.   E.g., perhaps cookies are relevant.

timothyclemansinsea:
The difference is /settings -> projects /#settings -> settings the easy fix
I think would be in sign out

PR #1492, opened 2017-01-12
For #1473: on sign out get rid of settings hashtag

timothyclemansinsea:
so on sign in the projects page shows


williamstein:
Of course -- I was really confused by this!  Thanks for figuring this out!

williamstein:
Works perfectly.

Issue #1506, opened 2017-01-16
sagews %octave plots are too small and octave plot fails pytest


DrXyzzy:
It looks as if scaling octave plots by 0.66 with %octave mode is no longer needed with recent updates.

DrXyzzy:
Jupyter message protocol with octave_kernel 0.20.0 is sending extra stdout messages with plot images.

DrXyzzy:
Also in one test run I got `sage_server start time 8.08109712601 sec`. I increased the startup time threshold from 8 to 10 seconds.

williamstein:
@DrXyzzy -- is this now fixed?

DrXyzzy:
Yes.

PR #1519, opened 2017-01-16
Issue/1506

DrXyzzy:
Ref. #1506 

To review:
1. Install new code in dev mode and run pytest suite.

    ```
    cd [current smc-repo]
    git checkout DrXyzzy:issue/1506 # or equivalent
    cd src/smc_sagews
    pip install --user --upgrade ./
    cd smc_sagews/tests
    python -m pytest # expect all 124 tests to pass
    ```

1. Create sagews file and do %octave plot. Verify plot size is correct, e.g.

    ```
    %octave
    x = -10:0.1:10;
    plot (x, sin(x));
    ```

### DIFF
--- a/src/smc_sagews/smc_sagews/sage_salvus.py
+++ b/src/smc_sagews/smc_sagews/sage_salvus.py
@@ -2090,7 +2090,7 @@ def octave(code=None,**kwargs):
     """
     if octave.jupyter_kernel is None:
         octave.jupyter_kernel = jupyter("octave")
-        octave.jupyter_kernel.smc_image_scaling = .66
+        octave.jupyter_kernel.smc_image_scaling = 1
     return octave.jupyter_kernel(code,**kwargs)
 octave.jupyter_kernel = None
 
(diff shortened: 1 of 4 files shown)


williamstein:
> Create sagews file and do %octave plot. Verify plot size is correct, e.g.

How do I know what "correct size" is?  I just tried, and to my older eyes, the plot seems pretty small and hard to see (e.g., small labels).  I'm not sure what to expect.

image

haraldschilly:
from my experience, octave plots always have a very small font size

haraldschilly:
well, I give this a positive review. looks sensible and all tests pass. 

(although, on first try it had a timeout, so I'm wondering if there are "soft errors" (aka "warnings"?) instead of such hard errors. my thought is, that it did load, but just took longer than expected. if it still works, it's just something we wouldn't desire but it's not really broken. e.g. I was running the tests while restarting my dev project and octave in there, so it's sensible to take longer than in a perfect situation)

DrXyzzy:
About "correct size", I should have indicated that the goal for this PR was maintaining the same resolution as the jupyter notebook running octave, which gives plot resolution as 560x420. The present PR achieves that.

PR #1532, opened 2017-01-17
webapp: sharing to social media buttons if shared file is public

haraldschilly:
I think there is not ticket, but this is something very simple we should have done long time ago. If in "commercial" mode and inside the sharing file dialog with a published file, it shows facebook, g+ and twitter action buttons. They point to sharing URLs and hence there are no additional dependencies.

testing: select a file and share it. 3 sharing buttons should appear after clicking on "make item public".


williamstein:
I'm merging this -- it's valuable right now and works.  Please switch the order and add email in another PR later.

Issue #1528, opened 2017-01-16
the terminal cursor on a dark background is very hard to see


williamstein:
1. In account settings, change your terminal preferences to something like "Solarized Dark" (a dark background).

2. Open a new terminal.  Then start rapidly typing, not pausing much.  The cursor will stay *always* black.   Instead, while "in action" the cursor should always be white if the background is black.

williamstein:
@haraldschilly this seems like the sort of thing you like to fix...

haraldschilly:
compensating for the border: 08f8992e71630e1e38202dc562d704216ce0b8b0

williamstein:
I'm closing this since it is no longer hard to see the cursor.   I don't necessary like the cursor now, but this is definitely fix.

PR #1533, opened 2017-01-17
webapp/console: cleanup cursor blinking and introduce a border for improved visibility on dark backgrounds

haraldschilly:
issue #1528

besides cleaning up the animation a bit, my only idea is to introduce a border. looks ok for me. I don't see a way to solve this in a more sophisticated way by working with the actual background color of the theme.


williamstein:
Significant improvement for "dark background" people.  It was unusable before.

Issue #1499, opened 2017-01-15
Add read receipt to chat


timothyclemansinsea:
When I'm requesting help I need to know if my helper has read my message. I'm marking this as M-medium because I imagine that instructors/TAs want to know if a student has read their message.

williamstein:
I'm marking this as high.  I'm also assigning you, because this is a really good idea, worth doing, and I think all information needed to implement this is *already* in the browser, so this might be pretty easy.   Look at what other chat programs do to indicate "read receipt" information for inspiration...

novoselt:
I would be in favour of closing it without implementation:
- it was opened for a while and I don't recall requests to have it
- it is actually tricky to implement "correctly", in particular Facebook and WhatsApp have this support and from personal experience sometimes people read stuff which is not shown as read because they read a push notification (or the email sent by CoCalc) and sometimes they don't read stuff which was shown on screen and thus deemed read. In some personal cases the consequences were quite substantial ;-)

williamstein:
I've never seen this issue before.  However, I think we have to implement this in order to be able to show users what *they* have and haven't seen, which is a feature request people do make.  E.g., if I want to know which messages i've missed in existing threads, we have to store somewhere whether I've read those threads.  It's exactly the same data as "add read receipt to chat".

PR #1552, opened 2017-01-19
For #1499: add read up to status to chats

timothyclemansinsea:
To test: open several chrome user sessions, log into same project with different smc users, and type messages and trigger mark_read. You should see Avatars if < all project collaborators read message or only two collaborators or collaborators have different read up tos and "All collaborators read" otherwise. 

A sender's Avatar does not show if they've only read up to their last message.

image

image

### DIFF
--- a/src/smc-util/misc.coffee
+++ b/src/smc-util/misc.coffee
@@ -1898,3 +1898,17 @@ exports.bind_objects = (scope, arr_objects) ->
                 return bound_func
             else
                 return val
+
+# Provide the index that a value would be inserted
+# into an array where the val is between the number before
+# and after it
+exports.array_bisect = (arr, val) ->
+    idx = undefined
+    if arr.length == 0
+        return 0
+    idx = 0
+    while idx < arr.length
+        if val < arr[idx]
+            return idx
+        idx++
+    idx
--- a/src/smc-util/misc.coffee
+++ b/src/smc-util/misc.coffee
@@ -1899,16 +1899,17 @@ exports.bind_objects = (scope, arr_objects) ->
             else
                 return val
 
-# Provide the index that a value would be inserted
-# into an array where the val is between the number before
-# and after it
-exports.array_bisect = (arr, val) ->
+# Assuming a sorted array of numbers return the index
+# where val is between the number in that index and
+# the number after it. Return the last index
+# if val is greater than the largest number in the array
+exports.find_smaller_number = (arr, val) ->
     idx = undefined
     if arr.length == 0
-        return 0
+        return -1
     idx = 0
     while idx < arr.length
         if val < arr[idx]
-            return idx
+            return idx - 1
         idx++
-    idx
+    idx - 1
--- a/src/smc-util/test/misc-test.coffee
+++ b/src/smc-util/test/misc-test.coffee
@@ -1389,3 +1389,8 @@ describe 'bind_objects', ->
 
         expect(b_obj1.func()).toEqual("cake")
         expect(b_obj1.val).toEqual("lies")
+
+describe 'array bisect function', ->
+    array_bisect = misc.array_bisect
+    it 'returns insertion index of 3 for a value of 11 for array of [0,3,5,15,30,90]', ->
+        array_bisect([0,3,5,15,30,90], 11).should.eql 3
--- a/src/smc-util/test/misc-test.coffee
+++ b/src/smc-util/test/misc-test.coffee
@@ -1390,7 +1390,11 @@ describe 'bind_objects', ->
         expect(b_obj1.func()).toEqual("cake")
         expect(b_obj1.val).toEqual("lies")
 
-describe 'array bisect function', ->
-    array_bisect = misc.array_bisect
-    it 'returns insertion index of 3 for a value of 11 for array of [0,3,5,15,30,90]', ->
-        array_bisect([0,3,5,15,30,90], 11).should.eql 3
+describe 'find smaller number function', ->
+    find_smaller_number = misc.find_smaller_number
+    it 'returns index of 2 for a value of 11 for array of [0,3,5,15,30,90]', ->
+        find_smaller_number([0,3,5,15,30,90], 11).should.eql 3
+    it 'returns index of -1 for a value of 3 for array of [5,10,15]', ->
+        find_smaller_number([5,10,15], 3).should.eql -1
+    it 'returns index of -1 for a value of 9 for array of []', ->
+        find_smaller_number([], 9).should.eql -1
--- a/src/smc-webapp/smc_chat.cjsx
+++ b/src/smc-webapp/smc_chat.cjsx
@@ -323,18 +323,7 @@ Message = rclass
                 {cols}
             </Row>
 
-Array::bisect = (val) ->
-    idx = undefined
-    if @length == 0
-        return 0
-    idx = 0
-    while idx < @length
-        if val < @[idx]
-            return idx
-        idx++
-    idx
-
-sortCopy = (arr) -> 
+sortCopy = (arr) ->
     return arr.slice(0).sort()
 
 are_dicts_equal = (a, b) ->
@@ -366,6 +355,7 @@ ChatLog = rclass
         messages     : rtypes.object.isRequired   # immutable js map {timestamps} --> message.
         user_map     : rtypes.object              # immutable js map {collaborators} --> account info
         read_dates_by_user : rtypes.object
+        message_dates: rtypes.array
         account_id   : rtypes.string
         project_id   : rtypes.string   # optional -- used to render links more effectively
         file_path    : rtypes.string   # optional -- ...
@@ -378,7 +368,7 @@ ChatLog = rclass
         set_scroll   : rtypes.func
 
     getInitialState: ->
-        read_indexes_to_user_ids: @find_read_tos(@props.read_dates_by_user, @props.messages.keySeq().sort(misc.cmp_Date).toJS())
+        read_indexes_to_user_ids: @find_read_tos(@props.read_dates_by_user, @props.message_dates)
         prev_read_dates_by_user: @props.read_dates_by_user
 
     mixins: [SetIntervalMixin]
@@ -389,14 +379,7 @@ ChatLog = rclass
     update_read_indexes_to_user_ids: ->
         if are_dicts_equal(@props.read_dates_by_user, @state.prev_read_dates_by_user)
             return
-        message_dates = @props.messages.keySeq().sort(misc.cmp_Date).toJS()
         read_indexes_to_user_ids = @state.read_indexes_to_user_ids
-        find_between = (read_date) ->
-            answer = message_dates.bisect(read_date)
-            answer -= 1
-            if answer == -1
-                return null
-            return answer
         # figure out which users to update
         diff = diffArray(Object.keys(@props.read_dates_by_user), Object.keys(@state.prev_read_dates_by_user))
         for account_id in Object.keys(@props.read_dates_by_user)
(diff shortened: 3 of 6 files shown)


src/smc-webapp/smc_chat.cjsx
```diff
@@ -323,27 +323,105 @@ Message = rclass
                 {cols}
             </Row>
 
+Array::bisect = (val) ->
```
haraldschilly: could you please move this block of code from smc_chat to smc-util/misc ? it looks very generic and could be useful anywhere. besides that, I'm not sure if it is so good to change the prototype. `bisect` is only used in one place, and there is nothing gained by doing this. Just define a normal function and call it `array_bisect`.

Also, once it is in misc.coffee, add some tests to `smc-util/test/misc-test.coffee`

src/smc-webapp/smc_chat.cjsx
```diff
@@ -362,6 +440,37 @@ ChatLog = rclass
                 else
                     @props.actions.set_editing(@props.messages.get(date), false)
 
+    find_read_tos: (read_dates_by_user, message_dates) ->
+        read_indexes_to_user_ids = {}
+        [user_ids, read_dates] = [Object.keys(read_dates_by_user), underscore.values(read_dates_by_user)]
+        find_between = (read_date) ->
```
haraldschilly: find_between appears twice. just make it its own function and use it here. makes it shorter and more readable :-)

src/smc-webapp/smc_chat.cjsx
```diff
@@ -362,6 +440,37 @@ ChatLog = rclass
...
+            answer -= 1
+            if answer == -1
+                return null
+            return answer
+        indexes_for_read_up_to = (find_between(date) for date in read_dates)
+        for index_for_read_up_to, i in indexes_for_read_up_to
+            user_id = user_ids[i]
+            if index_for_read_up_to is null
+                continue
+            if String(index_for_read_up_to) in Object.keys(read_indexes_to_user_ids)
```
haraldschilly: maybe just `if index_for_read_up_to of read_indexes_to_user_ids` ?

haraldschilly: or even simpler, two lines:

```
read_indexes_to_user_ids.index_for_read_up_to ?= []
read_indexes_to_user_ids.index_for_read_up_to.push(user_id)
```

timothyclemansinsea:
@haraldschilly I've cleaned this up per your review

src/smc-util/misc.coffee
```diff
@@ -1898,3 +1898,17 @@ exports.bind_objects = (scope, arr_objects) ->
                 return bound_func
             else
                 return val
+
+# Provide the index that a value would be inserted
+# into an array where the val is between the number before
+# and after it
+exports.array_bisect = (arr, val) ->
```
johnjeng: Takes an Array and a Val. Returns the first index where the value there is greater than `Val`. Else returns the length of the array.

Or you can say something about assumptions of sortedness, but the code does no such thing.

src/smc-util/test/misc-test.coffee
```diff
@@ -1389,3 +1389,8 @@ describe 'bind_objects', ->
 
         expect(b_obj1.func()).toEqual("cake")
         expect(b_obj1.val).toEqual("lies")
+
+describe 'array bisect function', ->
+    array_bisect = misc.array_bisect
+    it 'returns insertion index of 3 for a value of 11 for array of [0,3,5,15,30,90]', ->
+        array_bisect([0,3,5,15,30,90], 11).should.eql 3
```
johnjeng: Likewise, this description should be updated.

There should be tests for edge cases: Zero length array, smaller than all elements, larger than all elements.

src/smc-webapp/smc_chat.cjsx
```diff
@@ -323,27 +323,86 @@ Message = rclass
                 {cols}
             </Row>
 
+sortCopy = (arr) ->
+    return arr.slice(0).sort()
+
+are_dicts_equal = (a, b) ->
```
johnjeng: I have to imagine this is in misc somewhere or use Underscore. Unless this does something special in which case that needs to be documented.

src/smc-webapp/smc_chat.cjsx
```diff
@@ -323,27 +323,86 @@ Message = rclass
...
+    # assuming maps to lists
+    sorteda = {}
+    for key in sortCopy(Object.keys(a))
+        sorteda[key] = a[key]
+    sortedb = {}
+    for key in sortCopy(Object.keys(b))
+        sortedb[key] = b[key]
+    return JSON.stringify(sorteda) == JSON.stringify(sortedb)
+
+diffArray = (arr1, arr2) ->
```
johnjeng: Please comment what this does. Reading the code, it doesn't seem to do any of my first 3 guesses.

src/smc-webapp/smc_chat.cjsx
```diff
@@ -323,27 +323,86 @@ Message = rclass
...
         set_scroll   : rtypes.func
 
+    getInitialState: ->
+        read_indexes_to_user_ids: @find_read_tos(@props.read_dates_by_user, @props.message_dates)
+        prev_read_dates_by_user: @props.read_dates_by_user
+
+    mixins: [SetIntervalMixin]
+
+    componentWillMount: ->
+        setInterval(@update_read_indexes_to_user_ids, 300)
```
johnjeng: I don't remember, does this automatically clear?

src/smc-webapp/smc_chat.cjsx
```diff
@@ -323,27 +323,86 @@ Message = rclass
...
+
+    update_read_indexes_to_user_ids: ->
+        if are_dicts_equal(@props.read_dates_by_user, @state.prev_read_dates_by_user)
+            return
+        read_indexes_to_user_ids = @state.read_indexes_to_user_ids
+        # figure out which users to update
+        diff = diffArray(Object.keys(@props.read_dates_by_user), Object.keys(@state.prev_read_dates_by_user))
```
johnjeng: Why do we check if they're equal and then diff them? It seems we could diff them and based on that result know if they're equal or not.

src/smc-webapp/smc_chat.cjsx
```diff
@@ -323,27 +323,86 @@ Message = rclass
...
+        for account_id in Object.keys(@props.read_dates_by_user)
+            if @props.read_dates_by_user[account_id] != @state.prev_read_dates_by_user[account_id]
+                if account_id not in diff
+                    diff.push(account_id)
```
johnjeng: I'm curious why this isn't covered by the diffing operation? It seems to rely on exactly the same data.

src/smc-webapp/smc_chat.cjsx
```diff
@@ -323,27 +323,86 @@ Message = rclass
...
+
+    mixins: [SetIntervalMixin]
+
+    componentWillMount: ->
+        setInterval(@update_read_indexes_to_user_ids, 300)
+
+    update_read_indexes_to_user_ids: ->
+        if are_dicts_equal(@props.read_dates_by_user, @state.prev_read_dates_by_user)
+            return
+        read_indexes_to_user_ids = @state.read_indexes_to_user_ids
```
johnjeng: Does this make a copy?

src/smc-webapp/smc_chat.cjsx
```diff
@@ -362,6 +421,36 @@ ChatLog = rclass
                 else
                     @props.actions.set_editing(@props.messages.get(date), false)
 
+    find_between: (read_date) ->
+        answer = misc.array_bisect(@props.message_dates, read_date)
+        answer -= 1
+        if answer == -1
+            return null
+        return answer
```
johnjeng: This seems to work around your design of `array_bisect`.

src/smc-webapp/smc_chat.cjsx
```diff
@@ -406,6 +495,17 @@ ChatLog = rclass
                      set_scroll       = {@props.set_scroll}
                     />
 
+            if @state?.read_indexes_to_user_ids?[i]
+                if @state?.read_indexes_to_user_ids?[i].length > 0
```
johnjeng: When does `state` not exist? You have `getInitialState`.

johnjeng: Also, it shouldn't be two if statements.

src/smc-util/misc.coffee
johnjeng: Also this doesn't seem to be doing any bisecting. It's more like the index finder for an insertion sort. Maybe name it `find_larger_entry`. This naming would be analogous to [Array.prototype.find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find). Which maybe this should use?

src/smc-webapp/smc_chat.cjsx
timothyclemansinsea: diffArray does not work on associative arrays which is what `@props.read_dates_by_user, @state.prev_read_dates_by_user` are

src/smc-webapp/smc_chat.cjsx
timothyclemansinsea: IDK

src/smc-webapp/smc_chat.cjsx
timothyclemansinsea: because diffArray is only for arrays of values not associative arrays

williamstein:
Can you move the read information to the right hand side, like @johnjeng showed us with Facebook's Messenger?

timothyclemansinsea:
@williamstein now right aligned
image

williamstein:
Just testing this for a few minutes, I feel like it is broken and just doesn't work right.  For example, with three users and trying things for a bit, then just letting it sit a few minutes looks like the following:

screen shot 2017-01-23 at 11 00 25 am

1. The heads should all be at the bottom exactly like with Messenger (that thing @johnjeng showed us).
2. There should be exactly two heads, or a message like "All messages seen by everybody" (whatever that think @johnjeng showed us does).    As it is, often there are no small heads at all -- just no information.
3. In the picture the position of the one small head that does appear is inconsistent between clients.
4. I didn't test, but what happens if there are 100 heads?  In SMC there are projects with 100+ collaborators on the same project and even opening the same chatroom (e.g., Kedlaya's class)...  Which heads get shown?  

Anyway, ether the file_use stuff is broken or unsuitable for this purpose, or whatever you have implemented has bugs.

timothyclemansinsea:
@williamstein 
1. I just pulled up a large group chat in Facebook messenger on phone and see heads in various points, not all at bottom. I think there's an issue with auto scrolling not scrolling to the bottom. With Facebook messenger this model seems to make it easier to spot who hasn't read my last message.
2. Are you sure it's just not scrolled to bottom? Also as it's coded now if person last read right after they sent a message then I don't show their Avatar. Maybe I should only do this if its the last message.
For 3 and 4 I'll do more testing to figure out exactly what's going on.

williamstein:
> I just pulled up a large group chat in Facebook messenger on phone and
see heads in various points, not all at bottom. I think there's an issue
with auto scrolling not scrolling to the bottom. With Facebook messenger
this model seems to make it easier to spot who hasn't read my last message

Thanks for checking -- you're right, I was confused.

I honestly have never used Facebook messenger or really anything with this
sort of "read notification" before, so I'm not sure what to expect.  I
think it is inconsistent in the screenshot I gave.

timothyclemansinsea:
@williamstein OK studying the screenshot I do see the inconsistency. The immediate thing that comes to mind is the setinterval of 3000. I'll start testing in a few moments.

williamstein:
> ... setinterval of 3000.

1. That isn't the reason, since I waited more than 3 seconds.
2. Why the heck are you using setInterval?  You should make rendering depend on the file use store, so whenever it changes (e.g., because a user marks something read), you can update the output accordingly.  There is almost never any reason to use setInterval in SMC.  If you're using it, probably you're doing things wrong.  Almost everything is event/data driven.

timothyclemansinsea:
@williamstein I used setInterval to get around the can't change state during rendering issue. I need to save which index has been read by whom in order to efficiently modify the index to accounts mapping.

williamstein:
> to get around the can't change state during rendering issue.

:man_shrugging:   If you're trying to get around that as an issue, you're surely not understanding redux/react...

timothyclemansinsea:
Closing because I don't know react/redux enough to fix this

Issue #836, opened 2016-08-21
support editing makefiles


williamstein:
If a file is named `makefile` or `Makefile`, load it in a special codemirror mode for makefile editing.    In particular [tab key] --> really sends a tab!

PR #1555, opened 2017-01-19
webapp/editor/makefile: preliminary support for makefiles

haraldschilly:
issue #836

This works for files named `build.make` and `[Mm]akefile` <strike>but not for those without extensions. Not sure how to implement that, since it seems like there are only files with extensions or a catch-all if there is none. In any case, this is a first step and dealing with the filename is a neat follow up.</strike>

To test this, create a file `makefile` and check that tabs are always tabs for indentation -- regardless of your account editor setting.


williamstein:
NICE!

Issue #19, opened 2015-06-11
make it possible to click and drag to move the position of the document in the latex preview


williamstein:
Right now you can move the document by scrolling vertically or horizontally using your mouse, which is OK if you have a super-good mouse.  However, it would also make sense to be able to click and drag the preview to change its position, especially since click and drag currently does nothing.

This is a feature requested by Mani Chandra.

haraldschilly:
Quick search:
http://www.debiki.com/dev/utterscroll
or just http://hitconsultants.com/dragscroll_scrollsync/scrollpane.html

look like a good match for this ...

haraldschilly:
Utterscroll enables it everywhere, but just to get a feeling for it:

```
%coffeescript
$.getScript '//rawgit.com/debiki/utterscroll/master/jquery-scrollable.js', ->
    $.getScript '//rawgit.com/debiki/utterscroll/master/debiki-utterscroll.js', ->
        debiki.Utterscroll.enable scrollstoppers: '.CodeMirror, .ui-resizable-handle'
```

It's probably super trivial to extract the essential parts of utterscroll and just apply it dynamically to the div-scrollable areas (besides the latex preview, I'm also thinking about the pdf, image viewer, etc.)

williamstein:
still not fixed/done...

haraldschilly:
other candidates are https://www.npmjs.com/package/react-dragscroll or maybe even https://www.npmjs.com/package/react-scroll-box ... pending rectification of the latex editor (and all the viewers for pdf and images, too) of course

PR #1590, opened 2017-01-25
webapp/latex: drag preview with mouse, renaming output to pages

haraldschilly:
see issue #19

testing:

* click and drag in a latex preview, double click should still do the inverse search.
* make sure it doesn't behave strangely when going outside the viewport or outside the browser window (i.e. it might get stuck in the mouse-down state, etc.)
* it even works on my laptop's touch screen, but I haven't tested it with an actual tablet

### DIFF
--- a/src/smc-webapp/editor.coffee
+++ b/src/smc-webapp/editor.coffee
@@ -2342,8 +2342,8 @@ patchSynctex(\"#{@filename_tex}\");' | R --no-save"
 
     trash_aux_files: (cb) =>
         log = ''
-        EXT = ['aux', 'log', 'bbl', 'synctex.gz', 'sagetex.py', 'sagetex.sage', 'sagetex.sage.py', 'sagetex.scmd', 'sagetex.sout']
-        EXT = (E + '.' for E in EXT)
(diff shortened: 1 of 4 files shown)


williamstein:
It works as advertised.   However, it breaks a **critical** piece of functionality!   When I was writing the latex editor I surveyed people about what they care about, and several people (e.g., David Jao) said: "I want to be able to click on the preview, then hit the space bar (and arrows and pg-up/down) to scroll through the pages."   This PR completely breaks this functionality, which works perfectly in the live version and always has. 

Basically, you need to add back a notion of the right hand preview being "focus"ed.

I don't get the point of the "Preview" checkbox, to be honest.   Why is it there and what does it accomplish?   If you don't want a preview, just change to a different page?    It seems like now if I click the checkbox next to preview to disable it, then no compilation ever happens at all, e.g., if you switch to the "Issues" tab, then make an error on the left, then nothing happens on the right.
Thus the preview toggle is really a *build* toggle, so it's lying.  And an actual preview toggle would be better accomplished by just not computing the preview if that tab isn't visible.

So:

1. Fix that click to move preview should focus preview side and allow for the keyboard to page around.
2. Fix that the preview checkbox doesn't do what is claimed -- it toggles build completely.  Instead, maybe have a toggle that completely disables everything on the right... which of course means yet more jQuery development.

williamstein:
Wow  this seems PERFECT now.  So solid... that nobody will notice :-)

Issue #1586, opened 2017-01-24
singular jupyter kernel - add sagews pytest and %singular mode


DrXyzzy:
Ref: #1553

DrXyzzy:
Since there is already `singular(x, type='def')` among sage interfaces, I propose naming the jupyter bridge `singular_mode`.

williamstein:
What about `singular_jupyter` or `singular_kernel`.  

Keeping our existing singular interface makes a lot of sense, because it is actually really really good (it's a C++ interface).

DrXyzzy:
`singular_kernel` it is

williamstein:
I made a remark at the bottom here:  
https://cloud.sagemath.com/projects/4a5f0542-5873-4eed-a85c-a18c706e8bcd/files/support/singular.sagews

Remind me to update this when `singular_kernel` exists.

williamstein:
@DrXyzzy  I added you to the project, so you can update that worksheet.

DrXyzzy:
I updated the first cell of https://cloud.sagemath.com/projects/4a5f0542-5873-4eed-a85c-a18c706e8bcd/files/support/singular.sagews to be
```
%auto
%default_mode singular_kernel
```
and re-ran all the cells to verify output.

_Note: to pick up the new smc_sagews package, I moved `~/.local` to `~/.local0` in that project and restarted the sage_server process._

Issue #1553, opened 2017-01-19
Singular Jupyter Kernel


haraldschilly:
> The kernel needs a current Singular (>=4.1.0), as there is in Sage. To 
run it, the python packages
> 
> PySingular (https://pypi.python.org/pypi/PySingular) and
> 
> jupyter-kernel-singular (https://pypi.python.org/pypi/jupyter-kernel-singular/)
> 
> are needed. Both are available on PyPI. The first package is a C++ modulefor both Python2 and 
> Python3, interfacing the Singular parser via LibSingular. The second 
> package is the Jupyter kernel.

jdemeyer:
This is done, right?

haraldschilly:
oh yes, don't remember why it is still open.

PR #1592, opened 2017-01-25
issue/1586 sagews %singular_kernel mode and pytest cases

DrXyzzy:
Ref: #1586 #1553  

also: increase sage_server acceptable start latency from 10 to 12 sec in pytest suite

### DIFF
--- a/src/smc_sagews/smc_sagews/sage_salvus.py
+++ b/src/smc_sagews/smc_sagews/sage_salvus.py
@@ -1827,6 +1827,23 @@ def python3(code=None,**kwargs):
     return python3.jupyter_kernel(code,**kwargs)
 python3.jupyter_kernel = None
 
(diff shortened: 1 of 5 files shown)


williamstein:
Definitely positive review if it passes tests.  Code looks good to me. 

One thought -- maybe include a link to https://github.com/sebasguts/jupyter_kernel_singular in the docstring of the singular_kernel function...?

DrXyzzy:
Link suggested by @williamstein added to docstring.

To test this PR, run the pytest cases, which use the new mode to check version of the jupyter singular kernel and factor a polynomial:
```
# check out this code branch to e.g. ~/smc
cd ~/smc/src/smc_sagews
pip2 install --user --upgrade ./
cd tests
python -m pytest test_sagews_modes.py::TestSingularMode
```

Issue #1599, opened 2017-01-26
if you have no upgrades available, you can't edit the upgrades already assigned to a project to remove them!


williamstein:
This is absurd.  See:

<img width="399" alt="screen shot 2017-01-26 at 12 17 01 pm" src="https://cloud.githubusercontent.com/assets/1276278/22348644/6556d626-e3c1-11e6-9110-c5295573c32f.png">

Upgrading should only be disabled if you have no upgrades at all, not if you have none available.

williamstein:
This is just not true or a problem.   The problem is that if you have no purchased upgrades then you can't change anything.... which makes perfect sense.  

I tried making a new account, buying a subscription, maxing it out, and everything worked perfectly as expected.

PR #1600, opened 2017-01-26
For #1599: allow editing of upgrades even if no upgrades available

timothyclemansinsea:
You should see 
image
if you have no upgrades available. I tested both no upgrades available and upgrades available.

For #1599


williamstein:
This makes things worse....

timothyclemansinsea:
@williamstein fixed

Issue #1601, opened 2017-01-27
Remove little warning banner for courses that have $9 fee warning


timothyclemansinsea:

DrXyzzy:
Is there a requirements change with paid courses? I don't understand why the banner should be removed. How will students know they should pay, and how much time they have to do so?

williamstein:
I have no clue what Tim meant by this ticket.  It is too vague.  Closing.

PR #1603, opened 2017-01-27
For #1601: remove little warning banner if course fee required

timothyclemansinsea:
How I tested:
Create a course, require fee, add user to it, go to that user and see course fee warning but not the little warning banner at top
Also create a new project with no upgrades, should see little warning banner


johnjeng:
Looks good to me.

PR #1615, opened 2017-01-30
webapp: switching from the deprecated jade to pug html processor

haraldschilly:
jade had a naming issue and hence it's called [pug](https://pugjs.org) now. this updates all the deprecated aspects of this name transition.

test: the `static/index.html` page builds just like before.


williamstein:
I did `npm install`, etc., got the build to work fine, then actually tested by refreshing the page and the loading graphic of the sagemath cloud is gone:

screen shot 2017-01-30 at 9 44 21 am

haraldschilly:
good catch! I was only concerned that it loads and runs, but there were indeed some subtle syntax changes starting with version 2.0. so, this is also a maintenance update.

Issue #1447, opened 2017-01-03
allow sorting of students in course


kcrisman:
When you go to a list of students in a course, there isn't any obvious way to sort them.  But one should be able to sort by last name, first name, email address, and (crucially) by last usage time.  Perhaps also by points etc., but I haven't used that capability yet.

kcrisman:
Related to #108

aarontresham:
One issue I've run across is that students don't input their own names very well when they register for SMC. Some only use their first name, others include a middle name or initial, some have Jr., etc., some have compound names, some don't put a space between their first and last name, and so on.

I would like to see the student list numbered, with the option of the instructor manually changing the numbers and sorting by number. Then I could put my students in the same order as they appear on my roster. I could also keep track of different sections this way.

johnjeng:
For the sake of breaking this up into parts here's what I'll be working on:

- [ ] Sort by first/last name that students have used to sign up for
- [ ] Sort by email.
- [ ] Sort by last used
- [ ] Sort by a custom ranking.
- [ ] A "Save this ordering" ability.

williamstein:
> A "Save this ordering" ability.

No -- Just make the current sort preference always be saved in localStorage on that client browser.  That's a simple UI that is consistent with everything else.

And please make it so the sorting is all the same and just like in OS's -- just put in column headings and make clicking on them sort by that heading.

johnjeng:
Oh, that was confusing since really the last two on that list are the same thing.

johnjeng:
Closing in favor of #1623.

williamstein:
This is now mostly implemented and now live.

PR #1622, opened 2017-01-31
Add Student Sorting to Courses

johnjeng:
## Fixes most of #1447  

Things **not** in this PR:
- Custom ordering
- Changing default ordering

### Adds headers to students panel columns
adjusted
Also adjusts some spacing across tabs.

### Spacing on the Assignments tab (Handouts is similar)
assignments

### Settings (and shared project) gets adjusted to match Account Settings
settings

Account settings:
account settings

### Sort without spacing adjustments:
current sort

## Testing
1. Create students in a course
2. Clicking on any of the headers should sort by that column.
3. Clicking on the same header again should reverse the order.
4. Clicking on a different header should start with the arrow pointed down.
5. Any columns not being sorted by should not have any arrow next to the name
6. The arrow should initially appear next to `Last Name`
7. Arrow down should mean a -> z (caps ignored) for names and emails.
8. Arrow down for `Last Active` should mean most recent to never used
9. Arrow down for `Hosting Type` should put members before free servers.

### DIFF
--- a/src/smc-webapp/course/common.cjsx
+++ b/src/smc-webapp/course/common.cjsx
@@ -540,7 +540,7 @@ exports.FoldersToolbar = rclass
         @setState(add_search_results:immutable.List([]))
 
     render: ->
-        <Row>
+        <Row style={marginBottom:'-15px'}>
             <Col md=3>
                 <SearchInput
                     placeholder   = {"Find #{@props.plural_item_name}..."}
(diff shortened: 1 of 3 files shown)


williamstein:
NICE!

haraldschilly:
There was one support request today, where the course listing didn't load. I don't know if there was an exception, but my vague feeling is that it could be related. The browser was Safari.

johnjeng:
Interesting. Was it fixed?

williamstein:
We can't replicate the problem.  Dima thinks the user just has very slow internet....

Issue #1493, opened 2017-01-13
put a copy of the "Start/stop" all projects buttons at the top of the assignments page in a course, to greatly speed up grading


williamstein:
This would make it so opening a project takes about 1s for the random instructor (with no idle timeout upgrades) instead of possibly 15s or more, hence saving a lot of time and frustration when they are grading.

williamstein:
I wrote **grading** above, but it's really to speed up browsing what students are doing.  Grading happens entirely in the one single instructor project.

novoselt:
If technically possible (I think so) I would add this button next to course tabs, i.e. Students/Assignments/Handouts - probably to the right from them, with title "Start all" (why do we need Stop?) and a tooltip that explains what it does, WHY the instructor should press it, and maybe some warning for not abusing it.

This does not come up in support, but I think it also cannot come from people not familiar with how CoCalc works internally. They are just silently frustrated, that it is slow.

PR #1624, opened 2017-01-31
For #1493: add start/stop student projects to assignments panel

timothyclemansinsea:
image

For #1493

### DIFF
--- a/src/smc-webapp/course/assignments_panel.cjsx
+++ b/src/smc-webapp/course/assignments_panel.cjsx
@@ -33,9 +33,10 @@ exports.AssignmentsPanel = rclass ({name}) ->
         user_map        : rtypes.object.isRequired
 
     getInitialState: ->
-        err           : undefined  # error message to display at top.
-        search        : ''         # search query to restrict which assignments are shown.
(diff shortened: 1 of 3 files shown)


williamstein:
It looks like you took up half the screen with a nice big panel that will get rarely used.    I'm worried.  I was imaging something more subtle, e.g., a little button on the far right (or to the left of "add assignments"), labeled "Projects..." (?).  When clicked, then your panel would appear. 

@johnjeng what do you think?

johnjeng:
Closing this. I think it should be integrated into the student's panel and maybe a column which says if the project is running or not.

timothyclemansinsea:
@johnjeng I don't understand why you think this should be on the students panel when #1493 is specifically for speeding up grading which is not handled in the students panel. Please explain. For now I'm fixing this PR by doing what @williamstein suggests.

timothyclemansinsea:
image
image

johnjeng:
>specifically for speeding up grading

Sorry, my brain hadn't switched contexts from the student panel that I've been working on. 

Making start all student projects more robust is going to be next on my todo list.

PR #1633, opened 2017-02-01
markdown external links

haraldschilly:
right now, in the preview of md files, relative and absolute links inside SMC open a new tab .This fixes this. It also ports the enhancements from the now removed image and object source rewriter to this plugin.

testing: check that relative urls open inside smc inside the md preview. since the jquery plugin is touched, also check that images in `%md` cells pointing relatively to smc still work.


haraldschilly:
lol, ignore the changes to the queries file. they shouldn't be here ^^

Issue #1640, opened 2017-02-02
grade export -- separate first and last name as two columns


williamstein:
> "What would be really helpful is when it exports the grades to .csv that it would be first and last name into two different things, instead of combined, because then when I copy the grades into excel, I can't sort by last name. Well, I could write an excel function to split the names or something, but I wouldn't have to do that if the .csv just had it set that name normally. When using SMC for a class, this is the kind of thing that really does matter."

novoselt:
I am confused on the status of this - instead of splitting the name into two columns, which seems quite natural at least as an option, we now allow sorting by the last name, but then dump CSV as combined anyway? And everybody is happy?

williamstein:
Nobody has complained about this for 7 years, evidently.

novoselt:
Then let's close it - existing users can apparently do without, so no need to complicate the UI for them.

PR #1648, opened 2017-02-05
add first/last name columns to grades export

timothyclemansinsea:
For #1640 

image


haraldschilly:
I'm wondering what `store.get_student_name(student)` is really doing. Also querying the user-store? @johnjeng is working on renaming student names. So, I have the vague feeling, that renamed students will not show up with the renamed name here. If that's the case, this would add even more confusion.

timothyclemansinsea:
@haraldschilly I confirm that changed name doesn't appear. I marked this as needs work.

timothyclemansinsea:
@haraldschilly fixed it and confirmed it works with changed name

PR #1656, opened 2017-02-08
webapp error reporting

haraldschilly:
The main goal of this PR is to capture exceptions and errors and record them in the DB.

To properly deploy it, there needs to be a schema update in the prod DB! (for your dev project, just restart the hub)

Testing this is hard, because we have to make sure that it doesn't brick some browsers. I'm confident with FF and Chrome on Linux, though.

For actually testing that this does something, you have to either test it in debug mode, or to test prod mode comment out the `and DEBUG` in `smc-webapp/app_shared.cjsx`. Then, shift-click on the notification bell.

Due to all the minification the stacktraces aren't always that useful, and well, they're not the same as they show up in the console, but at least we get a good feeling about what's going on "out there". Improving on what's actually recorded and how the stacktraces are generated is for future tickets.


src/smc-hub/postgres-server-queries.coffee
```diff
@@ -105,6 +105,52 @@ class exports.PostgreSQL extends PostgreSQL
...
+        @_query
+            query       : 'INSERT INTO webapp_errors'
+            values      :
+                'id            :: UUID'      : misc.uuid()
+                'account_id    :: UUID'      : opts.account_id
+                'name          :: TEXT'      : opts.name
+                'message       :: TEXT'      : opts.message
+                'comment       :: TEXT'      : opts.comment
+                'stacktrace    :: TEXT'      : opts.stacktrace
+                'path          :: TEXT'      : opts.path
```
williamstein: What about project_id?  You have account_id and path, but not project_id.

src/webapp-lib/webapp-error-reporter.coffee
```diff
@@ -0,0 +1,276 @@
...
+#    GNU General Public License for more details.
+#
+#    You should have received a copy of the GNU General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+###############################################################################
+
+###
+Catch and report webapp client errors to the SMC server.
+This is based on bugsnag's MIT licensed lib.
```
williamstein: link?

src/smc-hub/postgres-server-queries.coffee
williamstein: Maybe path is the URL, so it includes the project_id, etc.   Confirm that, and I'm happy.

src/smc-util/db-schema.coffee
```diff
@@ -385,6 +385,42 @@ schema.client_error_log =
...
+        responsive   : type : 'boolean'
+        user_agent   : type : 'string'
+        path         : type : 'text'
+        smc_version  : type : 'string'
+        build_date   : type : 'string'
+        smc_git_rev  : type : 'string'
+        uptime       : type : 'string'
+        start_time   : type : 'timestamp'
+        time         : type : 'timestamp'
+    indexes:
```
williamstein: you can delete the indexes thing -- it's 100% only used by the rethinkdb stuff, which I'm going to complete delete asap.

src/smc-hub/postgres-server-queries.coffee
haraldschilly: good question, let me read up on this a bit. right now it's like `path: https://cloud.sagemath.com/projects/.../files/test/`

Issue #1658, opened 2017-02-09
IE11 compatibility: String.prototype.trimLeft


haraldschilly:
In `misc_page` for a codemirror plugin of the latex editor, there are a couple of `trimLeft`s in `CodeMirror.registerHelper "fold", "stex", (cm, start) -> ... `

We could setup this: https://www.npmjs.com/package/string.prototype.trimleft

PR #1659, opened 2017-02-09
webapp: using lodash.trimstart instead of String.prototype.trimLeft

haraldschilly:
See #1658

This is about codefolding in latex, i.e. test that

```
\section{asdf}

asdf

   \subsection{asdfasdf}

fdas

    \subsection{asdfasdf}
```

closing the first subsection still shows the second one, etc.

I'm against using that shim for IE, just adds some baggage and long term we'll transition to lodash anyways. (<strike>Regarding size, that  just adds that one function, not the whole deal</strike> we already ship lodash anyways ...)


williamstein:
I was using our code folding in firefox today to demo something for somebody (a professor here) and it was behaving 'oddly'.  This might be relevant...

PR #1662, opened 2017-02-10
webapp/sagews: hotfix for line breaks in stdout

haraldschilly:
I don't know what's happening. There has been a tiny change in the css and this patch here at least makes them blocks and hence there is a perceived line break. What was the problem (or what was fixed by the change) before? /cc @williamstein


williamstein:
My fault.  I discovered the unintended consequences of this **during** my talk :-(

PR #1664, opened 2017-02-11
Course scaling

johnjeng:
This patch upgrades the rendering of certain things when they become too numerous on the screen. It's only most apparent if you have many students (or many assignments) in a course. 

(Thinking about it now, the problem is solved in the file listing by pagination of the results... Courses could maybe use that.)

Nevertheless:

- Removes some uses of `@props.redux`
- Adds optional timing code for generation of components.
- Reduces usage of `OverlayTrigger` to mouseover. 
  - `OverlayTrigger` has some bad performance regarding its lifecycle methods. 
  - This improves rendering time of large (200+) courses by about 20% (.5 - 2 seconds)
- Adds a scroll optimization for large courses.
  - This should make a huge difference for lower end machines opening large courses (certain reasonable cases saw an increase from 8 fps to 45fps)

Test: 
1. Tips should still work everywhere. 
2. All buttons in courses should render things correctly still
3. Open a very large course.
4. Open some assignments
5. Scroll. Should not be much lag.

If you have a high end machine, consider using Chrome developer tools to throttle things down to see the difference.


williamstein:
Works!

PR #1670, opened 2017-02-12
webapp/latex: safeguard against error 9486e96a-342d-4acb-b1bf-2848645da0dc: Cannot read property 'clearGutter' of undefined at f.exports.LatexEditor.f._reset_inline_errors

haraldschilly:
yet another simple fix, based on an error found in the webapp_error table


williamstein:
Please rewrite this to instead make the function cms() only return editors that are defined.
There are several places in the code that use `for cm in @cms()`, and in future may be more, and they are all likely to run into this bug at some point.  Let's just fix the root cause.

PR #1671, opened 2017-02-12
Fix rare file use bug

johnjeng:
Per bug found by HSY webapp error reporter!

Would throw an error when someone with no file use notifications tried to submit a search inside the file use widget.


williamstein:
Looks solid and also the funtion that gets called does the right thing on undefined input.

PR #1674, opened 2017-02-13
webapp/projects: be explicit about which project one is not a collaborator of

haraldschilly:
This is just a slightly better logging and error message.

### DIFF
--- a/src/smc-webapp/projects.cjsx
+++ b/src/smc-webapp/projects.cjsx
@@ -77,7 +77,7 @@ class ProjectsActions extends Actions
 
     set_project_title: (project_id, title) =>
         if not @have_project(project_id)
-            console.warn("Can't set title -- you are not a collaborator on this project.")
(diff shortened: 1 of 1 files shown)


williamstein:
I'm not sure about this.  The project id likely means nothing to most users, even though we are ridiculously used to it.   @johnjeng  what do you think?   E.g., instead of giving the project_id in the message, why not give the title (maybe with trunc)?

haraldschilly:
My motivation and background of this happened when I opened a course with a list of projects and that error showed up. Hand-checking a few of the student courses showed that I was a collaborator there (I was also part of the instructor's course, etc.). So, my thought was, that I want to know over which course it stumbles over first. It would be enough, if only the console warning message changes, and the user facing warning stays the same. At the same time, the title in the warning would also be nice.

williamstein:
>  It would be enough, if only the console warning message changes, and the user facing warning stays the same. At the same time, the title in the warning would also be nice.

+1 to making these two changes -- put the title in the user facing warning, and the id in the console.

Issue #1668, opened 2017-02-12
Don't show the "Warning: Create a file with no extension?" message for filenames that don't have extensions


williamstein:
This screenshot should make the point of this issue clear:

<img width="1247" alt="screen shot 2017-02-11 at 6 52 39 pm" src="https://cloud.githubusercontent.com/assets/1276278/22859287/5d4622da-f08b-11e6-8184-9d5334697d2d.png">

timothyclemansinsea:
I'm a little unclear on what's wanted here. Is the idea to have a list of
filenames to not show the warning for? I'm not finding a list. I tried
Googling
https://www.google.com/search?q=common+file+names+without+extensions&oq=common+file+names+without+extensions&aqs=chrome..69i57j0.7456j0j7&sourceid=chrome&ie=UTF-8

williamstein:
fixed

PR #1691, opened 2017-02-16
For #1668 don't show warning for file associations starting with noext-

timothyclemansinsea:
For #1668 and depends on #1602


williamstein:
This will probably break when we switch to using fully react editors, so file_assocations in editor.coffee goes away.  However, this breaking isn't really that bad -- it's just an annoying confirm, and we'll do more later.  For now it works perfectly fine.

PR #1696, opened 2017-02-17
Draft of partial notes on SMC design

embray:
These notes are based on my personal notes taken while exploring the SMC source code.  As is these notes are incomplete--I still have more notes on the server side to organize.

The separate sections on SMC's dependencies and the source code walk through might make more sense, ultimately, to break up into separate documents.  I just wrote everything together to start with.


src/doc/design_overview/overview.rst
```diff
@@ -0,0 +1,961 @@
...
+
+React-Redux
+^^^^^^^^^^^
+
+SMC's ``smc-react.coffee`` modules also makes use of the `React Redux`_
+JavaScript module to tie Redux state objects to React containers (i.e. update
+displays when the state changes--abstracting the state itself from any given
+view of the state).  This is just a package for making it convenient to
+implement model / view separation in React components.  The web developers
+describe this as "container components" but really they're just reinventing MVC
```
williamstein: I don't agree with this.  Understanding and deciding on the architecture of where to put data, how it flows, what does logic, etc. is the sort of thing that takes a lot of deep thinking...  There's a very complicated world of approaches, and "MVC" etc. don't capture that complexity at all....

src/doc/design_overview/overview.rst
```diff
@@ -0,0 +1,961 @@
...
+correct order is tricky.
+
+For one, the modern ECMAScript supports features not supported by the
+JavaScript on browsers, such as the ``import`` statement for loading
+variables, classes, and functions into other files (without polluting the
+global namespace, as was necessary to share between JavaScript files in the
+bad old days).  Unfortunately, most (in fact no) browsers support this
+feature.  One also needs ways to find static resources relative to
+JavaScript modules, transform the development dialect into JavaScript that
+can run in the browser, minify and uglify the code, and put it all together
```
williamstein: I think uglify is a particular tool for minification...?

src/doc/design_overview/overview.rst
```diff
@@ -0,0 +1,961 @@
...
+usage--SMC currently actually generates three JS files (from three separate
+"entry points" to the dependency graph webpack generates).  It also generates
+the ``index.html`` file that is served at the root of the website (from the
+aforementioned ``index.jade`` template) into which webpack inserts ``<script>``
+tags that load its generated JS files.  It also does some other tricks, such as
+appending a hash to the JS filenames so that they can replace cached versions
+whenever the source changes.
+
+In practice it's less convenient to run ``webpack`` over and over
+again--instead one can run ``webpack --watch`` which watches all files for
```
williamstein: maybe ; instead of -- here?

src/doc/design_overview/overview.rst
```diff
@@ -0,0 +1,961 @@
...
+          |
+          |
+         \|/
+     HAproxy (load balancing...)HAproxy                  Admin     (monitor and control system)
+     /|\       /|\      /|\      /|\
+      |         |        |        |
+      |http1.1  |        |        |
+      |         |        |        |
+     \|/       \|/      \|/      \|/
+     Hub<----> Hub<---->Hub<---> Hub  <-----------> RethinkDB <--> RethinkDB  <--> RethinkDB ...
```
williamstein: RethinkDB --> PostgreSQL (and just one).

src/doc/design_overview/overview.rst
```diff
@@ -0,0 +1,961 @@
...
+               /|\      /|\      /|\
+                |        |        |
+       ---------|        |        | (tcp)
+       |                 |        |
+       |                 |        |
+      \|/               \|/      \|/
+     Compute<-------->Compute<-->Compute <--- rsync replication  to Storage Server, which has BTRFS snapshots
+
+(with the exception that William is in the process of replacing RethinkDB
+with PostgreSQL).
```
williamstein: totally done.

src/doc/design_overview/overview.rst
```diff
@@ -0,0 +1,961 @@
...
+between the Client and the backend happens through the Hub, whose HTTP server
+uses Express to route requests to different services (account management,
+project management, payment, etc.).  Each Hub also sets up a
+``ComputeServerClient`` which gives it access to all the running compute
+servers (discussed next).  The names and URLs of all the available compute
+servers live in a system table in the database.
+
+It also uses `node-http-proxy`_ to create an HTTP proxy server associated
+with each Hub (on port number one higher than the Hub's HTTP port).  If we
+understand correctly, the proxy handles all requests that are to be
```
williamstein: you do understand correctly.

williamstein: it's also used to proxy the websocket connection for jupyter right now...

src/doc/design_overview/overview.rst
```diff
@@ -0,0 +1,961 @@
...
+project's compute node::
+
+    Client <--> HAProxy <--> Hub Proxy <--> Compute
+
+Compute
+^^^^^^^
+
+Compute servers are where the real work gets done in SMC projects.  Every
+project is associated with a specific compute server where all their data
+is stored (by way of storage servers mounted on the compute node) and where
```
williamstein: The data is currently on the local filesystem.  It gets rsync'd from a remote server when the project opens.

src/doc/design_overview/overview.rst
```diff
@@ -0,0 +1,961 @@
...
+SMC, or one can SSH in directly.
+
+Each compute node also runs a simple socket server
+``smc-hub/compute-server.coffee`` that is used by the Hub to communicate with
+the compute node (using simple JSON messages). For example, one can make status
+inquiries on the node, or send commands to run a command in a project.
+
+Additionally, each *project* runs a couple per-project daemons when the
+project is created and running.  These include the the console server
+(``smc-project/console_server.coffee``) which provides the backend for the
```
williamstein: My top priority right now is to eliminate console_server.coffee, and instead just have that function done directly by the local_hub node.js process.  This makes things much more efficient.

src/doc/design_overview/overview.rst
```diff
@@ -0,0 +1,961 @@
...
+between software running in the project and the "global hub" (i.e. the Hub,
+through which the client is communicating).
+
+As we understand it, whereas the compute server is used to issue commands to
+the compute node on behalf of a project, such as starting the project's
+local hub, the local hub then takes care of the rest.  This all makes sense,
+but gets rather complicated as starting the actual local hub is buried under
+a pile of Python scripts.  Some of this may still be legacy from earlier
+versions of SMC that were written primarily in Python, and is need of
+cleaning up.
```
williamstein: There's an unfinished rewrite on top of kubernetes... that is even more complicated, but is the right way to go.  That rewrite, which finished, isn't going to be open source, since it'll only be needed by very large sites....

src/doc/design_overview/overview.rst
```diff
@@ -0,0 +1,961 @@
...
+  In fact, much of ``AppRedux``'s API mimics the Redux library's own top-level
+  API.  So instead of calling |redux.createStore()|_ for each sub-store, one
+  actually calls ``AppRedux.createStore()`` (the latter has some important
+  differences from the former, however, which we'll come to later).  In fact,
+  since SMC names the ``AppRedux`` singleton ``redux``, one *does* in fact
+  literally call ``redux.createStore()``, but it's important to be clear that
+  here ``redux`` is an instance of ``AppRedux``, not the Redux library itself.
+
+The whole thing is fairly smart, and almost nothing about this framework is
+particular to SageMathCloud--it could (and probably should) be factored out
```
williamstein: In the future SMC is going to be a more generic application hosting framework...

src/doc/design_overview/overview.rst
```diff
@@ -0,0 +1,961 @@
...
+
+Server stats
+^^^^^^^^^^^^
+
+Continuing to follow ``entry-point.coffee``, the next module that's loaded
+is one called ``smc-webapp/redux_server_stats.coffee``.  This actually sets
+up a "synchronized table"--a client side view of one of the database
+tables--and attaches this to the ``AppRedux``, which also carries around a
+collection of synchronized tables which are instances of a class called
+``Table`` dfined in ``smc-redux.coffee``.  The tables are actually *not*
```
williamstein: dfined

src/doc/design_overview/overview.rst
```diff
@@ -0,0 +1,961 @@
...
+state variable which defaults to ``'account'``.  At what point does this
+change from ``'account'`` to ``'projects'``?
+
+In fact it does start out on the account page.  If your browser is running
+slowly enough you can catch this briefly, while the page displays the
+"Signing you in" message.  This is part of a component called
+``<LandingPage>`` defined in ``smc-webapp/landing_page.cjsx``.  In fact this
+is the page that is displayed when you're not logged in (i.e. the
+``"remember_me"`` cookie is not set, with all the marketing content and
+account creation box.  But if your cookie is set, then it just shows the
```
williamstein: missing close paren

src/doc/design_overview/overview.rst
```diff
@@ -0,0 +1,961 @@
...
+One of these side-effects is that the ``ProjectsTable`` class is registered
+with ``AppRedux`` via the ``AppRedux.createTable()`` method.  This class is
+a client-side front-end to the ``"projects"`` table in the DB backend.  This
+creates an instance of a fairly complex object called ``SyncTable`` defined
+in ``smc-util/synctable.coffee``.  We won't go into detail of how this
+works, but basically it keeps a client-side copy of the results of queries
+to the Hub's database, and sets up change listeners that synchronizes those
+cached query results every time the real database changes (via a
+"changefeed", which asynchronously pushes an update to the client every time
+the query result changes--this is a feature around which RethinkDB was
```
williamstein: done

src/doc/design_overview/overview.rst
```diff
@@ -0,0 +1,961 @@
...
+designed, but this is now being rewritten on top of PostgreSQL).
+
+In effect, while the page has been loading, in the the client app has been
+sending a database query to the server for all the user's projects, and
+receiving the result.  The ``ProjectTable`` has a change handler that
+updates a prop called ``project_map`` which contains the result of the query
+for all the user's projects.  When the ``<ProjectsPage>`` component renders,
+it checks to see if ``project_map`` is undefined.  If it *is* undefined this
+means the database query hasn't completed yet (if the user has no projects
+the result of the query would be an empty list, but ``undefined``).  In this
```
williamstein: but **NOT** ``undefined``

src/doc/design_overview/overview.rst
```diff
@@ -0,0 +1,961 @@
...
+``<ProjectsPage>`` component consists of a number of sub-components,
+including one called ``<ProjectList>`` where the all the projects are looped
+over and displayed in a table.  There are some additional complications
+related to the ability to filter projects in various ways, but there's
+nothing special to this.
+
+Conclusion
+^^^^^^^^^^
+
+And that's it! We got from an empty browser window the user's projects
```
williamstein: window **to** the...

src/doc/design_overview/overview.rst
```diff
@@ -0,0 +1,961 @@
...
+routed by HAproxy not to the Hub, but to an associated Hub Proxy that
+redirects the requests straight to services running on the project's compute
+node.  For example, when running the Jupyter Notebook in an SMC project, the
+Notebook server is configured so that all its websocket requests go through
+a URL that starts with ``<project_id>/port/jupyter``.  This way the Jupyter
+Notebook can set up its own websockets as normal (without any special
+SMC-specific modifications) and the requests are proxied directly to the
+project's Notebook server.
+
+In principle the scheme supports other web services embedded in SMC as well.
```
williamstein: one could host anything that has base_url support...

src/doc/design_overview/overview.rst
williamstein: One other very interesting thing that is missed in all this is how realtime synchronized editing works...

embray:
Thanks for the detailed review.

src/doc/design_overview/overview.rst
```diff
@@ -0,0 +1,961 @@
...
+languages like C superseding the need to write assembly language for different
+hardware platforms.  But in this case the "assembly languages" are
+browser-specific JavaScript implementations.  This has led to an explosion in
+"high-level" languages over native JavaScript, as well as associated compilers
+and build tools.
+
+Node.js
+-------
+
+Although SMC's source tree contains a smörgåsbord of (mostly small) Python
```
embray: true, but the Swedish spelling is so much more fun.

src/doc/design_overview/overview.rst
```diff
@@ -0,0 +1,961 @@
...
+scripts for various management tasks, most / all of SMC's backend is built
```
embray: Yes, I was looking at that the other day.  I will mention it here.

src/doc/design_overview/overview.rst
```diff
@@ -0,0 +1,961 @@
...
+
+As mentioned in the section on pug, essentially all of SMC's web frontend is
+built using React.  Almost no HTML is ever sent from the server.  Instead
+the frontend is built up by React.  When user interactions with the UI need
+to be persisted, those are sent as event messages (typically over
+WebSockets) to the server, which may in turn respond with events that result
+in updating the UI in some appropriate way (the event messages are usually a
+JSON object of some kind).  This is still an over-simplification (see for
+example the section on Redux later), but that's the basic idea.
+
```
embray: My understanding though was that most of that has to do with the fact that you're using third-party tools, like codemirror.  That seems pretty hard to get around.  Are there any particular challenges that represents, other than some code not using the React interfaces?

src/doc/design_overview/overview.rst
embray: I see what you're saying here.  I'm being glib when I write that it's "[just] reinventing MVC".  There are plenty of different places to draw the line, for example, between presentation and data, and I've been in several debates about that with no clear right answer.

I think my comments here just stem from frustration with some of the jargon I encounter when reading about web frameworks--especially ones like React.  So when I read in some docs somewhere something like "this is so that you can do such and such with your container components".  Then I try to google about what they mean by "container components" and get some long-winded explanation that could have been summarized (to me) as something like "we do things like this sometimes to keep purely presentational components independent of the application state" which seems obvious to me.

That said, I should go through and remove little subjective comments like these--like I said, these started as my personal notes :)

src/doc/design_overview/overview.rst
embray: Yes, that's redundant.  I think I meant to say "obfuscate".

src/doc/design_overview/overview.rst
embray: Right, I'll mention that.  I know how that works.  At some point it would be good to generalize how the Jupyter server is handled too.

src/doc/design_overview/overview.rst
williamstein: > My understanding though was that most of that has to do with the fact that you're using third-party tools, like codemirror.

No.  That is not correct at all.  The issue is simply that it is a lot of work.  React is basically "functional programming" and writing using that approach to program is more time consuming and harder.  Maintenance and bug fixing and reading later is easier.  

>  That seems pretty hard to get around. 

No -- it's orthogonal to the problem.   You can embed CodeMirror (or any other third party javascript widget) as a wrapped react component...   That's the very beginning of getting going on such a project.
 
> Are there any particular challenges that represents, other than some code not using the React interfaces?

Yes, many interesting challenges...  With react/redux/etc. you have to very carefully think through your data model, one way flow of data, etc.     This is related to your oversimplification about React just being "MVC with different names".  Not appreciating the subtlety of the problems doesn't make them go away, unfortunately.    As with all programming, from high up it all looks the same, but when you actually dive in to really do something, you start seeing the real problems, which are deep and numerous.

src/doc/design_overview/overview.rst
embray: I see.  A couple weeks ago @fcayre and I were poking around in an SMC project and were confused, because we figured `/projects` would be mounted on some networked volume, though it wasn't clear that was the case (though I assume it is on some level within Google's SAN).  

Are projects ever moved to a different compute node, and if so how?

src/doc/design_overview/overview.rst
embray: I was a little confused as to why they were separate.  I assumed just separation of concerns, but they're both simple enough to combine it seems...

src/doc/design_overview/overview.rst
embray: Am I correct though in saying that much of the complication stems from historical rewrites, and not so much out of necessity?  I'm not sure what you mean is being rewritten here either. I'd be interested to learn how the kubernetes stuff works (I have gone through their tutorial but have never actually used it).

src/doc/design_overview/overview.rst
embray: That would definitely make a lot of sense :)

src/doc/design_overview/overview.rst
embray: Yes, I just started looking at that last Thursday (e.g. syncdoc.coffee).  I think this example was meant to be very simple, but it would be interesting when I have more time to write up a more complicated example involving, say, the Sage worksheet.  I think it's very clever how it's built on top of codemirror--that surprised me a lot.  There are some things about it that I think are a bit clunkier UI-wise than the Jupyter Notebook, but I can also see the advantages to this approach.  I think I could learn to like it if it just had a vi mode :)

src/doc/design_overview/overview.rst
williamstein: > /projects would be mounted on some networked volume, though it wasn't clear that was the case (though I assume it is on some level within Google's SAN).

SMC uses Google persistent disks, which are network-provided block devices, with similar speed to local HDD's.  These block devices can only be mounted read-write on one VM at a time.

> Are projects ever moved to a different compute node, and if so how?

Yes.  rsync.

src/doc/design_overview/overview.rst
williamstein: It does have vim mode, and in fact a better vim mode than you might expect...  The vim editor settings apply even to sage worksheets.

screen shot 2017-02-20 at 8 16 36 am

williamstein: Sage worksheets are far from done, but when they are done they will be uniformly much less clunky than the Jupyter notebook UI approach.

src/doc/design_overview/overview.rst
embray: > No -- it's orthogonal to the problem. You can embed CodeMirror (or any other third party javascript widget) as a wrapped react component... That's the very beginning of getting going on such a project.

That's basically what I thought you were doing.  In order to do much beyond that, codemirror would have to be rewritten in React as well...

> This is related to your oversimplification about React just being "MVC with different names"

That wasn't meant to be a statement about React in general.  More about React-specific jargon.

src/doc/design_overview/overview.rst
embray: But you also have to update some things in the database, such as which compute server the project lives on.  Do you have management scripts for that?

src/doc/design_overview/overview.rst
embray: Excellent! I'll have to try that out.

If nothing else, I really like that you're using essentially the same interface for editing notebooks as for any other text document (except that it's a sort of "hyper-rich-text" editor).

src/doc/design_overview/overview.rst
williamstein: >  In order to do much beyond that, codemirror would have to be rewritten in React as well...

Which is pointless.  It would presumably be identical from the outside as a black box (same API), so why write it "in react".    For comparison, Github's Atom.io editor *was* written in React for a while (with the same outside api), but then they rewrote it directly in Javascript to get a speedup.  With very careful attention and thought, direct Javascript will always be better than React, and Codemirror has already got years and  years of that careful attention, thought, and optimization.  (I think codemirror is significantly more optimized than atom.io by the way.)

williamstein: Rewriting the editors in react involves eliminating massive amounts of jquery code, mutation of state, button rendering logic, etc. and replacing it by React's approach....

embray: Right, I think we're in complete agreement.  The way I saw it, SMC's framework may be React-based, but any applications hosted in it are a black box controlled via whatever API they expose.  I was just confused by your statement

> We haven't rewritten the core functionality -- sage worksheets, jupyter, terminals, latex editor, etc.

which seemed to imply that you wish you could.

williamstein: I don't mean to rewrite every part from top to bottom in React, but instead just the parts that make sense to rewrite.  For all editors there are substantial parts that need to be redone.

src/doc/design_overview/overview.rst
williamstein: https://github.com/sagemathinc/smc/blob/master/src/smc-hub/storage.coffee and https://github.com/sagemathinc/smc/blob/master/src/smc-hub/compute-client.coffee do that.

src/doc/design_overview/overview.rst
williamstein: Originally -- for efficiency reasons (both coding and runtime) -- it was because I didn't want to use the same TCP connection for a terminal as for everything else (all the JSON control messages).  However, I just bit the bullet and added this extra support...

embray: Makes sense.

embray:
Thanks! Once I have more time I plan to add a few more examples.

PR #1730, opened 2017-02-27
Typo fixes to Erik's overview

defeo:


embray:
Thanks--as it's not too late I'll make these fixes in the copy included with the ODK report as well.

Issue #1597, opened 2017-01-26
if you don't set your last name, then you can't delete your account


williamstein:
1. Make an account with no last name
2. Try to delete it.

DrXyzzy:
- Account can be deleted if you type a space after the first name in the confirmation dialogue.
- Similar problem if there is no first name, but account can be deleted if space is entered before the last name in the dialogue.

Looks like this line around 498 in https://github.com/sagemathinc/smc/blob/master/src/smc-webapp/r_account.cjsx#L498
```
user_name     = {@props.first_name + ' ' + @props.last_name}
```

DrXyzzy:
PR deployed. Verified in production.

PR #1779, opened 2017-03-13
allow for blank first name or last name when deleting account

DrXyzzy:
Re: #1597
To reviewer: please check this. I'm unable to delete a test account in smc-in-smc, even before changing any code. But with code change in this PR, the `Confirm Account Deletion` button highlights properly when first name or last name is blank.


williamstein:
LGTM

Issue #1781, opened 2017-03-15
sagews: tachyon objects are not supported


haraldschilly:
Full example below. The whole point of this to detect the Tachyon object and render the image for display.

```
t = Tachyon(xres=1000, yres=800, camera_center=(2,7,4), look_at=(2,0,0), raydepth=4)
t.light((10,3,2), 1, (1,1,1))
t.light((10,-3,2), 1, (1,1,1))
t.texture('black', color=(0,0,0))
t.texture('red', color=(1,0,0))
t.texture('grey', color=(.9,.9,.9))
t.plane((0,0,0),(0,0,1),'grey')
t.cylinder((0,0,0),(1,0,0),.01,'black')
t.cylinder((0,0,0),(0,1,0),.01,'black')
E = EllipticCurve('37a')
P = E([0,0])
Q = P
n = 100
for i in range(n):
    Q = Q + P
    c = i/n + .1
    t.texture('r%s'%i,color=(float(i/n),0,0))
    t.sphere((Q[0], -Q[1], .01), .04, 'r%s'%i)
t.show()
```
gives: `<class 'sage.plot.plot3d.tachyon.Tachyon'>`

while changing the bottom line `t.show()` to
```
t.save_image('tachyon.png')
salvus.file('tachyon.png')
```
works.

DrXyzzy:
process: For small sagews issues, it usually takes me as much time to write pytest cases as it does to write and manually test the code change. In this case about 90 minutes each. I believe the tests are a good investment because 1) they build a suite that helps prevent re-bugging, and 2) creating tests usually leads to new insight, even when things are working as expected.

williamstein:
>  it usually takes me as much time to write pytest cases as it does to write and manually test the code change. 

Same for anything I write.   Often writing tests can even take longer than the thing being tested, as it can be quite challenging to figure out how to test something in the context of SMC (unlike say sage, where you just try some examples and copy them into a docstring).

DrXyzzy:
Fixed, verified in production.

PR #1789, opened 2017-03-17
Issue/1781

DrXyzzy:
Ref: #1781.

Add code for Tachyon plot and pytest cases. Patch passes all pytests with
```
= 132 passed in 193.53 seconds =
```

To test just the Tachyon plot, apply this PR, then:
```
python -m pytest test_graphics.py::TestTachyon
```


williamstein:
Wow, we already had the show_3d_plot_using_tachyon function...!

PR #1805, opened 2017-03-27
webapp: adding an react HTML component which is 90% of the already existing Markdown component

haraldschilly:
@williamstein few days ago you were asking about an `HTML` component. This patch creates such element, which is basically the already existing `Markdown` component except for the Markdown part.

I tested it by temporarily adding some code like this

    <HTML value="<a href='http://www.latex-project.org/'>LaTeX</a> is a system ... " />

and indeed, the rendered link opens a new tab (automatically added `target="_blank"`), renders a formula, etc. Hence it does all the extras `Markdown` usually does. Maybe merge it into your jupyter branch first...


williamstein:
This is awesome; and no code duplication!  very nice.

PR #1806, opened 2017-03-27
Clean up some differences between projects and files search behavior

johnjeng:
Changes in projects to match project files more closely.
- If you open a project `enter`, clear the project search
- Allow `ctrl + enter` to open a project without switching to it
- Do not clear the project search in the latter case.

Changes in project files to match projects more closely
- `ctrl + click` on a listing when searching should not clear the search.

Also remove some inappropriate usage of redux.getActions


williamstein:
NICE!

Issue #1810, opened 2017-03-28
upgrade client side codemirror


DrXyzzy:
Consider this example from coffeescript.org:
```
class Person
  constructor: (options) ->
    {@name, @age, @height = 'average'} = options

tim = new Person name: 'Tim', age: 4
```
It works if you put it in a file, say `foo.coffee` and compile and  run on command line with `coffee` and `node`. Adding `console.log tim` produces the expected output:
```
Person { name: 'Tim', age: 4, height: 'average' }
```

But if you put the above example in a sagews cell, starting with directive `%coffeescript`, the console log shows multiple error messages like the ones below, and more are added every time a cell is executed. And the coffeescript-mode function `print` produces no output.
```
[31h48m22.51s Δ 0s] Error rendering worksheet SyntaxError: unexpected =
Stack trace:
[stdin]:6:27: error: unexpected =
    {@name, @age, @height = 'average'} = options
                          ^  smc-08fa4586d2c69c034175.js:12:4104
[31h48m22.52s Δ 0.01s] Error rendering worksheet SyntaxError: unexpected =
Stack trace:
[stdin]:6:27: error: unexpected =
    {@name, @age, @height = 'average'} = options
                          ^  smc-08fa4586d2c69c034175.js:12:4104

```

williamstein:
There is a difference in the version of Coffeescript running on the server versus on the client:

<img width="1009" alt="screen shot 2017-03-28 at 4 10 37 pm" src="https://cloud.githubusercontent.com/assets/1276278/24431517/20831010-13d1-11e7-99d5-75a58225187e.png">

Please change the title to "upgrade client side codemirror"...

williamstein:
This is trickier...

src/smc-webapp/misc_page.coffee at line 1651 is where coffeescript gets loaded.

That looks like maybe the old way before we used npm much or webpack.

Yep, that’s from smc/src/webapp-lib/coffeescript and is actually a big file checked into the repo.

It may be best to assign this to @haraldschilly

DrXyzzy:
Note from @williamstein:
> It’s not so simple. https://cloud.sagemath.com/projects/10f0e544-313c-4efe-8718-2142ac97ad11/files/smc/src/smc-webapp/misc_page.coffee at line 1651 is where coffeescript gets loaded.

> That looks like maybe the old way before we used npm much or webpack.

>Yep, that’s from smc/src/webapp-lib/coffeescript and is actually a big file checked into the repo.

[...]

PR #1811, opened 2017-03-29
updating client-side code for %coffeescript cells

haraldschilly:
this updates the coffeescript code. I tried to figure out how to integrate this into the build process, but it has too many dependencies (i.e. one would have to compile from source, setup this cake builder, etc.). there is also no uptodate npm package.

upside: the mentioned code in #1810  works

screenshot from 2017-03-29 11-33-41

process: copy/paste a file, and failed to find a simple way to use the npm package directly.

time: 30 min


williamstein:
> I tried to figure out how to integrate this into the build process, but it has too many dependencies (i.e. one would have to compile from source, setup this cake builder, etc.). there is also no uptodate npm package.

We would definitely have to create our own npm package for this and publish it.  That's the right thing to do and would work.  However, coffeescript is so stable, there's no need right now, and we can do that next year...

PR #1813, opened 2017-03-29
webapp: pull up of latex editor's @cms to CodeMirrorEditor::codemirrors and some cleanup

### LINKED ISSUE #1809: webapp editor: restore_font_size not careful enough about codemirror
This is an easy one. 

```
TypeError: Cannot read property 'getWrapperElement' of undefined
    at n.set_font_size (https://cloud.sagemath.com/static/smc-74163f95a1c120d6d2ee.js?74163f95a1c120d6d2ee:21:28783)
    at n.set_font_size (https://cloud.sagemath.com/static/smc-74163f95a1c120d6d2ee.js?74163f95a1c120d6d2ee:21:742)
    at n.restore_font_size (https://cloud.sagemath.com/static/smc-74163f95a1c120d6d2ee.js?74163f95a1c120d6d2ee:21:28516)
    at n.restore_font_size (https://cloud.sagemath.com/static/smc-74163f95a1c120d6d2ee.js?74163f95a1c120d6d2ee:21:742)
    at new n (https://cloud.sagemath.com/static/smc-74163f95a1c120d6d2ee.js?74163f95a1c120d6d2ee:21:22032)
    at Object.exports.codemirror_session_editor (https://cloud.sagemath.com/static/smc-74163f95a1c120d6d2ee.js?74163f95a1c120d6d2ee:22:13118)
    at new f (https://cloud.sagemath.com/static/smc-74163f95a1c120d6d2ee.js?74163f95a1c120d6d2ee:208:20042)
    at Object.f (https://cloud.sagemath.com/static/smc-74163f95a1c120d6d2ee.js?74163f95a1c120d6d2ee:23:16036)
    at Object.init (https://cloud.sagemath.com/static/smc-74163f95a1c120d6d2ee.js?74163f95a1c120d6d2ee:190:14330)
    at Object.exports.initialize (https://cloud.sagemath.com/static/smc-74163f95a1c120d6d2ee.js?74163f95a1c120d6d2ee:52:747)
```

top frame, fails in the last `getWrapperElement`:

```
n.prototype.get_font_size=function(e){var t,n;return t=$(e.getWrapperElement()),null!=(n=t.data("font-size"))?n:this.default_font_size},
n.prototype.set_font_size=function(e,t){var n;if(t>1)return n=$(e.getWrapperElement()),n.css("font-size",t+"px")
```

since `e` is the `@codemirrorX` instance, the actual code to fix is

```
restore_font_size: () =>
        # we set the font_size from local storage
        # or fall back to the default from the account settings
        for i, cm of [@codemirror, @codemirror1]
             ...
```

idea: I would do the same as for the latex editor, i.e. all those 2-element lists should be replaced by a method `@cms()` like in `smc-webapp/editor_latex.coffee` →  method `cms: => ...`

time: 10 min

haraldschilly:
fixes #1809

test: check that editors and some features like saving/restoring font size still works

process: just following the idea to pull up the `@cms` from the latex editor into the `CodeMirrorEditor` class and using it in the latex editor and various other places.

time: 20 min

### DIFF
--- a/src/smc-webapp/editor.coffee
+++ b/src/smc-webapp/editor.coffee
@@ -1007,7 +1007,7 @@ class CodeMirrorEditor extends FileEditor
         @hide_startup_message()
         @element.find(".salvus-editor-codemirror-content").show()
         for cm in @codemirrors()
-            cm?.refresh()
+            cm.refresh()
 
     hide_startup_message: () =>
         @element.find(".salvus-editor-codemirror-startup-message").hide()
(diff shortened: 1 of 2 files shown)


src/smc-webapp/editor.coffee
```diff
@@ -1067,7 +1061,8 @@ class CodeMirrorEditor extends FileEditor
         cb?()
 
     codemirrors: () =>
-        return [@codemirror, @codemirror1]
+        c = [@codemirror, @codemirror1]
+        return underscore.filter(c, ((x) -> x?))
```
williamstein: Why do this when you can do: `(x for x in c when x?)`.  CoffeeScript has nice list comprehensions just like in Python (but with parens).

Also, it would be better to just define the list `[@codemirror, @codemirror1]` once, then change it once right when things are deleted, rather than filtering every single time.  

That said, this code is definitely all going to be deleted relatively soon, so I'm fine with it exactly as you wrote it.

Issue #1814, opened 2017-03-29
LatexEditor / update_preview: content sometimes undefined


haraldschilly:
`content` sometimes undefined, which is surprising ...

```
TypeError: Cannot read property 'match' of undefined
    at f.exports.LatexEditor.f.update_preview (https://cloud.sagemath.com/static/smc-74163f95a1c120d6d2ee.js?74163f95a1c120d6d2ee:209:5)
    at f.update_preview (https://cloud.sagemath.com/static/smc-74163f95a1c120d6d2ee.js?74163f95a1c120d6d2ee:208:16941)
    at Object.cb (https://cloud.sagemath.com/static/smc-74163f95a1c120d6d2ee.js?74163f95a1c120d6d2ee:208:31488)
    at https://cloud.sagemath.com/static/smc-74163f95a1c120d6d2ee.js?74163f95a1c120d6d2ee:11:17633
    at n._save (https://cloud.sagemath.com/static/smc-74163f95a1c120d6d2ee.js?74163f95a1c120d6d2ee:161:9101)
    at Object.f (https://cloud.sagemath.com/static/smc-74163f95a1c120d6d2ee.js?74163f95a1c120d6d2ee:160:31114)
    at n (https://cloud.sagemath.com/static/smc-74163f95a1c120d6d2ee.js?74163f95a1c120d6d2ee:11:16993)
    at function.e._wrapper.e._wrapper (https://cloud.sagemath.com/static/lib-74163f95a1c120d6d2ee.js?74163f95a1c120d6d2ee:117:25719)
    at https://cloud.sagemath.com/static/lib-74163f95a1c120d6d2ee.js?74163f95a1c120d6d2ee:117:27239
````

time: 3 min

PR #1815, opened 2017-03-29
webapp/latex: content sometimes undefined

haraldschilly:
see issue #1814

process: this is a simple defensive solution, the real question is what's going on with `@_get()`

time: 5 min


williamstein:
> the real question is what's going on with @_get()

Since update_preview is called in the save callback -- which could happen long after everything just got closed -- this is a reasonable fix.  This is precisely like the "cannot set state of unmounted component" class of react bugs we hit whenever we don't track mount state and have a callback that modifies state...

PR #1821, opened 2017-03-29
Clean up SearchInput and extra markup

johnjeng:
Cleans up what I think are the final holes for SearchInput.

- Removes extraneous markup and unifies search inputs fixing various small bugs.
  - For example the project search icon now correctly becomes orange when non-empty 
- Opening a directory now clears the search input since in most cases, you are not looking for something with the same name as the directory you searched for.
- Allows `↓` to trigger selection highlighting in project files.
- Improve readability of selected_file_index manipulations.


williamstein:
This is very good.

Issue #1836, opened 2017-04-05
document in show background option for 3d plotting.


williamstein:
1. Do this in a Sage worksheet:
```
show(icosahedron(color='green', opacity=.5, mesh=5, spin=10), background='rgb(8,3,1)')
```

2. Type `show?`. This background option is not documented.  This is a docstring in the smc_sagews python library.

So document it.

Also, if there is a way to make the background transparent that would probably be a better default.  If so make that another issue.

DrXyzzy:
Will update docstring.

Based on this posting http://stackoverflow.com/questions/20477951/make-a-renderers-background-transparent-but-not-shapes-three-js, it's probably not feasible to specify transparent background without setting the second parameter (alpha) to a value less than 1 in line 180 of 3d.coffee: 
```
@renderer.setClearColor(@opts.background, 1)
```
FWIW, I tested values less than 1 in the alpha parameter above attaching a .css file to a test worksheet with
```
.sagews-output-cm-wrap {
    background-color: cyan;
}
```
and was able to get transparent background when showing a 3d object, although with only that change, the value of `background` was ignored except for a small strip of color at the bottom of the image.

When is it useful to have transparent background for a 3d plot?

williamstein:
> When is it useful to have transparent background for a 3d plot?

Set your editor's theme (in account settings) to something with a dark background.

DrXyzzy:
ok will investigate further

PR #1837, opened 2017-04-06
add viewer and background to docstring for show(), enable transparent bg for three.js

DrXyzzy:
Ref: #1836 

time: 2:45
mostly researching feasibility of transparency option


DrXyzzy:
Ref: #1840.

Code added to enable transparent bg for three.js rendering & make it default.

williamstein:
nice!

haraldschilly:
works as it claims to.

time: 5 min

Issue #1802, opened 2017-03-23
chat preview does not display local images


williamstein:
E.g.,

1. Make an image called local.png somehow.
2. Type this in a chat: `<img src="local.png">`
3. Don't see it in preview (after clicking the preview button)
4. Do see it after submitting.

This is a bad bug since preview is precisely when you *do* expect to see how things will look...

PR #1839, opened 2017-04-06
webapp/chat: simple fix for #1802

haraldschilly:
unexciting fix for #1802

process: see what happens when a message is displayed

time: 10 min


haraldschilly:
merging, since completely harmless

Issue #1565, opened 2017-01-21
%latex mode in sage worksheets -- several of the buttons do nothing


williamstein:
1. Create a sage worksheet
2. Type `%latex` to see the latex bar
3. Many of the buttons on the right side, e.g., "center text", do absolutely nothing.

DrXyzzy:
Done, see #1842 for verification results.

PR #1842, opened 2017-04-07
webapp/sagews: making the %latex buttonbar a bit smarter and some enhancements

haraldschilly:
issue #1565

This isn't pretty, but it hides the undefined latex buttons. There are also new features like font size, and also striking through text. That in turn needs the `sout` package. Hence my question: @DrXyzzy do you know if the line `...latex.add_package_to_preamble_if_available('soul')` in `smc_sagews/sage_salvus.py` makes sense? It works for my dev project, so I'm quite sure it is at least heading into the right direction.

process: untangling the labyrinth of buttonbar commands, their special cases and fallbacks, reading what CodeMirror's `edit_selection` in misc_page does and adding more code to handle more latex commands.

test: in a sagews, make a `%latex` cell and click all the buttons. bonus feature: in debug mode, `edit_selection` will give a warning if it doesn't know what to do (no `how?`).

time: 75 min

### DIFF
--- a/src/smc-webapp/buttonbar.coffee
+++ b/src/smc-webapp/buttonbar.coffee
@@ -250,6 +250,18 @@ exports.commands =
             wrap :
                 left  : '\\underline{'
                 right : '}'
+        strikethrough :       # requires the soul package
+            wrap :
+                left  : '\\st{'
+                right : '}'
+        equation :
+            wrap :
+                left  : "$"
+                right : "$"
+        display_equation :
+            wrap :
+                left  : "$$"
+                right : "$$"
         insertunorderedlist :
             wrap :
                 left  : "\\begin{itemize}\n    \\item\n"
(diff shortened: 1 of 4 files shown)


williamstein:
Happy to see improvements here!!

haraldschilly:
oh yes, thank you for the feedback. most of your points (I guess image, fuzzy, etc.) are also true for `%latex` mode right now. We should make separate tickets for this or just accept them as they are for now.

* font is empty because no font is setup (I thought, it's better to show none than defunct ones ... adding latex fonts there could be a ticket of its own)

* hyperlink, I have to check (maybe it needs the `url` package?), and strikethrough gives an error unless that `soul` package is required. I guess, you didn't install that smc_sagews, or did you? That line there is a requirement to make this work :-)

DrXyzzy:
@haraldschilly - I ran tests in smc-in-smc off branch issue/1565 but somehow was not running dev code for server side. I tried with dev code is running on both front and back ends. Verified this by doing `latex.extra_preamble()` in the test sagews. Before processing any cells, it's empty, but after running any `%latex` cell, the preamble shows 'soul'.

1. strikethrough works
1. url still causes an error. The error goes away if I add 'hyperrerf' to the preamble, but there is no usable link because, at least in Safari, the entire output cell is a hyperlink to the image of the output.
1. image still causes file not found, even if I add 'graphicx' to the preamble

williamstein:
I'm going to merge this since it's a major improvement.  @DrXyzzy please open additional tickets/issues for problems you note that need to be addressed.

PR #1843, opened 2017-04-07
hub/api: blueprint for a general purpose API

haraldschilly:
blueprint for an `/api` endpoint in the hub


williamstein:
Is this something that should be refereed and merged, or?

haraldschilly:
I'll add something more useful first, then I'll signal.

williamstein:
@haraldschilly I just put some hard thought into this API problem, and this is not the right approach.  There is already an API, which is even mostly REST-ful, which is what is defined in 

   https://github.com/sagemathinc/smc/blob/master/src/smc-util/message.coffee

and what is implemented mostly via the (massive 2000+-line class) called Client here:

   https://github.com/sagemathinc/smc/blob/master/src/smc-hub/hub.coffee#L129

via the messages that start with `mesg_`.  

We need to:

1. Refactor the hub.coffee code a bit (for sanity) so that Client is in another file.
2. Provide a method via https to authenticate (can just fake it for testing/dev)
3. Expose *exactly* the api defined by message.coffee and Client via http rest.  With maybe 3 or 4. Exceptions, every single one of those messages is really a JSON object that gets one JSON object as a response -- i.e., it's an API.

Once that is done, it's already even documented (by message.coffee), and Minerva (or whoever) can just get to work.   It's a robust proven API as well, since we built everything on top of it.

What doesn't it do?  
  - push updates for sync tables
  - terminals

But it does provide 100% full ability to query any synctables.

williamstein:
Closing in favor of https://github.com/sagemathinc/smc/tree/json-api/src

Issue #1845, opened 2017-04-08
use alternative CDN for mathjax


DrXyzzy:
This issue is needed because The MathJax CDN hosted at cdn.mathjax.org will be shutting down on April 30, 2017.
https://www.mathjax.org/cdn-shutting-down/

There is one match for 'cdn.mathjax.org' in the smc repo, in src/smc-webapp/printing.coffee: https://github.com/sagemathinc/smc/blob/e7744160a30d3806a20e7c8dd81047ef22f2c9fd/src/smc-webapp/printing.coffee#L297

Replace as recommended in the online announcement:
`https://cdn.mathjax.org/mathjax/latest/MathJax.js?...`
becomes
`https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.0/MathJax.js?...`

Also:
> Note If you have been using https://cdn.mathjax.org/mathjax/latest/, you have to switch to a dedicated version going forward. That is, you will have to change the address manually to a higher number when new versions become available; the automatic switching provided by /mathjax/latest/ is not available on the cdnjs server.

DrXyzzy:
Done.

PR #1846, opened 2017-04-09
change url for mathjax CDN #1845

DrXyzzy:
Updated url per #1845.

Tested by introducing deliberate typo in the MathJax CDN URL and confirming that clicking print button at upper right for a sage worksheet did not produce LaTeX output, fixing the typo and seeing correct LaTeX in the HTML produced by clicking print.

time: 30 minutes


haraldschilly:
very good. I didn't think about that!

Issue #1835, opened 2017-04-04
sage worksheet bug -- ending a block of code with ? shouldn't always submit introspection request


williamstein:
1. Put this code in a worksheet:
```
a = 5 # what?
```
2. Hit shift enter and get introspection.

<img width="757" alt="screen shot 2017-04-04 at 10 07 49 am" src="https://cloud.githubusercontent.com/assets/1276278/24669250/9d6b9500-191e-11e7-9530-0bdef5f23e54.png">

williamstein:
[Here's how jupyter does it](https://github.com/jupyter/notebook/blob/master/notebook/static/notebook/js/codemirror-ipython.js) -- they make a custom python mode that treats `?` in a special way.  I don't like this approach at all.  

---

Comment I put on a PR

Sorry, I don't think this will help much with #1835 because:
(1) that needs to run on the frontend, which is when doing introspection is determined,
(2) that also needs to support **all languages**, not just python.  What you've done here must be only python.

The right solution for #1835  is to use Codemirror's own parser, and run CodeMirror.runMode over the code to be submitted first.  See, e.g., 
 https://github.com/sagemathinc/smc/blob/jupyter2/src/smc-webapp/jupyter/parsing.coffee#L7

That's not to say that we don't want this on the backend like you have it.  It's just not going to solve #1835.

williamstein:
This is easy now, since parsing.coffee is in master.

DrXyzzy:
The approach in `parsing.coffee` will work for a block ending with a comment ending with one or two question marks, but I don't see a simple solution for dealing with this input, which also evokes the error:
```
a = 5 # what?
b = 6
```

I've looked briefly at the python parser in codemirror at [src/smc-webapp/codemirror/mode/python.js](https://github.com/sagemathinc/smc/blob/jupyter2/src/smc-webapp/codemirror/mode/python.js), and don't see a simple way of extracting individual blocks, as is done in [src/smc_sagews/smc_sagews/sage_parsing.py](https://github.com/sagemathinc/smc/blob/jupyter2/src/smc_sagews/smc_sagews/sage_parsing.py)

Am I missing something?

williamstein:
> Am I missing something?

No, I forgot that the block parser is only implemented in the backend.  Dang.

DrXyzzy:
There is a fix for this bug that is commented out... Can we restore that fix and close this bug? 
[...]

PR #1849, opened 2017-04-10
sagews: use a lexer to parse codeblocks for stripping comments and "understanding" strings

haraldschilly:
see issue #1835

This is one way how how to deal with this. Part of this is to come up with a list of code-strings to test:

* `sage_server?` → docstring
* `x = random?   # bar` → docstring for random
* `x = random?   # plot?` → still docstring for random, not plot
* `x = 1 # plot?` → no docstring, x is 1
* `x="random?" # plot?` → assign x, no docstring for plot
* ```
  x = """
  salvus?
  """
  ``` 
  x is this multiline string, no help for salvus.

@DrXyzzy apart from the obvious fact that this is in the need for proper test cases, please check if I have missed something. I am also not sure if it doesn't break something else. I.e. note that the `sage_parsing.introspect` do get `b` now, and no longer the `block`. That's necessary, because otherwise it doesn't introspect correctly.

process: investigating a few techniques and tuning shlex in a way that it covers many (all?) cases.

time: 1 hour

### DIFF
--- a/src/smc_sagews/smc_sagews/sage_server.py
+++ b/src/smc_sagews/smc_sagews/sage_server.py
@@ -961,12 +961,19 @@ class Salvus(object):
             sys.stdout.reset(); sys.stderr.reset()
             try:
                 b = block.rstrip()
+                # get rid of comments at the end of the line -- issue #1835
+                from shlex import shlex
+                s = shlex(b)
+                s.commenters = '#'
+                s.quotes = '''"\''''
+                b = ''.join(s)
+                # e.g. now a line like 'x = test?   # bar' becomes 'x=test?'
                 if b.endswith('??'):
-                    p = sage_parsing.introspect(block,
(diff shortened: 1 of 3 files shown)


DrXyzzy:
@haraldschilly: Nice. I'd like to add test cases later today as part of review. Maybe include strings containing '#' and nested strings.

williamstein:
Sorry, I don't think this will help much with #1835 because:
(1) that needs to run on the frontend, which is when doing introspection is determined,
(2) that also needs to support **all languages**, not just python.  What you've done here must be only python.

The right solution for #1835  is to use Codemirror's own parser, and run CodeMirror.runMode over the code to be submitted first.  See, e.g., 
 https://github.com/sagemathinc/smc/blob/jupyter2/src/smc-webapp/jupyter/parsing.coffee#L7

PR #1851, opened 2017-04-10
util/misc: making "transform_get_url" smarter and adding tests

### LINKED ISSUE #1818: in +New, pasting in a URL to github should clone that repo.
This used to work, but is totally gone now.  

1. Click +New
2. Enter https://github.com/plotly/python-user-guide and hit enter or "download from network".
3. Get nothing good!

haraldschilly:
see issue #1818

this is work in progress, but at least I've added tests and expanded what it does a bit. I haven't actually checked if it fixes #1818.

time: 30 min

### DIFF
--- a/src/smc-util/misc.coffee
+++ b/src/smc-util/misc.coffee
@@ -1818,10 +1818,15 @@ exports.ticket_id_to_ticket_url = (tid) ->
 
 exports.transform_get_url = (url) ->  # returns something like {command:'wget', args:['http://...']}
     URL_TRANSFORMS =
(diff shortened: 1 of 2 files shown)


haraldschilly:
oh of course, that's definitely a file without extension when viewed as a "file". I hope searching for that error message string helps to find where this heuristic sits. My guess is, it isn't taking this into account. What it should do is to do the url transformation first and ignore this path when the command is "git" or "wget".

haraldschilly:
ok, this is more complex. the function to look at is `create_file` in `smc-webapp/project_store.coffee`

williamstein:
I'm merging this, but not marking it as fixing #1818.

Issue #1669, opened 2017-02-12
make it possible to edit the displayed email address for a student in a course


williamstein:
This is just a natural followup to https://github.com/sagemathinc/smc/pull/1645

Picture of where editing email would go:

<img width="654" alt="screen shot 2017-02-11 at 7 00 12 pm" src="https://cloud.githubusercontent.com/assets/1276278/22859322/818048e6-f08c-11e6-8c46-b3c99c40614b.png">

This will also make it possible add emails of students that you may have added by name, which is an often requested feature!

johnjeng:
This is done.

PR #1857, opened 2017-04-12
Allow course emails to be edited by instructors

johnjeng:
Addresses #1669.
edit_email
Works just like first and last name forms. 
- `Tab` to next input works. 
- `Enter` on any field submits the form. 
- `Esc` on any input cancels all changes. 
- The save button is disabled unless changes have been made

Does **not** display the student's prior or actual email.


williamstein:
Works fine; however, it might be natural to also not allow the user to use the same email address for multiple students?  
screen shot 2017-04-12 at 6 20 48 pm

That said, I'm merging this as is -- we only use the email right now for display purposes, anyways...

Issue #1238, opened 2016-11-16
Support drag and drop of images onto chatroom messages


timothyclemansinsea:
Many programs, including Slack, Google Hangouts, Facebook, and Gitter, support copy and paste of images into messages, issues, etc.   Let's do the same.

For this ticket, in particular, we should implement drag and drop of image files onto chat messages.  When the image is dropped, it should either be uploaded to the blob store or uploaded to a hidden subdirectory (not sure which), and a link `<img src=...>` should be inserted.   (What Github does, which involves changing the markdown for the link as the file uploads is interesting, but probably overkill.)   I think probably sending the file to the blob store would be best, since otherwise moving the .sage-chat file might break all embedded images.

johnjeng:
Done for dedicated chat room files. There's an issue open for side chat.

Issue #1237, opened 2016-11-16
For drag and drop of files; don't require user to move cursor to box


timothyclemansinsea:
When I tried out drag and drop of files I thought I had to do it twice. I later realized that if I dragged to below where the box appeared that I had to keep holding down the left mouse key and drag to the box. Google Drive allows me to drag to anywhere.

Issue #935, opened 2016-09-24
Make it possible to drag and drop images on chats, .md files, latex files, jupyter notebooks, sage worksheets


timothyclemansinsea:
I just experienced the coolest thing. I was making a Github issue and was able to drag and drop my cropped screenshot image. 

On drag and drop on one of these file editors SMC would save the file and then based on the editor add the correct markup code for embedding it. 

These are the areas where this might work. There should be more concrete tickets for each of them:
- [x] chat: that's https://github.com/sagemathinc/cocalc/issues/1238 ?
- [ ] text editor (md, html, ...)
- [ ] latex
- [ ] jupyter
- [ ] sagews

williamstein:
>  SMC would save the file 

For precision -- it would save the file to the database blob store.

timothyclemansinsea:
@williamstein is this database blob store used for anything else right now? I would have assumed this would be saved as a regular file.

williamstein:
>  is this database blob store used for anything else right now?

Yes, it is used for all output in Sage worksheets.  That's why you can copy/paste content of a worksheet **with images** from one project to another no problem, and why there aren't image files cluttering up the filesystem. 

Likewise, here, by making drag and drop work this way you don't even have to worry about what the image is called or where it is stored.  

The blob store is automatically moved to Googel cloud storage after a week (or so) and served from there, so it is very scalable.

haraldschilly:
I've added the label metabug [in the hopes it means what I think], since it covers many different areas.

williamstein:
> I've added the label metabug [in the hopes it means what I think]

I'm not sure I know what it means either.  However, this is definitely not a bug, but a feature/enhancement, and a "metabug" should be a bug at least...

williamstein:
What about a new label: "metafeature" or "meta-enhancement"?

haraldschilly:
or just "metaissue" to cover everything? We could also edit the very first
description to include `- [ ]` checkboxes in a list. Then, github will show
a progress bar for partially completed issues and we categorize these as
"metaissue". from that list there, we could easily reference more concrete
issues.

williamstein:
Most of this has been implemented years ago...

PR #1858, opened 2017-04-12
Fix file drop notification

### LINKED ISSUE #1478: get rid of the "File drop disabled To upload a file, drop it onto the files listing or the "Drop files to upload" area in the +New tab." warning message when using codemirror
Codemirror supports drag and drop of text.  However, I often see the message

screen shot 2017-01-09 at 2 08 09 am

when dragging around text.  Disable this on file editing page...

johnjeng:
Fixes #1478, #1237. 

Provides stepping stone to #1238 and #935 by creating an initial wrapper component that directly uses Dropzone.js. It is basically a slightly altered coffeescript rewrite of `react-dropzone-component` that is still used on the +New page.

List of other changes below
- Dragging non-files (like icon links) across the page now does nothing instead of giving the user an error.
- Dragging a file onto a non-drop-enabled area now says "File Rejected" instead of "File Drop Disabled"
- Dragging a file onto the files area now works automatically without having to hit some specific dropzone that gets rendered.
  - Also works on a directory with no files.
- Clicking Upload on the Files page now opens a file upload dialogue.
- Upload status on the Files page is now only as big as it needs to be
upload status

Tested:
- Clicking X on files page uploader clears the status area
- Dragging multiple files works
- Uploader form results in upload status identical to dragging onto files
- Dropping non-files anywhere does nothing unless defined elsewhere (ie, text onto a text field)
- Dropping files only does some
- Dropzone on the +New Page stays the same

### DIFF
--- a/src/smc-webapp/desktop_app.cjsx
+++ b/src/smc-webapp/desktop_app.cjsx
@@ -152,11 +152,12 @@ Page = rclass
             console.log "react desktop_app.drop", e
         e.preventDefault()
         e.stopPropagation()
-        {alert_message} = require('./alerts')
-        alert_message
-            type     : 'info'
-            title    : 'File drop disabled'
-            message  : 'To upload a file, drop it onto the files listing or the "Drop files to upload" area in the +New tab.'
+        if e.dataTransfer.files.length > 0
+            {alert_message} = require('./alerts')
+            alert_message
+                type     : 'info'
+                title    : 'File Drop Rejected'
+                message  : 'To upload a file, drop it onto the files listing'
 
     render: ->
         style =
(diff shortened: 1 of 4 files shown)


williamstein:
# **WOW**

williamstein:
This broke building the static pages on the backend, because you put a bunch of code in r_misc that can't be imported anymore.  I'll move it out.

williamstein:
This is how you can see the error: `~/smc/src$ scripts/update_react_static `

williamstein:
Instead of `Object.prototype.toString.call(handlers) == '[object Array]'` just do `misc.is_array(handlers)`

williamstein:
See https://github.com/sagemathinc/smc/commit/5bcf851986a5de00bd232beba02f02166f732ccd

Issue #1738, opened 2017-03-03
webapp: Cannot read property 'collection' of undefined in react-sortable-hoc


haraldschilly:
This is the exception which happens roughly most frequently. I wasn't able to track it down to anything specific. `react-sortable-hoc` is only used for the navigation, so maybe it's possible to replicate it somehow.

```
TypeError: Cannot read property 'collection' of undefined
    at e.value (https://cloud.sagemath.com/static/smc-fb285af1a52913f90652.js?fb285af1a52913f90652:220:27818)
    at n.t.handlePress (https://cloud.sagemath.com/static/smc-fb285af1a52913f90652.js?fb285af1a52913f90652:220:30167)
    at HTMLUListElement.t.handleMove (https://cloud.sagemath.com/static/smc-fb285af1a52913f90652.js?fb285af1a52913f90652:220:29933)
    at HTMLUListElement.function.e._wrapper.e._wrapper (https://cloud.sagemath.com/static/lib-fb285af1a52913f90652.js?fb285af1a52913f90652:118:5579)
```

top frame is in `react-sortable-hoc`, the `this.active.collection`
```
{key:"getActive",value:function(){var e=this;return(0,s.default)(this.refs[this.active.collection],function(t){var n=t.node;return n.sortableInfo.index==e.active.index})}},
```

bottom: fails at `handlePress` here, which is most likely in `node_modules/react-sortable-hoc/dist/commonjs/SortableContainer/index.js`

```
,t.handleMove=function(e){var n=t.props.distance;if(!t.state.sorting&&t._touched){t._delta={x:t._pos.x-e.clientX,y:t._pos.y-e.clientY};var r=Math.abs(t._delta.x)+Math.abs(t._delta.y);n?r>=n&&t.handlePress(e):t.cancel()}},
```

Maybe there is also an update available (no idea) and we should upgrade.

williamstein:
The exception can't be where you say in handePress, since active is undefined, but is successfully accessed the line before in handlePress. It has to be here:
```
                        _this.handleSortEnd = function (e) {
                                var _this$props3 = _this.props;
                                var hideSortableGhost = _this$props3.hideSortableGhost;
                                var onSortEnd = _this$props3.onSortEnd;
                                var collection = _this.manager.active.collection;
 
                                // Remove the event listeners if the node is still in the DOM
```

williamstein:
[...]

PR #1872, opened 2017-04-18
Only sort tabs when dragging a sortable item

johnjeng:
Fixes case for #1738 for which I could reproduce.

Reproduce case: Drag on the space near a file or project tab and you get the error. 

Fix: Dragging on that space no longer triggers react-sortable.


williamstein:
thanks!

PR #1882, opened 2017-04-21
Fix sales tax from WA DOR programatically

johnjeng:
Changed a long with all current subscriptions.

Generated via http://dor.wa.gov/Content/FindTaxesAndRates/RetailSalesTax/DestinationBased/ClientInterface.aspx


haraldschilly:
Well, in such a case I'm a fan of line breaks, but ok, doesn't matter that much here. I've added a simple test to confirm that a zip code exists, but maybe there should be something better to check all values are in a reasonable range?

I've no idea about us zip codes, but are there **all** numbers in a given range that should exist as keys? A test could do this:

1.  iterate through all expected zip code numbers and check for their existence in that map
2. The value should be between 0.0001 and 0.2000

I've added a commit here on your branch, which fails because there are missing zip codes. So, what I wrote above is probably wrong.

haraldschilly:
well, I added a list of zip codes from another source, but the test still fails. this is probably not useful at all ...

johnjeng:
I used this list http://www.zipcodestogo.com/Washington/
Some of them you don't get a value from Wasington's DOR.

src/smc-util-node/test/misc_node-test.coffee
```diff
@@ -48,4 +48,25 @@ describe "sanitizing HTML", ->
...
+
+describe 'check sales tax work', ->
+    sales_tax = misc_node.sales_tax
+
+    it 'knows about 98122', ->
+        expect(sales_tax('98122')).toBe 0.10100
```
williamstein: change to use parens for consistency

src/smc-util-node/test/misc_node-test.coffee
```diff
@@ -48,4 +48,25 @@ describe "sanitizing HTML", ->
...
+    it 'knows all expected zip codes', ->
+        for zip in WA_zips
+            v = sales_tax(zip)
+            expect(v).toBeGreaterThan(0)
```
williamstein: sales tax is the sum of the state rate and a local rate.  The WA state rate is 6.5%.
See http://www.taxrates.com/state-rates/washington/?referrer=https://www.google.com/&lastReferrer=www.taxrates.com&sessionId=1492826583854
So, change this > 0 to >= 6.5%.

src/smc-util-node/misc_node.coffee
```diff
@@ -847,10 +847,10 @@ exports.loadSync = (filename) ->
     JSON.parse(fs.readFileSync(filename).toString())
 
 
-# WA state sales tax rates, as of August 2015.
-# Generated scripts/sales_tax.py
+# WA state sales tax rates, as of April 20 2016.
+# Generated via http://dor.wa.gov/Content/FindTaxesAndRates/RetailSalesTax/DestinationBased/ClientInterface.aspx
```
williamstein: 1. Did you write a script to run through all zip codes and call the API a the URL, but **NOT** include that script with this patch?

2. Why are these rates different than the ones you included into Sage a few weeks ago?  Did you make typos, or what happened?

johnjeng: 1. Yes. Patching with the script.
1. For whatever reason, these rates are different from another search tool that the DOR provides. This one returns lower values so I used this one.

src/scripts/sales_tax.py
```diff
@@ -1,16 +1,71 @@
...
+    result_code = parse_result_code(text)
+
+    if int(result_code) != 2:
+        return None
+    elif tax_rate < 0:
+        raise ValueError("Result code was 0 but tax rate was negative for zip code {}".format(zip_code))
+    else:
+        return float(tax_rate)
+
+with open('wa_zips.txt') as inputfile:
```
johnjeng: @williamstein  How do I upload a file to aws so that it is accessible like the old taxrates file?

src/smc-util-node/test/misc_node-test.coffee
johnjeng: as in `toBe(0.10100)`?

src/scripts/sales_tax.py
williamstein: What is "aws"?

src/smc-util-node/test/misc_node-test.coffee
williamstein: yep

src/scripts/sales_tax.py
johnjeng: Oh, Amazon web services. I presume we put up https://s3-us-west-2.amazonaws.com/taxrates.csv/TAXRATES_ZIP5_WA201508.csv

src/smc-util-node/test/misc_node-test.coffee
johnjeng: That seems inconsistent with the rest of our tests.
They are largely of the form
`thing.should.be "value"`

eg.
`l({}).should.be.exactly 0`
`should.be.exactly 47`
`expect(ret).toBe "<p>hello</p>"`

src/scripts/sales_tax.py
johnjeng: Or should I just put the file in the scripts folder?

williamstein: I've never used AWS S3.   Maybe jlee long ago.   But yes, just put the file in our repo somewhere.

src/smc-util-node/test/misc_node-test.coffee
williamstein: smc-util/test might be that way, but that was mostly written by @haraldschilly as his first thing on SMC long ago, and conflicts with our codebase style.   We don't really have time to worry about such stuff though, so forgot about this for now.

src/smc-util-node/misc_node.coffee
williamstein: OK.

src/smc-util-node/misc_node.coffee
```diff
@@ -847,10 +847,10 @@ exports.loadSync = (filename) ->
     JSON.parse(fs.readFileSync(filename).toString())
 
 
-# WA state sales tax rates, as of August 2015.
-# Generated scripts/sales_tax.py
+# WA state sales tax rates, as of April 20 2016.
```
williamstein: Surely you mean **April 20, 2017** ???

williamstein:
What about Redondo?  -- https://www.area-codes.com/zip-code/zip-code-98054.asp   It's a real place and missing. 

The test currently fails exposing zip codes that are 100% legit, but not in our table.  E.g.,
```
~/smc/src/smc-util-node$ mocha test/misc_node-test.coffee
  check sales tax work
    ✓ knows about 98122
98054 0
    1) knows all expected zip codes
 
  8 passing (633ms)
  1 failing
 
  1) check sales tax work knows all expected zip codes:
     Error: Expected 0 to be greater than 0.065
```

johnjeng:
Entering 98054 into http://dor.wa.gov/content/findtaxesandrates/salesandusetaxrates/lookupataxrate/ returns the error `The zip code entered is an invalid zip code or not in Washington state`.

Using the map from the same tool and clicking on a spot in the area that other maps give me, it says 0.86% so I'll use that?

src/smc-util-node/misc_node.coffee
johnjeng: Yes.

williamstein:
> Using the map from the same tool and clicking on a spot in the area that other maps give me, it says 0.86% so I'll use that?

The minimum is 6.5% or so, right, since isn't tax the state tax (of 6.5% or whatever) plus a local rate.  Maybe by 0.86% you mean 8.6%?

johnjeng:
Oh, yes. I mean 8.6%.

PR #1892, opened 2017-04-24
Disabling require.js

haraldschilly:
Two PRs for one!

This builds on top of #1891 and #1889 and disables require.js. Github says it can't be merged? Ohh, there is already a comment in the pug file. Comments are ususally `//-`, otherwise with `//` it's a silly `<!-- .... -->` comment.

Despite that, this also needs some testing. It changes the bootbox (overlay dialog box, e.g. for terminal paste) and updates spin.js (see referenced PRs)


johnjeng:
Works for me. Tested usage of the terminal in smc. Does spin.js governs the dotted spinner like on the loading page?

Issue #1896, opened 2017-04-25
dropzone → @dropzone.options → Cannot read property 'options' of undefined


haraldschilly:
all public pages are broken.

```
TypeError: Cannot read property 'options' of undefined
    at Object.componentWillUpdate (https://cloud.sagemath.com/static/smc-3abb158ef8abcbae2d76.js?3abb158ef8abcbae2d76:193:9856)
    at u._performComponentUpdate (https://cloud.sagemath.com/static/smc-3abb158ef8abcbae2d76.js?3abb158ef8abcbae2d76:187:5558)
    at updateComponent (https://cloud.sagemath.com/static/smc-3abb158ef8abcbae2d76.js?3abb158ef8abcbae2d76:187:4946)
    at u.receiveComponent (https://cloud.sagemath.com/static/smc-3abb158ef8abcbae2d76.js?3abb158ef8abcbae2d76:187:4057)
    at Object.receiveComponent (https://cloud.sagemath.com/static/smc-3abb158ef8abcbae2d76.js?3abb158ef8abcbae2d76:30:20189)
    at Object.updateChildren (https://cloud.sagemath.com/static/smc-3abb158ef8abcbae2d76.js?3abb158ef8abcbae2d76:186:30952)
    at h._reconcilerUpdateChildren (https://cloud.sagemath.com/static/smc-3abb158ef8abcbae2d76.js?3abb158ef8abcbae2d76:187:30032)
    at h._updateChildren (https://cloud.sagemath.com/static/smc-3abb158ef8abcbae2d76.js?3abb158ef8abcbae2d76:187:30797)
    at h.updateChildren (https://cloud.sagemath.com/static/smc-3abb158ef8abcbae2d76.js?3abb158ef8abcbae2d76:187:30695)
    at h._updateDOMChildren (https://cloud.sagemath.com/static/smc-3abb158ef8abcbae2d76.js?3abb158ef8abcbae2d76:187:17353)
```

i.e.

```
return this._create_dropzone()},componentWillUpdate:function(){return this.dropzone.options=$.extend(!0,{},this.dropzone.  [BANG]  options,this.get_djs_config())},
```

which is in `smc-webapp/smc-dropzone`

```
    componentWillUpdate: ->
        @dropzone.options = $.extend(true, {}, @dropzone.options, @get_djs_config())
```

Didn't we already fixed that bug? Was https://github.com/sagemathinc/smc/pull/1869 never merged?

time: 20 min

[...]

Issue #1895, opened 2017-04-25
display of mathjax in safari (in markdown output) is completely broken


williamstein:
1. Open a chatroom (or anything involving markdown) using Safari.
2. Type `hello $x^3$`
3. See nothing where the math should be.  It might appear for a moment then vanish.

If we don't fix this in the next 1-2 days, we should put a big red "don't use safari" error that appears when you try to use SMC with safari...

DrXyzzy:
Could this be related: mathjax/MathJax#1732 `MathJax fails silently when run locally with RequireJS` ? Doesn't explain why we only see the problem with Safari, but thought I'd mention it.

williamstein:
I disabled requirejs already and the problem persists, I think.  It may have been a problem for a while....?

haraldschilly:
I don't have safari, so hard to test. Same holds true for math output in sagews? As far as I know there was no change to how it is setup or even Mathjax's version. Maybe Safari was updated recently?

haraldschilly:
one hint: [mathjax 2.7.1](https://github.com/mathjax/MathJax/releases/tag/2.7.1)

the only weird thing is it was updated on npm only a few hours ago. but maybe it went in earlier with the latest update? or maybe there was a broken npm release and updating again will fix it? we should first check such a safari browser against 2.7.1 before drawing conclusions on our smc setup here.

haraldschilly:
is it safari 10.1? there are upstream bugs mentioned at mathjax, e.g. https://github.com/mathjax/MathJax/issues/1709

DrXyzzy:
As of latest merge, Mathjax in markdown displays in Safari 10.1 logged-in sessions. But it still does not display when the file is published and viewed with Safari 10.1 from a non-logged-in session; note the test is done in regular viewing mode, not private browsing mode. Private browsing mode has issues with local storage and should not be used with SMC.

williamstein:
Technically we could enable public viewing (when not logged in) without localStorage.   I just haven't yet, due to lack of effort.

williamstein:
Assigning me and unassigning @haraldschilly , since duh -- he doesn't have a mac or safari so can't test this.

williamstein:
> Same holds true for math output in sagews? 

NO -- everything works fine there.  It's a bug in our complicated markdown rendering pipeline. Should be easy to isolate by commenting out things.

williamstein:
time: 22 minutes

process: comment out in the process flow for rendering the HTML component

PR #1897, opened 2017-04-25
webapp/dropzone: hotfix for #1896 based on #1869

haraldschilly:
see #1896 based on #1869

test: dropzone works in public view

process: no idea what's going on, seems like the previous changes to that did revert some earlier bugfixes -- or -- some code is missing and wasn't committed.

time: 30 min


haraldschilly:
ok, this 067b3ed is live, including the updated mathjax (maybe related to #1895 )

PR #1904, opened 2017-05-01
safe html sanitization

haraldschilly:
* introduces `sanitize_html_safe` which removes script tags and possibly dangerous attributes
* switches webapp to use it
* updates back-end jquery to 3+ (required to make this work)
* some tests on the back-end to demonstrate what it does

test: put this in a chat message and check via inspector that the script and onclick attribute are gone.

```
<pre onclick="javascript: alert(321);">click me</pre>

<script>alert(123);</script>
```

for the review, probably the most important detail is how react's HTML and Markdown tags are extended. I didn't change the default behavior of the sanitize_html function, but IMHO those tags default should change.

time: 1 hour


williamstein:
Question: where is the **backend html sanitation** ever actually used?  I can't remember or find it used anywhere, and can't think of any reason for it....?

williamstein:
time: 18 minutes

Issue #1899, opened 2017-04-25
video chat "Enter room" button doesn't work in chromebooks and chrome browser


DrXyzzy:
> [@williamstein] Kevin lui reports (and I confirm) that video chat (via appear.in) doesn’t work on chromebooks. In particular, clicking the “Enter room” button leads to a traceback:
```
ScreenShareAction: check-extension message: undefined
4libs.facb87b2.js:2 Unhandled error from JOIN_ROOM: already in a room
(anonymous) @ libs.facb87b2.js:2
serviceworker.js:1 Service Worker termination by a timeout timer was canceled because DevTools is attached.
```
> Probably best solution is add a link to the top of our window that pops up saying: if room doesn’t work, click to open room here…

DrXyzzy:
"Enter room" button also does not appear when using Google Chrome 57.0.2987.133 (64-bit) on OS X 10.12.4. Clicking "Video Chat" button in a chat panel pops up a blank window.

DrXyzzy:
Updated chrome to 58.0.3029.81 (64-bit) and problem is still there.

haraldschilly:
with some details how we embed it we could email them at feedback@appear.in

On Apr 25, 2017 20:27, "Hal Snyder" <notifications@github.com> wrote:

"Enter room" button also does not appear when using Google Chrome
57.0.2987.133 (64-bit) on OS X 10.12.4. Clicking "Video Chat" button in a
chat panel pops up a blank window.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/sagemathinc/smc/issues/1899#issuecomment-297122718>,
or mute
the thread

williamstein:
Problem isn't there for me on chrome, and never has been. Weird.

DrXyzzy:
Might be two different problems, since there are no tracebacks or console error log messages with Google Chrome. Just a blank window. I tried whitelisting sagemath.com and appear.in in the popup list, but no change.

DrXyzzy:
This is repeatable for me with Chrome:
1. Open a chat file.
1. While `Video Chat` button is white, hover cursor over it, wait for it to change to gray backgound, then wait for the "Toggle Video Chat" pop-up.
1. Click `Video Chat`. Pop-up appears and __`Enter room` button is visible__ (and it works, but don't enter the room right now).
[...]

PR #1911, opened 2017-05-03
webapp/video-chat: workaround for #1899

haraldschilly:
see #1899

test: (i.e. what I checked in chrome and firefox) open video chat, indicator is "1", close the window, ~1 sec later the indicator is "0".

process: testing some approaches, in the end http://stackoverflow.com/questions/3291712/is-it-possible-to-open-a-popup-with-javascript-and-then-detect-when-the-user-clo was most helpful.

time: 25 min

### DIFF
--- a/src/smc-webapp/video-chat.cjsx
+++ b/src/smc-webapp/video-chat.cjsx
@@ -33,19 +33,20 @@ VIDEO_UPDATE_INTERVAL_MS = 30*1000
 VIDEO_CHAT_LIMIT         = 8       # imposed by free appear.in plan
 
 # The pop-up window for video chat
-video_window = (title, url) ->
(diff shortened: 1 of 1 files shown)


williamstein:
time: 7 minutes

DrXyzzy:
Reviewer note: looks great.

One thing I noticed is that the popup seen on mouseover of the `(n) Video Chat` button, with the toggle message and chat occupants, does not appear in firefox.

haraldschilly:
> One thing I noticed is that the popup seen on mouseover of the (n) Video Chat button, with the toggle message and chat occupants, does not appear in firefox.

Hmm, interesting @DrXyzzy ... I haven't even looked what code that is, maybe it isn't working everywhere? We should make a ticket for this and make a note to investigate it.

PR #1912, opened 2017-05-03
jupyter2: tablewidth auto, otherwise it looks ridiculous for some R dataframes

haraldschilly:
e.g. `data.frame(list(x=c(1,2,4), y = c(9,8,2)))`


williamstein:
Ridiculous indeed!

Issue #1948, opened 2017-05-11
get rid of the "### reloading attached file compute_lambda.sage modified at 21:25:11 ###" message when using %attach


williamstein:
... or at least display it somehow else.  It's really obnoxious.

PR #1949, opened 2017-05-11
move annoying attach/reload message from cell output to log

DrXyzzy:
See #1948 

Verified that pytests for %attach still work.


williamstein:
If somebody wants a message on attach reload, they can easily put a print statement in the file itself.

PR #1950, opened 2017-05-11
Add file listing headers

johnjeng:
Adds sortable headers to the file listing.

file sorting
**Type**
Sorts from the end of the file in chunks delimited by a `.`
type sort

**Name**
Sorts alphabetically ignoring case

**Date Modified**
Sorts by time last modified

**Size**
Folders are treated as size = -1 so they always come first (or last) when sorting by size.

**Other notes**:
- Removed the old button for toggling between sort by time and sort by name.

## Tested
- Sorting order remains when navigating away
- Sorting order resets after closing and reopening the project
- Sorting order respects account defaults
- Sorts *all* files in a directory, not just the ones shown on the page


src/smc-webapp/project_store.coffee
```diff
@@ -648,24 +645,34 @@ class ProjectActions extends Actions
             opts?.finish_cb?()
         )
 
+    # Sets the active file_sort to next_column_name
+    set_sorted_file_column: (column_name) =>
+        current = @get_store()?.active_file_sort
```
williamstein: It makes no sense to put a ? here and then use `current.column_name` in the next line.  You absolutely need `current.column_name` if there is any possible reason to use `@get_store()?`.

williamstein: I just fixed this myself.  Everything else looks very good.

PR #1995, opened 2017-05-23
Fix vertical spacing when chat is open

johnjeng:
Notice how the file tab floats above the editor.
offset

Shrinking the video button a little bit fixes this.
fixed

### DIFF
--- a/src/smc-webapp/video-chat.cjsx
+++ b/src/smc-webapp/video-chat.cjsx
@@ -178,10 +178,10 @@ exports.VideoChatButton = rclass
 
     render: ->
         num_users_chatting = @video_chat.num_users_chatting()
+        style = {height : '31.4px'}
         if num_users_chatting > 0
-            style = {color: '#c9302c'}
-        else
-            style = {}
+            style.color = '#c9302c'
+
         <Button onClick={@click_video_button} style={style}>
             <Tip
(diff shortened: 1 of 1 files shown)


src/smc-webapp/video-chat.cjsx
```diff
@@ -178,10 +178,10 @@ exports.VideoChatButton = rclass
...
     render: ->
         num_users_chatting = @video_chat.num_users_chatting()
+        style = {height : '31.4px'}
         if num_users_chatting > 0
-            style = {color: '#c9302c'}
-        else
-            style = {}
+            style.color = '#c9302c'
+
         <Button onClick={@click_video_button} style={style}>
```
haraldschilly: maybe we can make this a small button (`.btn-sm`)? looks odd to me to have  fraction px height value ... or is this how it is done now?

haraldschilly:
For the record, an hour ago I wasn't even at home. Now I understand why it's weird that you se  **me** doing that label :upside_down_face:

src/smc-webapp/video-chat.cjsx
johnjeng: I'm not sure. I think it's not definitively the best way to do it but the rest of the spacing for file tabs is sort of written this way with very specific values.

Issue #1987, opened 2017-05-23
document the API


williamstein:

haraldschilly:
@DrXyzzy is this done? If so, please close it. (I just closed the last remaining issue about the introduction)

DrXyzzy:
@haraldschilly,  Last part of API doc (& test) is waiting on #2185. Will you please have a look at that?

DrXyzzy:
Done after merge & release of multiple PRs, latest being #2185. Will open new issues for edge cases and stripe API calls.

PR #1997, opened 2017-05-24
first draft of in-source doc on CC api

DrXyzzy:
Added to message.coffee:
- general comments about API messages
- specific docs and examples for `ping`, `create project`, and `get_usernames`.

### DIFF
--- a/src/smc-util/message.coffee
+++ b/src/smc-util/message.coffee
@@ -226,7 +226,7 @@ API message
 ###
 Note: Options for the 'get_usernames' API message must be sent as JSON object.
 example (reformatted for readability):
-  curl -X POST -u sk_abcdefQWERTY090900000000: -H "Content-Type: application/json" \
+  curl -u sk_abcdefQWERTY090900000000: -H "Content-Type: application/json" \
     -d '{"account_ids":["cc3cb7f1-14f6-4a18-a803-5034af8c0004","9b896055-920a-413c-9172-dfb4007a8e7f"]}' \
     https:// cocalc.com/api/v1/get_usernames
   ==>  {"event":"usernames",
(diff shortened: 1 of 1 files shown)


src/smc-util/message.coffee
```diff
@@ -195,10 +217,21 @@ message
 #############################################
 
 API message
-    event       : 'get_usernames'
+    event       : 'get_usernames' # get first and last names for a list of account ids
     id          : undefined
     account_ids : required
 
+    # Note: Options for this messsge must be sent as JSON object.
```
williamstein: typo: ' messsge' -->  message

src/smc-util/message.coffee
```diff
@@ -195,10 +217,21 @@ message
...
+    # example (reformatted for readability):
```
williamstein: Use ### for multiline coffeescript comments, e.g.
###
a bunch
of lines
###

src/smc-util/message.coffee
```diff
@@ -448,7 +481,7 @@ message
 # starting_bundle_number.
 #
 # client --> hub --> project_server
-API message
+API message # unimplemented
```
williamstein: Why do you say "unimplemented?"  It is too implemented.

src/smc-util/message.coffee
```diff
@@ -775,10 +813,24 @@ message
 ###
 Ping/pong -- used for clock sync, etc.
 ###
+
 API message
-    event : 'ping'
+    event : 'ping' # test connection to API back end
```
williamstein: Returns time as ISO string when server responds to ping

src/smc-util/message.coffee
```diff
@@ -798,12 +850,12 @@ logged in.
 # return a JSON object with all data that is
 # meant to be publically available about this project,
 # who owns it, the title/description, etc.
-API message
+API message # unimplemented
```
williamstein: why  unimplemented ?

If anything is genuinely unimplemented, then it should just be deleted.

src/smc-util/message.coffee
DrXyzzy: > Why do you say "unimplemented?" It is too implemented.

I got unknown message event errors and didn't see the message event type elsewhere in the source using `git grep`.
```
 {"error":"unknown message event type 'public_get_project_info'"}
 {"error":"unknown message event type 'public_project_info'"}
 {"error":"unknown message event type 'save_project'"}
```

williamstein: OK, they must have got removed from the implementation; I think they should get deleted entirely then, and that functionality is now provided in a different way (i.e., via the query message type).

DrXyzzy:
oops, forgot this: Ref: #1987

williamstein:
What's the status of this?  I could just merge it; my only fear is that we were wrong about deleted messages (but I don't think we are).  Are you going to rewrite it to use message2 today, or later...?

DrXyzzy:
See new pr #2018 which uses API message2 and includes recent changes from master. This PR shouldn't be merged.

PR #1999, opened 2017-05-24
documentation and examples for message protocol as data

haraldschilly:
This introduces a version 2 as `message2`.

The way I tested this is minimal, in `coffee`:

```
coffee> m = require('./message')

coffee> m.query(query: "test")
{ query: 'test', multi_response: false, event: 'query' }

coffee> m.documentation
{ query:
   { description: 'This queries directly the database (sort of Facebook\'s GraphQL)',
     fields:
      { id: 'A unique UUID for the query',
        query: 'The actual query (required)',
        changes: '',
        multi_response: ' (default: false)',
        options: '' } } }

coffee> m.examples
{ query: [ [ [Object], [Object] ], [ [Object], [Object] ] ] }
```

### DIFF
--- a/src/smc-util/message.coffee
+++ b/src/smc-util/message.coffee
@@ -45,22 +45,26 @@ message = (obj) ->
 # TODO document it, for now just search for "message2" to see examples
 message2 = (obj) ->
 
-    mk_descr = (val) ->
-        descr = val.descr 
-        descr += ' (required)' if val.init == required
-        return descr
+    mk_desc = (val) ->
+        desc = val.desc 
+        if val.init == required
+            desc += ' (required)'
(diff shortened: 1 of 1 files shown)


src/smc-util/message.coffee
```diff
@@ -1034,13 +1068,33 @@ message # client ← hub
...
+        changes:
+             init  : undefined
+             desc  : ''
+        multi_response:
+             init  : false
+             desc  : ''
+        options:
+             init  : undefined
+             desc  : ''
+    examples: [
```
williamstein: Maybe examples should be commented out or have a comment that these are not valid examples?

PR #2002, opened 2017-05-25
font files and webpack

haraldschilly:
For some reasons, webpack did create small one-line wrapper scripts disguised as font files. Of course, that's not going to work out well, if the content type of e.g. a *.ttf file is binary but contains a small js script. 

This here changes the configuration and well, the most important aspect is to test it. I.e. when webpack-production compiles, that there are no errors about unknown file types (that happens, when e.g. the regex `\.ttf(\?[a-z0-9\.-=]+)?$` doesn't match and it falls back to a default). The second aspect to test is that when the page loads the newer font-icons for the symbols show up. (right at the moment when I wrote this, https://test.cocalc.com/ works with that newer config)


williamstein:
Looks good to me.

Issue #2001, opened 2017-05-25
zoom out some (and in) not working on newly-opened .tex file


DrXyzzy:
User report:

>After creating a brand new .tex file, if I click on either the "zoom in a little" button or the "zoom out a little" button, the result is the same: it zooms in significant. The second time pressing either button, the behavior is normal. 

I am able to reproduce the problem using Firefox and Chrome.

PR #2003, opened 2017-05-25
zoom width for preview was uninitialized in latex editor #2001

DrXyzzy:
Ref: #2001 

To test
1. install the PR
1. create a new .tex file and wait for preview to appear in right pane.
1. before doing any edits, click "zoom out a little". Preview should shrink to show black margins at sides
1. save and reload the .tex file and confirm preview size is remembered
1. create another new .tex file and confirm that "zoom in a little" works as expected before any edits are done

### DIFF
--- a/src/smc-webapp/editor_latex.coffee
+++ b/src/smc-webapp/editor_latex.coffee
@@ -93,10 +93,7 @@ class exports.LatexEditor extends editor.FileEditor
             @spell_check()
 
         @latex_editor.syncdoc.on 'connect', () =>
-            if @load_conf().zoom_width?
-                @preview.zoom_width = @load_conf().zoom_width
-            else
-                @preview.zoom_width = 100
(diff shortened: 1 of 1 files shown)


src/smc-webapp/editor_latex.coffee
```diff
@@ -93,7 +93,10 @@ class exports.LatexEditor extends editor.FileEditor
             @spell_check()
 
         @latex_editor.syncdoc.on 'connect', () =>
-            @preview.zoom_width = @load_conf().zoom_width
+            if @load_conf().zoom_width?
+                @preview.zoom_width = @load_conf().zoom_width
+            else
+                @preview.zoom_width = 100
```
haraldschilly: what about `@preview.zoom_width = @load_conf().zoom_width ? 100` ?

DrXyzzy: testing now, thanks for coffeescript lesson

Issue #2005, opened 2017-05-26
API call `create_account` succeeds, but returns gateway timeout


DrXyzzy:
1. Issue API call to create an account.
    ```
    curl -u sk_....: -d first_name=John -d last_name=Doe \
      -d email_address=jd@test... -d password=....... -d agreed_to_terms=true \
      https://cocalc.com/api/v1/create_account
    ```

1. Response is an html page with the following info:
> Error 504 Ray ID: ... • ... UTC
> **Gateway time-out**
> You
> **Browser**
> Working
> Chicago
> **Cloudflare**
> Working
> cocalc.com
> **Host**
> Error
> **What happened?**
> The web server reported a gateway time-out error.
> 
> **What can I do?**
> Please try again in a few minutes.

In each case, the account is created, but the response from the API (including the new account id) is not received. It is possible to log into the new account. Trying the same command again elicits an error of `This e-mail address is already taken`. Other API calls such as `query` complete normally.

The error code of 504 is reported in the body of the response as well as in the HTTP/1.1 status code.

williamstein:
time: 45 minutes

PR #2011, opened 2017-05-29
Issue/2005

DrXyzzy:
Ref: #2005 

This is PR is only a start at a fix. It adds two mocha tests in src/smc-hub/test/api/account.coffee for `get_usernames` and two for `create_account`. The two `create_account` tests fail without the changes in this PR and pass when the changes are applied.

But something is missing, because the API call is still not succeeding with CC-in-CC, i.e., this curl call is still failing with 504 gateway timeout:
```
curl -u sk_...: -d first_name=... -d last_name=... -d email_address=... -d password=... \
  -d agreed_to_terms=true https://cocalc.com/api/v1/create_account
```

I'm submitting the PR in case it helps solve the other part of the bug.

Something else to consider: do we want to use [validator](https://www.npmjs.com/package/validator) to check for valid email address and uuid formats during testing?


williamstein:
> Something else to consider: do we want to use validator to check for valid email address and uuid formats during testing?

The more validation the better!

DrXyzzy:
Added test file, account.coffee, is now in a separate PR #2014.

williamstein:
closing this...

PR #2013, opened 2017-05-29
Hub stats opened files

haraldschilly:
This adds some stats about files on the landing page. Most importantly, it requires a schema update!

There is also some compactification of the code for those stats tasks in the hub; and in case the query looks scary, it isn't so bad as it looks.

Test:
* calling `/stats` should still contain the old entries and of course, doesn't cause exceptions in the hub.
* The landing page should shows more data in the stats table.

### DIFF
--- a/src/smc-hub/hub_http_server.coffee
+++ b/src/smc-hub/hub_http_server.coffee
@@ -268,10 +268,12 @@ exports.init_express_http_server = (opts) ->
             return
         opts.database.get_stats
             cb : (err, stats) ->
+                res.header('Cache-Control', 'private, no-cache, no-store, must-revalidate')
                 if err
                     res.status(500).send("internal error: #{err}")
                 else
-                    res.json(stats)
+                    res.header("Content-Type", "application/json")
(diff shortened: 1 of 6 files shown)


src/smc-hub/hub_http_server.coffee
```diff
@@ -268,10 +268,12 @@ exports.init_express_http_server = (opts) ->
             return
         opts.database.get_stats
             cb : (err, stats) ->
+                res.header('Cache-Control', 'private, no-cache, no-store, must-revalidate')
                 if err
                     res.status(500).send("internal error: #{err}")
                 else
-                    res.json(stats)
```
williamstein: Harald, why is
```
res.header("Content-Type", "application/json")
res.send(JSON.stringify(stats, null, 1))
```
better than `res.json(stats)`?   Is it required because you set the headers above?  Genuinely curious.

haraldschilly: It's just for introducing line breaks in the output. I'm really bad at reading json in a single line, that's all :eyes:

Issue #2028, opened 2017-05-31
make https://embed.cocalc.com/ default to full screen mode


williamstein:

haraldschilly:
Besides the PR, I thought about this, and I still don't understand why there is a secondary subdomain. I am kind of against hardcoding any domain handling. I would rather suggest that "embedding" CoCalc means to point to `/app?fullscreen`.

williamstein:
>  I still don't understand why there is a secondary subdomain.

I'm happy to eliminate it!

williamstein:
Closing this in favor of @haraldschilly's much better ?fullscreen approach.

PR #2034, opened 2017-06-01
fullscreen query parameter

haraldschilly:
Ref #2028 -- this introduces a query parameter `?fullscreen` to toggle it on.

process: mainly finding my way around how various redirects work and how `smc_target` is used, etc.

test and edge cases:
* `/app?fullscreen`
* `/projects/ ... /files/dir/name.ipynb?fullscreen` → opens up that ipynb file in fullscreen (and does not stupidly add the `?...` part to the filename)
* all normal paths still work, i.e. `/projects`, `/`, `/projects/.../files/...`, etc. (e.g. that when opening a file, `*.ipynb` isn't mangled into `*.ipyn` or so)

deployment: don't miss that it also touches the hub

time: 1 hour 10 min

### DIFF
--- a/src/smc-hub/hub_http_server.coffee
+++ b/src/smc-hub/hub_http_server.coffee
@@ -259,7 +259,9 @@ exports.init_express_http_server = (opts) ->
 
     # Save other paths in # part of URL then redirect to the single page app.
     router.get ['/projects*', '/help*', '/settings*'], (req, res) ->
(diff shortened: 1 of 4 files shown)


williamstein:
time: 18 minutes

process: read code, nodejs manual, test, email people who care.

williamstein:
and complain about the lack of unit testing!

Issue #1988, opened 2017-05-23
unit test the api


williamstein:
Relevant code goes here:

https://github.com/sagemathinc/cocalc/tree/master/src/smc-hub/test/api

To run, you need postgres setup (e.g., source dev/project/postgres-env) and do
```
cd src/smc-hub
mocha test/api
```

williamstein:
@haraldschilly did this!

PR #2040, opened 2017-06-03
Issue1987d

DrXyzzy:
Ref: #1987 and #1988.

Doc and tests for API calls for private and public text files.
- write_text_file_to_project
- read_text_file_from_project
- public_get_directory_listing
- public_get_text_file

Note that the test for `write_text_file_to_project` can take up to 5 seconds.


williamstein:
It's incredible it is even possible to automatically test using projects like this!  We'll have to  increase some of the timeouts to avoid spurious fails, and deal with the too-much logging.  But this is fantastic.

Issue #2048, opened 2017-06-05
test sagews jupyter bridge kernels without using sage_server


DrXyzzy:
Per suggestion of @williamstein. Some testing of sage server could be done in-process without the client/server overhead. It could be faster & more robust, similar to api mocha testing at `src/smc-hub/test/api`.

DrXyzzy:
Done. 
```
~/cocalc/src/smc_sagews/smc_sagews/test_direct$ python -m pytest
...
==== 14 passed in 47.42 seconds ====
```

PR #2050, opened 2017-06-06
test some or all jupyter kernels

DrXyzzy:
Ref: #2048 

To test:
```
cd cocalc/src/smc_sagews/smc_sagews/test_without_server

# test jupyter_client launch of all 15 non-sage kernels
python -m pytest

# test selected kernels
python -m pytest --kname=anaconda3,singular
```

@haraldschilly, please have a look at naming for test reports in `conftest.py` and `README.md`. You might want to change file names and `sagews-jclient` labels in the reports.


DrXyzzy:
Note: at present, `apache_toree_scala` kernel consistently fails to start. It is the only one.

src/smc_sagews/smc_sagews/test_without_server/conftest.py
```diff
@@ -0,0 +1,88 @@
...
+    """default is to test all non-sage kernels listed with `jupyter kernelspec list`"""
+    option_value = metafunc.config.option.kname
+    if 'kname' in metafunc.fixturenames and option_value is not None:
+        knames = option_value.split(',')
+        metafunc.parametrize("kname", knames)
+    # nsk = list of available non-sage kernel names
+    # skip first line of command output, "Available kernels"
+    else:
+        v = map(str.strip, os.popen("jupyter kernelspec list").readlines())
+        nsk = [x.split()[0] for x in v[1:] if not x.startswith('sage')]
```
haraldschilly: This `v[1:]` isn't python 3 ready, is it? change the `map( ... )` to comprehension like `[ x.strip() for x in ... ]`

```
$ python3
>>> x = [1,2,3]
>>> v = map(lambda z : z+22, x)
>>> v
<map object at 0x7f8bdd94f0b8>
>>> v[1:]
  File "<stdin>", line 1, in <module>
TypeError: 'map' object is not subscriptable
```

haraldschilly:
ok, we should get rid of this failing one, and only introduce it once we know it works well.

besides the naming, I'm wondering what exactly is tested. just that the kernels start?  Maybe we should check something more meaningful, too?

src/smc_sagews/smc_sagews/test_without_server/README.md
```diff
@@ -0,0 +1,57 @@
...
+python -m pytest
+
+# test selected kernels
+python -m pytest --kname=anaconda3,singular
+
+# look for failed tests
+grep fail ~/jclient-test-report.prom
+
+# output:
+# sagews_jclient_test{name="compute[apache_toree_scala]",outcome="failed"} 16.109400034 1496753157000
```
haraldschilly: just a silly detail, but here I don't know where this `compute[...]` is coming from.

haraldschilly:
well, we should get rid of this failing kernel and only introduce it once we know it works well. will happen with the new compute image anyways.

besides the naming, I'm wondering what exactly is tested. just that the kernels start?  Maybe we should check something more meaningful, too? i.e. `feature = <start, help, autocomplete, ... >` ? So, this is more like a general question about where this will be heading at, merging this is in any case fine.

src/smc_sagews/smc_sagews/test_without_server/conftest.py
DrXyzzy: map changed to list comp, thanks

DrXyzzy: also test_without_server changed to test_direct

DrXyzzy:
Merged, as advised by @haraldschilly.

PR #2052, opened 2017-06-06
Fix some visual edge cases in courses

johnjeng:
## Remove headers when there are no results to sort
**Before**
no display at all before

**After**
no display at all change

## Change margin for show/hide deleted button when there are no displayed items
**Before**
no display before

**After**
no display change

## Bonus
- Also refactors part of the settings panel.

# Test
- Delete all student projects works


williamstein:
time: 7 minutes

PR #2059, opened 2017-06-07
Api Doc

haraldschilly:
Very ugly, but all basics are in place and it's fully data driven.

cc-api-doc


DrXyzzy:
@haraldschilly I am finished pushing markdown edits to this PR branch

PR #2065, opened 2017-06-09
Persist Markdown Inputs

### LINKED ISSUE #1546: peer-grading: editing grading guidelines does not save state
When editing the peer-grading guidelines for a particular assignment, navigating to another file and then back to the course file seems to have the same effect as clicking "cancel", which has caused me to lose progress on several occasions (when I was trying to inspect another file while composing the guidelines). What would be better is if upon return, the editor returned to the same state as it was left in (neither saved nor cancelled), at least as long as nothing drastic has happened.

johnjeng:
- Fixes #1546.
- Moves save/cancel buttons to the bottom of the markdown input
- Allows MarkdownInputs to be "attached" to stores instead of having to write the saving code in stores/actions for every possible location. 

TODO:
- [x] More Comprehensive Testing
- [x] Refactor MarkdownInput into its own file (especially actions/store).

## Test
- Toggle edit mode, change the value, hit cancel, then hit edit. The edit value should be what the display value was.
- Toggle edit mode, navigate away, come back, editing should remain.
- Toggle edit mode, change the value, close the file, reopen the file, edit state should be lost
- Open multiple MarkdownInputs, each one should maintain their state independently of each other.

### DIFF
--- a/src/smc-webapp/course/assignments_panel.cjsx
+++ b/src/smc-webapp/course/assignments_panel.cjsx
@@ -233,6 +233,7 @@ Assignment = rclass
             <Col xs=10>
                 <MarkdownInput
                     persist_id    = {@props.assignment.get('path') + @props.assignment.get('assignment_id')}
+                    attach_to     = {@props.name}
                     rows          = 6
                     placeholder   = 'Private notes about this assignment (not visible to students)'
                     default_value = {@props.assignment.get('note')}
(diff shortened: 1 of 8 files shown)


williamstein:
Nice!

williamstein:
My comment is +1.  

Also, I don't see a reason not to merge this now, and have a follow-up for refactoring, etc.

PR #2075, opened 2017-06-09
Clean up smc-react Stores/Actions abstraction a little

johnjeng:
Stores should clean up after themselves
Neither needs to know they are within a `ReduxApp`
Actions should not have any state or listeners to destroy

# Test
- Search codebase for any relevant uses of `destroy` that I may have missed
- Confirm that MarkdownInputs still work.


williamstein:
I'm completely reverting this code -- i.e., another commit that undoes what it did.    It caused numerous tracebacks in user code running in the browser, as @hsy noticed automatically.   You really just can't make changes to the basic REDUX level stuff (stores/actions) without first implementing a proper test suite.

Issue #2077, opened 2017-06-10
no left margin in rendering of public .md files


DrXyzzy:
Note from @williamstein:

> Relevant file: https://github.com/sagemathinc/cocalc/blob/master/src/smc-webapp/public/editor_md.cjsx

haraldschilly:
I'm closing this, since it is merged

PR #2078, opened 2017-06-10
add left and right margins for public view of .md files

DrXyzzy:
See #2077 

Hard to test public view of a project from cc-in-cc due to cookies (needed to access the project, but don't want for public view). I tested the change using live update of html using browser tools page editor (Inspect with Chrome and Firefox).

Before the change:
no-margins

After:
margins

### DIFF
--- a/src/smc-webapp/public/editor_md.cjsx
+++ b/src/smc-webapp/public/editor_md.cjsx
@@ -45,8 +45,12 @@ PublicMarkdown = rclass ({name}) ->
         else if not @props.content?
             <Loading />
         else
-            <div className="webapp-editor-static-html-content" style="margin-left:2em;margin-right:2em">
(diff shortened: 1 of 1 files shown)


haraldschilly:
I suggest to keep `editor-static-html-content` as it is, and add this to the `style` property of the `Markdown` component.

DrXyzzy:
Thank you for feb56cb. Looks like a better solution. I'm closing this PR & suggest not committing my version of the change.

DrXyzzy:
Wait, I see your @haraldschilly revision was committed to the PR branch. I agree with merging the second commit.

haraldschilly:
screenshot from 2017-06-10 11-35-20

DrXyzzy:
+1! thank you

PR #2081, opened 2017-06-10
Enable persistence on all markdownInput widgets in course

### LINKED ISSUE #1546: peer-grading: editing grading guidelines does not save state

johnjeng:
Actually apply the code required to fix #1546...

Also save the open state of the peer grading configuration dialogue


williamstein:
time: 9 minutes

Issue #1610, opened 2017-01-29
multilingual support for TeX docs (utf8 encoding)


dimpase:
For people who write TeX files in some unusual encoding (e.g. `latin2`, or cyrillic encodings could be
even harder to deal with)---I guess on Windows or some other systems that support such
encodings natively---and then try them on SMC TeX it might be tricky to realise that they are supposed to do
everything in UTF-8, cf. e.g.[ this thread](https://groups.google.com/d/msg/sage-support/lPk6QUyrDb4/vwRdviUcBgAJ) - they start thinking that their specific language support is lacking etc. It ought to be documented that they should do `\usepackage[utf8]{inputenc}`.

haraldschilly:
There could be a fourth tab on the right hand side of the latex editor titled "help". that could be a short intro and pointing to some links ... some pages maintained by smc, some others are like that wikibook latex guide. What do you think about this idea?

And regrading language support: one of the harder parts are special characters in code listings and so on. For that, one has to switch using xelatex. I don't really know it, but in some circumstances there is no way around it.

dimpase:
there are particularly bad cases where only xelatex, or only pdflatex, work (Hebrew support comes to mind).
But the thing in the linked thread is OK with pdflatex.

slel:
I think this is fixed by the (January to March 2017) commits to the file
https://github.com/sagemathinc/cocalc/commits/master/src/smc_pyutil/smc_pyutil/templates/linux/default.tex

PR #2083, opened 2017-06-12
webapp/latex: adding help tab

haraldschilly:
@DrXyzzy this is about documentation and fixing my language. It also addresses #1610 a little bit, at least I think so.

### DIFF
--- a/src/smc-webapp/_editor.sass
+++ b/src/smc-webapp/_editor.sass
@@ -132,6 +132,11 @@
       &.btn-danger
         color: white
 
+.webapp-editor-latex-help
+  padding            : 20px
+  background-color   : white
+  overflow-y         : auto !important
+
 .smc-latex-inline-error
   padding            : 10px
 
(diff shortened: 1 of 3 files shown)


DrXyzzy:
@haraldschilly I made some minor edits and pushed to this PR branch. Additional comments:

- maybe missing link at line 154 about inserting image: `Follow [these instructions]...`
- is the plan to put this under a button called `Help` at upper right? If so, that might make support conversations confusing because there is already a `Help` button in a higher level toolbar at upper right. Looks as if they will have different icons, but might also want different button label.

williamstein:
> is the plan to put this under a button called Help at upper right?

No since the vast majority of our users (99.99% ?) would never want to see this...

williamstein:
There's a link right now to the API docs from the "get an api key" part of account settings...

haraldschilly:
> is the plan to put this under a button called Help at upper right?

This patch already integrates it into the UI, in the latex editor. It looks similar, but once you see it, it's clear because of the context where it is placed. 

You're also right about the missing link.

haraldschilly:
@DrXyzzy I've fixed the link and merged with master, so, IMHO this is ready to go in. Maybe you want to add one or another paragraph with useful information? I don't even know if we have latex related information in our wiki. So, one remaining task could be to either copy over some of the important bits and pieces from the wiki or to link to some of them.

Otherwise, this is something we can do later, too.

PR #2084, opened 2017-06-12
Hub Prometheus Metrics

haraldschilly:
This changes the metrics of the hub to produce prometheus status updates. It's refactored based on a very old branch, so there are hopefully no breaks. This is also just the first step to get this into the hub. Once we actually collect them and see what's missing, this will be refined. That's also why most of the old code is commented and will be gone once it's clear we don't need it.

The only questionable change I made is in `postgres-base` → `_query` → commenting `delete opts.table`. The idea is to record the query time and use the table name as a label. It can't be recorded when it is deleted...

Test: the `/metrics` endpoint shows prometheus stats and the hub itself doesn't crash...


williamstein:
time: 50 minutes

Issue #2076, opened 2017-06-10
Generate intro part of API HTML doc from message.coffee


DrXyzzy:
API endpoint documentation is generated from [message.coffee](https://github.com/sagemathinc/cocalc/blob/master/src/smc-util/message.coffee) and viewable [here](https://cocalc.com/doc/api.html).

1. Expand the introductory doc on the API that is found at the top of message.coffee. Format it with markdown and wrap in appropriate code for the second step:
1. Expand code that generates api.html to process the introductory markdown and add to the current `What's this API about?` placeholder text.

See also: #1987

DrXyzzy:
@haraldschilly, I figure we can pair up on this the way we did with the individual API message2 docs. If you agree with this approach, what coffeescript code do you want me to use to wrap the Introduction markdown?

haraldschilly:
Yes, we should do this together. My approach would be to have everything in that `exports.documentation` object. Problem is, right now the entries are directly the events. To avoid name clashes, we should add another layer to make this clearer:

```
exports.documentation =
    events  = {}
    intro = {}
    ...  # something else?
```

and then in the `message2` function: `exports.documentation.events[obj.event]  = ... `

Accordingly, the only other change is in [api.pug](https://github.com/sagemathinc/cocalc/blob/master/src/webapp-lib/doc/api.pug).

haraldschilly:
I've pushed this to a new branch, see c370ccb

haraldschilly:
Closing this, since I have the impression all of this is done. The intro section is already much longer than initially, too.

PR #2087, opened 2017-06-13
expanded API intro doc section

DrXyzzy:
See #2076 #1987.


DrXyzzy:
Passes all mocha api tests `72 passing (16s)`.

Issue #2111, opened 2017-06-14
SMC --> CoCalc rename


williamstein:
There is still a message "Your connection is unstable or SMC is not available".  Make that "SMC" be "CoCalc".

PR #2112, opened 2017-06-15
webapp: various smc → cocalc renamings and licensing code

haraldschilly:
issue #2111

process: replacing text and comments

time: 15 min

### DIFF
--- a/src/smc-webapp/billing.cjsx
+++ b/src/smc-webapp/billing.cjsx
@@ -1209,9 +1209,9 @@ FAQS =
            prioritize supporting paying users.
            We very strongly encourage you to make an account and explore our product for free!
            There is no difference in functionality between the free and for-pay versions of
-           CoCalc; everything is still private by default for free users, and you can
+           <SiteName/>; everything is still private by default for free users, and you can
            make as many projects as you want.  You can even fully start teaching a course
-           in SMC completely for free, then upgrade at any point later so that your students
+           in <SiteName/> completely for free, then upgrade at any point later so that your students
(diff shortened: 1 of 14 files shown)


DrXyzzy:
first pass review, looks as if the 5 instances of boilerplate:
```
#    This program is distributed in the hope that it will be useful,
#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
#    GNU General Public License for more details.
```
are missing a line. Correct text is visible at [src/misc.py](https://github.com/sagemathinc/cocalc/blob/master/src/misc.py):
```
#    This program is distributed in the hope that it will be useful,
#    but WITHOUT ANY WARRANTY; without even the implied warranty of
#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
#    GNU General Public License for more details.
```

haraldschilly:
Oh, well, don't bother with the licensing text. They are probably GPL2+ artefacts. At some point I'll use some off-the-shelf script or come up with something on my own to fix all of them.

haraldschilly:
and actually, both are wrong. See https://www.gnu.org/licenses/agpl-3.0.en.html at the bottom.

We deliberately have a mix of licenses, so it's not always clear what's going on. Maybe we should start a dedicated licensing issue and classify our files first.

williamstein:
Irregardless of what FSF says, we should just have a single one-line statement in every file, which explicitly refers to another file, which has the canonical correct statement.  Doing anything else is madness.

PR #2115, opened 2017-06-15
Latex refactor

haraldschilly:
Motivated by today's hide and seek game, this reactors the latex editor and consolidates of what it consists of. I haven't tested it thoroughly, so this is mainly to inform everyone about this.


haraldschilly:
Ok, @DrXyzzy I think this is also a nice one for you to review. Looks complicated, but it's actually just about moving some content to different files. Whatever and whenever more work will happen with latex editor, this saves a lot of hairs from being pulled out.

In essence, I've only added a few more words to the help tab page and enhanced remembering the preview position by also taking scrolling from the left into account.

testing: the main task is to check if there are any stacktraces caused by undefined references. I.e. does sagetex work, forward/inverse search, are errors shown as line markers, etc. I've tested all this and also made sure nothing is broken for .rnw files, but it doesn't hurt to have a second opinion. So, when you run it, open the javascript console ...

DrXyzzy:
Looks good. I found a couple unusual things, reported below, but I suspect both are out of scope for this PR.

Test results:
- ✓ remember Preview scroll position in .tex file after clicking away to other browser tab and to other CoCalc tab
- ✓ remember PDF scroll position in .tex file after clicking away to other browser tab
- ✗ remember PDF scroll position in .tex file after clicking away to other CoCalc tab - not restored
- ✓ render .Rnw checks ok, test file http://kbroman.org/knitr_knutshell/assets/knitr_example.Rnw
- ✓ render .Rmd checks ok, test file https://gist.githubusercontent.com/jeromyanglim/2716336/raw/574783e141a99bd315c49647a3b308bd7de64cf3/example-r-markdown.rmd
- ✓ sagetex test file https://raw.githubusercontent.com/sagemath/sagetex/master/example.tex
- ✓ multiple files checks ok, test folder at https://github.com/sagemathinc/cocalc-example-files/tree/master/latex/multiple-files

Note: contents for multiple-file example appears as below, but it appears this way without the PR, too:
- 1 Introduction
- 2 Setup
- 1 Makefile
- 2 Subfiles Package
- 3 KOMA-Script
- 4 SageTex
- 5 Conclusion

DrXyzzy:
One more thing. The .pdf view repositions at top of documents spontaneously. First time is 1-2 seconds after clicking `PDF`. You see small view, then it resizes to width of the pane, then re-renders a second or two later. Repositioning of PDF happens again approximately every 3 minutes.

Issue #1226, opened 2016-11-14
In the host down message, don't show payment info for member projects


haraldschilly:
when running a member project, don't show message for paying a member project.

> Host down
> The host for this project is down, [...] please become a paying customer (US $7/month) [...]
> "timeout"

williamstein:
Still true?

DrXyzzy:
Looks as if message still includes text about becoming a paying member: https://github.com/sagemathinc/smc/blob/master/src/smc-webapp/project_files.cjsx#L2102

haraldschilly:
This is somehow similar to https://github.com/sagemathinc/smc/issues/1628 ... not the same issue, but the same spot in the code

haraldschilly:
I pushed to https://github.com/sagemathinc/smc/tree/issue/1226 but that's not done (untested, and I'm not sure what exactly should happen in all the cases). also, this should  ideally  also fix #1628 in one go.

PR #2120, opened 2017-06-17
webapp/projects: do not show payment info if quotas are not even known

haraldschilly:
ProjectsStore::get_total_project_quotas could return undefined and `not quotas?` is hence true, even if `member_host` isn't even known.

see issue #1226, this is a second attempt


haraldschilly:
giving myself a positive review. yay...

Issue #2103, opened 2017-06-14
chat comments disappear if file is renamed


zeigerpuppy:
To replicate:
1. create a file
2. start a chat and add comment
3. close document
4. rename document and reopen
5. the chat will have disappeared
6. renaming the file back to the original restores the chat.

This may be quite confusing to users if comments disappear.  I guess that the chat loading must be coupled to the filename somehow.  Perhaps when a file is created, a unique identifier could be created for a document and the chat linked to that identifier even if the file is moved or renamed.
I guess this would also require some form of persistent metadata at the file level.

Alternatively it may make sense to have a hidden file storing the chat data (eg .filename.chat) which is automatically renamed on a GUI rename function.  This would have the benefit of being relatively transparent when doing git operations too.

haraldschilly:
In "files", when you click the eye to see hidden files, too, there is a chat file for any name `foo`, it's `.foo.sage-chat`. So, I think it's reasonable that when you rename the file in the UI with the rename button, this chat file should also be renamed.

johnjeng:
Closing in favor of #2119 since renaming is fixed.

PR #2127, opened 2017-06-19
Rename/move Side Chat files with the source files.

johnjeng:
Fixes a portion of #2103 
Also fixes `setState(...)` warnings of Side chat

# Testing
Things with chats should have their chats be moved/renamed with them

Be aware of #2119.
Do each of the following with with hidden files and visible files
Do each of the following with files in the home directory and a subdirectory
Do the same from a subfolder to the home folder
1. Test moving 1 file without a chat into subfolder
2. Test moving 2 files without chats into a subfolder
3. Test moving 1 file with a chat into a subfolder
4. Test moving 2 files with chats in to a subfolder
5. Test moving 2 files into a subfolder where one has a chat and the other does not.
6. Test renaming a file without a chat
7. Test renaming a file that has a chat

Additionally, trying to copy associated chat files with their files manually should not throw an error

### DIFF
--- a/src/smc-util/misc.coffee
+++ b/src/smc-util/misc.coffee
@@ -499,14 +499,17 @@ exports.path_split = (path) ->
     v = path.split('/')
     return {head:v.slice(0,-1).join('/'), tail:v[v.length-1]}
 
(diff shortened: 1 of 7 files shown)


williamstein:
BRAVO!!!!!

Issue #2067, opened 2017-06-09
API message `jupyter_port` results in error


DrXyzzy:
API message `jupyter_port` is listed in [src/smc-util/message.coffee](https://github.com/sagemathinc/cocalc/blob/master/src/smc-util/message.coffee#L791), but attempting to use it results in an error:
```
curl -u sk_... -d project_id=... https://cocalc.com/api/v1/jupyter_port
 ==>
{"error":"unknown message event type 'jupyter_port'"}
```

Is it ok to remove this API message definition from message.coffee?

See also: #2057

DrXyzzy:
Fixed by #2148 - remove jupyter_port from the API

PR #2148, opened 2017-06-23
doc & test API copy between projects

DrXyzzy:
Ref: #1987 and #1988.

API calls `copy_path_between_projects` and `copy_public_path_between_projects`.
```
~/cocalc/src/smc-hub$ mocha test/api
... 
  90 passing (32s)
  1 pending
```

### DIFF
--- a/src/smc-hub/test/api/copy_between_projects.coffee
+++ b/src/smc-hub/test/api/copy_between_projects.coffee
@@ -1,0 +1,157 @@
+###
+Testing API functions that copy from one project to another
+###
+
+api   = require('./apitest')
+{setup, teardown} = api
+misc = require('smc-util/misc')
+expect = require('expect')
+
+describe 'testing copy between projects -- ', ->
+    before(setup)
+    after(teardown)
+
+    project_id  = undefined
+    project_id2 = undefined
+    
+    path1 = 'doc1.txt'
+    path2 = 'B2/doc2.txt'
+    path3 = 'FOO/abc.txt'
+    path2dir = path2.split('/')[0]
+
+    content  = 'TEST CONTENT'
+    src_dir  = 'TEST_DIR'
+
+    it "creates source project", (done) ->
+        api.call
+            event : 'create_project'
+            body  :
+                title       : 'TEST1'
+                description : 'Source Project'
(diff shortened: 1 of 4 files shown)


haraldschilly:
In general, positive review, I also checked that the variables in the test are fine and really check what they say. However, I'm wondering if the second case for copying a subdirectory shouldn't actually use a directory as path argument? I.e. in that test there is `path2`, but at the top you define `path2dir = path2.split('/')[0]`  and never use it.

haraldschilly:
and I have my own little issues running the actual tests. it says `error: unknown option '--compilers'`

haraldschilly:
Above is also a tiny commit to reorganize the npm run test command in smc-hub.

DrXyzzy:
Thanks for review. `pathdir` is dead code and can be deleted.
I'm not sure what happens if `target_path` ends with a slash. I can test it.

haraldschilly:
Just to clarify, I think we should test copying directories here. Hence I think all it needs is to change `path2` to `path2dir` in that test case?

williamstein:
Ping -- this seems forgotton.  @DrXyzzy

DrXyzzy:
Not forgotten. I'm working on it now. Took a detour to go after process/mem leaks from mocha test runs.

DrXyzzy:
@haraldschilly, test cases added for 'file to directory' and 'directory to directory'

DrXyzzy:
I just added the 4-character change in message.coffee to remove `jupyter_port` from the API.
Ref: #2067

DrXyzzy:
> [@haraldschilly] it says error: unknown option '--compilers'

I am getting the `unknown option '--compilers'` error now suddenly too, and I don't see anything in git diffs to explain it. Did we switch versions of node or npm? Anyway, workarounds are either
- use one of the npm rules, e.g. `npm run testapi`, or
- add explicit path to opts file, e.g.
```
~/cocalc/src/smc-hub$ mocha --opts test/mocha.opts test/api
```

haraldschilly:
yes, it's weird, `--compilers` appears in `smc-hub/test/mocha.opts` ... but why does it work when you pass it as `--opts` ? it must be related to a newer version of mocha.

Issue #2150, opened 2017-06-26
refreshing a page on a given URL often doesn’t open that URL


williamstein:
I’ve frequented noticed the last few days that if you visit a URL in CoCalc, then refresh the page, you end up in the project, not the file.  

This is either caused by @haraldschilly‘s ?fullscreen, my ?auth, or some change @johnjeng made to the store or actions recently.   My instinct says it’s @johnjeng, but I’m not sure.

Replicate:

1. Open a file in CoCalc.
2. Refresh your browser.

Often it comes up in the project, not the file.

If this is happening for published or ?auth url’s (like Minerva will use), then it will be a horrendous bug.

haraldschilly:
I'm working on this. So far, everything looks good, i.e. the path does end up in 

ProjectsActions.load_target → open_project is getting called with <project_id> files/x.md true

Therefore reading the path, dissecting it, looking for "projects" and a valid UUID, all works.

haraldschilly:
Ok, I know the problem, but no idea how to solve it.

In `ProjectStore::fetch_directory_listing`, there is a lock-barrier to avoid excessive directory fetching. Problem is, first, the directory is read without the callback, then, it is called with a callback but blocked right at the barrier. That callback is necessary in order to continue the waterfall in `ProjectStore:load_target`.

I'll attempt to make the lock itself a stack of callbacks and run them when it finishes. Altough:

time: 70min

haraldschilly:
Above is a commit with a lot of debug statements (necessary to see what's going on upon loading)

Also, there are one or two small changes. E.g. in `redux_account.coffee` I'm commenting out a line opening up the projects page. I didn't figure out why this is there, because there is clearly also another event listener to open that page when the event happens. So, that also needs to be thought through at some point ... (if it is necessary, there should be a comment why. at least I struggle to see it)

maarksman:
Fixed

johnjeng:
@maarksman Was this fixed by Harald or is there an unlinked commit/PR?

haraldschilly:
Yes, I fixed it. It's just that my feeling while fixing it -- with the changes above -- was that I made it work but I'm neither sure why exactly nor if this is the proper solution. I guess we can leave this as it is, though, and revisit this either by completely rethinking and rewriting how this works or when we discover a new problem. 

[...]

PR #2154, opened 2017-06-27
webapp: attempted fix for target paths not opening files properly

haraldschilly:
see issue #2150

This still contains debug lines, so please someone has to clean this up a bit afterwards. I've also no idea if my contraption is any good, or if there is a much better fix. I also didn't catch how and where the problem was actually introduced.

Open question: why is `redux_account.coffee`  opening up the projects list? this could interfere with other callbacks reacting to "sigend_in". I've commented the line, maybe erroneously ...


williamstein:
> Open question: why is redux_account.coffee opening up the projects list? this could interfere with other callbacks reacting to "sigend_in". I've commented the line, maybe erroneously ...

Removing it seems sensible to me now.

PR #2160, opened 2017-06-28
webapp: let copy operation write to clipboard (if allowed)

haraldschilly:
adding this line here, because it's quite odd to me to not write to the clipboard

### DIFF
--- a/src/smc-webapp/copy-paste-buffer.coffee
+++ b/src/smc-webapp/copy-paste-buffer.coffee
@@ -15,4 +15,7 @@ exports.get_buffer = ->
 
 exports.set_buffer = (s) ->
     buffer = s ? ''
+    try
+        # https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand
+        document.execCommand('copy')
     return


williamstein:
I only didn't do this, since I didn't know how.

williamstein:
I merged this since I don't think it is wrong to do.  However, is there any context in which it would actually work???

williamstein:
Of course this doesn't work....  That's why github has some funny flash button for copying the url of a git repo to the clipboard.  I think I've spent days in the past investigating this sort of thing...

haraldschilly:
For me this works. I can select some text (e.g. in codemirror, or some sagews output text) and press copy. Then, in another application like a text editor, I can paste it.

Github shows you a flash icon? That's interesting. I have a normal svg icon and I think it uses the selection api to copy it. AFAIK, this only works for such selectable text boxes, though.

Here is a library generalizing all this https://github.com/zenorocha/clipboard.js/

williamstein:
Maybe Macs are more locked down/secured for this particular thing (copy/paste).  In any case, if it works somewhere, I see no harm in including it.  Thanks!

– William

Issue #1153, opened 2016-10-28
(smc-in-smc dev only) stop all projects also stops the instructor's project


williamstein:

johnjeng:
This doesn't seem to happen on the live servers.

williamstein:
priority low since not on the live servers...

williamstein:
There's nothing really having to do with courses here.  It's just some bug in how closing projects work with smc-in-smc.  I've seen that with two projects open, closing one closes both, etc.   I think the code for closing/controlling projects for smc-in-smc is completely different than in live smc, and it's maybe more likely to be buggy.  Maybe there is some funny scoping error or something in some smc-hub code...

williamstein:
This is an issue with the forever command which daemonizes the project local_hub.  It's really a pain to fix without switching to something else entirely (not forever).  I did try to fix this recently for an hour.  Basically any time any project is stopped all of them are killed.

williamstein:
Was an issue with how we were using forever, which I recently finally fixed!

PR #2161, opened 2017-06-28
add mocha_clean.sh for after mocha api run

DrXyzzy:
Ref: #2162

sample use:
```
~/cocalc/src/smc-hub$ mocha test/api
...
90 passing (35s)
1 pending
~/cocalc/src/smc-hub$ mocha_clean.sh
process count 28 -> 8
memory use 966MB -> 393MB
```


DrXyzzy:
Compare #1153.

PR #2183, opened 2017-07-05
Copy menu update

maarksman:
Pull request for copy menu bug

### DIFF
--- a/src/smc-webapp/projects.cjsx
+++ b/src/smc-webapp/projects.cjsx
@@ -502,7 +502,7 @@ class ProjectsStore extends Store
                     return pay
 
     is_deleted: (project_id) =>
-        return !!@getIn(['project_map', project_id, 'deleted']) #getIn returns nested data form list pointers. e.g. 'project_map' points to project_id, which points to 'deleted. 
+        return !!@getIn(['project_map', project_id, 'deleted']) 
 
     is_hidden_from: (project_id, account_id) =>
         return !!@getIn(['project_map', project_id, 'users', account_id, 'hide'])
@@ -525,6 +525,7 @@ class ProjectsStore extends Store
             return 0
         others = []
         for i in v
+            # deleted projects have a map node " 'deleted': true' standard projects do not have this property
             if (not (i.get('deleted') == true )) and (show_hidden or not i.get('users').get(account_id).get('hide'))
                 others.push(id:i.get('project_id'), title:i.get('title'))
         list = list.concat others
--- a/src/smc-webapp/projects.cjsx
+++ b/src/smc-webapp/projects.cjsx
@@ -509,7 +509,6 @@ class ProjectsStore extends Store
 
     get_project_select_list: (current, show_hidden=true) =>
         map = @get('project_map')
-        console.log(map)
         if not map?
             return
         account_id = webapp_client.account_id
@@ -526,11 +525,8 @@ class ProjectsStore extends Store
             return 0
         others = []
         for i in v
-            console.log("(not (i.get('deleted') == 'true' ))")
-            console.log((not (i.get('deleted') == true )))
             if (not (i.get('deleted') == true )) and (show_hidden or not i.get('users').get(account_id).get('hide'))
                 others.push(id:i.get('project_id'), title:i.get('title'))
-                console.log("i.get('title') was just pushed to list")
         list = list.concat others
         return list
 
--- a/src/smc-webapp/projects.cjsx
+++ b/src/smc-webapp/projects.cjsx
@@ -513,7 +513,7 @@ class ProjectsStore extends Store
             return
         account_id = webapp_client.account_id
         list = []
-        if current? and map.has(current)   #current is for current project
+        if current? and map.has(current)   # current is for current project_id value
             list.push(id:current, title:map.get(current).get('title'))
             map = map.delete(current)
         v = map.toArray()
--- a/src/smc-webapp/projects.cjsx
+++ b/src/smc-webapp/projects.cjsx
@@ -502,18 +502,19 @@ class ProjectsStore extends Store
                     return pay
 
     is_deleted: (project_id) =>
-        return !!@getIn(['project_map', project_id, 'deleted'])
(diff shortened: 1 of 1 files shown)


src/smc-webapp/projects.cjsx
```diff
@@ -502,7 +502,7 @@ class ProjectsStore extends Store
                     return pay
 
     is_deleted: (project_id) =>
-        return !!@getIn(['project_map', project_id, 'deleted'])
+        return !!@getIn(['project_map', project_id, 'deleted']) #getIn returns nested data form list pointers. e.g. 'project_map' points to project_id, which points to 'deleted.
```
williamstein: In the comment, `‘deleted` has an opening quote but no quote.  Also, I find the comment hard to read.  Finally, it makes no sense to explain what getIn is in this *one* place.   It’s used in hundred of places; it’s better to not have a comment here, and instead have a comment in some dev guide...?

src/smc-webapp/projects.cjsx
```diff
@@ -513,7 +513,7 @@ class ProjectsStore extends Store
             return
         account_id = webapp_client.account_id
         list = []
-        if current? and map.has(current)
+        if current? and map.has(current)   #current is for current project
```
williamstein: Our style for comments is to put a space after the hash, so `# current is for current project`.  But might be to be more descriptive, e.g., `# current is the project_id of the current project`.

src/smc-webapp/projects.cjsx
```diff
@@ -525,7 +525,7 @@ class ProjectsStore extends Store
             return 0
         others = []
         for i in v
-            if not i.deleted and (show_hidden or not i.get('users').get(account_id).get('hide'))
+            if (not (i.get('deleted') == true )) and (show_hidden or not i.get('users').get(account_id).get('hide'))
```
williamstein: Change `not (i.get(‘deleted’) == true)` to `not i.get(‘deleted’)`.

src/smc-webapp/projects.cjsx
maarksman: Took the comment out since it was for personal use

src/smc-webapp/projects.cjsx
maarksman: Fixed

src/smc-webapp/projects.cjsx
maarksman: Any reason for the preferred syntax? I wasn't sure what that would return since projects which aren't deleted simply don't have a 'deleted' key/property in their map.  If 'deleted' isn't true this is avoided.

src/smc-webapp/projects.cjsx
```diff
@@ -502,7 +502,7 @@ class ProjectsStore extends Store
                     return pay
 
     is_deleted: (project_id) =>
-        return !!@getIn(['project_map', project_id, 'deleted'])
+        return !!@getIn(['project_map', project_id, 'deleted'])
```
williamstein: This part of the diff does not but add a blank space at the end of the line.  Please enable “Strip trailing whitespace: remove whenever file is saved” in CoCalc account settings, then open the file, then save it (with cursor not in live 505).  Then make another commit.

src/smc-webapp/projects.cjsx
```diff
@@ -525,7 +525,8 @@ class ProjectsStore extends Store
             return 0
         others = []
         for i in v
-            if not i.deleted and (show_hidden or not i.get('users').get(account_id).get('hide'))
+            # deleted projects have a map node " 'deleted': true' standard projects do not have this property
```
williamstein: This is still sloppy:

 - There is a single double quote just sitting there in the middle of the line.   Maybe it is supposed to match the single quote after true?   
 - Also, this seems to be two sentences “deleted projects have...” then another sentence “standard project do not...”, but there is not punctuation between them.  Use a period or semicolon.

src/smc-webapp/projects.cjsx
```diff
@@ -525,7 +525,8 @@ class ProjectsStore extends Store
...
+            if (not (i.get('deleted') == true )) and (show_hidden or not i.get('users').get(account_id).get('hide'))
```
williamstein: The only actual non-comment request I made was related to this line, but you seem not to have seen that.

williamstein:
> -            if not i.deleted and (show_hidden or not i.get('users').get(account_id).get('hide'))
> +            if (not (i.get('deleted') == true )) and (show_hidden or not i.get('users').get(account_id).get('hide'))
> Any reason for the preferred syntax? I wasn't sure what that would return since projects which aren't deleted simply don't have a 'deleted' key/property in their map. If 'deleted' isn't true this is avoided.
> 
Doing 

     if (not (thing == true))

 is completely equivalent in CoffeeScript (and most languages) to doing 

    if not thing

The latter is more readable.   When you’re unsure about what will happen in some case, it’s useful to do some tests or ask.   In this case, if deleted isn’t a key, then `i.get(‘deleted’)` is `undefined`.    Undefined is falsey for the purposes of if/not/etc., not undefined is true.

src/smc-webapp/projects.cjsx
```diff
@@ -525,7 +525,8 @@ class ProjectsStore extends Store
             return 0
         others = []
         for i in v
-            if not i.deleted and (show_hidden or not i.get('users').get(account_id).get('hide'))
+            # Deleted projects have a map node " 'deleted': true ". Standard projects do not have this property.
+            if (not i.get('deleted')) and (show_hidden or not i.get('users').get(account_id).get('hide'))
```
williamstein: So to summarize, the net relevant change is to replace `i.deleted` by `i.get(‘deleted’)`, which was needed because i is a list of immutable.js objects?

maarksman: Yes. The variable 'i' is a map element in a list and since 'i.deleted' didn't exist 'not i..deleted' was always true. The deleted property was working in other parts of the site so I figured it probably was not being called properly here.

Issue #2184, opened 2017-07-06
curl API log_client_error results in 504: Gateway time-out


DrXyzzy:
Try the following:
```
curl -u sk_xxx: -d error=foo_error https://cocalc.com/api/v1/log_client_error
```

After about 7 seconds, there is an HTML response with title
> cocalc.com | 504: Gateway time-out

I have looked into the problem a bit and notice the following which may be relevant:
1. In src/smc-hub/client.coffee, function `mesg_log_client_error` does not use a callback for error status when it writes to the database, nor does it call e.g. `@success_to_client(id:mesg.id)`.
1. In src/smc-util/message.coffee, the definition of message `log_client_error` does not include the usual optional `id` parameter for message id.

Note that a successful mocha test of a call to `log_client_error` is possible - see #2185 - including read-back of the logged error, but it does not use a callback on the logging step.

williamstein:
This is intentional at the level of the websocket message API.  It’s a waste of bandwidth to send back an acknowledgement message to the user – log_client_error is just fire and forget.  The user doesn’t even know it happened.

That said, naturally the http API, if it also has log_client_error, it should respond with a confirmation that the message was received.

And all that said, I’m fine with changing the websocket api in this case to be a little wasteful and send back an ack, just to keep everything clean and uniform and consistent.

DrXyzzy:
Verified fixed with #2185

Issue #2136, opened 2017-06-20
fix API unit test for inviting non-cloud users to project


DrXyzzy:
Per comments at `it.skip "invites non-cloud collaborators"` in src/smc-hub/test/api/account.coffee:

> \# The test fails because api.last_email isn't set until after this test runs.
> \# See smc-hub/client.coffee around L1220 where cb() is called before
> \# email is sent.

DrXyzzy:
Verified fixed by #2185.

PR #2185, opened 2017-07-06
log_client_error test without `done` on logging step

DrXyzzy:
Ref: #1988

~~__WARNING__: I'm requesting a review of this test in relation to #2184, but this PR should probably not be merged until that issue is resolved.~~

Please see my last comment below. The code looks good, but I was not able to complete testing using cc-in-cc (details in the comments). If @haraldschilly can do the two suggested tests in a normal build, it should be good for a merge.

### DIFF
--- a/src/smc-hub/test/api/log_client_error.coffee
+++ b/src/smc-hub/test/api/log_client_error.coffee
@@ -1,0 +1,37 @@
+###
+API test of log_client_error
+###
+
+api   = require('./apitest')
+{setup, teardown} = api
+
+misc = require('smc-util/misc')
+
+expect = require('expect')
+
+
+describe 'log_client_error', ->
+    before(setup)
+    after(teardown)
+
+    logged_event = 'error'
+    logged_error = 'API error FOO'
+
+    it "logs an error", ->
+        api.call
+            event : 'log_client_error'
+            body  :
+                error   : logged_error
+            cb    : ->
+    it 'gets error log using database', (done) ->
+        api.db.get_client_error_log
+            event : logged_event
+            cb    : (err, log) ->
+                expect(log.length).toBe(1)
+                expect(log[0]).toEqual
+                    event:logged_event
+                    error:logged_error
+                    account_id:api.account_id
+                    id:log[0].id
+                    time:log[0].time
+                done(err)
--- a/src/smc-hub/test/api/log_client_error.coffee
+++ b/src/smc-hub/test/api/log_client_error.coffee
@@ -17,12 +17,15 @@ describe 'log_client_error', ->
     logged_event = 'error'
     logged_error = 'API error FOO'
 
-    it "logs an error", ->
+    it "logs an error", (done) ->
         api.call
             event : 'log_client_error'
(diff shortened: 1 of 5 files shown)


williamstein:
LGTM

DrXyzzy:
I updated message definition, client code, and unit test to add a callback to eliminate the gateway timeout. This change is probably good for production. It tests ok with mocha, but I have not been able to test with curl. See below.

```
~/cocalc/src/smc-hub$ mocha test/api/log_client_error.coffee
...
  2 passing (356ms)
```

Questions:
- What is the `cc-in-cc` setup for testing changes to src/smc-hub/client.coffee via `curl`? Is start_apy.py necessary?
- What is the base URL to use for API tests in this situation?

Depending on which variations I use for the URL and security key, I have seen the following:
- runs production code for src/smc-hub/client.coffee, not the code under test
- curl: (35) gnutls_handshake() failed: The TLS connection was non-properly terminated.
- user must be signed in before accessing projects
- user ... does not have write access to project ...

williamstein:
Bravo, exactly the right move!  LGTM.

DrXyzzy:
added workaround for #2136 in src/smc-hub/test/api/account.coffee to get rid of 1 pending test on api run

DrXyzzy:
Added doc for view/set project upgrade. Added a timeout.

`99 passing (48s)`

DrXyzzy:
Tests ok with mocha, but the following should be tested in a normal build, not cc-in-cc, before release:
- normal client error processing such as hub disconnect
- curl for log_client_error, e.g.
```
curl -u sk_...: -d id=...  -d "error=client error test" https://cocalc.com/api/v1/log_client_error
```

DrXyzzy:
also relates to #1987

haraldschilly:
Uhm, I think we should merge this. In particular, I don't see how any of these changes could have "dangerous" side effects.

I also don't understand your questions. E.g. `start_apy.py`, you mean `start_api.py`? I don't know what it does,  and besides that I don't think we have any such setup (regardless of what you really mean). When there is something to test in the hub that changes what endpoints do, I open them directly in the browser and see what happens (e.g. for the `/stats` endpoint).

src/smc-hub/test/api/log_client_error.coffee
```diff
@@ -0,0 +1,40 @@
...
+        api.db.get_client_error_log
+            event : logged_event
+            cb    : (err, log) ->
+                expect(log.length).toBe(1)
+                expect(log[0]).toEqual
+                    event:logged_event
+                    error:logged_error
+                    account_id:api.account_id
+                    id:log[0].id
+                    time:log[0].time
```
haraldschilly: testing `log[0].time` to equal `log[0].time` is not so exciting, but I understand that you just want to see if everything is there. it's ok ;-)

williamstein:
start_api.py is in dev/project; it’s for running a separate webserver that listens on localhost, which you can use directly from within the project.  It’s not necessary, but can be useful for some level of testing/dev.

PR #2188, opened 2017-07-07
Refactor chat actions

johnjeng:
Initial refactor of chat actions without touching the rendering components.
Contains a README with some in-progress notes.

## Test
- Open and close a standalone chat
- Send messages, edit messages, and click all the buttons
- Open and close a side chat
- Open and close side chat's file with and without the side chat open


src/smc-webapp/chat/REAME.md
```diff
@@ -0,0 +1,48 @@
...
+## Overview
+CoCalc has several chat view considerations.
+- Side chat associated with files
+- Primary chat rooms (also a file)
+- Mobile and Desktop versions of both
+
+The constricting factors are primarily keyboard related or screen size related.
+ie. You cannot use certain hotkeys without a physical keyboard and certain things don't fit well on a smaller screen.
+
+opening side chat on mobile should "slide over" the main document. Basically, don't try to split the screen.
```
williamstein: I wonder if we should start saying “mobile phone”, since laptops and iPads are mobile, but have very different constraints.   Don’t change this – just a comment.

Issue #2211, opened 2017-07-19
Make dead keys work in the online terminal


slel:
This is a follow-up to https://github.com/sagemathinc/cocalc/issues/2205.

French keyboard has some [dead keys](https://en.wikipedia.org/wiki/Dead_key), eg `~`, `^`, `¨`, `´`.
These allow to type accented characters, eg you would type `^` followed by `u` to get `û`.
To get the standalone "accent" without any letter underneath, you type it followed by space,
eg hitting `~` then spacebar gives you `~`.

Currently, in CoCalc's online terminal, dead keys are ignored, so using a French keyboard
there is no way to input `~`, besides copy-paste.

CoCalc's online terminal is currently based on [term.js](https://github.com/chjj/term.js/)
whose README file has this warning:

> **This project is no longer maintained.**
> **For a maintained fork take a look at [sourcelair/xterm.js](https://github.com/sourcelair/xterm.js).**

The website https://xtermjs.org points to https://www.sourcelair.com
where one can get a 30-day free trial. The terminal there lets one input
`~` in a French keyboard by typing `~` followed by spacebar, as expected.

So it might be worth considering replacing the use of term.js by that of xterm.js.

slel:
The workaround at https://github.com/sagemathinc/cocalc/issues/2205#issuecomment-316501988 would be nice to have in the meantime.

haraldschilly:
Are you sure this is even related to term.js and not the locale setup in linux?

slel:
I think so. See similar discussion, with distinction between accented characters
that can be typed with a single key vs accented characters involving a dead key:
https://github.com/jeremyramin/terminal-plus/issues/28.

williamstein:
> Are you sure this is even related to term.js and not the locale setup in linux?

Yes, definitely.  We tested on Gonzalo Torn\’{a}ria’s laptop in exactly the same shell with the same setup via ssh’ing versus the graphical UI and had things work via ssh, but not via term.js.   We used multiple browsers.

haraldschilly:
Ok, I also played around with this and can confirm it.

tornaria:
A possible strategy, that seems to work in firefox is to read the composed text from the hidden textarea itself, namely:
* After handling a keypress event (for a "normal" key), prevent the default handler so that this won't go into the hidden textarea
* When a dead-key combination is produced, there's no keypress event, hence the default handler won't be supressed and the resulting character will be inserted into the hidden textarea
[...]

PR #2219, opened 2017-07-21
term.js: fix handling of dead characters -- issue #2211

haraldschilly:
that's the patch from the ticket


williamstein:
Oops – I already pushed (live an hour ago) a version of this with a better comment.

williamstein:
He explicitly stated in a comment on the issue that he makes the code available under BSD or whatever, so  I’m OK with this.  We just don’t want AGPLv3 code that we **can’t** relicense.

PR #2220, opened 2017-07-21
Drag n drop chat images

johnjeng:
First version of dragging items onto the chat text area.

Images will embed as `` tags with width='60%'. This can be made smarter later.
Files will embed as `[file_name.ext](file_name.ext)`, making a link which opens as a new file tab.

Text is inserted where the cursor is at the time of dropping. 
If you have highlighted text, it will replace that text.

Items get uploaded to the current directory of the `project_store`. Right now this is always the directory of the open file (in this case, chat).

**Not** implemented for side chats yet.

# Testing
Needs testing on more platforms
- [ ] Firefox
  - [x] OSX
  - [x] Linux
  - [ ] Windows
- [ ] Safari
- [ ] Chrome
  - [x] OSX
  - [x] Linux
  - [ ] Windows

### DIFF
--- a/src/smc-webapp/smc-dropzone.cjsx
+++ b/src/smc-webapp/smc-dropzone.cjsx
@@ -92,14 +92,14 @@ exports.SMC_Dropwrapper = rclass
         config           : rtypes.object               # All supported dropzone.js config options
         event_handlers   : rtypes.object
         preview_template : rtypes.func                 # See http://www.dropzonejs.com/#layout
-        show_upload      : rtypes.bool
+        show_upload      : rtypes.bool                 # Whether or not to show upload area
         on_close         : rtypes.func
         disabled         : rtypes.bool
 
     getDefaultProps: ->
         config         : {}
-        hide_previewer : false
         disabled       : false
+        show_upload    : true
 
     getInitialState: ->
         files : []
@@ -182,7 +182,7 @@ exports.SMC_Dropwrapper = rclass
         @setState(files : [])
 
     render_preview: ->
-        if not @props.show_upload and @state.files.length == 0
+        if not @props.show_upload or @state.files.length == 0
             style = display : 'none'
         box_style =
             border       : '2px solid #ccc'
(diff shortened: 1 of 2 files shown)


williamstein:
You absolutely MUST add a way to add images to the chat that does not use drag and drop, as drag and drop basically doesn’t exist in iOS (it’s a new experimental feature only in iOS 11 beta, which hardly works).   Look at github comments – there is a message at the bottom: “Attach files by dragging & dropping or (underlined link)selecting them.”  We should do something very similar.

johnjeng:
>You absolutely MUST add a way to add images to the chat that does not use drag and drop

In this PR specifically?

williamstein:
On Sat, Jul 22, 2017 at 3:06 PM John Jeng <notifications@github.com> wrote:

> You absolutely MUST add a way to add images to the chat that does not use
> drag and drop
>
> In this PR specifically?
>

When I wrote that you would need to in order for me to provide any
feedback, since I only have my ipad on this trip.

For this PR, if you can wait until next week, then definitely do not need
that extra feature.  As is, is fine.

-- 
-- William Stein

williamstein:
I don't like this as is since it clutters the directory containing the chatroom with image files. There are two reasonable options: 
  1. Change this to upload the image to a subdirectory, e.g.,  `.pics/`. 
  2. Upload the images to our blob store.

2 is vastly better in pretty much every way; no clutter, and also everything will work if the chatroom is moved to another directory...  But you probably don't know how to do that, so I'll have to do it (in fact, I'm not even sure how to do it immediately -- would take some thought!). 

I'm sure you could easily do 1.  Can you please just do 1?

johnjeng:
Updated to do 1. Images are now uploaded to `.chat-images/`. You can still drag and drop a file and it will be put in the directory. The reasoning is that if you're sharing a file, you're more likely to want to open it from the file listing.

williamstein:
Your design makes very good sense -- I think it's better than either of my suggestions above!

PR #2227, opened 2017-07-26
Issue/1835b updates to sagews pytest

### LINKED ISSUE #1835: sage worksheet bug -- ending a block of code with ? shouldn't always submit introspection request

DrXyzzy:
Ref: #1835 

There are no changes to production code in this PR, only changes to sagews pytest.

- This does not fix #1835 but updates pytest to get ready for that fix.
- Also skips test for sagemath kernel in jupyter bridge, which we no longer support.

```
python -m pytest
...
137 passed, 6 skipped in 206.32 seconds
```


williamstein:
looks safe to me.

Issue #2236, opened 2017-07-28
sage_server (for sage worksheets) does not start under sage-8.0


DrXyzzy:

DrXyzzy:
With the workaround in PRxxxx, another bug related to upgrade to sage-8.0 is unmasked, that `import graphics` fails in sage_salvus.py, so that calls to `plot()` and related functions fail.

DrXyzzy:
Above `import graphics` failure, as well as jupyter bash kernel failing to start under sage_server, were fixed by moving aside the sage-8.0 python3 binary to
  `/projects/sage/sage-8.0/local/bin/python3.bkb`.

DrXyzzy:
fixed under #2238

PR #2238, opened 2017-07-31
Issue/2236 sage_server with sage-8.0

DrXyzzy:
Ref: #2236.

These changes, plus @hsy fix hiding sage binary for python3 so that sage_server uses /usr/bin/python3, allow sage worksheets to run with sage-8.0.

```
# testing with sage-8.0:
~/cocalc/src/smc_sagews/smc_sagews/tests
python -m pytest
...
== 137 passed, 5 skipped in 195.90 seconds ==

# testing with sage-7.6:
...
== 137 passed, 5 skipped in 207.19 seconds ==

# testing with sage-7.5:
== 137 passed, 5 skipped in 210.55 seconds ==
```

### DIFF
--- a/src/smc_sagews/smc_sagews/sage_server.py
+++ b/src/smc_sagews/smc_sagews/sage_server.py
@@ -1770,7 +1770,17 @@ def serve(port, host, extra_imports=False):
         log("imported sage.")
 
         # Monkey patch the html command.
+        try:
+            # need the following for sage_server to start with sage-8.0
+            # or `import sage.interacts.library` will fail
+            import sage.repl.user_globals
+            sage.repl.user_globals.set_globals(globals())
+            log("initialized user_globals")
+        except RuntimeError:
+            # may happen with sage version < 8.0
+            log("user_globals.set_globals failed, continuing",sys.exc_info())
         import sage.interacts.library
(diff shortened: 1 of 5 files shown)


haraldschilly:
Well, after getting it to run (installed smc_sagews locally in my project) I got one strange error:

`TestShowGraphs.test_issue594`:

```
[...]
        # 5 stdout BEFORE PLOT 1
        typ, mesg = sagews.recv()
>       assert typ == 'json'
E       assert 'blob' == 'json'
E         - blob
E         + json
```

haraldschilly:
Weird, running it again (just the test_graphics.py file), it works.

DrXyzzy:
Thanks for review and merge. I believe the error you saw with `TestShowGraphs.test_issue594` is because order of response messages from sage_server can vary. The less common variant seems to occur when the sage_server has been inactive for awhile. Will open a test bug for this error.

haraldschilly:
@DrXyzzy yes, after reading it, I believed this is the reason. Still odd, though. Maybe we have to come up with some stress test running a simpler variant of that hundreds of times in a loop to see how often it happens and then gather more data to see what the previous results were, etc.

Issue #2240, opened 2017-07-31
3d plot in sagews ignores frame=False option


DrXyzzy:
Try the following in a sage worksheet and notice that `frame=False` is ignored:
```
t,x,y,z = var('t,x,y,z')
helix = (cos(t), sin(t), t)
helix_plot = parametric_plot3d(helix, (t,-4*pi,4*pi), frame=False)
helix_plot
```

Frame does not appear in file created if `helix_plot.save("img.png")` is done and file is viewed.
See also heart and trefoil examples from [here](http://doc.sagemath.org/html/en/reference/plot3d/sage/plot/plot3d/parametric_plot3d.html).

DrXyzzy:
Looks as if the problem comes from a change in rendering routines in [sage_salvus.py](https://github.com/sagemathinc/cocalc/blob/master/src/smc_sagews/smc_sagews/sage_salvus.py#L2596):
```
                salvus.threed(obj, **kwds)
                # graphics.show_3d_plot_using_threejs(obj, **kwds)
```
The newer (uncommented) function, `salvus.threed`, does not pick up `frame` and other options from `obj._extra_kwds`. The commented-out function, `graphics.show_3d_plot_using_threejs` did this.

haraldschilly:
@DrXyzzy I've merged your patch and rolled it out.

PR #2242, opened 2017-08-01
process frame and other extra kwds in threed()

DrXyzzy:
Ref: #2240.

To test:
1. Install the patch (1 file changed, sage_salvus.py).
1. Try any 3d plot with `frame=False` and notice that a frame is not drawn, for example:
    ```
    sphere(frame=False)
    line3d([(0,0,0),(1,1,1)],frame=False)
    ```
1. Try other options that were ignored before this fix:
    ```
    icosahedron(frame={'color':'darkgreen','fontsize':20,'thickness':5})
    cube(viewer='tachyon')
    ```

### DIFF
--- a/src/smc_sagews/smc_sagews/sage_salvus.py
+++ b/src/smc_sagews/smc_sagews/sage_salvus.py
@@ -2590,6 +2590,14 @@ def show(*objs, **kwds):
         elif isinstance(obj, Animation):
             show_animation(obj, **kwds)
         elif isinstance(obj, Graphics3d):
+
+            # _extra_kwds processing follows the example of
+            # src/smc_sagews/smc_sagews/graphics.py:show_3d_plot_using_threejs()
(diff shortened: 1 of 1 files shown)


src/smc_sagews/smc_sagews/sage_salvus.py
```diff
@@ -2590,6 +2590,14 @@ def show0(obj, combine_all=False):
         elif isinstance(obj, Animation):
             show_animation(obj, **kwds)
         elif isinstance(obj, Graphics3d):
+
+            # _extra_kwds processing follows the example of
+            # src/smc_sagews/smc_sagews/graphics.py:show_3d_plot_using_threejs()
+            extra_kwds = {} if obj._extra_kwds is None else obj._extra_kwds
```
haraldschilly: Just FYI, you could also write `extra_kwds = obj._extra_kwds or {}`

haraldschilly:
tested it with a few examples, looks all good.

Issue #2243, opened 2017-08-02
add mocha tests for gap and pari/gp jupyter kernels


DrXyzzy:
See #160 and #1706.

Add tests to `src/smc-project/jupyter/test/supported-kernels.coffee`.

slel:
Is this solved by #2244 and ready to be closed?

haraldschilly:
no, because that PR was closed, not merged

PR #2244, opened 2017-08-02
add mocha tests for gap and pari kernels

DrXyzzy:
See #2243.

Needs work, but opening PR in case others have ideas. The test for `pari_jupyter` follows the pattern for other kernel mocha tests, but fails about 60% of the time. DEBUG output for failure cases shows kernel output ending when jupyter message channels are created.

I left the timeout for the pari test at 5 seconds because there is nothing to gain by increasing it. Either the test succeeds, usually in under 3 seconds, or it times out indefinitely. Here's sample failure output:

```
~/cocalc/src/smc-project/jupyter$ DEBUG=1 mocha --opts test/mocha.opts --grep "pari kernel" test/supported-kernels.coffee
DEBUG = true
 
 
  test the pari kernel --
Client.jupyter.Kernel('pari_jupyter',path='').constructor:
Client.jupyter.Kernel('pari_jupyter',path='')._process_execute_code_queue:  state='off'
Client.jupyter.Kernel('pari_jupyter',path='')._process_execute_code_queue:  queue has 1 items; ensure kernel running
Client.jupyter.Kernel('pari_jupyter',path='').spawn:  spawning kernel...
Client.jupyter.Kernel('pari_jupyter',path='').spawn:  spawned kernel; now creating comm channels...
[kernel_info_request message on iopub channel should happen next but does not appear]
```

### DIFF
--- a/src/smc-project/jupyter/test/supported-kernels.coffee
+++ b/src/smc-project/jupyter/test/supported-kernels.coffee
@@ -213,6 +213,43 @@ describe 'test the scala kernel --', ->
     it 'closes the kernel', ->
         kernel.close()
 
+describe 'test the gap kernel --', ->
+    @timeout(10000)
+    kernel = undefined
+    it 'evaluates 3^74', (done) ->
+        kernel = common.kernel('gap')
+        kernel.execute_code
+            code : '3^74'
+            all  : true
+            cb   : (err, v) ->
+                if err
+                    done(err)
+                else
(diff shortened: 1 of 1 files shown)


williamstein:
That is very, very worrisome.  I could change the backend to retry to spawn things a couple of times before just hanging there, or try to look more deeply into what is wrong.  It's very good having tests uncover such problems!

Issue #2234, opened 2017-07-28
new mode like ?fullscreen=true called maybe ?embed=true or ?fullscreenlock=true


williamstein:
- would remove fullscreen toggle
- clicking anything that would open a new tab/project, just opens that in the usual way in a new tab with no fullscreen at all

williamstein:
REQUESTED BY: MINERVA

haraldschilly:
Ok, just to clarify, the main goal is to disable the fullscreen toggle? (and other means of getting rid of fullscreen?)

That's usually called "kiosk" mode, i.e. where you can browse on a page in a web-browser, but you can't switch to another website. I think that's the idea. Hence `?kiosk` might be appropriate.

Besides that, we could consider to remove other UI elements too (my first thought is the project setting, or even all project tabs) ... as follow up tickets.

wikipedia: https://en.wikipedia.org/wiki/Kiosk_software

williamstein:
yes, EXACTLY -- kiosk mode.

haraldschilly:
the referenced implementation realizes this via `?fullscreen=kiosk` (in addition to the usual `?fullscreen`)

williamstein:
Live in a moment!

PR #2251, opened 2017-08-04
webapp: introduce ?fullscreen=kiosk mode, as a no-way-back to normal fullscreen state

haraldschilly:
see issue #2234

Test: kiosk-mode is triggered in two ways
* append `?fullscreen=kiosk` to the URL
* shift-click on fullscreen
* no regressions: i.e. appending `?fullscreen` still shows the toggle button


johnjeng:
Since the url usage is `?fullscreen=kiosk`, I think `fullscreen` in the code should be a string. If it's `kiosk`, then do one thing, if it's `default`, do the normal.

Issue #2259, opened 2017-08-06
rewrite our scripts in smc_pyutil to not use the standard Python command line scripts approach in setup.py because IT SUCKS


williamstein:
This PISSES ME OFF about Python!  We made all those smc- and cc- scripts by "doing things right" in setup.py? Well there is an intensly stupid nearly 1s overhead just to run any of them!  This slows down, e.g., smc-ls, which gets directory listings. Try this and be pissed;
```
time echo "from pkg_resources import load_entry_point" | /usr/bin/python
```
Note that pkg_resources is one of the lines in the scripts that are generated.

Just directly running, e.g., our smc-ls python script, i.e., not doing things right, is 10x faster!

I hope someone can just rewrite that part of our setup to generate our own scripts that don't import load_entry_point -- or just rewrite it, as it is pretty trivial, probably...

haraldschilly:
that's the bug: https://github.com/pypa/setuptools/issues/510

there is a possible solution: https://github.com/ninjaaron/fast-entry_points

haraldschilly:
This fastentrypoints works, here is my timing from withing my dev project. I'll test-deploy on compute0 now.

```
⋊> ~/s/s/smc_pyutil on issue/2259 ⨯ which cc-ls
/projects/14eed217-2d3c-4975-a381-b69edcb40e0e/.local/bin/cc-ls

⋊> ~/s/s/smc_pyutil on issue/2259 ⨯ time cc-ls > /dev/null
0.03user 0.00system 0:00.03elapsed 92%CPU (0avgtext+0avgdata 11388maxresident)k
0inputs+0outputs (0major+2985minor)pagefaults 0swaps

vs.

⋊> ~/s/s/smc_pyutil on issue/2259 ⨯ time /usr/local/bin/cc-ls > /dev/null
0.37user 0.04system 0:00.42elapsed 99%CPU (0avgtext+0avgdata 19476maxresident)k
0inputs+0outputs (0major+7092minor)pagefaults 0swaps
```

haraldschilly:
wanna guess which ones has it deployed?

```
ansible compute -m shell -a 'time cc-ls > /dev/null'                       
compute0-us | SUCCESS | rc=0 >>
0.02user 0.00system 0:00.03elapsed 84%CPU (0avgtext+0avgdata 11240maxresident)k
0inputs+0outputs (0major+3082minor)pagefaults 0swaps

compute1-us | SUCCESS | rc=0 >>
[...]

PR #2260, opened 2017-08-06
smc_pyutil: use fastentrypoints.py to speed up starting commands

haraldschilly:
see issue #2259

I should add that I haven't tested all the commands, but my compute0 test project seems to work well and I don't see a reason why there should be some failing while others work.


haraldschilly:
well, I've even tested rmd2html and no issues. I think I should just go ahead and merge+deploy it.

Issue #2254, opened 2017-08-04
kiosk hardening -- too easy to break out for Jupyter


williamstein:
Followup to https://github.com/sagemathinc/cocalc/issues/2234

1. Open a jupyter notebook url and put `?fullscreen=kiosk` at the end.
2. Go into the file menu and click on some of the options (e.g., New), and sort of break out...

Since embedded Juypter is the motivation for kiosk in the first place, it would be good to make some changes to how the File menu is rendered in kiosk mode.

There may be other editors with issues... e.g., "export sage worksheet to jupyter" opens a jupyter notebook after exporting.

This is NOT a security issue and doesn't have to be perfect.  It's just about making the UX good.

williamstein:
Lower priority since no user actually seems to care yet.

benepo:
This functionality would be helpful for our K-12 classrooms use case where students are running through a pre-prepared notebook and we want to limit ways in which they could break out of it or get distracted or click the wrong thing and take up educator time fixing it.

haraldschilly:
@benepo I just saw your comment right now. There is already a kiosk mode, this ticket is just collecting ideas how to harden it.

Related to notebooks: in case you talk about jupyter, we support edit and/or delete protected cells in our own jupyter implementation. This helps with limiting random damage that can be done :-)

williamstein:
Interestingly, this is no longer broken, i.e., clicking new stays in kiosk mode (and silently does nothing).   Yeah.

PR #2262, opened 2017-08-08
webapp: prevent files from being opened in kiosk mode -- except for the initial load of the url target

haraldschilly:
part of issue #2254

this is hardening one aspect of kiosk mode. in particular, the action to open a file. to test the effect, create a markdown file with the content:

```
open [new.ipynb](./new.ipynb)
```

usually, clicking the link in the preview opens it. with this patch, you get an error.

process: most of the code deals with still being able to open the target url (introducing `ignore_kiosk`)


williamstein:
conflicts...

williamstein:
I'm scared to merge this as is.  The first thing I tried was:
1. Open a Jupyter notebook,
2. Click File --> Open.

The result is seeing a file listing.  However, anything I click on shows the RED error saying you can't open files -- which is making a database entry, and really not something that should happen in normal usage.  Also, there is no way to even open the ipynb file I was just looking at (short of digging up the URL), so this is somehow worse than nothing at all.

williamstein:
Basically, this is all good, but I think it needs to be more finished before we make it live.

haraldschilly:
The real effect of this PR is best seen by opening the files listing for a project, i.e. `.../port/.../projects/...uuid.../files/?fullscreen=kiosk` and then clicking has no effect. Most of the code is about ignoring that for the initial URL, such that opening a file still works.

Unrelated, during resolving the merge conflict I changed a plain `window.open` to use `misc_page.open_new_tab`.

src/smc-webapp/project_store.coffee
```diff
@@ -357,14 +357,21 @@ class ProjectActions extends Actions
...
+        if (not opts.ignore_kiosk) and (redux.getStore('page').get('fullscreen') == 'kiosk')
+            return
+
         if opts.new_browser_window
             # options other than path don't do anything yet.
             url  = (window.app_base_url ? '') + @_url_in_project('files/' + opts.path)
             url += '?session=' + misc.uuid().slice(0,8)
             url += '&fullscreen=default'
-            window.open(url, '_blank', 'height=640,width=800')
+            misc_page.open_new_tab(url, true)
```
williamstein: Revert this or improve open_new_tab to have more options.  The new_browser_windo thing is used for implementing the shift+click on a tab to create a popout window.   Read the code of `misc_page.open_new_tab(url, true)` -- it's different than ` window.open(url, '_blank', 'height=640,width=800')`, in that it doesn't have the height/width params, which I think make the window appear as a new pop-up rather than new tab.

williamstein:
> Unrelated, during resolving the merge conflict I changed a plain window.open to use misc_page.open_new_tab.

That breaks it though...  Maybe there should be a comment explaining why -- sorry.

haraldschilly:
Hmm, I have trouble following your thoughts. I still pushed a change to make these settings fully customizable. Here are the two variants:

> before: `window.open(url, '_blank', 'height=640,width=800')`

vs

> after: `window.open(url, '_blank', 'menubar=yes,toolbar=no,resizable=yes,scrollbars=yes,height=640,width=800')`

see [this line in the master branch](https://github.com/sagemathinc/cocalc/blob/9e8377ae67a79a4e63aa96fb0b3ef6537f10ed5d/src/smc-webapp/misc_page.coffee#L1844)

The actually weird part is that neither variant does open up a window with a width of 800px for me. I'm investigating this now, maybe it's simply ignored by chrome...

williamstein:
Looks good!

PR #2263, opened 2017-08-08
webapp: ignore query strings like "?session=local" in internal links

haraldschilly:
This is a small detail for queries like session=local in the URL. When someone copies that link into a file for internal linking, the `?session=local` part should be ignored. 

Test: external links should still work as they are. Example, create a test.md file:

```
open [new.ipynb](new.ipynb?session=local)

open [external](https://www.google.at/search?q=cocalc)
```

There is still a design question open, namely if this logic should be in the `ProjectsActions::load_target` function instead.

### DIFF
--- a/src/smc-webapp/process-links.coffee
+++ b/src/smc-webapp/process-links.coffee
@@ -5,7 +5,15 @@ when we fully switch to react.
 
 misc = require('smc-util/misc')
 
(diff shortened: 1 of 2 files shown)


haraldschilly:
Oh, and additionally, being in fullscreen mode, and opening a link with `?fullscreen=default` or not, keeps fullscreen enabled...

PR #2264, opened 2017-08-08
webapp: fix newline wrapping of upload button in files listing

haraldschilly:
Hi @johnjeng ... thank you for reviewing the last one, but your little change reverted a problem I'm usually seeing in my files listing. The upload button wraps into the next line, hence the nowrap style.

This fix reverts it, and to show you what I mean, here are before/after screenshots:

files-whitespace-button-wrapping

vs.

fix-button-wrapping-afterwards

### DIFF
--- a/src/smc-webapp/project_files.cjsx
+++ b/src/smc-webapp/project_files.cjsx
@@ -1974,7 +1974,8 @@ exports.ProjectFiles = rclass ({name}) ->
             actions      = {@props.actions} />
 
     render_new_file : ->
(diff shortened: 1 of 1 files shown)


johnjeng:
But nowrap makes this possible:
overlap

johnjeng:
Opened #2266

haraldschilly:
ohh! well. bummer. so, all this looks like they're just bandaids and we
have to rethink how we should restructure this horizontal line.

PR #2265, opened 2017-08-08
webapp: prefix local session with app base url

haraldschilly:
there is another detail I just saw while testing: the local session info is shared with the main site and hence there are interferences when doing cc-in-cc dev. this patch prefixes they key with the app base url. it doesn't change the already existing key.


williamstein:
Oh wow, good point -- I totally forgot to do this....

Issue #2266, opened 2017-08-08
Buttons in File Listing Sometimes Overlap


johnjeng:
Like so:
![overlap](https://user-images.githubusercontent.com/618575/29081996-109326d0-7c19-11e7-875a-ae6cf3ce4bbd.png)

Take note of #2264 when fixing

haraldschilly:
I'm closing this. Exactly this doesn't happen any more and other problematic details should have their own issue.

PR #2270, opened 2017-08-09
webapp: files action button row, switching path to breadcrumb and some fine tuning

haraldschilly:
here is another approach to clean this up. related to #2266 and the button objections ...

unfortunately, the more I look at this, the more problems I see. Regarding this patch, I've tested it in chrome and firefox, at different window sizes and with and without zoom (also, using myself as a second user to see the active-heads area). It's mostly ok, and #2266 didn't happen for me, but I think we have to redo this using flexbox css and rethink this a bit.


williamstein:
WOW, this is soooo good!

PR #2272, opened 2017-08-10
Space warning

johnjeng:
Shows up when disk space used is within 5mb of the quota limit.
- No `x` to remove the warning. Delete files, increase quota, or suffer the warning.
- Disappears automatically when quota is adjusted
annoying

# Testing
I tested on smc-in-smc by using admin edit to reduce a project quota to under 5mb since "usage" is always undefined/zero.

### DIFF
--- a/src/smc-webapp/project_page.cjsx
+++ b/src/smc-webapp/project_page.cjsx
@@ -271,10 +271,10 @@ DiskSpaceWarning = rclass ({name}) ->
         if not quotas?.disk_quota? or not project_status?
             return null
         else
-            disk = project_status.get('disk_MB')
+            disk = project_status.get('disk_MB') ? 0
             if disk?
                 disk = Math.ceil(disk)
-        if not disk or quotas.disk_quota - 5 > disk
+        if quotas.disk_quota - 5 > disk
             return null
 
         styles =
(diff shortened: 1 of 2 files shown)


src/smc-webapp/project_page.cjsx
```diff
@@ -247,6 +247,52 @@ FreeProjectWarning = rclass ({name}) ->
...
+    propTypes :
+        project_id : rtypes.string
+
+    shouldComponentUpdate: (nextProps) ->
+        return @props.project_map?.get(@props.project_id) != nextProps.project_map?.get(nextProps.project_id)
+
+    render: ->
+        if not require('./customize').commercial
+            return null
+        if @props.free_warning_closed
```
williamstein: This line is wrong.  It won't cause trouble, but should just be deleted.

src/smc-webapp/project_page.cjsx
```diff
@@ -247,6 +247,52 @@ FreeProjectWarning = rclass ({name}) ->
...
+            return null
+        quotas = @props.get_total_project_quotas(@props.project_id)
+        project_status = @props.project_map?.get(@props.project_id)?.get('status')
+        if not quotas?.disk_quota? or not project_status?
+            return null
+        else
+            disk = project_status.get('disk_MB') ? 0
+            if disk?
```
williamstein: This line makes absolutely no sense at all, since disk is DEFINITELY defined, based on the previous line.

williamstein:
I'm going to merge this and fix the two tiny problems myself.

PR #2274, opened 2017-08-10
Project files action bar flex layout

haraldschilly:
... yet another approach, again, works for me in chrome and FF, this time I haven't seen any of the corner cases with overlapping elements. still, no idea what safari is capable of ;-)


williamstein:
It looks better, but the dropdown button the left now breaks and wraps...
img_0077

williamstein:
I'll merge; you can further polish in a followup.

PR #2275, opened 2017-08-10
webapp/files: no newline wrap for create-dropdown button when narrow

haraldschilly:
I got angry with that stupid line break.

process: less is more (getting rid of a div) for more control and the wrapping itself can only be dealt with dynamically reassigning flex layout to the element. I think under the hood some element-juggling-magic happens ... 

also, removing an erroneously set bsSize argument.

tested in Chrome and FF, no impact on the other parts of this horizontal row.

ugly-but-ok

### DIFF
--- a/src/smc-webapp/_projects.sass
+++ b/src/smc-webapp/_projects.sass
@@ -58,3 +58,7 @@
 
 .highlight
   background-color : yellow
+
+.cc-project-files-create-dropdown
+  > .dropdown.btn-group
+    display: flex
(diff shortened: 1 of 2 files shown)


johnjeng:
I added a commit because it's rather busted on mobile. I think it works acceptably now. Someone sanity check this because the amount of times we've patched this in the last 48 hours feels absolutely ridiculous.

johnjeng:
At half screen (a common use case for students), 1080p, it grows like this.
half

grow

longer

wrap

haraldschilly:
good idea about mobile. I'll test this here, if it also wraps on my desktop. (if it does, we can make this row-wrapping responsive).

I've also seen another tiny issue, I'll push it here, too...

PR #2276, opened 2017-08-10
project/jupyter: update nteract dependencies

haraldschilly:
in particular to get the fix for https://github.com/nteract/kernelspecs/issues/28

... I haven't tested it, this is more or less just a reminder that we need to update this.

### DIFF
--- a/src/smc-project/jupyter/package.json
+++ b/src/smc-project/jupyter/package.json
@@ -12,10 +12,10 @@
   "author": "",
   "license": "AGPL-3.0",
   "dependencies": {
-    "enchannel-zmq-backend": "^2.6.1",
-    "expect": "^1.20.2",
+    "enchannel-zmq-backend": "^4",
+    "expect": "^1.20",
     "kernelspecs": "^2",
-    "node-cleanup": "^2.1.2",
+    "node-cleanup": "^2.1",
     "spawnteract": "^4"
   }
 }
--- a/src/smc-project/jupyter/package.json
+++ b/src/smc-project/jupyter/package.json
@@ -14,8 +14,8 @@
   "dependencies": {
     "enchannel-zmq-backend": "^2.6.1",
     "expect": "^1.20.2",
-    "kernelspecs": "^1.1.0",
+    "kernelspecs": "^2",
     "node-cleanup": "^2.1.2",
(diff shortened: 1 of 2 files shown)


haraldschilly:
not good, running the tests with the updated libs gives 4 errors:

```
1) compute 2/3 using the python3 kernel --  evaluate 2/3:
     Error: Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
  
 
  2) get input using the python3 kernel --  do it:
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
  
 
  3) test the non-sage R kernel -- evaluate sd(c(1,2,5)):
     Error: Timeout of 20000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
  
 
  4) test the sage R kernel -- evaluate sd(c(1,2,5)):
     Error: Timeout of 20000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

src/smc-project/jupyter/package.json
```diff
@@ -14,8 +14,8 @@
   "dependencies": {
     "enchannel-zmq-backend": "^2.6.1",
```
rgbkrk: You'll want to bring this up to `"enchannel-zmq-backend": "^4.0.0",`

rgbkrk: (I just published this too).

We're about to have this setup where these dependent projects are together in one monorepo, though still separate node modules.

haraldschilly: ok, thank you, and a consistent monorepo sounds good (to eliminate the combinatorial explosion)

haraldschilly:
I updated `enchannel-zmq-backend` as well, increased some timeouts, and now only one test fails: `test the sage R kernel -- evaluate sd(c(1,2,5))`

not sure what's going on, but since all the other (similar) tests pass, it look ok.

also, going back to the old version, 12 tests fail...

williamstein:
That R test used to fail a lot...

PR #2278, opened 2017-08-11
Changed download behavior for multiple files

bryantpq:
Now compresses the files for the user and downloads the resulting zip file, rather than showing the notice


src/smc-webapp/project_files.cjsx
```diff
@@ -1525,44 +1525,66 @@ ProjectFilesActionBox = rclass
...
             log : true
         @props.actions.set_file_action()
 
-    render_download_link: (single_item) ->
+    download_multiple_click: ->
+        destination = ReactDOM.findDOMNode(@refs.download_archive).value
+        @props.actions.zip_files
+            src  : @props.checked_files.toArray()
+            dest : misc.path_to_file(@props.current_path, destination)
+        @props.actions.download_file
```
haraldschilly: I'm a bit surprised that these two commands in a row work. Aren't they asynchronous? So, imagine the zip-operation takes a few seconds. The download will continue to start immediately but fails with a file not found error. The unfortunate situation is, in these props-actions (`smc-webapp/project_store.coffee`) the callback is already some default file-listing & cleanup operation.

bryantpq: It worked for files I tested with, but they were really small so maybe that's why it worked fine - I will change it just to be safe. 
> the callback is already some default file-listing & cleanup operation
What do you mean by this?

haraldschilly: > What do you mean by this?

look at `zip_files` in `webapp_client.exec`. not only is there no callback, but the one used in `webapp_client.exec` is set to `@_finish_exec(id)`. So, to me this looks like `@_finish_exec(id)` needs a second argument to be able to trigger an additional callback after the finish-exec callback. :arrow_right:  :stuck_out_tongue_winking_eye: 

(If this makes no sense to you, don't worry -- and there are maybe better ideas, too...)

williamstein: Yes, it seems like this code will result in silently corrupted downloads...

PR #2280, opened 2017-08-11
dev/docker: some updates to the dockerfile

haraldschilly:
I made some updates to the dockerfile

* python3-yaml
* locales-all (hence I think we do not have to generate any locale, at the bottom)
* some reformatting

tested: nope, I just edited it


williamstein:
Looks good; I'll merge, then do a test build and push the image if it works.

Looking over the code, I think we should switch to getting postgres from apt, since our Ubuntu is new enough..

williamstein:
This did work perfectly.  Now pushing build to dockerhub...

Issue #2279, opened 2017-08-11
jupyter classic: wrong/old/deprecated startup parameter?


haraldschilly:
Everytime the classical notebook is started, the logfile says

```
Config option `base_kernel_url` not recognized by `NotebookApp`.  Did you mean one of: `base_project_url, base_url`?
```

Should we simply remove this unrecognized option? Code location is `smc_pyutil/smc_pyutil/jupyter_notebook.py` → the function `command`.

DrXyzzy:
Looking at src/smc_pyutil/smc_pyutil/jupyter_notebook.py, I see [the following](https://github.com/sagemathinc/cocalc/blob/master/src/smc_pyutil/smc_pyutil/jupyter_notebook.py#L63), but no definition for `command_line_setup()`. What am I missing?
```
if __name__ == '__main__':
    command_line_setup()
```

williamstein:
> What am I missing?

What you are missing is that jupyter_notebook.py is never run directly from the command line. It is only imported somewhere else.  See setup.py.  You can delete the "if name equals main" lines from the file.

DrXyzzy:
Looks ok to remove setting for `base_kernel_url`. See [ipython 2.0 backwards-incompatible-changes](http://ipython.readthedocs.io/en/stable/whatsnew/version2.0.html?highlight=base_kernel_url#backwards-incompatible-changes):
> Simplifying configurable URLs
> - base_project_url is renamed to base_url (base_project_url is kept as a deprecated alias, for now)
> - __base_kernel_url configurable is removed__ (use base_url)
> - websocket_url configurable is removed (use base_url)

PR to follow.

DrXyzzy:
It looks as if the jupyter_notebook.py `run` option is no longer supported. I suggest removing it.

1. `smc-jupyter run` fails with
    ```
    UnboundLocalError: local variable 'cmd' referenced before assignment
    ```
2. If I add code so that `cmd` is defined, then `smc-jupyter run` succeeds, but pointing browser at `https://cocalc.com/...project-id.../port/jupyter/` results in "Error 502 Bad gateway". Refreshing the browser kills the foreground jupyter, starts the daemon, and shows the classic notebook.

PR #2282, opened 2017-08-11
Issue/2279

DrXyzzy:
See #2279.


haraldschilly:
Well, I don't know about the implications removing "run". Looks like dead code? And well, the command line option is just gone, so it's ok to get rid of it.

PR #2284, opened 2017-08-11
longer sage_server startup timeout for kucalc & sage-8.0

DrXyzzy:
SAGE_SERVER_MAX_STARTUP_TIME_S from 30 → 60

Tested by adding 35 sec delay to `src/smc_sagews/smc_sagews/sage_server.py:run_server()` and verifying that worksheet startup timed out before the change and did not timeout after the change.


haraldschilly:
thank you, looks harmless, I'll merge it

PR #2294, opened 2017-08-16
Docker image: some improvements and cleanup

tornaria:
Several improvements and cleanup for the docker image:
 * avoid warnings in haproxy
 * use haproxy for the http->https redirect
 * set up acme-challenge to allow using an acme client to obtain certificates from letsencrypt
 * move some runtime files into /run, for consistency (intent is to put /run in a tmpfs)
 * in the Dockerfile, have the configuration files be last so changes don't spoil the docker cache
 * use "cocalc" for the image name in makefile, instead of "smc"

### DIFF
--- a/src/dev/docker/Dockerfile
+++ b/src/dev/docker/Dockerfile
@@ -130,13 +130,6 @@ RUN \
   && cp -rv /usr/local/sage/local/share/texmf/tex/latex/sagetex/ /usr/share/texmf/tex/latex/ \
   && texhash
 
-COPY login.defs /etc/login.defs
-COPY login /etc/defaults/login
-COPY nginx.conf /etc/nginx/sites-available/default
-COPY haproxy.conf /etc/haproxy/haproxy.cfg
-COPY run.py /root/run.py
-COPY bashrc /root/.bashrc
-
 RUN echo "umask 077" >> /etc/bash.bashrc
 
 # Install R Jupyter Kernel package into R itself (so R kernel works)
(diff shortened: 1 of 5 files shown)


src/dev/docker/Dockerfile
```diff
@@ -149,7 +149,7 @@ COPY kernels /usr/local/share/jupyter/kernels
 COPY kernels/ir/Rprofile.site /usr/local/sage/local/lib/R/etc/Rprofile.site
 
 # Build a UTF-8 locale, so that tmux works -- see https://unix.stackexchange.com/questions/277909/updated-my-arch-linux-server-and-now-i-get-tmux-need-utf-8-locale-lc-ctype-bu
-RUN apt -y install locales && echo "en_US.UTF-8 UTF-8" > /etc/locale.gen && locale-gen
+RUN echo "en_US.UTF-8 UTF-8" > /etc/locale.gen && locale-gen
```
haraldschilly: About the locales: I know there are many out there with other languages, e.g. you and me :-)

What about the idea to install `locales-all` instead? That package installs a set of pre-generated locales

tornaria: I think locales-all is already installed.

The issue is that locales have to be compiled (it's what locale-gen does) and that takes a lot of space, that's why it is done on a request basis.

Compiling `en_US.UTF-8` takes 1.68M, I'm not sure if other languages may take more space. There are almost 500 locales, so I guess at least 500-1000 M in total.

If you were to try, I would replace the above by something like
```
RUN sed -i 's/^#//' /etc/locale.gen && locale-gen
```
and see how much space that takes.

I think it's more useful to have it just like that and add a section to the documentation on how to compile more locales, I guess something like:
```
docker exec -it cocalc "echo 'es_UY.UTF-8 UTF-8'> /etc/locale.gen && locale-gen"
```
(UNTESTED)

tornaria: PS: the change you quoted above is only removing the 'apt -y install locales' because locales is already installed at the first step of the docker file.

Also, I think using "apt" is discouraged for scripts, should use apt-get instead which has a stable UI.

One could also considered changing the locale to something like "C.UTF-8" which feels more "neutral" although I guess it would be equivalent in terms of translations (but e.g. sorting and stuff is old-school, that's what 'ENV LC_ALL C.UTF-8' is doing).

At least installing and using a locale with the UTF-8 part means unicode characters work in the terminal. For me that's indispensable, while translations are important but a bit less so.

williamstein: > Also, I think using "apt" is discouraged for scripts, should use apt-get instead which has a stable UI.

Yes, it says that in the docs.

tornaria:
Closed because of https://github.com/tornaria/cocalc-docker/pull/1

PR #2295, opened 2017-08-16
Docker image: Install Julia 0.6 and the IJulia kernel

tornaria:
NOTE: this is intended to be a single commit on top of PR #2294, but I don't know how to make dependent pull requests.

The relevant commit is b037aa2, which in two steps:
 * installs Julia 0.6 from the official binary
 * installs IJulia kernel using julia package management

It may be desirable to preinstall other julia packages, that should be very easy copying what I did to install IJulia, something like:
```
# Install julia packages
RUN echo '\
ENV["JULIA_PKGDIR"] = "/opt/julia-903644385b/share/julia/site"; \
Pkg.init(); \
Pkg.add("package1"); \
Pkg.add("package2"); \
... 
Pkg.add("packageN");' | julia
```

### DIFF
--- a/src/dev/docker/Dockerfile
+++ b/src/dev/docker/Dockerfile
@@ -144,6 +144,22 @@ COPY kernels/ir/Rprofile.site /usr/local/sage/local/lib/R/etc/Rprofile.site
 # Build a UTF-8 locale, so that tmux works -- see https://unix.stackexchange.com/questions/277909/updated-my-arch-linux-server-and-now-i-get-tmux-need-utf-8-locale-lc-ctype-bu
 RUN echo "en_US.UTF-8 UTF-8" > /etc/locale.gen && locale-gen
 
+# Install Julia
+RUN cd /tmp \
+ && wget https://julialang-s3.julialang.org/bin/linux/x64/0.6/julia-0.6.0-linux-x86_64.tar.gz \
+ && echo 3a27ea78b06f46701dc4274820d9853789db205bce56afdc7147f7bd6fa83e41 julia-0.6.0-linux-x86_64.tar.gz | sha256sum -c \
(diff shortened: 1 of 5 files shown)


williamstein:
We are focused on a major new backend release right now, and aren't doing much else until that is done.

haraldschilly:
I'm wondering if we should promote this docker dev subdirectory to its own github repo, e.g. `cocalc-docker`.

williamstein:
Yes, that's a good idea -- there's very little coupling between it and cocalc itself.

tornaria:
https://github.com/tornaria/cocalc-docker

tornaria:
I created the above repo doing:
```
git filter-branch --prune-empty --subdirectory-filter src/dev/docker/ master
```
so it preserves the history of the subdirectory.

haraldschilly:
yes, looks good, but can we transfer your repository to us here? for that, you have to transfer ownership (in the settings) to me `haraldschilly` and then I transfer it here. The issue you created will move along the way.

williamstein:
We should close the two PR's here, and you should make them to the new repo, when it is setup.

williamstein:
And I like everything you did in this PR...

tornaria:
I've just transfered my new repo 'cocalc-docker' to haraldschilly, you can pick up from there.

Note that in that repo I made 3 PR of which I merged two into a branch "devel", but the branch "master" is pristine as it was produced by the filter-branch command.

The remaining PR is about having letsencrypt integration, it's working for me and I think it should work fine in general, but I haven't done extensive testing.

Let me know if you have any problem with the transfer (and if you prefer to rerun the above "filter-branch" command, that's fine with me, the way it works it will result in the exact same commits so the repos will be compatible, in that case I will redo all my PR in your repo).

tornaria:
Closed because of https://github.com/tornaria/cocalc-docker/pull/2

PR #2296, opened 2017-08-16
project/local_hub: idea for a simple health check

haraldschilly:
This adds two endpoints to the local hub raw server in a project.

* only if `IN_KUCALC` is true (so, when testing, I did comment out the check in the init function)
* health: returns 200 and OK, should be useful for k8s
* metrics: returns the prometheus stats, for now not much there, but that's for later.

How I tested it in cocalc is to query the endpoint from within a terminal, i.e.:

```
$ curl http://localhost:`cat ~/.smc/local_hub/raw.port`/metrics
kucalc_project_bugs_total 0

$ curl http://localhost:`cat ~/.smc/local_hub/raw.port`/health
OK
```

### DIFF
--- a/src/smc-project/kucalc.coffee
+++ b/src/smc-project/kucalc.coffee
@@ -114,4 +114,19 @@ exports.init_gce_firewall_test = (logger, interval_ms=60*1000) ->
     setInterval(test_firewall, interval_ms)
     return
 
+# called inside raw_server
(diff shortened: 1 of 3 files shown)


williamstein:
Instead of making a PR against master, can you just merge this into the kucalc-master branch and push it there?

haraldschilly:
ok

PR #2298, opened 2017-08-16
SSH UI

johnjeng:
Work so far on the UI.
Testing https://cocalc.com/92234d52-8a1c-4e63-bde3-f2727f5ab8b1/port/41480/settings/ssh-keys


src/smc-webapp/account_ssh_keys.cjsx
```diff
@@ -22,6 +22,13 @@ exports.SSHKeysPage = rclass
             grant access only to a project in the settings for that project.
         </div>
 
+    help: ->
+        <div>
+            To SSH into a project, use the following <span style={color:'#666'}>username@host:</span>
+            <pre>[projectIdWithoutDashes]@ssh.cocalc.com </pre>
+            A project's Id can be found in it's settings panel.
```
williamstein: "it's" means "it is", and "in it is settings panel" is wrong.

src/smc-webapp/project_settings.cjsx
```diff
@@ -618,7 +618,7 @@ ProjectControlPanel = rclass
                 <div>
                     SSH into your project: <span style={color:'#666'}>First add your public key to <a onClick={@open_authorized_keys} href=''>~/.ssh/authorized_keys</a>, then use the following username@host:</span>
                     {# WARNING: previous use of <FormControl> here completely breaks copy on Firefox.}
-                    <pre>{"#{misc.replace_all(project_id, '-', '')}@#{host}.sagemath.com"} </pre>
+                    <pre>{"#{misc.replace_all(project_id, '-', '')}@ssh.cocalc.com"} </pre>
```
williamstein: Change this back?  This code doesn't get run on kucalc, right, so it is exactly the code that should use the @{host} stuff.

williamstein: E.g., imagine if I merged this into master and make it live right now, where host is set?

williamstein:
I merged this into kucalc-master.

PR #2317, opened 2017-08-22
Upgrade project restart warning

johnjeng:
In project settings:
project settings

In courses:
courses

In remove all upgrades:
remove all upgrades


williamstein:
Thanks!!  Merged into kucalc-master.

PR #2321, opened 2017-08-23
DO NOT MERGE!  Kucalc switch banner

haraldschilly:
This is based on #2320 and adds a hard coded timestamp. It must be set to the actual time of the switch ... but there is an additional check to not show it prior to the given date.


haraldschilly:
I've resolved the merge conflicts here in github, and note, this is a PR against the kucalc-master branch. If github is doing what this is supposed to be, it won't show up on the live site until the kucalc-master is merged :relaxed:

PR #2328, opened 2017-08-25
Close race condition hole in student creation

### LINKED ISSUE #2323: Adding students to a .course file may (i) unexpectedly create a project for the instructor (ii) create duplicate projects for a student
Using a teacher's account, **T**, create a project called **P**, then create a .course file called **C** and add a few students (I think adding multiple students here is key to the problem). From what I've observed, the system does the following when adding students:

1. creates multiple projects with the same name as **C** for **T**
2. renames the projects individually to accommodate a student's name and adds the student as a collaborator
3. hides it for **T**
4. repeats 2 and 3 for however many students were added

The first problem (EXPERIENCED BY THE INSTRUCTOR) is that SOMETIMES all projects created for students have already been renamed accordingly BUT there are some leftover created projects from step 1 with the name **C** that are stuck without a student to be assigned to. 

The second problem (experienced by the student) is that sometimes projects created in step 1 would assign two projects to a student. If the instructor tries to access the student's project from the .course file (By selecting the student's dropdown and pressing "Open student project"), the instructor would only open one of them. But both projects would still be hidden and accessible for the instructor through the projects listing.

johnjeng:
Fixes #2323. @bryantpq Please test.

The problem may still occur on disconnects.

### DIFF
--- a/src/smc-webapp/course/actions.coffee
+++ b/src/smc-webapp/course/actions.coffee
@@ -493,7 +493,7 @@ exports.CourseActions = class CourseActions extends Actions
                     project_id     : project_id
                     table          : 'students'
                     student_id     : student_id
-                @configure_project(student_id)
+                @configure_project(student_id, undefined, project_id)
             delete @_creating_student_project
             queue.shift()
             if queue.length > 0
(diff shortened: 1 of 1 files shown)


bryantpq:
There is a case that when you quickly switch between the project listings tab and the project containing the .course file used to add students WHILE the projects for students are being made, the system may not hide a project for the instructor. Otherwise no problems found.

PR #2332, opened 2017-08-28
db query fix

haraldschilly:
fix  string-interpolation that should result in a string for postgres containing a "-string for the key in a json object.

This is how I tested it:

```
coffee> opts = {}
{}
coffee> opts.to = 'test@email'
'test@email'
coffee> "SELECT invite#>'{\"#{opts.to}\"}' AS to FROM projects"
'SELECT invite#>\'{"test@email"}\' AS to FROM projects'
coffee> console.log("SELECT invite#>'{\"#{opts.to}\"}' AS to FROM projects")
SELECT invite#>'{"test@email"}' AS to FROM projects
```

### DIFF
--- a/src/smc-hub/postgres-server-queries.coffee
+++ b/src/smc-hub/postgres-server-queries.coffee
@@ -1672,7 +1672,7 @@ class exports.PostgreSQL extends PostgreSQL
             cb         : required
         if not @_validate_opts(opts) then return
         @_query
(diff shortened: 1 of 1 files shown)


haraldschilly:
and well ... that's how the query should look like

`select invite#>'{"...email@address..."}' as to from projects where project_id = '...uuid...';`

PR #2341, opened 2017-08-30
Changed behavior when downloading multiple files

bryantpq:


src/smc-webapp/project_files.cjsx
```diff
@@ -1536,44 +1537,76 @@ ProjectFilesActionBox = rclass
...
-    render_download_link: (single_item) ->
+    download_multiple_click: ->
+        destination = ReactDOM.findDOMNode(@refs.download_archive).value
+        async.series([
+            (cb) =>
+                @props.actions.zip_files
+                    src  : @props.checked_files.toArray()
+                    dest : misc.path_to_file(@props.current_path, destination)
+                    cb   : cb
+            (cb) =>
```
williamstein: This second block isn’t actually asynchronous.

Basically you want
```
@props.actions.zip_files
    cb: (err) ->
        if err 
            ....
            return
        download file
        ...
```
It’s so simple in this case that async isn’t useful.

src/smc-webapp/project_store.coffee
```diff
@@ -829,18 +829,30 @@ class ProjectActions extends Actions
...
-        webapp_client.exec
-            project_id      : @project_id
-            command         : 'zip'
-            args            : args
-            timeout         : 50
-            network_timeout : 60
-            err_on_exit     : true    # this should fail if exit_code != 0
-            path            : opts.path
-            cb              : @_finish_exec(id)
+        if (opts.cb?)
```
williamstein: Don’t copy/paste the same code twice like this; doing that is not pretty and hard to mainting.

The entire thing you’ve done here is equivalent to just doing it once and making the input for cb be

 `opts.cb ? @_finish_exec(id)`

which means `opts.cb` if it is defined, and `@_finish_exec(id)` if it isn’t.

bryantpq: I choose to write it like that for [lines 845 & 846](https://github.com/sagemathinc/cocalc/pull/2341/files#diff-e6453b90cf71dc76e152191a2985bafdR845), which would have no clean up if the callback was already given. What would be the best way to handle that?

williamstein: With an if statement.  There’s absolutely no possible reason ever — except for one-off code (so not part of a production codebase) — to write copy that has an exact copy of itself later on.

PR #2347, opened 2017-09-02
support: replace or add ?session=support to the support requests or included URLs in the body

haraldschilly:
(this ticket also enabled hub/misc tests!)

This is mainly to help us with support, and I gave it a brief dry run and added some tests (4 postgres tests fail for me, though, so that's why `npm run testmisc` is all there is for this)


williamstein:
I merged this, and life is too short to complain, but wouldn’t it have been better to do this on the client?

Issue #2348, opened 2017-09-03
fix failing api mocha tests for public_get_text_file


DrXyzzy:
Mocha runs of src/smc-hub/test/api/text_files.coffee fail on the two tests for `public_get_text_file`:
- it "reads a public text file in a project"
- it "reads a folder as public text file and gets zip archive"

Both tests fail reporting the target project id as unrecognized argument:
```
resp: {"event":"error","id":"666b49c1-ab17-4c5a-9585-f3eb01fad81e","error":"usage: smc_compute.py [-h] [--dev] [--single] [--kucalc] [--projects PROJECTS]\n
{start,status,state,disk_quota,compute_quota,create_user,delete_user,killall,stop,restart,directory_listing,read_file,copy_path,mkdir}\n
    ...\nsmc_compute.py: error: unrecognized arguments: a06e1052-...-bf0d47f1f167\n"}
```

All other tests pass in src/smc-hub/test/api/ at this time, doing the following:
```
# ... start test postgres in other shell, e.g. src/dev/proj/start_postgres.py
~/cocalc/src/smc-hub$ mocha --opts test/mocha.opts test/api
```

haraldschilly:
Hmm, I tried to understand this, but I don't really see what's going on. It seems to be harmless, but maybe there is some negative side effect...

DrXyzzy:
Problem is gone. There have been various updates to code, npm, and node.

To check that problem is gone:
1. in one .term, start test instance of postgres:
    ```
    ~/cocalc/src/dev/project
    . postgres-env
    ./start_postgres.py
    ```
1. in a second .term, run the tests
    ```
    ~/cocalc/src/smc-hub
    # do next line if mocha is absent from node_modules/.bin
    npm install --only=dev
    npm run testapi 2>&1 | grep -v debug
    ==>
    > smc-hub@1.0.0 testapi /home/user/cocalc/src/smc-hub
    > echo 'TEST API'; SMC_DB_RESET=true SMC_TEST=true node_modules/.bin/mocha --reporter ${REPORTER:-progress} test/api/*.coffee
 
    TEST API
    ...
      [▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬]
 
      100 passing (33s)
    ```

PR #2349, opened 2017-09-03
change nargs for `public_get_text_file` mocha test

DrXyzzy:
Please review. I am have confirmed that this 1-character change fixes the problem in #2348, e.g.
```
# start test postgresql in background...
~/cocalc/src/smc-hub$ mocha --opts test/mocha.opts test/api/text_files.coffee
 ==>
12 passing (6s)
```

If I understand the change to `nargs` correctly, it _should_ be benign for other cases, e.g. kucalc testing, but it should be reviewed.


williamstein:
Also, smc-compute is I think not even used at all anymore in the kucalc backend — only for the docker image and project dev.

Issue #2233, opened 2017-07-27
prolog jupyter kernel


williamstein:
Requested by: Minerva

williamstein:
https://github.com/Calysto/calysto_prolog

haraldschilly:
It's installed, but not working 100%. (and I'm by far not an expert on prolog ...)

* In the new jupyter2, the question mark at the end triggers an autocompletion (resulting in the error: `{"error":"completion failed"}` in a help box opening up at the bottom)
* It does something more useful when switching to the classical one.
* There seem to be no "False" answers? Not sure about their syntax though ... Maybe there is even some implementation issue with the kernel itself.
* Answers always show `Use '%continue' for more results.` and then in the next cell, `%continue` gives another such warning and shows one more ... that's maybe something for upstream, too.

Link to see current status: https://cocalc.com/projects/94169f5c-20cd-43f0-9bfd-8c62c9c56e6c/files/prolog.ipynb

Console:

```
$ jupyter console --kernel=calysto_prolog
Jupyter console 5.1.0
 
Prolog kernel - evaluates Prolog programs
 
 
In [1]: has(a, b).
      : 
Rule added to database.
 
In [2]: has(a, b)?
      : 
Use '%continue' for more results.
Out[2]: True

In [3]: has(a, c)?
      : 
No more results.
 
```

haraldschilly:
note about completions: `smc-webapp/jupyter/parsing.coffee` is clearly the culprit of breaking prolog.

williamstein:
? help is fixed.

williamstein:
@haraldschilly -- what's current status on this?

haraldschilly:
I don't remember much about this. I think you mentioned that this should be somehow fixed differently. Also, it's weird for me that there is any parsing going on regarding question marks at all. Besides that, this prolog kernel is working again. Hence this issue from my first comment is resolved.

williamstein:
OK, I'll close it.

PR #2355, opened 2017-09-06
jupyter2: special case regarding processing ?/?? help for the language "prolog"

haraldschilly:
se issue #2233

process: the main feature of prolog is to answer questions about some believes. in calisto, these questions end with a questionmark. this jupyter implementation triggers introspection when the code line ends with a questionmark regardless of which programming language we deal with ... that's not going to work. (https://github.com/Calysto/calysto_prolog)

test:
* `has(a, b).`
* `has(a, b)?` → true

### DIFF
--- a/src/smc-webapp/jupyter/actions.coffee
+++ b/src/smc-webapp/jupyter/actions.coffee
@@ -703,7 +703,9 @@ class exports.JupyterActions extends Actions
         switch cell_type
             when 'code'
                 code = @_get_cell_input(id).trim()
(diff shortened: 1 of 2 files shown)


williamstein:
I’m ok with making this change.  However, I think the right fix is to rearchitect things so the kernel itself decides whether an evaluation is requesting an introspection or not.   I didn’t do this since it’s so vastly different than what I was used to with sagews/sagenb/etc., and didn’t even realize it is how jupyter works until it was very late.

Issue #307, opened 2015-12-02
in the course editor, add a setting to not require entering a grade before returning assignments


williamstein:
Requested by: Henning Ulfarsson

This should, of course, be really easy to implement...

haraldschilly:
skip grading was merged a year ago

PR #2357, opened 2017-09-06
webapp/course: add a "skip grading" button

haraldschilly:
This branch contains some work towards issue #307. It's much harder than I thought, quite chaotic when I don't know what all the supposed data structures and implications are.

What's missing: offset for peer grading in the header and figuring out how to fix the "return" confirmation dialog (and also the logic behind it, because instead of only returning graded assignments it should return all collected ones)


haraldschilly:
The scope of this PR might likely also cover https://github.com/sagemathinc/cocalc/issues/2362 (otherwise merge conflicts)

haraldschilly:
Ok, this is ready and  someone should look into this. It's confusing for me to work on this because the logic behind this does counterintuitive things. So, please really review :-)

For testing itself, there is a combinatorial set of cases: 
* collected assignments: some are, or all are collected. check what the return button says when skipping grading
* some/all graded, but skipping: no grade file is created
* peer grading: check with and without, e.g. what to do if there is a grade, grading skipped, and returning all of the graded documents? (right now it shows the grade in the grade file if there is a grade)
* in the student list, the return button should be there and work when the student's assignment is collected and grading is skipped

haraldschilly:
oh, and this ticket is also solved by this: https://github.com/sagemathinc/cocalc/issues/1630 (it makes no sense to give everyone a dummy grading, when you actually want to skip this step)

src/smc-webapp/course/actions.coffee
```diff
@@ -314,7 +314,7 @@ exports.CourseActions = class CourseActions extends Actions
 
     # Takes an item_name and the id of the time
     # item_name should be one of
-    # ['student', 'assignment', 'peer_config', handout']
+    # ['student', 'assignment', 'peer_config', handout', 'skip_grading']
```
haraldschilly: self reviewing this, ha, well, this 'skip_grading' shouldn't be here in the comment. leftover from early attempts.

johnjeng:
I don't think it makes any sense to have a skip peer grading option. You can already opt out of peer grading. It's opt in only. Furthermore, peer grading can always be collected regardless of a peer grade. 

Also, currently checking `skip grading` allows for the next step in peer grading even though this breaks.

haraldschilly:
yes, skipping peer grading is obviously not an issue because it's opt in. the point is to skip the grading step to return the graded files without filling in a grade. there is this grading step regardless of peer grading.

johnjeng:
Ah, I see. The second point still stands. Also `skip grading` appears in the wrong column in this case.

haraldschilly:
Hmm, well, could be the wrong column. Also, initially I assumed that with "peer grading" the whole "grading" column should go away, but then I saw some mentions of grading a peer graded assignment. 

Hence in summary, this change should change:
*  non-peer-grading -> grading is mandatory to optional (no grade file created, unless to temporarily disable skip grading, add a grade, and return it)
* peer grading -> grading is mandatory to optional (and if there is a grade, stil add it to the grade file, because the grade file is always created in this case) <- this is of course an arbitrary decision I made.

To make this even more complicated you might want to come up with a "partial grading" scheme. I don't think this makes much sense, though. That's why I made it to hide the grade buttons in skip mode.

williamstein:
Wouldn’t a much simpler solution to all this be to just always allow returning assignments, even if no grade was explicitly entered (like all the other steps in the workflow)?  I don’t see why we even need a “skip grading” button.  Every other step of the workflow allows that step to just happen without any condition except the previous step happened.  Wouldn’t a 1-line change somewhere allow returning of assignments without the condition that a grade was entered, and that would solve the whole problem without making anything more complicated?  Perhaps I’m wrong and just not thinking of some use case?

Just don’t include a GRADE.txt file with the students that didn’t have a grade assigned.

(This would also address the case where 2 or 3 students are gone, so don’t get a grade, but are still listed.)

johnjeng:
Dang, @williamstein beat me to it. I was just wondering the same thing.

We don't have to assume all or any students need a grade before passing back assignments.

haraldschilly:
I assumed it's a feature to have such a "mandatory" grading by default. That way it's not overlooked if a grade is missing. It's not that complex what I added, just a couple of cases to consider and the codebase is new to me. I also lean towards not changing the existing behavior. 

Also, changing this makes it inconsistent with peer grading. There, the peer-assignment step is only possible if all assignments are collected.

Regarding missing students, they need to be deleted, is this the idea? Maybe "delete" should more like an "ignore" category? delete sounds destructive, while "ignore" fits more with the behavior we have right now.

williamstein:
Since what you’ve done is already done... I am OK with going with it; we could always change it to my idea later.

> Not overlooked...

That cloud be addressed with a warning message (like we do now at each step when there is a potential issue). 

```
Return...

[To all students even the 4 without grades]
[To only the 18 students who have assigned grades]
```

johnjeng:
There's still the issue with peer grading: it allows for doing a next step which results in an error.
screen shot 2017-09-10 at 12 22 18 pm

haraldschilly:
Oh, good catch! Thank you! I'll fix this.

johnjeng:
Oh, should probably change the GRADE.txt file to say something like
```
Either your instructor has made the grade is not available inside CoCalc or there is no grading for this assignment
```

PR #2366, opened 2017-09-08
Don't show extension warning for urls in new file

### LINKED ISSUE #2322: putting a URL e.g., to github into the New file box displays an invalid warning message
See:

screen shot 2017-08-23 at 12 49 43 pm

It does still work fine and clone the git repo.  But the warning is just totally wrong in this case.

Solution: don't show warning if it is a URL.

johnjeng:
Fixes #2322. 

Shows no warning since the text already offers URLs as a valid option to enter. You hit enter or create and you'll be redirected to the project files page and a `downloading [url]...` notification in the upper right.

### DIFF
--- a/src/smc-util/misc.coffee
+++ b/src/smc-util/misc.coffee
@@ -1776,10 +1776,13 @@ exports.peer_grading_demo = (S = 10, N = 2) ->
 exports.ticket_id_to_ticket_url = (tid) ->
     return "https://sagemathcloud.zendesk.com/requests/#{tid}"
 
+# Checks if the string only makes sense (heuristically) as downloadable url
+is_only_downloadable = (string) ->
+    string.indexOf('://') != -1 or exports.startswith(string, 'git@github.com')
+
 # Apply various transformations to url's before downloading a file using the "+ New" from web thing:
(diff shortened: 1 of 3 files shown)


williamstein:
Please change this to actually be a little more intelligent about what is and what is not a URL.  E.g., your code would match the filenname `woahttpeak`.

Two thoughts:

 - elsewhere in the code there is obviously some sort of check whether the file is a url or not, since that is used to decide whether or not to use wget.  Use/refactor that code.  Maybe that code is dumb, but I’m sure it is better than indexOf(‘http’)
 - proper solutions: https://stackoverflow.com/questions/3809401/what-is-a-good-regular-expression-to-match-a-url

johnjeng:
It won't match `woahttpeak` but I will make a better URL checker.

williamstein:
Sorry it would match `httpeepers`...

johnjeng:
That code was better than the ones in the stackoverflow question for our purpose. Mostly because I don't think we want to match `www.website.com`. Making a folder with that name is totally reasonable.

williamstein:
@johnjeng  -- You didn't test this, or maybe didn't commit the change properly (the function you defined in misc isn't exported).  I'm fixing it.

Issue #2370, opened 2017-09-12
api 'public_get_directory_listing' returns redundant struct if no such path


DrXyzzy:
Example of value returned from api call when path requested is "nonexist":
```
{'event': 'public_directory_listing',
 'id': '217b4e1b-2599-4185-ae44-e6b5271ba7ef',
 'result': {'error': "command 'cc-ls' (args=/home/user/nonexist) exited with "
                     "nonzero code 1 -- stderr='error: no such path "
                     "'/home/user/nonexist''",
            'files': {'files': []}}}
```

Last line should be:
```
            'files': []}}
```

DrXyzzy:
Output structure looks ok if public path requested exists and is empty:
```
{'event': 'public_directory_listing',
 'id': 'b6df1763-d79b-44dc-88b2-2c6982a33c9e',
 'result': {'files': []}}
```

And looks ok if public path requested exists and is not empty:
```
{'event': 'public_directory_listing',
 'id': 'eb9a2763-8604-460a-b219-f084bba2321c',
 'result': {'files': [{'mtime': 1505191879, 'name': 'hello.txt', 'size': 6}]}}
```

DrXyzzy:
Fixed by #2371.

PR #2371, opened 2017-09-12
fix for empty files list

DrXyzzy:
Ref: #2370.
All api mocha tests pass with this change. There is something odd, though. Please see second item.

1. The change at line 1564, `listing.files = []`, looks ok to me, but...

1. After adding the above change, but without the change at line 1563, another test fails that should pass. If the public path requested exists and contains a file, the value of listing just before line 1563 is (correctly) displayed by console.log as

    ```
    {"files":[{"size":11,"name":"doc1.txt","mtime":1505195811}]}
    ```
but the test for `if not misc.is_array(listing?.files?)` succeeds, causing the API call to return an empty files list. This problem is eliminated by the change to line 1563, `if not listing?.files?[0]?`.

### DIFF
--- a/src/smc-hub/postgres-server-queries.coffee
+++ b/src/smc-hub/postgres-server-queries.coffee
@@ -1560,8 +1560,8 @@ class exports.PostgreSQL extends PostgreSQL
                     # some files in the listing might not be public, since the containing path isn't public, so we filter
                     # WARNING: this is kind of stupid since misc.path_is_in_public_paths is badly implemented, especially
                     # for this sort of iteration.  TODO: make this faster.  This could matter since is done on server.
-                    if not misc.is_array(listing?.files?)  # There is no telling what listing actually is, so be careful.
-                        listing.files = {files:[]}
+                    if not listing?.files?[0]?  # There is no telling what listing actually is, so be careful.
+                        listing.files = []
                     else
                         listing.files = (x for x in listing.files when \
                             misc.path_is_in_public_paths(misc.path_to_file(opts.path, x.name), public_paths))


haraldschilly:
In general, this looks ok to me. However, one api test fails: ` testing calls relating to creating user accounts --  "before all" hook:` Does it work for you?

Also, running `npm run testpg`, I get 4 failing postgress errors. They are unrelated to this here, but we should try to get rid of these, too.

DrXyzzy:
@hsy, thank you for the review.

I only get the 'before all' error with `npm run testapi`, not with
```
cd ~/cocalc/src/smc-hub
mocha --opts test/mocha.opts test/api
```

I also see other errors with `npm run test*` at smc-hub and parent directory to that and agree they should be fixed. And I'm concerned about `misc.is_array()` returning an apparently wrong result.

haraldschilly:
Ahh! right, this needs to be run differently, I forgot. We should probably fix this `testapi` in the package.json, too.

About misc.is_array, I understand it is true for an empty one? That's ok IMHO. Maybe the actual test should be (e.g. for `x = []`) `misc.is_array(x) and x.length > 0` (which gives false)

haraldschilly:
and noteworthy, this fails with that single timeout.

```
SMC_DB_RESET=true  node_modules/.bin/mocha --opts test/mocha.opts test/api
```

so, the question I'm wondering about is, what exactly does this reset? why is it triggering this? Maybe there is some broken edge case with some test setup routine.

DrXyzzy:
No, misc.is_array() is returning false for an argument of `{"files":[{"size":11,"name":"doc1.txt","mtime":1505195811}]}?.files?`, but _only if the line after it is changed as above_. In the file before this patch, the is_array anomaly does not occur.

DrXyzzy:
Above commit passes all api mocha tests:
```
~/cocalc/src/smc-hub$ mocha --opts test/mocha.opts test/api/
→
  99 passing (28s)
```

PR #2376, opened 2017-09-14
webapp: reporting wait_until_project_is_open time

haraldschilly:
This is a proof of concept, I haven't figured out how to measure the project's start time. 

During my tests, this does indeed send an integer value to GA, but I only see a way to get an average value. That's not much.

There is no real need to merge this.

### DIFF
--- a/src/smc-webapp/projects.cjsx
+++ b/src/smc-webapp/projects.cjsx
@@ -28,7 +28,7 @@ underscore = require('underscore')
 
 misc = require('smc-util/misc')
 {required, defaults} = misc
-{html_to_text} = require('./misc_page')
+{html_to_text, analytics_event} = require('./misc_page')
 {SiteName, PolicyPricingPageUrl} = require('./customize')
 
 markdown = require('./markdown')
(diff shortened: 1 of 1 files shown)


williamstein:
> I haven't figured out how to measure the project's start time

Don’t do that.  Measure what **matters to the user**, which is the time from when they try to open the project, until it “works” as perceived by them.  E.g., directory listing or a file opens successfully.  It doesn’t matter one bit to them how long the project took to start up, or even if it started up.  What matters to them is they want the project to work... and then it works.

haraldschilly:
yes, that's what I mean.

and regarding reporting values to GA, @johnjeng, don't do what I did and send them the ms values. reporting the millisecond value doesn't help us much. Instead, do a bin-count with thresholds. e.g. [0 ... 50], [0 .... 100], [0 ... 500], [0 ... +inf] (give them labels like "50ms", "100ms", ...). Each one counts "1" or isn't reported. that way we can easily see the ratios and we can answer a question like "how often do project open times take longer then our objective threshold of Xms?")

haraldschilly:
well, this is not a good idea at all, not enough information and we don't need this external dependency.

PR #2383, opened 2017-09-20
Added error handling when converting sagews to ipynb

bryantpq:


williamstein:
Thanks for trying, but there is no possible way this can work to show an error.  I mean look at it...

Issue #2400, opened 2017-09-27
Subdirectories with a "#" in their names stopped existing.


diataha:
I have a CoCalc project where I wrote lecture notes. It has subdirectories named things like "Homework #1" and "Lecture #1". When I now try accessing those I get an error message:

> No such directory
> The path Lecture #1 does not exist.

Only the subdirectories with a # sign in their names have that problem.

haraldschilly:
Hello, that must be a bug. Do you need help renaming these directories back to have no "#"? In the meantime we're working on a bugfix.

williamstein:
fix will be live in a few minutes; only will applies to newly (re-)started projects.

diataha:
Works like a charm. Thanks!

PR #2401, opened 2017-09-27
project: handling hash signs in URLs for directory listings

haraldschilly:
hotfix for #2400

This needs a clear mind to think through this little change not causing other harm.

### DIFF
--- a/src/smc-project/directory-listing.coffee
+++ b/src/smc-project/directory-listing.coffee
@@ -22,7 +22,9 @@ exports.directory_listing_router = (express) ->
 directory_listing_http_server = (base, router) ->
 
     router.get base + '*', (req, res) ->
-        path = decodeURI(req.path.slice(base.length).trim())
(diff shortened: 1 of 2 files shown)


haraldschilly:
Looking at `smc-util/misc.coffee`'s `encode_path`, I think this should be `encodeURIComponent` and we should remove these string-substitution extras. It's usually the "component" variant to use for URI encoding, when it is meant for the url, an argument in the query, or the fragment part after "#" ...

In the project's raw server, express is already dissecting the components of the URI. So, that's why I think we definitely want to encode this path as a component.

(URI's are a tricky mess)

williamstein:
Thanks -- this was entirely introduced by me -- at least it isn't a longtime bug.

williamstein:
Your changes to misc.coffee completely break everything and are wrong.  I'm reverting those, but keeping the project change, which seems fine.  No matter what (even if it could be made to work)  we can't possibly make that change to misc because it would require restarting all projects, since all existing url requests would be broke with the old code (since forward slashes / are encoded in a way that decodeURI won't decode).

PR #2406, opened 2017-09-28
fix building error

lwintermelon:
`install.py all --web` doesn't work since added debug argument  detection to function `install_webapp` in `install.py`
https://github.com/sagemathinc/cocalc/blob/e30fe5e7a5684dcda58ed14d5b9f75e21cab8e2b/src/install.py#L84

### DIFF
--- a/src/install.py
+++ b/src/install.py
@@ -157,7 +157,7 @@ def main():
     # add argument '--action'(just a flag) and '--debug' to uniformly deal with 'install.py webapp ***' and 'install.py all --web ***' for install_webapp 
     parser_all.add_argument('--action', default='build')
     parser_all.add_argument("--debug", action="store_true", help="if set and, build debug version of code (rather than production) for webapp")
-    parser_all.set_defaults(func = lambda args: install_all(compute=args.compute, web=args.web))
+    parser_all.set_defaults(func = lambda args: install_all(args))
 
     args = parser.parse_args()
     args.func(args)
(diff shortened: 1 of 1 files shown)


src/install.py
```diff
@@ -81,7 +81,7 @@ def install_webapp(*args):
         # update term.js
         cmd("cd webapp-lib/term; ./compile")
         wtype = 'debug' if args[0].debug else 'production'
```
williamstein: Thanks for your PR!  

It looks like there is still an args[0] in this line?

lwintermelon: I'm so careless...

williamstein: Pretty much everybody makes lots of mistakes when writing code no matter how long they try.   It's great to be writing code where one has the luxury of making mistakes (e.g., I hate writing code where i have to push it into production immediately and it must work or things get way worse -- that sucks).  Relax and enjoy.

williamstein:
This doesn't seem to work:
```
~/smc/src$ git branch
...
* lwintermelon-patch-1
...
~/smc/src$ ./install.py all --web
  File "./install.py", line 166, in <module>
    main()
  File "./install.py", line 163, in main
    args.func(args)
  File "./install.py", line 160, in <lambda>
    parser_all.set_defaults(func = lambda args: install_all(compute=args.compute, web=args.web))
TypeError: install_all() got an unexpected keyword argument 'compute'
```

lwintermelon:
Apologies, i should have tested it. it's ok now.

Issue #2398, opened 2017-09-26
add an extra input box next to the grade box for a comment.


williamstein:
Requested by Kristin McCully:

> "Currently we've been putting both a numerical score & comments in the grade box. That's difficult to parse out in the resulting CSV. It would be great to have a small grade box just for the grade and a separate larger comment box for more text that would be separate columns in the CSV, so we can upload them to our gradebook as separate columns in CSV. Thus far, we've either had to enter scores in our gradebook manually or use complicated code/Excel function to parse out numerical scores. I hope that's specific enough ..."

(whoever does this may as well also make returning without a grade optional -- just don't do that check and include the return button -- it would probably just involve commenting out a line of code.)

johnjeng:
So, this will be a breaking change for people parsing their CSV files. Do we want to warn them?

williamstein:
Is there a way to do this that is backward compatible?  E.g., put the new column at the very end?

johnjeng:
Yeah, we can do that. Organizationally it makes less sense to me.

johnjeng:
Oh, except it mirrors the organization of the Python export. Cool.

williamstein:
Done!

williamstein:
<img width="1365" alt="screen shot 2017-10-02 at 2 08 32 pm" src="https://user-images.githubusercontent.com/1276278/31099578-56ec022c-a77b-11e7-9a11-b3a2844e7178.png">

PR #2409, opened 2017-09-28
Issue/2398

johnjeng:
Implements issue #2398 
screen shot 2017-09-28 at 4 17 15 pm
screen shot 2017-09-28 at 4 17 38 pm
screen shot 2017-09-28 at 4 17 50 pm

# Test
- Export with CSV
- Export with Python
- Empty comment box
- Hitting `esc` while the cursor is in either box
- Saving

### DIFF
--- a/src/smc-webapp/course/actions.coffee
+++ b/src/smc-webapp/course/actions.coffee
@@ -799,6 +799,17 @@ exports.CourseActions = class CourseActions extends Actions
         obj.grades = grades
         @_set(obj)
 
+    set_comments: (assignment, student, comments) =>
+        store = @get_store()
+        return if not store?
+        assignment    = store.get_assignment(assignment)
+        student       = store.get_student(student)
+        obj           = {table:'assignments', assignment_id:assignment.get('assignment_id')}
+        comments_map = @_get_one(obj).comments ? {}
+        comments_map[student.get('student_id')] = comments
+        obj.comments = comments_map
+        @_set(obj)
+
     set_active_assignment_sort: (column_name) =>
         store = @get_store()
         if not store?
--- a/src/smc-webapp/course/assignments_panel.cjsx
+++ b/src/smc-webapp/course/assignments_panel.cjsx
@@ -756,13 +756,14 @@ StudentListForAssignment = rclass
     render_student_info: (student_id) ->
         store = @props.redux.getStore(@props.name)
         <StudentAssignmentInfo
-              key     = {student_id}
-              title   = {misc.trunc_middle(store.get_student_name(student_id), 40)}
-              name    = {@props.name}
-              student = {student_id}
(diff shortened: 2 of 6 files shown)


williamstein:
## Critical:

- [ ] If you make a comment with a newline in it, the exported .py file contains syntax errors due to not properly making mulitline strings (use repr properly).  I think the .csv file is fine though.
- [ ] When returning the homework assignment, the comment should be included in the GRADE.txt file along with the grade.
- [ ] Hitting shift+enter in the input box should save its contents.  Everybody using cocalc will be used to and assume this, since this is how Jupyter and Sagews (and even Github!) work.  Not doing this will drive our users crazy.

## Could be done after an initial release:

- [ ] The input box should expand vertically as you enter more text.  This is a pretty easy trick to do.  Or, at the worst, make it 5 lines tall when it is focused orr visible.  2 lines is not enough.
- [ ] When the comment is not being edited, only show the first 40 characters (?) of the comment, then `...`; i.e., truncate it.  I typed a long comment, and now that line for that student will always take up tons of space.  Clicking on the truncated comment could open it for editing.
- [ ] Display the comment using markdown.  In particular, note that there are ALREADY two notes facilities for students and for assignments -- check them out. Everything should be as consistent as possible.  I just noticed that the label for assignments is "Private assignment notes", but for a student it is just "Notes" -- this is a longtime mistake, I assume, so can you fix this too.   In any case, these are all rendered as markdown and that they are is clear.
- [ ] I hit another bug when testing -- the .py is named the same as the .course file. However, if the course filename has spaces or dashes in it (etc.) then [course name].py is NOT a valid python file and can't be imported.  We should canonicalize the .py file.  Please open a ticket...

johnjeng:
I did not make the box auto-grow but everything else ought to be working.

Newlines correctly get converted to `\n` for exporting.
Markdown correctly truncates on rendered text. 
Comments are truncated at 80 characters.
`Shift-enter` works while in either input box.
On display, Grade and Comments labels are bolded for easy visual separation.

williamstein:
@j3 -- that html truncate is pretty disturbing -- it just 100% silently truncates with no ... or anything else.  Otherwise this looks very good.

It's also sort of dumb how the returned file is .txt containing markdown.  The students never see the formatting.  However, that's not possible to nicely solve without a lot of work, and is really best solved with a .student file that will render everything better...

williamstein:
In fact, I'm really not happy with this html truncation; it makes more sense to use CSS right?  HTML truncation seems pretty weird, and also it doesn't control at all the actual vertical height -- a comment could still take a very unevent amount of vertical height.    I'll try making this change.

williamstein:
Otherwise, this is really, really good.

johnjeng:
Oh yeah, just let the parent truncate with overflow. Durrr. +1

Issue #2415, opened 2017-10-04
mocha api config needs updating


DrXyzzy:
Deprecation warning:
```
~/cocalc/src/smc-hub$ node_modules/.bin/mocha --opts test/mocha.opts test/api/basic.coffee

(node:44619) DeprecationWarning: "--compilers" will be removed in a future version of Mocha; see https://git.io/vdcSr for more info
```
As noted in https://git.io/vdcSr in the warning, along with changing the coffeescript option in `mocha.opts`, it's also necessary to add a glob in the `testapi` test directory script in `packages.json`.

nemanjacosovic:
@DrXyzzy Check out the solution > https://github.com/mochajs/mocha/wiki/compilers-deprecation

Everything I needed to do was to switch from --compilers to --require.

DrXyzzy:
fixed in #2416

PR #2416, opened 2017-10-04
eliminate deprecation for mocha test on compiler option

DrXyzzy:
Ref #2415.

Note: most tests in src/smc-hub/test/api hang after the tests pass. That is a separate issue independent of #2415


haraldschilly:
nitpicking: isn't this glob pattern also necessary for the tests in the lines right above and below it? maybe even `**/*` to catch subdirectories...

DrXyzzy:
lines for testpg and testmisc updated as well

PR #2447, opened 2017-10-26
webapp: expose uptime and total cpu usage in project settings

haraldschilly:
this is one of the kucalc specific tasks, to expose more about project usage.

for testing, well, I hope the queries work and it needs to be run on ws1


haraldschilly:
tested on ws1, see screenshot

screenshot from 2017-10-27 15-10-37

Issue #2453, opened 2017-10-30
invitations say 'SageMathCloud Invitation to Course'


DrXyzzy:
Should say "CoCalc"

https://github.com/sagemathinc/cocalc/blob/master/src/smc-webapp/course/actions.coffee#L523

PR #2454, opened 2017-10-30
invitation from CoCalc, not SMC

DrXyzzy:
Ref: #2453.


williamstein:
lgtm

haraldschilly:
That's not good and should not have been merged. I'll create a follow up change.

I'm also still puzzled why there are these leftovers. There must have been a severe merge inconsistency.

williamstein:
> That's not good and should not have been merged.

What's not good and should not have been merged.  That = ?

PR #2460, opened 2017-11-01
First steps

haraldschilly:


src/smc-webapp/project_files.cjsx
```diff
@@ -423,6 +426,47 @@ TerminalModeDisplay = rclass
...
+    get_first_steps: ->
+        {LIBRARY} = require('./project_store')
+        @props.actions.copy_from_library(src:LIBRARY.first_steps)
+
+    dismiss_first_steps: ->
+        @props.actions.touch_file(dest: exports.NO_FIRST_STEPS_SENTINEL_FILE)
+        @props.actions.setState(show_first_steps:false)
+
+    render: ->
+        <Row>
```
williamstein: All of the rest of the source code uses 4-space indentation for cjsx.

src/smc-webapp/project_store.coffee
```diff
@@ -890,6 +893,62 @@ class ProjectActions extends Actions
...
+            project_id      : @project_id
+            command         : 'rsync'
+            args            : ['-rltgoDxH'].concat(opts.src).concat([opts.dest])
+            timeout         : 120   # how long rsync runs on client
+            network_timeout : 120   # how long network call has until it must return something or get total error.
+            err_on_exit     : true
+            path            : '.'
+            cb              : (err, output) =>
+                (@_finish_exec(id))(err, output)
+                @open_file(path: 'first-steps/first-steps.tasks')
```
williamstein: 1. Don't call open_file if there is err is true-ish.  Otherwise you'll just create a blank file.

2. Open something involving opts.dest not first-steps/first-steps.tasks, or eliminate dest.    Otherwise this will be buggy.

williamstein:
It seems strange to me that whether or not to show first steps, etc., is a function of the PROJECT and managed via creating files on the filesystem.    It should be a function of the *user*.   For example, every time I create a new project, I'll see it repeatedly.

We also know from looking that a lot of users use tons of very small projects...

1. First steps should I think be something that is always available *somehow* (possibly via account settings to reset it).
2.  By making whether or not first steps should be shown as an option to a client a function of the *user*, not the project, you can delete all of the code about touching files to indicate whether or not first steps is shown, and replace it by a field in accounts (or some json structure in there).
3. Aside from that, it is also very good to store status of user onboarding in the *database* and relative to the user.  This makes anlaytics easier and gives us a better sense of the status of a user.  It could be pretty hard to do analytics on onboarding if it involves files being set in the filesystem (which is not just a database query away).

I would make a new JSONB field in accounts called onboad, which would be a map storing status info about onboarding of a user.  For starters it could just be whether or not (or when) they've selected to se exactly this beginner tour we're talking about here.  Later it could have a lot more info about their onboarding progress.

haraldschilly:
I agree with your comments. The only detail I didn't change is this "onboarding" field. Here, it is simply a boolean in "other settings".

PR #2466, opened 2017-11-02
webapp/project: calculate a "slowdown" estimate depending on quotas.core and normalized load 5

haraldschilly:
The purpose of this is to make "free users" aware of the current load and the impact on running any calculations. It shows an estimate about how much longer a computation takes compared to having no elevated load or cpu core limits.

The problem is, I'm sure what I coded is at the wrong spot and can be done much better. I think what I wrote is at least safe to do.

I've assigned j3 because I think he knows exactly how to properly trigger this upon project startup.

### DIFF
--- a/src/smc-webapp/project_page.cjsx
+++ b/src/smc-webapp/project_page.cjsx
@@ -246,7 +246,7 @@ FreeProjectWarning = rclass ({name}) ->
 
         if @props.free_compute_slowdown? and @props.free_compute_slowdown > 0.0
             pct = Math.round(@props.free_compute_slowdown)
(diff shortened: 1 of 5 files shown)


johnjeng:
UI and display logic tested. Did not test correctness of the `exec` calls.

johnjeng:
I like that it's a positive statement. I guess there are no minimums? My bs meter always goes off whenever I read "up to X% faster". It's like when cable companies say you're paying for "up to X mbps".

williamstein:
@johnjeng -- is this ready to go?  If not, can the label be changed to "needs work"...?

johnjeng:
Yeah, I think this is good to go.

Issue #2481, opened 2017-11-05
update api docs


DrXyzzy:
1. Indicate which api calls may use a blank security key.
1. Add query example for editor_settings['jupyter_classic'].
1. Add query example to list all projects.

PR #2482, opened 2017-11-05
Issue/2481

DrXyzzy:
Ref: #2481.

### DIFF
--- a/src/smc-util/message.coffee
+++ b/src/smc-util/message.coffee
@@ -2252,7 +2252,7 @@ This change not usually needed, because accounts
 default to "Modern Notebook".
 
 It is not necessary to specify the entire `editor_settings` object
-if you are only setting the `jupyter_classic` attribute.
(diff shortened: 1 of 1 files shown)


src/smc-util/message.coffee
```diff
@@ -2176,6 +2246,24 @@ read, with `cpu_shares` increased to 1024 = 4 * 256.
          "id":"ec822d6f-f9fe-443d-9845-9cd5f68bac20"}
 ```
 
+Set present user to open Jupyter notebooks in
+"Modern Notebook" as opposed to "Classical Notebook".
+This change not usually needed, because accounts
+default to "Modern Notebook".
+
+It is not necessary to specify the entire `editor_settings` object
+if you are only setting the `jupyter_classic` attribute.
```
williamstein: because changes are merged in.

DrXyzzy: "because changes are merged in" added.

PR #2483, opened 2017-11-06
Compute Environment Information

haraldschilly:
This exposes the gathered information about the computation environment of CoCalc (from the test suite) in the static pages and the UI. Technically, it is completely optional and needs the env variable `CC_COMP_ENV` set to `true` to be active → implications for kucalc/static!

Combined, this makes the offer of CoCalc in Kucalc much more discoverable for users and by exposing it in static pages visible to the world.

### Test

`env CC_COMP_ENV=true ./install.py webapp build`

### webapp: section in "About"

comp-env-ui

### Static pages

comp-env-static


williamstein:
This is awesome!  It also seems extremely valuable and low risk.

Any risk at all?

haraldschilly:
status update: made some additional small tweaks, fixed edge cases, and deployed it

PR #2484, opened 2017-11-07
Coupons

johnjeng:
# Description
Screenshots are at the bottom.

Non-coupon Billing UI changes
- Multiple subscription warning only fires if you try to add multiple of the **same kind**.
- Activity spinner on billing page floats in the top right so you see it even if you're scrolled down

Coupons
- Coupons are validated when you add them *and* when you create the subscription.
  - The latter won't be relevant when done via the UI
- Coupons default to single use per account
  - Can set `max_per_account` in metadata if we want a multi-use coupon for some reason...
- Coupon descriptions are set on their metadata via Stripe
- Subscriptions only take one coupon
- No implementation for adding a coupon to the account

# Test
Use Stripe test data.

1. Make some coupons
2. Add any metadata eg. `description : "Takes $30 your first 3 billing cycles"`
3. Try to use the coupons
4. Get errors or successes as appropriate

**Empty**
screen shot 2017-11-06 at 6 51 58 pm

**Typed In** 
Notice the `No coupons applied` becomes relevant incase they think typing it in is sufficient
screen shot 2017-11-06 at 6 55 34 pm

**Successfully Added**
screen shot 2017-11-06 at 6 55 41 pm

**Non-existant Coupon**
screen shot 2017-11-06 at 6 52 19 pm

**Already Used**
screen shot 2017-11-06 at 6 52 29 pm

**Expired Coupon**
This happens any time Stripe says the coupon is invalid.
screen shot 2017-11-06 at 6 53 07 pm

### DIFF
--- a/src/smc-hub/client.coffee
+++ b/src/smc-hub/client.coffee
@@ -1763,6 +1763,7 @@ class exports.Client extends EventEmitter
                 cb(err); return
             cb(undefined, customer_id)
 
+    # id : user's CoCalc account id
     stripe_get_customer: (id, cb) =>
         dbg = @dbg("stripe_get_customer")
         dbg("getting id")
@@ -2113,6 +2114,33 @@ class exports.Client extends EventEmitter
                     @stripe_error_to_client(id:mesg.id, error:err)
                 else
                     @push_to_client(message.stripe_subscriptions(id:mesg.id, subscriptions:subscriptions))
+
+    mesg_stripe_get_coupon: (mesg) =>
+        dbg = @dbg("mesg_stripe_get_coupon")
+        dbg("get the coupon with id == #{mesg.coupon_id}")
+        if not @ensure_fields(mesg, 'coupon_id')
+            dbg("missing field coupon_id")
+            return
+
+        @stripe_get_customer mesg.id, (err, customer) =>
+            if err
+                return
+            async.waterfall([
+                (cb) =>
+                    dbg("retrieve the coupon.")
+                    @_stripe.coupons.retrieve(mesg.coupon_id, cb)
+                (coupon, cb) =>
+                    dbg("check account coupon_history")
+                    # TODO: Limit coupon use. Need a db entry
+                    #if coupon_history[mesg.coupon_id] >= (coupon.metadata.max_per_account ? 1)
+                    #    cb("You have already used this coupon the maximum number of times")
+                    cb(undefined, coupon)
+            ], (err, coupon) =>
+                if err
+                    @stripe_error_to_client(id:mesg.id, error:err)
+                else
+                    @push_to_client(message.stripe_coupon(id:mesg.id, coupon:coupon))
+            )
 
     mesg_stripe_get_charges: (mesg) =>
         dbg = @dbg("mesg_stripe_get_charges")
--- a/src/smc-hub/client.coffee
+++ b/src/smc-hub/client.coffee
@@ -2026,6 +2026,7 @@ class exports.Client extends EventEmitter
                         @validate_coupon options.coupon, (err, coupon, coupon_history) =>
                             if err
                                 cb(err)
+                                return
                             coupon_history[coupon.id] += 1
                             @database.update_coupon_history
                                 account_id     : @account_id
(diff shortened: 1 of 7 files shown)


src/smc-hub/client.coffee
```diff
@@ -2001,7 +2002,7 @@ class exports.Client extends EventEmitter
             options =
                 plan     : mesg.plan
                 quantity : mesg.quantity
-                coupon   : mesg.coupon
+                coupon   : mesg.coupon_id
```
williamstein: can you make it "coupon: mesg.coupon" or "coupon_id:mesg.coupon_id" for consistency?

src/smc-hub/client.coffee
```diff
@@ -2114,6 +2136,48 @@ class exports.Client extends EventEmitter
...
+    # Checks these coupon criteria:
+    # - Exists
+    # - Is valid
+    # - Used by this account less than the max per account (hard coded default is 1)
+    # Calls cb(err, coupon, coupon_history)
+    validate_coupon: (coupon_id, cb) =>
+        dbg = @dbg("validate_coupon")
+        @_stripe = get_stripe()
+        async.series([
+            (local_cb) =>
```
williamstein: So you know, if you just call this (cb) rather than (local_cb) there is no problem whatever with scoping.  It does NOT conflict with the cb in lie 2156 above. I've done this about 10K times, at least....

src/smc-hub/client.coffee
johnjeng: The call to stripe takes it as `coupon : "theCodeString"` but I want to clarify in our code when it's just a `coupon_id` vs an entire `coupon` object. 

I could change it to only right before the stripe call.

src/smc-hub/client.coffee
williamstein: (not saying you should change anything -- just letting you know in case)

johnjeng: I know. I just hate reading it when it's a long function (it was a long function during writing).

src/smc-hub/client.coffee
```diff
@@ -2114,6 +2136,48 @@ class exports.Client extends EventEmitter
...
+                dbg("retrieve the coupon")
+                @_stripe.coupons.retrieve(coupon_id, local_cb)
+            (local_cb) =>
+                dbg("check account coupon_history")
+                @database.get_coupon_history
+                    account_id : @account_id
+                    cb         : local_cb
+        ], (err, [coupon, coupon_history]) =>
+            if err
+                cb(err)
```
williamstein: **MAJOR BUG** You need a return after
```
cb(err)
```
Otherwise, coupon.valid will crash the server, since coupon is undefined.   Harald and I used to cause bugs like this all time, but now very rarely do...

src/smc-hub/client.coffee
```diff
@@ -2114,6 +2136,48 @@ class exports.Client extends EventEmitter
...
+            if not coupon.valid
+                cb("Sorry! This coupon has expired.")
+            times_used = coupon_history[coupon.id] ? 0
+            if times_used >= (coupon.metadata.max_per_account ? 1)
+                cb("You've already used this coupon.")
+
```
williamstein: Same as above.  NO return.  Scaring me...

src/smc-hub/postgres-server-queries.coffee
```diff
@@ -666,6 +666,34 @@ class exports.PostgreSQL extends PostgreSQL
...
+            where : 'account_id = $::UUID' : opts.account_id
+            cb    : one_result("coupon_history", opts.cb)
+
+    update_coupon_history: (opts) =>
+        opts = defaults opts,
+            account_id     : required
+            coupon_history : required
+            cb             : undefined
+        @_dbg("Setting to #{opts.coupon_history}")
+        async.series([
```
williamstein: get rid of the async.series.  It does nothing here.

src/smc-util/client.coffee
```diff
@@ -1541,15 +1541,15 @@ class exports.Connection extends EventEmitter
...
+            coupon_id : undefined
+            cb        : required
         @call
             message : message.stripe_create_subscription
-                plan     : opts.plan
-                quantity : opts.quantity
-                coupon   : opts.coupon
+                plan      : opts.plan
+                quantity  : opts.quantity
+                coupon_id : opts.coupon_id
```
williamstein: Yes, nice.

PR #2489, opened 2017-11-08
Update client stripe version

johnjeng:
Based on my testing of our Stripe calls, this [changelog](https://stripe.com/docs/upgrades?since=2015-07-07#api-changelog), and their api [docs](https://stripe.com/docs/api).

How to tell the new version is running? The way I did it was using `subscriptions.list`
>You can now view canceled subscriptions by specifying status=canceled or status=all when listing subscriptions. In addition, you can now retrieve a canceled subscription by its ID

### DIFF
--- a/src/smc-hub/stripe/connect.coffee
+++ b/src/smc-hub/stripe/connect.coffee
@@ -10,8 +10,10 @@ misc                 = require('smc-util/misc')
 {defaults, required} = misc
 
 stripe  = undefined
+DEFAULT_VERSION = '2017-08-15'
 
-exports.get_stripe = ->
+exports.get_stripe = (version) ->
+    stripe.setApiVersion(version ? DEFAULT_VERSION)
     return stripe
 
 # TODO: this could listen to a changefeed on the database


johnjeng:
- A new value, not_required, has been added to the enum of redirect.status on the source resource. Previously not required redirects were marked as succeeded.
  - Not used
- A new value, under_review, has been added to the enum of verification.disabled_reason on the account resource. under_review was previously shown as other.
  - Not used
- The managed property on account objects has been replaced by the type property, and is now required for account creation. Possible values are standard, express, or custom.
  - Not used
- Event objects (and webhooks) now show entire sub-arrays in previous_attributes when those arrays have changes. Previously those sub-arrays only contained the specific fields that were changed.
  - Not used
- Event objects (and webhooks) will now render a request subobject that contains a request ID and idempotency key instead of just a string request ID.
  - Not used
- The user_id field of Connect-related event objects (and webhooks) has been renamed to account.
  - Not used
- MajorPayouts were moved out of Transfers into their own resource. /v1/payouts now represents moving money from a Stripe account to a bank account or debit card, and /v1/transfers now represents Connect platforms moving money between Stripe accounts. For more details, see https://stripe.com/docs/transfer-payout-split.
  - Only affects my parsing of stripe for accounting. Already changed.
- Charge objects will now render the ID of their dispute instead of the full Dispute object. You may expand this property to render the full Dispute as before.
  - We don't use charges
- Charges that have been blocked by Radar rules will now render the ID of the rule that blocked them under the outcome field instead of the full Rule object. You may expand this property to render the full Rule as before.
  - We don't use charges
- API responses to /v1/balance/history endpoints will no longer return a sourced_transfers property on balance transaction objects.
  - Not used
- The API will now return status code 403 instead of 401 when you make a request that has insufficient permissions.
  - We just report these errors directly so doesn't make a difference
- You can now view canceled subscriptions by specifying status=canceled or status=all when listing subscriptions. In addition, you can now retrieve a canceled subscription by its ID.
  - Not used. We list subscriptions from the customer object.
- When you set the active flag on a product to false, its SKUs will no longer automatically be marked as inactive.
  - Flag not used
- API responses to /v1/accounts endpoints no longer return a currencies_supported property on account objects. Currency information varies based on an account’s country, and this information is still available via the new /v1/country_specs API. For details on the Country Specs API, see https://stripe.com/docs/api#country_specs.
  - Not used
- Add strict checks for postal codes on the account update and account creation endpoints. See https://stripe.com/docs/api#update_account-legal_entity-address-postal_code for more details.
  - Not used
- When an order is changed from paid or fulfilled to canceled, returned, the charge on the order is now automatically refunded. Previously, an attempt to change an order to state canceled or returned would raise an error if the order’s charge had not already been refunded.
  - Not used
- An error is now returned when trying to add over 250 invoice items to an invoice.
  - Not used
- The name prop on the BankAccount resource is now deprecated. All values that were returned under name will now be returned under account_holder_name.
  - Not used
- The legal_entity hash within the Accounts endpoint now only contains properties that are applicable for the account’s country. legal_entity properties that have been filled in will not be hidden.
  - Not used
- An error is now returned if a tax_percent is provided without a plan during customer update and create.
  - Not used. We apply tax_percent on subscription creation
- `Major` An error is now returned if invalid parameters are passed in the card or bank account parameters hash during token, customer source, or external account creation. In addition, the error code for missing required parameters in these bank account or card hashes is now 400 instead of 402.
  - We just report these errors directly so doesn't make a difference
- bank_accounts is no longer a field in the Account object. Use external_accounts instead. Also, bank_account will be replaced by external_account in fields_needed.
  - Not used
- An invoice’s latest charge, if any, now always appears as the charge attribute of the invoice. Previously, a non-card charge appeared as the payment attribute, which no longer exists.
  - We only have card charges. In-face our invoice generator assumes all charges to be `charge`
- `Major` The Charges endpoint previously returned only card charges. Charges now returns all charges, such as card charges and bank account charges. See the endpoint’s documentation for information on filtering by charge source type. The deprecated offset parameter is now only supported when filtering by charge source type. To simultaneously paginate charges from all sources, see https://stripe.com/docs/api#pagination.
  - We don't use charges
- API rate limit errors now return a HTTP status code of 429 instead of 400. They also no longer return a rate_limit error code.
  - Not used
- An error is now returned if a request reuses an idempotency token with different parameters from the original request sharing the token. (Previously, errors were returned only when idempotency tokens were reused across different API endpoints.)
  - We don't retry transactions.
- Balance transactions corresponding to refunds and disputes now show the refund ID or dispute ID instead of the charge ID as the source.
  - Only affects accounter
- The tos_acceptance[date] field for accounts now performs date validation. Dates are expected to be integers, measured in seconds, not in the future, and after 2009.
  - Not used
- The balance.available webhook is now sent when immediate transfers are processed.
  - Not used
- The contacted boolean under account verification is now replaced with a disabled_reason string, describing why a certain account is unable to make transfers and/or charges.
  - Not used

johnjeng:
Actions to test:

- Adding payment method
- Removing payment method
- Buy a subscription
- Cancel a subscription
- Download a payment invoice
- Apply a coupon to a subscription

DrXyzzy:
Test results so far:

Test results so far:
- [x] Adding payment method
- [x] Removing payment method
- [x] Buy a subscription
- [x] Cancel a subscription
- [x] Download a payment invoice

williamstein:
How does this work?   I see that this patch changes something, which i guess impacts something?  Is the actual api change done by ALSO flipping a switch on the stripe website?  In other words how do I release this?

johnjeng:
This patch makes all of our calls to the specified API. 
It's easier to roll this back than from Stripe. So this should be pushed live and if we don't see any issues, we should flip the switch from Stripe.

Flipping the switch in stripe makes all non-specified calls into the upgraded api.

williamstein:
Oh, very nice.  Thanks.

So (1) merge, (2) restart hubs with new code, (3) done.  Then we do something later...

williamstein:
And it turns out that the way I wrote code against the stripe api was so luckily basic/simple that nothing needs to be changed at all?

johnjeng:
Yup. We don't even use Charges which is hilarious. In terms of touching the API surface, we tentatively poke it with our pinky.

PR #2490, opened 2017-11-08
webapp + db schema: brand new "notifications" setting

haraldschilly:
* minimal UI, nothing on the back-end yet.


williamstein:
Yikes.

PR #2501, opened 2017-11-16
Library2

haraldschilly:


src/smc-webapp/library.cjsx
```diff
@@ -0,0 +1,319 @@
...
+                if tag_extra_info.length > 0
+                    info = tag_extra_info.join(' ')
+                    <p style={color: COLORS.GRAY_D}>
+                        <Icon name='exclamation-triangle' style={color:COLORS.YELL_L} /> {info}
+                    </p>
+            }
+            <Button
+                bsStyle  = "success"
+                onClick  = {=> @copy()}
+                disabled = {@state.copy}
```
johnjeng: This copy state should go into the store so that if a user goes navigates elsewhere the copying status will still be correct.

src/smc-webapp/project_store.coffee
```diff
@@ -932,26 +930,99 @@ class ProjectActions extends Actions
...
+
+            webapp_client.read_text_file_from_project
+                project_id : @project_id
+                path       : require('./library').examples_path + '/index.json'
+                cb         : (err, response) =>
+                    if err
+                        cb(); return
+                    try
+                        data = misc.from_json(response.content)
+                        library = @get_store().library.set('examples', data)
```
johnjeng: Per comment above, call `immutable.fromJS(data)` here so we don't have a mutability leak.

src/smc-webapp/project_store.coffee
```diff
@@ -932,26 +930,99 @@ class ProjectActions extends Actions
...
-        if not lib?
-            @setState(error: "Library entry '#{opts.entry}' unknown")
+        async.series([
+            (cb) -> async.eachOfSeries(LIBRARY, check, cb)
+        ])
+
+    init_library_index: ->
+        if _init_library_index_cache[@project_id]?
+            data = _init_library_index_cache[@project_id]
+            library = @get_store().library.set('examples', data)
```
johnjeng: Same here. Or already have the cache saved as immutable data.

src/smc-webapp/library.cjsx
```diff
@@ -0,0 +1,319 @@
...
+        sorted_docs : undefined
+        metadata    : undefined
+
+    componentDidMount: ->
+        # TODO this isn't working
+        #@scroll_into_view = _.debounce((=> $(ReactDOM.findDOMNode(@refs.selector_list)).find('.active').scrollintoview()), 50)
+
+    componentDidUpdate: (props, state) ->
+
+    init_state: (props) ->
```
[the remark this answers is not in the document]
johnjeng: The sorting part is probably preferable to put in the store as a selector
```
sorted_docs : depends('library') ->
```
Which will additionally memoize the result based on library. Then it's accessible as if it were a data value in the store.

If it should be private, for some reason, don't bother.

src/smc-webapp/library.cjsx
```diff
@@ -0,0 +1,319 @@
...
+            sortfn = (doc) -> [
+                meta.categories[doc.category].weight ? 0
+                meta.categories[doc.category].name.toLowerCase()
+                doc.title?.toLowerCase() ? doc.id
+            ]
+            sdocs  = _.sortBy(docs, sortfn)
+            @setState
+                copy        : false
+                sorted_docs : sdocs
+                metadata    : meta
```
johnjeng: Having `metadata` accessible from props as well as state is an anti-pattern (mostly egregious in a large component like this) because it violates a "single source of truth". If I don't know that `@state.metadata` mirrors `library.getIn(['examples', 'metadata'])` I may easily update the wrong one.

johnjeng: Additionally, it's unclear if this means that metadata is *supposed* to be a copy of `library.getIn(['examples', 'metadata'])` that the component *does* modify locally. So then I have to understand all places `@state.metadata` is used and hunt down `library.getIn(['examples', 'metadata'])`. If the intent is just to mirror for convenience, a comment in `getInitialState` should suffice.

src/smc-webapp/project_store.coffee
```diff
@@ -932,26 +930,99 @@ class ProjectActions extends Actions
...
+                url     : index_json_url
+                timeout : 5000
+                success : (data) =>
+                    #if DEBUG then console.log("init_library/content", content)
+                    try
+                        library = @get_store().library.set('examples', data)
+                        @setState(library: library)
+                        _init_library_index_cache[@project_id] = data
+                        cb()
+                    catch e
```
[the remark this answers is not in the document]
haraldschilly: yes, but interestingly, data is already json, so I'm removing this try/catch

src/smc-webapp/projects.cjsx
```diff
@@ -231,7 +231,10 @@ class ProjectsActions extends Actions
         if opts.target?
             redux.getProjectActions(opts.project_id)?.load_target(opts.target, opts.switch_to)
         redux.getActions('page').save_session()
-        project_actions.check_library()
+        # init the library after project started.
+        # TODO write a generalized store function that does this in a more robust way
```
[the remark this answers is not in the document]
haraldschilly: Hmm, I think you started to look at older commits. I'm not sure how all this will end up. My thoughts here where that the parsed json data should also be validated in some way. Here, the only test is if a file exists on the disk.

src/smc-webapp/r_misc.cjsx
```diff
@@ -185,11 +185,15 @@ exports.Loading = Loading = rclass
     displayName : 'Misc-Loading'
 
     propTypes :
-        style : rtypes.object
+        style  : rtypes.object
+        text   : rtypes.string
+
+    getInitialState : ->
+        text   : @props.text ? 'Loading...'
```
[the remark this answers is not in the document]
haraldschilly: well, I don't fully get that, but ok. So, if it is done with `getInitialState`, changing this props.text won't change the text?

Because I've mostly copied over what I saw in the file and when I scrolled down  I saw `getInitialState: ->  number : @props.number` in `NumberInput`. So, the idea here is, to derive the value of an input widget and make it robust against changes from the parent? If so, I think I understand it.

src/smc-webapp/project_store.coffee
```diff
@@ -1753,6 +1824,22 @@ create_project_store_def = (name, project_id) ->
         if @public_paths?
             return immutable.fromJS(misc.copy_without(x,['id','project_id']) for _,x of @public_paths.toJS())
 
+
+    library_docs_sorted: depends('library') ->
```
haraldschilly: @johnjeng I hope this is what you had in mind!

johnjeng: Yup, this looks great! I have no idea if the caching will have any impact, but this is what I have in mind for computed values.

williamstein:
@johnjeng  -- what's the review status of this?  It says "needs review" and you said "Yup, this looks great! " about something above.

src/smc-webapp/library.cjsx
```diff
@@ -0,0 +1,284 @@
...
+    copy: (doc) ->
+        @props.actions.set_library_copy_active(false)
+        doc = @props.library_selected
+        @props.actions.copy_from_library
+            src    : doc.get('src')
+            target : @target_path()
+            title  : doc.get('title')
+            docid  : doc.get('id')
+            start  : doc?.get('start') ? '/'
+            cb     : => @props.actions.set_library_copy_active(true)
```
johnjeng: I read `library_copy_active` as "The library is actively copying" but reading the places where the condition is used, it seems to mean "The copy button is active". I recommend changing the name to `library_is_copying` and swapping the truthiness check in relevant situations. 

It could be `library_copy_allowed` but I think that condition might rely on other data and is more likely to muck up the data semantics.

johnjeng:
I need to run manual tests as well.

johnjeng:
Code looks good to me. @DrXyzzy can you test this?

Issue #2503, opened 2017-11-18
edge case with rendering mathjax


haraldschilly:
chrome: `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/52.0.2743.116 Safari/537.36 Edge/15.15063`

smc_git_rev: f7fe4a24687b7f02ebda6fc09c751ce0fbecbe76

Date: 2017-11-18 09:28:54.392362

`Object doesn't support property or method 'mathjax'`

```
TypeError: Object doesn't support property or method 'mathjax'
   at _update_mathjax (https://cocalc.com/static/smc-72f23d32e9774e2d7912.cacheme.js?72f23d32e9774e2d7912:13:12322)
   at update_content (https://cocalc.com/static/smc-72f23d32e9774e2d7912.cacheme.js?72f23d32e9774e2d7912:13:12856)
   at componentDidMount (https://cocalc.com/static/smc-72f23d32e9774e2d7912.cacheme.js?72f23d32e9774e2d7912:13:13074)
   at Anonymous function (https://cocalc.com/static/smc-72f23d32e9774e2d7912.cacheme.js?72f23d32e9774e2d7912:109:13018)
   at commitLifeCycles (https://cocalc.com/static/smc-72f23d32e9774e2d7912.cacheme.js?72f23d32e9774e2d7912:116:7145)
   at r (https://cocalc.com/static/smc-72f23d32e9774e2d7912.cacheme.js?72f23d32e9774e2d7912:116:11262)
   at c (https://cocalc.com/static/smc-72f23d32e9774e2d7912.cacheme.js?72f23d32e9774e2d7912:116:12906)
   at u (https://cocalc.com/static/smc-72f23d32e9774e2d7912.cacheme.js?72f23d32e9774e2d7912:116:13423)
   at m (https://cocalc.com/static/smc-72f23d32e9774e2d7912.cacheme.js?72f23d32e9774e2d7912:116:15457)
   at f (https://cocalc.com/static/smc-72f23d32e9774e2d7912.cacheme.js?72f23d32e9774e2d7912:116:14958)
```

`_update_mathjax`  is in `smc-webapp/r_misc.cjsx`. 

```
    _update_mathjax: (cb) ->
        [...]
        if @props.has_mathjax
            $(ReactDOM.findDOMNode(@)).mathjax
                cb : () =>
                    ...
```

williamstein:
[...]

PR #2504, opened 2017-11-18
webapp: mathjax might not be available after all

haraldschilly:
I think this should take care of issue #2503


williamstein:
I have been refactoring the frontend code in support of share.   That is likely causing this bug.  The attached code is definitely not the right fix.   Mathjax is a jquery plugin that I wrote, and what is happening in 2503 is that for some reason it hasn't been initialized even though something is trying to use it.  The fix in this PR will only make things worse by randomly (according to a race condition) making mathjax just not work.

PR #2509, opened 2017-11-20
Jupyter landingpage

haraldschilly:


williamstein:
How does we review this?  You set it to "needs review"...

haraldschilly:
Well, main aspect is to review my language. I also found a webpack configuration issue, which is sort of harmless but could cause troubles upon some refactoring.

Testing: build the webapp and then go to `[base_url]/doc/jupyter-notebook.html`. 

I've assigned @DrXyzzy because this is mainly text.

DrXyzzy:
I like the new page a lot. Suggested changes in #2597.

Thank you for the testing instructions.

PR #2513, opened 2017-11-21
sagews: warn explicitly about wrong quotes

haraldschilly:


haraldschilly:
hmm, I'll try to merge your changes. Technically, these quote characters are different. Hence both of them need to be replaced.

However, I am working on another twist to that, too. I.e. to detect implicit multiplication ...

DrXyzzy:
I added a PR with 3 more small changes. https://github.com/sagemathinc/cocalc/pull/2516

haraldschilly:
ha, worked. your 2nd level PR is now part of this :dizzy:

PR #2516, opened 2017-11-22
Sagews warn quotes

DrXyzzy:
- 5(1+2) generates TypeError, not SyntaxError; do you want to catch these?
- another case added to end of pytest test_unicode.py
- unrelated, but found during testing: increase sagews pytest timeout because of slow scala launch


haraldschilly:
yes, let's also check for these type errors. It's quite common for newcomers to type in "1 + 2(3+x)" and so on

PR #2522, opened 2017-11-24
webapp and hub/share: same config for macros, etc.

haraldschilly:
This reuses the existing webapp mathjax configuration for the static renderer in shared.Not all settings can be used, though ,and well, no idea how to fix all the remaining erroneous attempts to render a formula in the codemirror boxes.

screenshot for the QQbar macro:

screenshot from 2017-11-24 15-29-00

### DIFF
--- a/src/smc-hub/share/mathjax-support.coffee
+++ b/src/smc-hub/share/mathjax-support.coffee
@@ -8,8 +8,24 @@ async   = require('async')
 console.log("loading mathjax-node...")
 mathjax = require('mathjax-node')
 
-# TODO: see
-# mathjax.config(...)
+# Deriving MyMathJaxConfig from the webapp's MathJaxConfig
+# It's the same, execpt for the entries that reference additional .js files
+# Otherwise the rendered page wouldn't even show up, so I assume there is no support for that
+{MathJaxConfig} = require('smc-util/mathjax-config')
+MathJaxConfig = require('smc-util/misc').deep_copy(MathJaxConfig)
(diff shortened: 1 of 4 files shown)


haraldschilly:
hmm, reading much of the code that renders sagews later, well, I don't see a way around pushing down mathjax rendering to be more specific to the part of the document. I saw your comment about being more targeted, yes, I agree. But ok, so, this change here doesn't touch much about that, at least in the way it is used right now. It mainly helps with these predefined macros.

williamstein:
> I don't see a way around pushing down mathjax rendering to be more specific to the part of the document.

Yep, there is 100% definitely not way (or reason to) want to get around that.  I just hadn't got to it yet.  With streaming rendering it is critical, of course.  Thanks for the config stuff, especially since it is done right!

src/smc-hub/share/mathjax-support.coffee
```diff
@@ -8,8 +8,24 @@ async   = require('async')
 console.log("loading mathjax-node...")
 mathjax = require('mathjax-node')
 
-# TODO: see
-# mathjax.config(...)
+# Deriving MyMathJaxConfig from the webapp's MathJaxConfig
+# It's the same, execpt for the entries that reference additional .js files
+# Otherwise the rendered page wouldn't even show up, so I assume there is no support for that
```
williamstein: This all happens on the backend in a jsdom that mathjax-node creates.  It's probably possible to have it load extensions, but I don't know.

PR #2523, opened 2017-11-24
webapp: switch TimeAgo to alternatively show absolute times

haraldschilly:
This bugged me for a while, too.

However, I'm wondering if there is a better way than this redux-double-wrapping of the element?

### DIFF
--- a/src/smc-webapp/r_account.cjsx
+++ b/src/smc-webapp/r_account.cjsx
@@ -1109,6 +1109,14 @@ OtherSettings = rclass
             Offer to setup the "First Steps" guide (if available).
         </Checkbox>
 
+    render_time_ago_absolute: ->
+        <Checkbox
+            checked  = {@props.other_settings.time_ago_absolute}
+            ref      = 'time_ago_absolute'
+            onChange = {(e)=>@on_change('time_ago_absolute', e.target.checked)}>
+            Display timestamps as absolut points in time – otherwise they are relative to the current time.
+        </Checkbox>
+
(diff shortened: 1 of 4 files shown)


src/smc-webapp/r_account.cjsx
```diff
@@ -1109,6 +1109,14 @@ OtherSettings = rclass
             Offer to setup the "First Steps" guide (if available).
         </Checkbox>
 
+    render_time_ago_absolute: ->
+        <Checkbox
+            checked  = {@props.other_settings.time_ago_absolute}
+            ref      = 'time_ago_absolute'
+            onChange = {(e)=>@on_change('time_ago_absolute', e.target.checked)}>
+            Display timestamps as absolut points in time – otherwise they are relative to the current time.
```
williamstein: Typo -- "absolut" -- "absolute".   Unless you're talking about "absolut vodka!"

williamstein:
This works for react components.  However, I'm not making this live unless:

1. The checkbox instructions also say "(This toggle does not work yet for some views, e.g., TimeTravel and task lists.)", or 
2. We make the jquery plugin that does timeago also obey this toggle.

The problem is is just that the react timeago is only used for things rewritten using react.

williamstein:
The sad part is that EXACTLY the places that Kiran Kedlay wants this functionality is precisely the parts where this doesn't work yet.

williamstein:
I'll merge this with (1) above, but keep open the issue.

PR #2524, opened 2017-11-24
webapp/sagews2html: process *all* mesg types, not just one

haraldschilly:
* [x] instead of picking only the first message type, it appends all of the encountered ones to a long html-string
* [x] some types are missing: clear, delete_last


williamstein:
I found this bug when testing this: https://github.com/sagemathinc/cocalc/issues/2560

This is not caused by this, but is there on the live site.

Issue #2530, opened 2017-11-29
correction to API docs example updating project upgrades


DrXyzzy:
Existing example at [query](https://cocalc.com/doc/api.html#query) API doc for "Add an upgrade to a project" is needlessly verbose, provides entire `upgrades` object to update `cpu_shares` attribute; only the attribute being updated is needed.

DrXyzzy:
merged

PR #2531, opened 2017-11-29
correction to api doc for updating project upgrades

DrXyzzy:
ref: #2530


haraldschilly:
ok, merging it. text only change brings no danger.

PR #2541, opened 2017-12-01
Hub loglines metric

haraldschilly:
This is sort of finished, but it definitely  needs testing (maybe some other uses of the hub are broken?) and some thoughts. E.g. what do we really want to log and at what level? Maybe it should simply be "debug" everywhere.

Technically, this updates to 2.4 (with a slightly different API) and instantiates winston loggers explicitly wherever they're configured. In particular, "winston" is then what the documentation calls "logger".

doc: https://github.com/winstonjs/winston/tree/2.x

Also, now that metrics transporter for winston works. I.e. here is then how it looks, a counter for each logger and the level.

```
# HELP cocalc_hub_log_lines_total counts the number of printed log lines
# TYPE cocalc_hub_log_lines_total counter
cocalc_hub_log_lines_total{name="hub",level="debug"} 275
cocalc_hub_log_lines_total{name="hub",level="info"} 2
cocalc_hub_log_lines_total{name="postgres",level="debug"} 1766
cocalc_hub_log_lines_total{name="support",level="debug"} 3
cocalc_hub_log_lines_total{name="hub_http_server",level="debug"} 20
```


williamstein:
I started deploying this on share server, thinking it should work, but it completely breaks that:
```
[prod.default] kucalc-prod-ctl-ws:~/kucalc/cluster2> c logs rnsrq
module.js:471
    throw err;
    ^
 
Error: Cannot find module 'colors/safe'
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.Module._load (module.js:417:25)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/cocalc/src/smc-hub/node_modules/winston/lib/winston/config.js:9:14)
```

Thus I'm guessing you didn't test this on ws1.   I'll try to fix this, or revert.

williamstein:
> Thus I'm guessing you didn't test this on ws1. I'll try to fix this, or revert.

Current guess -- colors is a module that needs to be updated when winston updates, and since the dependence isn't explicit it didn't happen.  So I'm trying a clean build from scratch of the hub image.  If that doesn't work, I'll be more explicit about the colors package ver.

williamstein:
Yep, complete rebuild fixed it.

Issue #2542, opened 2017-12-01
stacktrace: embedded cocalc in an iframe causes mathjax init exceptions


haraldschilly:
```
Uncaught DOMException: Blocked a frame with origin "https://cocalc.com" from accessing a cross-origin frame.
    at new ye.fn.init (https://cocalc.com/static/lib-a7e8cd08ee8c5f367412.cacheme.js?a7e8cd08ee8c5f367412:22:11429)
    at ye (https://cocalc.com/static/lib-a7e8cd08ee8c5f367412.cacheme.js?a7e8cd08ee8c5f367412:21:18997)
    at HTMLDocument.<anonymous> (https://cocalc.com/static/smc-a7e8cd08ee8c5f367412.cacheme.js?a7e8cd08ee8c5f367412:126:3450)
    at c (https://cocalc.com/static/lib-a7e8cd08ee8c5f367412.cacheme.js?a7e8cd08ee8c5f367412:22:15353)
    at u (https://cocalc.com/static/lib-a7e8cd08ee8c5f367412.cacheme.js?a7e8cd08ee8c5f367412:22:15668)
    at function.e._wrapper.e._wrapper (https://cocalc.com/static/lib-a7e8cd08ee8c5f367412.cacheme.js?a7e8cd08ee8c5f367412:61:10974)
    at https://cocalc.com/static/lib-a7e8cd08ee8c5f367412.cacheme.js?a7e8cd08ee8c5f367412:61:12494
```

haraldschilly:
well, the purpose of that line is still not clear to me, but this issue itself is fixed. hence closing.

PR #2543, opened 2017-12-01
webapp: hotfix for #2542

haraldschilly:
not sure what this does yet, but it breaks for iframe embedding.


haraldschilly:
I traced this line back to: https://github.com/sagemathinc/cocalc/commit/bc4daad7cddce139aeba511604b0e06fb8c649ff#diff-c6429b99ea286602fe43ed65b7efad81L281

I mean, this indicates it has something to do with responsive mode. But why and how? Maybe we can just delete `$(parent).trigger('initialize:frame')` ?

Issue #2552, opened 2017-12-06
in project settings there is a "Plain Jupyter Server" link -- make it work even the first time it is clicked


williamstein:
Right now, clicking the "Plain Jupyter Server" gives a 504 error the first time, due to how long it takes jupyter to start.  CHANGE this to work.  

Ideas:

- Make the proxy server return something besides 504 within 3s (?) if the jupyter server doesn't respond; that returned content would be a small html page that says "loading" and uses javascript (?) to refresh the page again in 3s (say).

haraldschilly:
I suggest we should do this in a general way, e.g. 

* add a minimal static page at the location `/static/loading.html`
* pass the URL you want to open as a parameter, e.g. `?target=<encodeURIComponent(url)>`
* write vanillaJS for loading the target via an ajax call, which retries it with some feedback, etc.

I've assigned this to me, because I'm pretty sure how to do the above.

williamstein:
A much simpler more general approach is to make a special react component:
```
<LinkOpenWhenWorking href={...}>Plain ...</LinkOpenWhenWorking>
```
that just does an async javascript request to a given url (up to a given time) in the browser until the result doesn't 504 (or whatever), and only then redirects the user (via javascript) to the target.  

This is much simpler, generic, and doesn't require any extra infrastructure (so would work in our docker image, etc.).

PR #2554, opened 2017-12-06
deferred loading of an URL

haraldschilly:
This is a simple proof-of-concept for issue #2552 

example for my dev project, note the `?`: `https://cocalc.com/14eed217-2d3c-4975-a381-b69edcb40e0e/port/56754/static/loading.html?https://cocalc.com/14eed217-2d3c-4975-a381-b69edcb40e0e/port/jupyter/`

what's definitely missing is  updating the URL

### DIFF
--- a/src/smc-webapp/project_settings.cjsx
+++ b/src/smc-webapp/project_settings.cjsx
@@ -44,6 +44,8 @@ misc                 = require('smc-util/misc')
 
 {AddCollaborators} = require('./collaborators/add-to-project')
 
+{COLORS} = require('smc-util/theme')
+
 URLBox = rclass
     displayName : 'URLBox'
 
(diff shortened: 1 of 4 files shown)


williamstein:
Cool!

I don't quite understand:

1. Why is it not `?target=https://cocalc.com/14eed217-2d3c-4975-a381-b69edcb40e0e/port/jupyter/` or something like that?

2. What does "what's definitely missing is updating the URL" mean?

Also, maybe don't hardcode the timeout interval `setTimeout((-> load_url(url)), 2 * 1000)`.  Instead, pass that as a GET param.   Different targets might benefit from different intervals.

haraldschilly:
Well, I didn't write a parser for the URL parameters, so far this is just for a test if this works at all.

> Updating the URL

the page for the jupyter notebook loads, but the URL is still the original one with the loading page and the query.

williamstein:
Closing since I addressed https://github.com/sagemathinc/cocalc/issues/2552 in a different way.

Issue #2533, opened 2017-11-29
improve stats table in r_help


haraldschilly:
the stats table on the landing page shows more data (in a maybe convoluted mess of digits?) while the one in r_help has less information (no file types, etc.)

they should match each other.

PR #2556, opened 2017-12-07
webapp/about: showing the opened file stats in the table

haraldschilly:
this is #2533

for testing in cc-in-cc,  make sure to run `hub --update_stats` to have recent data in the stats table.

changes:
- round numbers to two significant digits to avoid that pile of digits
- don't display count of distinct files
- add markdown files
- also record stats for lean, rnw and rtex files

about page in webapp:

screenshot from 2018-09-20 14-31-48

stats table on landing page, with some larger, random numbers

screenshot from 2018-09-20 14-34-34


williamstein:
Harald, I don't understand what "total (and unique) opened or edited" means.

- What is the definition of "opened"?  What is the definition of "edited"?  Are they different? 
- What is the definition of total?  What is the definition of unique?  How are the different?

I honestly really don't know.  I would suggest providing only one number who meaning is clear, if possible.   The one and only thing I think we really want to report is: 

   how many distinct documents where edited (=actually changed) during that period of time. 

To me, nothing else even makes much sense to report to users. 

I think all other numbers we report in the tables make sense.  E.g., "number of accounts / projects created" is very clear.   Number of projects modified is probably slightly wrong, but at least the phrase "Number of projects modified" is at least very clear, even if what we count isn't really exactly that.  However, I genuinely don't even know what "total (and unique) opened or edited" means, and I'm probably the one who wrote much of the code...

williamstein:
ping

haraldschilly:
> What is the definition of unique?

A "single file" is a path in a specific project. In a given period of time, that "single file" could appear once or more in the activity table. Unique means that it only counts once when it appears once or more.

PR #2559, opened 2017-12-07
Sharing 404s

johnjeng:
Do not show a link if the path is not public in the relevant project.

screen shot 2017-12-07 at 1 54 57 pm

# Testing
1. [x] Public file at root of project
1. [x] Public file nested inside some non-public folders
1. [x] Public folder at root of project with children
1. [x] Public folder nested inside some non-public folders
1. [x] Public file with a private parent but a public file in a different project has a public parent with the same name

My test server: https://cocalc.com/92234d52-8a1c-4e63-bde3-f2727f5ab8b1/port/2000/share/


src/smc-webapp/share/page.cjsx
```diff
@@ -103,13 +107,19 @@ CoCalcLogo = ({base_url}) ->
...
-    site_name  : rtypes.string
+    viewer       : rtypes.string
+    path         : rtypes.string # The share url. Must have a leading `/`. {base_url}/share{path}
+    project_id   : rtypes.string
+    base_url     : rtypes.string
+    site_name    : rtypes.string
+    public_paths : rtypes.immutable.Map
+
+TopBar = ({viewer, path, project_id, base_url, site_name, public_paths}) ->
+    list_of_paths = public_paths
```
williamstein: @johnjeng , what if there are 1 million public paths?  This linear search through all data does not scale.  Already, there are about 30K public paths, and having to do a linear search through all of them like you've done here is not reasonable.

We need to solve this problem differently, e.g, by maintaining another data structure that is built right when the public_paths synctable is constructed, and that gets updated whenever changes come in.

src/smc-webapp/share/page.cjsx
```diff
@@ -123,21 +133,23 @@ TopBar = ({viewer, path, project_id, base_url, site_name}) ->
...
+                segments.push(<span key={t}>{val}</span>)
             if not t
                 if path.slice(-1) == '/'
                     t = '..'
                 else
                     t = '.'
             else
                 t += '/..'
-        segments.reverse()
+        segments.reverse()g
```
williamstein: You've introduced a syntax error (I assume)  here with that g that got accidentally put at the end of the line?

src/smc-webapp/share/page.cjsx
johnjeng: Oh, in fact, you've already built this. Changing the PR to use that.

src/smc-webapp/share/page.cjsx
johnjeng: Yeah, typo.

PR #2563, opened 2017-12-10
sagews: ignore string literals for implicit multiplication

haraldschilly:
uses `strip_string_literals`

PR #2565, opened 2017-12-11
add test cases for false positive implicit mult error

DrXyzzy:
test cases for #2563


haraldschilly:
thank you, yes, that's a straightforward test

PR #2567, opened 2017-12-11
Email new collabs

haraldschilly:
in it's essence, this is a simple addition to inviting someone to a project. it shows the invitation template text and sends out the emails. that way, invited users know that they're invited.

implementation wise, this is a bit of a code duplication, but some aspects are slightly different from the case where a non-signed-up user is invited. so, I hope that's ok. we can clean that up later.

for testing, you need to make a second account and invite yourself to see the invitation. it also uses the same "invite" field in the project table as with all the other invites. reset it via `UPDATE projects SET invite = NULL WHERE project_id = '<UUID of your cc-in-cc dev project>';` in your cc-in-cc project.


src/smc-hub/client.coffee
```diff
@@ -1162,20 +1162,123 @@ class exports.Client extends EventEmitter
...
+
+                    # INFO: for testing this, you have to reset the invite field each time you sent yourself an invitation
+                    # in psql: UPDATE projects SET invite = NULL WHERE project_id = '<UUID of your cc-in-cc dev project>';
+                    @database.when_sent_project_invite
+                        project_id : mesg.project_id
+                        to         : locals.email_address
+                        cb         : (err, when_sent) =>
+                            if err
+                                cb(err)
+                            else if when_sent - 0 >= misc.days_ago(7) - 0 # successfully sent < one week ago -- don't again
```
williamstein: I think Date objects compare fine when using >= -- no need to convert lhs and rhs by subtracting 0.

src/smc-hub/client.coffee
```diff
@@ -1162,20 +1162,123 @@ class exports.Client extends EventEmitter
...
-                account_id : mesg.account_id
-                group      : 'collaborator'  # in future will be "invite_collaborator", once implemented
-                cb         : (err) =>
-                    if err
-                        @error_to_client(id:mesg.id, error:err)
+            async.series([
+                (cb) =>
+                    @database.add_user_to_project
+                        project_id   : mesg.project_id
+                        account_id   : mesg.account_id
```
williamstein: I'm suddenly really, realliy worried about what will happen with Minerva (say) when they use the API to dynamically add hundreds of students to 20 projects for a session.    It seems like whether or not to send an email should be an option in the invite mesg.  The option is off by default for the API (or on, but we tell minerva to turn it off), but on for the browser client.  What do you think?

williamstein:
This looks really good!  I have two minor comments.  It would be good to merge this tomorrow or saturday.

src/smc-hub/client.coffee
haraldschilly: yes, I didn't think about that. It should be off by default for the API.

haraldschilly:
I have added an idea how to deal with this, but I haven't tested it again. Will report back when done.

haraldschilly:
hmm, I think it's ok, but I tried to run these `testapi` tests but they fail. Maybe just a timeout problem... The point is, there is one for the `invite_collaborator` endpoint to be sure this is backwards compatible. I should probably add a second `invite_collaborator` test with these additional message fields for sending the invite emails

haraldschilly:
Short summary:

* If there is no `mesg.email`, no email is sent [see this change here](https://github.com/sagemathinc/cocalc/pull/2567/commits/2ba2889c33f9950ecb59a70a4805dc94d59cca58#diff-bcb26f5549425a95ddd26d95f357b183R1185) ... that way, there is no email sent in the API-call case 
* For testing, I am using sinon spy stubs to track when the `send_email` function is called ... not all cases are covered, though, e.g. one that's missing is inviting someone again shouldn't trigger a second invite email.

williamstein:
So is this potentially ready to be merged?

haraldschilly:
Short update, since this is still open. The functionality itself looks ok to me, but someone should manually test it besides me.

My main struggle with all this is to also write a unit test for it. Today I spent more time on figuring out what's wrong, but I don't get it. My attempts went even so far, that `smc-hub/compute-server.coffee` complains about unknown arguments. I.e. the ones defined in `smc-hub/test/mocha.opts`, e.g. `--reporter` and so on. Mocking them as additional arguments in `compute-server.coffee` makes them go away, but that sounds like a terrible workaround. Also, as it is committed right now when I wrote the message, it doesn't show up and is just stuck somewhere in a callback. 

To spare someone having to check the other tests, this is the line I used to pick out just one test  (tun it in the `smc-hub` directory)

`SMC_DB_RESET=true SMC_TEST=true node_modules/.bin/mocha  --report verbose test/api/account.coffee --only-async -g "inivtes a collaborator to a project via an email message"`

williamstein:
We should launch this functionality this week, no matter what.  

What's the worse that can realistically happen?   Serious question.

haraldschilly:
hmm, invites do not work and api causes hub crashes?

however, I've deployed this on ws1. The hub logs did help me to find a missing callback (which doesn't solve the testing-issues, but still important). I also tried the old api-invites (not sending the email) and newer ones with more arguments:

```
curl -u ...: \
    -d account_id=... \
    -d project_id=... \
    -d title=test111 \
    -d email='test2 email body' \
    -d subject='test2 subject' \
    -d link2proj='https://ws1.cocalc.com/test/test/test' \
    -d replyto='reply@to.com' \
    -d replyto_name="reply to" \
    https://ws1.cocalc.com/api/v1/invite_collaborator
```

does work as expected.

williamstein:
I'm going to just merge and release this today.

PR #2583, opened 2017-12-15
Simplify share rendering

johnjeng:
>[ws] A concern I just had. All editors in normal Cocalc assume they are in a fixed height div with flex column layout. That's basically why the share server is that way -- so they can keep that assumpion. Just a thought.

I think this is mostly for scroll position reasons. But on the share server, the whole page scrolls. Anyways, for the implemented editors so far it doesn't seem to be an issue.


williamstein:
Ready for review/merge?

Issue #2589, opened 2017-12-18
make sure sage-8.1 runs correctly as default sage in cocalc


DrXyzzy:
I have pytest changes, will add PR for that for this issue.

@haraldschilly  - `pluggy` Python package is not available in `sage -sh` environment in a .term. Can that be added? I've been using it for pytest + doctest.

haraldschilly:
Ok, consider this done on ws1. I also made 8.1 the default "sage" there.

This is what I get when I start the tests in the smc_sagews subdirectory. Note, there is `pluggy-0.6.0`.

```
(sage-sh) user@project-844369fc-b302-4b77-8968-ea16f8a74d73:smc_sagews$ pytest
========================================================================================= test session starts ==========================================================================================
platform linux2 -- Python 2.7.14, pytest-3.3.1, py-1.5.2, pluggy-0.6.0
rootdir: /home/user/cocalc/src/smc_sagews, inifile:
plugins: variables-1.7.0, timeout-1.2.1, repeat-0.4.1, param-0.1.1, mock-1.6.3, base-url-1.4.1
collected 162 items                                                                                                                                                                                    
 
test_direct/test_jupyter.py .....
```

PR #2590, opened 2017-12-18
Issue/2589

DrXyzzy:
Ref: #2589 

sagews pytest changes for sage-8.1.

with this patch, all tests pass or are skipped with sage-8.0 and sage-8.1:
```
python -m pytest
```

Will remove skip julia and bash jupyter kernels when issues with those are resolved.


williamstein:
Seems safe to merge.   Should we open a ticket for re-enabling tests of julia/bash/etc.?

PR #2593, opened 2017-12-19
Add receipt description

johnjeng:
Makes room for an invoice description on receipts.


williamstein:
Nice. I won't make this live immediately because I would have to restart the hubs, which conflicts with some major database maintenance I'm doing...

PR #2597, opened 2017-12-21
minor wording changes

DrXyzzy:
among other things, I didn't capitalize "notebook" unless it was in a heading or title


haraldschilly:
Ok, this is all fine with me. However, when you do "git grep Notebook" in the smc-webapp, you'll see that there are other cases where it is uppercase. So, maybe the question is if someone refers to it as "the product" or if it is used to mention the notebooks as "the files".

PR #2633, opened 2018-01-12
webapp: log project stop/restart requests by a user

haraldschilly:
this isn't finished, but could still be merged right now. it only logs the *requests* to stop/restart a project. more work needs to be done to then actually log the "stop" by the backend. the only tiny complication is that the backend has no `account_id`. hence, the front-end part needs to deal with that, too.

### DIFF
--- a/src/smc-webapp/project_log.cjsx
+++ b/src/smc-webapp/project_log.cjsx
@@ -119,8 +119,13 @@ LogEntry = rclass
         </span>
 
     render_start_project: ->
-        <span>started this project {@render_took()}
-        </span>
+        <span>started this project {@render_took()}</span>
+
+    render_project_restart_requested: ->
+        <span>requested to restart this project</span>
+
+    render_project_stop_requested: ->
+        <span>requested to stop this project</span>
 
     render_miniterm_command: (cmd) ->
         if cmd.length > 50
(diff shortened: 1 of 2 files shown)


haraldschilly:
Second commit deals with the case where there is no `account_id`. It assumes it is some sort of system process.

Attached a screenshot, where these two (synthetically created) project_log entries appear. Later, the back-end will create events like that.

```
smc=# select * from project_log  WHERE account_id IS NULL ORDER BY time DESC limit 3;
-[ RECORD 1 ]----------------------------------------------------
id         | 529dd503-6d7a-4a22-ae14-daf3ef0fec28
project_id | bc6f81b3-25ad-4d58-ae4a-65649fae4fa5
time       | 2018-01-12 16:44:32.939065
account_id |
event      | {"by": "manage-project", "event": "project_stopped"}
-[ RECORD 2 ]----------------------------------------------------
id         | 529dd503-6d7a-4a22-ae14-ddf3ef0fec28
project_id | bc6f81b3-25ad-4d58-ae4a-65649fae4fa5
time       | 2018-01-12 16:33:49.118997
account_id |
event      | {"event": "project_stopped"}
```

screenshot from 2018-01-12 18-02-38

williamstein:
> "System service manage-project stopped this project..." 

does not make sense to our users.  We need to communicate in consistent clear language, not technical jargon for ourselves.    E.g., 

> Project stopped after [30 minutes with no user interaction](link to wiki page about idle timeout and how to pay to increase it).

williamstein:
Also, precisely how is recording this going to be *useful*.  How will it improve cocalc for users or us?  

Knowing how *long* stopping/starting/restarting are perceived to take by users *is* useful, since we can use it to improvethe user experience.  Just knowing that the user clicked the stop button and nothing more... isn't as useful.

williamstein:
I'll merge, just because if nothing else it'll nice to see these in the log... Can't hurt.

PR #2643, opened 2018-01-17
mocha mods

DrXyzzy:
- export `_internals` from src/smc-webapp/smc-react.coffee for use by smc-react-test.coffee
- increase timeouts in src/smc-project/jupyter/test/supported-kernels.coffee for slow-starting kernels (some of these run much faster on second test run immediately after first test)
- modify package.json for easier debugging
  - following precedent in smc-hub, split multi-directory tests in smc-project and smc-webapp
  - more use of --reporter ${REPORTER:-progress}
  - add ${BAIL} to a couple tests; allow `BAIL=-b npm run testapi` etc

### DIFF
--- a/src/smc-webapp/smc-react.coffee
+++ b/src/smc-webapp/smc-react.coffee
@@ -592,7 +592,7 @@ exports.ReactDOM = require('react-dom')
 if DEBUG
     smc?.redux = redux  # for convenience in the browser (mainly for debugging)
 
-_internals =
+exports._internals =
         AppRedux                 : AppRedux
         harvest_import_functions : harvest_import_functions
         harvest_own_functions    : harvest_own_functions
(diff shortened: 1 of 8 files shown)


src/smc-webapp/smc-react.coffee
```diff
@@ -592,7 +592,7 @@ exports.ReactDOM = require('react-dom')
 if DEBUG
     smc?.redux = redux  # for convenience in the browser (mainly for debugging)
 
-_internals =
+exports._internals =
```
haraldschilly: I understand this is only used in https://github.com/sagemathinc/cocalc/blob/9fa215ca64f6ec824e807ac21fbef8bf5bb509c5/src/smc-webapp/test/smc-react-test.coffee#L2 

and well, otherwise it won't work. 

I don't think there are any negative implications exporting it. Still, in case there is paranoia about this, we could conditionally do it via `if process?.env.SMC_TEST == 'true'` …

DrXyzzy: done

DrXyzzy:
conditional added so `_internals` only exported when` SMC_TEST` is defined

DrXyzzy:
also added src/smc-hub/test/README.md to provide reminder about test db setup

PR #2654, opened 2018-01-20
Issue/2646

DrXyzzy:
If I understand the `reset()` code in sage_salvus.py starting around line 3100, this PR is consistent with the intent.

Also added pytest for the issue, which fails without the patch and succeeds with it.

With patch applied, all pytest tests pass.
```
~/cocalc/src/smc_sagews/smc_sagews$ python -m pytest
...
= 142 passed, 5 skipped in 524.71 seconds =
```


williamstein:
looks good!

Issue #2534, opened 2017-11-29
latex preview lists only an incomplete list of preview pages


haraldschilly:
This is a bit hard to explain, but it is definitely a bug drastically reducing the overall usability:

Consider you have a multi page long latex document (e.g. 20 for a presentation). Then, when it renders the image-preview, it is supposed to show all 20 pages and renders the visible ones with a low resolution image and then replaces them with a high resolution image. 

The problem is, sometimes it only loads the first 3 pages. The remaining 17 aren't there, so you can't scroll down. Interestingly, after 1-2 secs, the 4th page shows up (in full resolution), then, after scrolling down again, the 5th, and so on.

Toggling the "build" preview or closing&reopening the tab could fix this. Also, sometimes I wasn't even able to replicate it. Hence "heisenbug".

It's also rather new, because I know for sure that 3 or 4 weeks ago this did work all fine. 

my suspicion is that the underlying command for computing the total number of pages has changed, or there is some issue with updating some status.

haraldschilly:
some thoughts

* the pdfinfo command is alright
* the preview window is initialized twice, which isn't good. i.e. LatexEditor → init PNG_Preview. and then shortly afterwards, again LatexEditor → init PNG_Preview.
* that's why I think the updates no longer really match, e.g. it changes from num_pages 20 to 0 again, etc.

also, once the document is edited, the compilation goes through and in the end triggers an update, it number of pages is fine again. 

so, why is there this duplicate initialization?

williamstein:
We should rewrite this damned editor in React.

williamstein:
I did rewrite it in react, and also react-virtuoso.  I'm sure this is long since fixed.

PR #2658, opened 2018-01-24
webapp/latex: timing issue when opening a tex file that does not need to be compiled again

haraldschilly:
see issue #2534, this did bug me again and again when testing in the past few days.

In the end, all this does is delay the update rendering by 1 second after the syncdoc says that it is connected. I suspect this bug started when opening files was speeded up. I am also aware that this is not a perfect fix, but let's postpone this after a rewrite...

Testing: I checked with a normal document and upon Rnw documents.

Here is a document with a couple of pages for testing. The way to reproduce the problem in prod is to open it, wait until compiled, close the tab, wait a few secs, then open the tab again. The .tex file is saved while closed and there is no need to re-render it when opening. And without that additional rendering, it is somehow not properly initialized. That's why it's a "heisenbug".

```
\documentclass{article}
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{lmodern}
\usepackage[english]{babel}
\usepackage{blindtext}
\begin{document}
\Blinddocument[1]
\end{document}
```

### DIFF
--- a/src/smc-webapp/latex/editor.coffee
+++ b/src/smc-webapp/latex/editor.coffee
@@ -91,13 +91,13 @@ class exports.LatexEditor extends editor.FileEditor
             saved_cb = =>
                 if @_current_page == 'pdf-preview'
                     @preview_embed?.update()
-            @update_preview(saved_cb)
+            @update_preview(cb:saved_cb)
             @spell_check()
 
         @latex_editor.syncdoc.on 'connect', () =>
             #if DEBUG then console.log("LatexEditor. syncdoc.on 'connect'")
             @preview.zoom_width = @load_conf().zoom_width ? 100
-            setTimeout((=> @update_preview(undefined, false, false)), 1000)
+            setTimeout((=> @update_preview()), 1000)
             @spell_check()
 
         @latex_editor.print = () =>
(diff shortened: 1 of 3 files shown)


src/smc-webapp/latex/editor.coffee
```diff
@@ -513,16 +515,20 @@ class exports.LatexEditor extends editor.FileEditor
...
             if not err
-                @update_preview (force=force) =>
+                sf =  =>
                     if @_current_page == 'pdf-preview'
                         @preview_embed?.update()
+                @update_preview(sf, force)
                 @spell_check()
 
-    update_preview: (cb, force=false, only_compile=false) =>
+    update_preview: (cb, force, only_compile) =>
```
williamstein: for what it is worth:

(1) I think what you had originally does work fine in CS.  What doesn't work is calling it like `@update_preview(cb, only_compile=true)`.

(2) I would GREATLY prefer is you use defaults/opts here.  It's used about 10K other places in our codebase, solves precisely this problem, leads to very clear code, etc. Please use that rather than `force ?=` etc.  It's easy.

williamstein:
>  I suspect this bug started when opening files was speeded up.

The changes I made yesterday to the hubs and the protocol negotiation when users first connect could also siginificantly speed up initial connect/reconnect/load, which could also lead to such race conditions.  🤷‍♂️

haraldschilly:
ok, hmm, seems like it drops the key altogether.

PR #2664, opened 2018-01-26
webapp/course: "overwrite & delete" for copying over files to students

haraldschilly:
see #1483

caveat:  this isn't solving #1483 completely, but at least the aspect of sending out the files using rsync using the delete mode is being triggered.

test: I checked various cases I could think of: student having other files (should be gone), having edits (should be reverted in the file), editing the file right now (replaced with new content), etc.


williamstein:
Hold on!?  Reading this code, it looks to me like you are adding a new button right next to an existing button, where if you just accidentally click on the new button (by clicking a few pixels off from where you should), then you immediately delete all your students work with no backups!  And this is basically 100% to prevent the situation where your students accidentally got a copy of an answer key (this does cannot solve that problem, due to filesystem snapshots.  Once you push files out, the students have them forever. )

Sorry, but this makes me very nervous.  

At a bare minimum, after they click the button, make them type a random word or something... (like we do with account deletion).   I'm also a little worried about cluttering the UI.

williamstein:
@haraldschilly  -- can you resolve the merge conflict.  It's with another thing you did, so will be easier for you....

haraldschilly:
yes!

haraldschilly:
ok, this was indeed broken and in conflict with that due date only-write-once fix

haraldschilly:
@johnjeng I think the "core" aspect in understanding what this does here is to look [at this function call](https://github.com/sagemathinc/cocalc/pull/2664/files#diff-a56929e572f2c11ad7e9837fe52be041R1234). All the other stuff here is for the UI. ...  and well, if you really want to go down the rabbit hole of the client API at that point, you should resurface at parameter settings for rsync. (there are two exit holes: one for "normal" and one for "kucalc")

PR #2677, opened 2018-02-01
course assignment progress indicators

haraldschilly:
The goal here is to make progress information about the assignment steps more visible to the teacher. The status string is similar to how it looks like for handout distributions and the button is green when all are done. The following screenshot shows assignment steps in various progress states.

cocalc-course-progress-info

### DIFF
--- a/src/smc-webapp/course/store.coffee
+++ b/src/smc-webapp/course/store.coffee
@@ -274,7 +274,7 @@ exports.CourseStore = class CourseStore extends Store
         return x.get('time')
 
     has_grade: (assignment, student_id) =>
-        return @get_assignment(assignment)?.get("grades")?.get(student_id)
+        return !!assignment.get("grades")?.get(student_id)
 
     get_assignment_status: (assignment) =>
         #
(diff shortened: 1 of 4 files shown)


src/smc-webapp/course/store.coffee
```diff
@@ -274,7 +274,7 @@ exports.CourseStore = class CourseStore extends Store
         return x.get('time')
 
     has_grade: (assignment, student_id) =>
-        return @get_assignment(assignment)?.get("grades")?.get(student_id)
+        return !!assignment.get("grades")?.get(student_id)
```
haraldschilly: oddly enough, this was always wrong, because `assignment` is here already the assignment object (not a key). Given it's name, it should also be a boolean.

williamstein: Sorry, but change this back!  

It's not broken, it's not wrong, and it is not always wrong.  Everything that takes a var called "assignment" is designed to also work with just an `assignment_id` as input instead.  That's the reason for `@get_assignment` -- it just returns it if already an assignment, otherwise fetches the assignment from the id.

williamstein:
I'll merge and fix, since I need to work on this code, and don't want massive merge conflicts.

PR #2687, opened 2018-02-05
Use Katex (except in Sagews)

johnjeng:
Here's the branch for switching to Katex.
It should be easy to understand and adjust.

Tested in Jupyter, Chat, and our markdown editor.


haraldschilly:
Hmm, I briefly played around with this. Yes, we should submit a PR for adding the macro option [right here](https://github.com/Rokt33r/remark-math/blob/master/packages/rehype-katex/index.js#L47) ... unless I miss something, I think that's the spot in their code.

Regarding macros, we can easily reuse what we have configured right now, e.g. something like

```
{MathJaxConfig} = require('smc-util/mathjax-config')
macros = {}
for k, v of MathJaxConfig.TeX.Macros
    # only those without arguments, until KaTeX knows how to deal with them
    if v[1] == 0
        macros["\\#{k}"] = v[0]
```

and then

```
.use(katex, {macros: macros})
```

johnjeng:
Yeah, it seems fairly trivial.

johnjeng:
Made a pull request to add macros. https://github.com/Rokt33r/remark-math/pull/19

williamstein:
Unfortunately, as @DrXyzzy found, this PR can't go forward without implementing a full fallback to mathjax mode, and (thus) switching back to an async API :-(

Discussed privately [here](https://cocalc.com/projects/369491f1-9b8a-431c-8cd0-150dd15f7b11/files/work/2018-03-07.sage-chat?session=default).

johnjeng:
Okay, this PR now leaves the share server unaltered essentially. The frontend though gets many formulas rendered much faster using katex. For the rest, katex will throw an error to the console (we get these right?) and mathjax will render that instead.

williamstein:
>  For the rest, katex will throw an error to the console (we get these right?) and mathjax will render that instead.

Please clarify --  Do you really want every single attempt to show a formula that katex can't render by mathjax can to:
 
(1) liter the console.log,
(2) put an entry in our database?

Why?  Everybody is very aware that there are a lot of things that mathjax can do that katex can't, so what will (1) and (2) actually accomplish?   If nothing, do not log anything.  If something, what?

williamstein:
You add a number of packages, e.g., `markdown-it`, which have dependencies presumably.    Did you check that they don't increase the size much of our final bundle?  It's a good idea to check that every once in a while when adding a bunch of new packages...

williamstein:
> Okay, this PR now leaves the share server unaltered essentially. 

I'm curious -- are you planning to improve the share server to use katex at some point, or just giving up on that?

johnjeng:
>so what will (1) and (2) actually accomplish?

Let us know how much of user's math is unrenderable by katex. Then we can decide in the future to do a switch if it's low enough. Unless we just always want to always keep both mathjax and katex. 

> `markdown-it`

I have not been checking. I should do that.

>are you planning to improve the share server to use katex at some point

Well, I have an idea of a naive way to do this and no ideas beyond that. I'm guessing that katex is fast enough that it will be worth it. I'll append that to this PR.

williamstein:
> Unless we just always want to always keep both mathjax and katex.

I can't see any situation in which we switch, unfortunately, since there is a ton that katex doesn't support.   When I looked at this when you asked, I was very much misled into thinking katex was much closer to latex than it actually is.   (Also, we could get the same info by just directly scan documents, rather than checking render error logs...)

williamstein:
For me in Docker the build fails with some katex font issues... (?):
```
[1625] ^[[1m./smc-webapp/entry-point.coffee^[[39m^[[22m 2 kB {^[[1m^[[33m3^[[39m^[[22m}^[[1m^[[32m [built]^[[39m^[[22m
    + 1661 hidden modules
 
^[[1m^[[31mERROR in ./smc-webapp/node_modules/katex/dist/fonts/KaTeX_Math-Italic.ttf
Module build failed: Error: Cannot find module 'file-loader'
    at Function.Module._resolveFilename (module.js:538:15)
    at Function.Module._load (module.js:468:25)
    at Module.require (module.js:587:17)
    at require (internal/module.js:11:18)
    at Object.module.exports (/cocalc/src/node_modules/url-loader/index.js:35:24)
 @ ./node_modules/css-loader?{"minimize":true,"discardComments":{"removeAll":true},"mergeLonghand":true,"sourceMap":false}!./smc-webapp/node_modules/katex/dist/katex.min.css 6:2937-2977
 @ ./smc-webapp/node_modules/katex/dist/katex.min.css
 @ ./webapp-css.coffee^[[39m^[[22m
 
^[[1m^[[31mERROR in ./webapp-lib/fonts/droid/DroidSansMono.ttf
Module build failed: Error: Cannot find module 'file-loader'
    at Function.Module._resolveFilename (module.js:538:15)
    at Function.Module._load (module.js:468:25)
    at Module.require (module.js:587:17)
    at require (internal/module.js:11:18)
    at Object.module.exports (/cocalc/src/node_modules/url-loader/index.js:35:24)
 @ ./node_modules/css-loader!./node_modules/sass-loader/lib/loader.js?{"minimize":true,"discardComments":{"removeAll":true},"mergeLonghand":true,"sourceMap":false}&indentedSyntax!./smc-webapp/index.sass 6:163-217
 @ ./smc-webapp/index.sass
 @ ./webapp-smc.coffee^[[39m^[[22m
...
```

It does continue for a long time, but this alraedy seems bad.

williamstein:
Build log is here: https://cocalc.com/projects/f8cf98ed-299e-4423-a167-870e8658e081/files/ws1/cluster2/build.log?session=default

src/smc-webapp/r_misc.cjsx
```diff
@@ -921,16 +908,21 @@ exports.HTML = rclass
...
+                    {html, is_complete} = math_katex.render(html)
+                    @_needs_mathjax = not is_complete
+
             {__html: html}
         else
             {__html: ''}
 
 
     render: ->
-        # the random key is the whole span (hence the html) does get rendered whenever
```
williamstein: Why did you delete this -- I even left a comment about why it's needed... ;-)

johnjeng: I somehow read the next line as "Otherwise does NOT render when the value changes" and I tested that lol

Issue #2694, opened 2018-02-09
file not distributed to all students when folder is assigned


DrXyzzy:
Reported by: Brian Brady

A folder with two Jupyter notebooks is assigned, with due date, in a course with 54 students. All student projects receive the two notebooks, but 7 projects do not receive the DUE_DATE.txt file.

PR #2700, opened 2018-02-10
webapp/course: when assigning to students, create the due date file only once

haraldschilly:
The idea here is to write this due-date file only once when copying out an assignment. That's why there is a new `copy_assignment_create_due_date_file` and it is used in two places. Either for single student distributions, or when iterating over all students. It also checks if it was written correctly, which is motivated by #2694.

### DIFF
--- a/src/smc-webapp/course/actions.coffee
+++ b/src/smc-webapp/course/actions.coffee
@@ -1180,7 +1180,7 @@ exports.CourseActions = class CourseActions extends Actions
     #    assignment.last_assignment[student_id] = {time:?, error:err}
     #
     # where time >= now is the current time in milliseconds.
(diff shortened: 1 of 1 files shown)


williamstein:
if this works, positive review.

src/smc-webapp/course/actions.coffee
```diff
@@ -1244,13 +1239,64 @@ exports.CourseActions = class CourseActions extends Actions
...
+                    project_id : locals.project_id
+                    path       : locals.path
+                    content    : locals.content
+                    cb         : (err) =>
+                        if err
+                            @clear_activity(locals.due_id)
+                            cb(err)
+                        else
+                            cb()
+            (cb) =>
```
williamstein: any evidence ever that this can possibly fail?

PR #2708, opened 2018-02-15
available_upgrades is api response, not a query

DrXyzzy:
use `APi message[2]` for api endpoints, omit `API` for internal messages


williamstein:
Cool -- thanks for fixing this!  I obviously got confused/messed up.

PR #2716, opened 2018-02-21
add wiki link to sharing text

DrXyzzy:
sharing-files


williamstein:
Should we delete the help button?  It links to the same place.  Thoughts?

DrXyzzy:
Agree. Added a commit to remove help button.

PR #2723, opened 2018-03-01
webapp/support dialog: debouncing props updates to fix performance issue

haraldschilly:
There was a comment about a performance problem when typing into the support dialog. I can confirm this when the course editor is. In particular, I have a test course with about 50 students. When I open up an assignment and disclose the list of students (with all the buttons, etc.) and then click help for the dialog, I can't really type in. It lags a few seconds, and even more, when holding down a key in its pressed state, this doesn't really produce a sequence of characters... it just halts.

The change here is to use the state of the dialog and debounce updates to the "props". 

Alternative theory: this here is just a band aid against a another bug how courses render...

### DIFF
--- a/src/smc-webapp/support.cjsx
+++ b/src/smc-webapp/support.cjsx
@@ -1,4 +1,4 @@
-###############################################################################
+##############################################################################
 #
 #    CoCalc: Collaborative Calculation in the Cloud
 #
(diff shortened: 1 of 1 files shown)


williamstein:
> Alternative theory: this here is just a band aid against a another bug how courses render...

You're right that this is almost certainly totally the wrong approach to solving this problem.  All such problems get solved properly by (1) enabling profiling in smc-webapp/react (?), and (2) adding an appropriate `shouldComponentUpdate`.   

There are other examples of bad code like this -- e.g., when typing in a large chat it is noticeable.  It always happens when people write react code, but don't fully understand the "render every time something changes" model.... and even when they do, but don't profile the code.

Now-a-days I simply write an explicit shouldComponentUpdate with EVERY component I ever write -- see my code in tasks/.  It's very explicit, but addresses such problems usually.

It's also possible there's a mistake with using controlled versus not-controlled components and where state is stored, which can cause a lot of waste...

williamstein:
More precisely -- to do this right (or all react perf stuff):

Change COUNT=true on line 541 of `smc-webapp/smc-react.coffee`. 

Do stuff.

Type `get_render_count()` and `reset_render_count()` in terminal...

I'll do this right in this case.

williamstein:
Closing this instead of merging.

PR #2724, opened 2018-03-02
Fix mobile ui

johnjeng:
screen shot 2018-03-02 at 3 22 04 pm
Adjusts some styles so that the dropdown menu on mobile isn't totally broken.


williamstein:
ready?

PR #2727, opened 2018-03-05
webapp/course: update state of form box upon changes in supplied props data

haraldschilly:
This fixes stale information of the "grade comment" for a student in an assignment.

Test: open the same course assignment twice in two tabs and edit a comment of a student grade. After saving in on of them, the changes should show up in the other instance, too.

### DIFF
--- a/src/smc-webapp/course/common.cjsx
+++ b/src/smc-webapp/course/common.cjsx
@@ -138,6 +138,12 @@ exports.StudentAssignmentInfo = rclass
         edited_grade    : @props.grade ? ''
         edited_comments : @props.comments ? ''
 
+    componentWillReceiveProps: (nextProps) ->
+        @setState(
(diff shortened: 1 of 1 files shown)


src/smc-webapp/course/common.cjsx
```diff
@@ -138,6 +138,12 @@ exports.StudentAssignmentInfo = rclass
         edited_grade    : @props.grade ? ''
         edited_comments : @props.comments ? ''
 
+    componentWillReceiveProps: (nextProps) ->
+        @setState(
+            edited_grade    : nextProps.grade ? ''
+            edited_comments : nextProps.comments ? ''
```
williamstein: you could do a 3-way merge here.

williamstein:
Save state when component will be unmounted.

PR #2730, opened 2018-03-06
Mimic other coupon-promo adders

johnjeng:
Mimic other coupon adders as suggested.

In particular, 
- The button now says `Apply`
- The button is closer to the coupon text. 
- It lights up when you put anything in the box
- It submits onBlur

screen shot 2018-03-05 at 6 12 17 pm
screen shot 2018-03-05 at 6 12 35 pm
screen shot 2018-03-05 at 6 12 43 pm


johnjeng:
Also just added a more vivid confirmation. I don't see a downside.

screen shot 2018-03-05 at 6 23 11 pm

Issue #2728, opened 2018-03-05
duplicate file shows trailing slash in log


williamstein:
1. Duplicate a file.
2. See this: `William Stein copied work/2018-03-05-all-hands.md to work/2018-03-05-all-hands-1.md/ 3 minutes ago`

haraldschilly:
maybe [this commit](https://github.com/sagemathinc/cocalc/commit/a2d29d5784b5908ae2e1511cf5439fc2c5403047) is relevant.

PR #2735, opened 2018-03-08
webapp: special case for logging duplicating files as a copy operation

haraldschilly:
See #2728 

testing is easy: copy one or more files to a directory vs. duplicating a file.


williamstein:
Merged, but still not perfect -- for a directory it now does not include a trailing slash.  Still works though.

screenshot 2018-03-08 at 9 50 18 am

PR #2742, opened 2018-03-11
React verbose debugging

haraldschilly:
this adds more info to the verbose debugging, at least it helps me to figure out what exactly is being rendered and how often.

The change in "misc" is a bit unrelated, but I noticed that `is_different_array` doesn't take into account if both arguments are undefined. My stance is that this is then the same array (and shouldn't trigger render updates)


williamstein:
It would probably be better to just replace is_different_arrays by the 1-liner: 
```
not underscore.isEqual(a,b)
```

src/smc-webapp/smc-react.coffee
```diff
@@ -557,8 +557,9 @@ react_component = (x) ->
 MODE = 'default'  # one of 'default', 'count', 'verbose', 'time'
 #MODE = 'verbose'
 #MODE = 'count'
-if not smc?
+if not window.smc?
```
williamstein: This will break importing everywhere that window isn't defined (so surely in the local-hub project, which imports this code for backend jupyter).  Either don't do this (for no reason?) or do `window?.smc?`.

src/smc-webapp/smc-react.coffee
```diff
@@ -587,11 +588,35 @@ switch MODE
                 console.log r.displayName, "took", t1 - t0, "ms of time"
             return r
     when 'verbose'
+        renderings   = []
+        time_window  = 10 * 1000 # we take the last ~10 seconds into account
```
williamstein: could you make it `time_window_ms`; I'm tired of being bit by units.

src/smc-webapp/smc-react.coffee
```diff
@@ -587,11 +588,35 @@ switch MODE
...
         rclass = (x) ->
             x._render = x.render
             x.render = () ->
-                console.log x.displayName
-                return @_render()
+                # measure time
+                t0 = performance.now()
```
williamstein: This is much, much more complicated than any of the other profiling code I added.   Can you put it in another file?  Also, make sure it doesn't just reproduce https://reactjs.org/docs/perf.html ?

williamstein:
It looks like you just completely deleted the "verbose" mode that I wrote, and replaced it with a different verbose mode that does something completely different (e.g., not printing anything in some cases).  I find the verbose mode I wrote very useful -- please don't kill it.    If you can't think of a better name for yours, can you call it 'verbose2', 'filtered-verbose' or something, and leave the previous one too?

Issue #2714, opened 2018-02-19
in a course, selecting "skip assignment" should make it so collect all is enabled.


williamstein:
This screenshot shows what currently happens, which is of course not very useful since there is no "collect all" button:

![screenshot 2018-02-19 at 7 34 43 am](https://user-images.githubusercontent.com/1276278/36385477-88f2b484-1547-11e8-804c-71443aee8388.png)

johnjeng:
Fixed by #2746.

PR #2746, opened 2018-03-13
Fix #2432 and #2714

### LINKED ISSUE #2432: failed assignments in a course should not be in the state "done"
When an assignment fails due to an error, it makes little sense to let the UI show it as if the files are copied. 

That makes reassigning the failed ones hard, because the UI thinks all are done. Also, the collect button would collect from nowhere...
screenshot from 2017-10-18 12-12-06

johnjeng:
Fixes #2432 and #2714 
Also improves UI involved with Skipped Steps.
 - Don't show errors involved in skipped steps
 - Display `Skipped`
 - If step is skipped, show all UI for next step even if previous step has errors

screen shot 2018-03-13 at 4 31 32 pm

screen shot 2018-03-13 at 4 32 03 pm

screen shot 2018-03-13 at 4 32 28 pm

### DIFF
--- a/src/smc-webapp/course/common.cjsx
+++ b/src/smc-webapp/course/common.cjsx
@@ -363,7 +363,7 @@ exports.StudentAssignmentInfo = rclass
                     <Col md={width} key='last_collect'>
                         {@render_last('Collect', @props.info.last_collect, 'collected', @props.info.last_assignment? or skip_assignment,
                            "Copy the assignment from your student's project back to your project so you can grade their work.",
-                           "Open the copy of your student's work in your own project, so that you can grade their work.") if not @props.info.last_assignment?.error}
+                           "Open the copy of your student's work in your own project, so that you can grade their work.") if skip_assignment or not @props.info.last_assignment?.error}
                     </Col>
                     {@render_peer_assign()  if peer_grade and @props.info.peer_assignment and not @props.info.last_collect?.error}
                     {@render_peer_collect() if peer_grade and @props.info.peer_collect and not @props.info.peer_assignment?.error}
--- a/src/smc-webapp/course/common.cjsx
+++ b/src/smc-webapp/course/common.cjsx
@@ -27,6 +27,7 @@ misc = require('smc-util/misc')
 
 # React libraries
 {React, rclass, rtypes, Actions, ReactDOM}  = require('../smc-react')
+Fragment = React.Fragment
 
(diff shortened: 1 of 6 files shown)


src/smc-webapp/course/common.cjsx
```diff
@@ -27,6 +27,7 @@ misc = require('smc-util/misc')
 
 # React libraries
 {React, rclass, rtypes, Actions, ReactDOM}  = require('../smc-react')
+Fragment = React.Fragment
```
williamstein: If we'll use this a lot, it would be nice to just do 
```
{Fragments, React, rclass, rtypes, Actions, ReactDOM}  = require('../smc-react')
```

src/smc-webapp/course/common.cjsx
```diff
@@ -303,7 +304,10 @@ exports.StudentAssignmentInfo = rclass
             error = "Try to #{name.toLowerCase()} again:\n" + error
         <ErrorDisplay key='error' error={error} style={maxHeight: '140px', overflow:'auto'}/>
 
-    render_last: (name, obj, type, enable_copy, copy_tip, open_tip) ->
+    render_last: (name, obj, type, enable_copy, copy_tip, open_tip, opts) ->
```
williamstein: I think this is the first time in all our codebase that opts is used this way.  I would rather you use a different name, e.g., extra_opts, *or*, put all the inputs into opts.  I think everywhere else in the codebase, if opts is used, then it is the only thing used, i.e.,
```
render_last: (opts) ->
    opts = ...
```

Definitely my preference is **very strongly** that if there are so many arguments that there is any possible chance that things could get confusing... then just use opts for all of them.  Combining using opts with a bunch of positional params is asking for trouble.  Positional params are already asking for trouble when we don't have strong typing (like in typescript) or  named params (like in python).

williamstein:
Code looks fine and works fine.  I just have two trivial and purely stylistic requests.  I think I'll merge this as is, and you can make them (or not) tomorrow.

src/smc-webapp/course/common.cjsx
johnjeng: Yeah, I'm going to investigate the actual fragment syntax today. It should work like this:
```
<>
  <Component />
  <AnotherComponent />
</>
```

src/smc-webapp/course/common.cjsx
johnjeng: It's been used in a few places but usually with less than 3 positional arguments (ie. very few args and the function has very clear default behavior). I think I will convert this over to all opts. It's already hard to map in my head what this random boolean in the middle of the function call is doing.

PR #2747, opened 2018-03-14
Add in place adjustor

johnjeng:
Add an in place adjustor to the Upgrades Page.

screen shot 2018-03-13 at 5 10 21 pm

screen shot 2018-03-13 at 5 10 28 pm


williamstein:
This looks good and works fine.  However, replace "Adjust...", which is cluttered (repeating the same thing over and over) by turning the content of the "Upgrades you have applied to this project" column into that link.  I.e., 

1. Delete the "Adjust..." column.
2. Change the rightmost column to be a link that does exactly the same thing as Adjust.

This eliminates clutter and makes sense: if you want to change something, click on it.

williamstein:
OK, I'm merging since it's functionally fine.  But please make a followup with the UI change.

PR #2748, opened 2018-03-14
course: rename settings to configuration to avoid mixup with project settings

haraldschilly:
I have the impression it's confusing to have two :wrench: + "Settings"  next to each other. At least it would help me with explaining it if it is immediately clear if "Setting" or "Configuration" are the place to look.

screenshot from 2018-03-14 17-36-31

screenshot from 2018-03-14 17-36-41


haraldschilly:
oh yes, and the redux name as well. I only looked that the user visible string. I'll fix this.

johnjeng:
Nope. Gonna leave it to you.

williamstein:
I have some concerns -- I'm worried this is a bad temporary bandaide to get around a more fundamental issue:

 - *A prioiri*, why change the course "Settings" rather than the project "Settings"?  Both have an equal right to be called "Configuration".

- "Configuration" is just a synomym for Settings, so you're just adding multiple words for the same exact concept, where there is no genuine distinction between them.  That's annoying and frustrating to users.

- This PR beaks existing documentation/guides/etc.   For example [the course tutorial](https://tutorial.cocalc.com/tips_and_tricks/tips_and_tricks.html) emphasis the distinction between the two types of settings; if you just change this without changing that guide (etc.), it's going to cause even more frustration.

- What about other editors in the future that have their own "Settings"?  Should they be called "configuration" arbitrarily?  Or something else?   For example, I plan to add a notion of settings for tasks, which will where labels are colored, etc.

- We have "Account Preferences", where "Preferences" is yet another synomym for Settings.  Why is this course thing "Configuration" instead of "Preferences", say?

- Any other ideas about how to address this problem?  I'm suspiciuos of choosing a solution without even listing at least 2 or 3 other ones and saying why one is better than the others.

DrXyzzy:
Alternatives:
- Group all settings under a global `Preferences` (or some other name) tab with subsections for each type of setting. This follows the example of common web browsers etc.
- Breadcrumbs. Also use these in docs to refer to which toolbar & tab we are referring to - e.g. `Account / Preferences / Editor settings`

williamstein:
> Group all settings under a global Preferences (or some other name) tab with subsections for each type of setting. This follows the example of common web browsers etc.

What is meant by "all settings"?   If you mean both project settings and course settings, this is not what browsers do, since that would be analogous to having the configuration for the github *web page* inside the globa chrome settings.

> Breadcrumbs.

I don't know what this means.

williamstein:
Probably the closest analogue with browsers might be to move the project's File/New/Log/Find/Settings to a single dropdown menu.    Like this, in chrome:

screenshot 2018-03-16 at 12 49 30 pm

johnjeng:
Well, one could arbitrarily describe levels or categories of these items. ie. We have account preferences, project settings, and any file editor has some configuration.

Calling the same thing different names annoying and frustrating to users but so is calling different things the same name. Matching UI to key words is exactly how users might navigate.

So another solution is to call them all the same thing but with clearer UI differentiation. If instructed to go to account settings, I've seen many people click on the only item that says settings which is actually project settings (Why didn't they pick up on "account"? Well, probably because english reads left to right). Furthermore, they didn't immediately realize they were in the wrong place since we should probably rename the title in project settings from `Settings and configuration` to `Project Settings`

Even knowing the UI, there's times when I click on course settings when I want project settings or vice versa.

williamstein:
Another issue is that our current "project settings", is actually mostly **control** and only a little bit settings... and other stuff to, e.g., the link for the plain jupyter server.  

If we completely got rid of "Settings" for a project, and replaced it by something else -- e.g., a menu with several options that bring up some of the panels, then the need for this current PR goes away.

williamstein:
Compromise:

- Settings in project stays "settings" and title becomes "Project Settings"
- Account "preferences" stays as is...  but
  - Account -> Preferences --> User
- **Everywhere else in editors** where it would be something like "settings", it is now **"Configuration"**.

williamstein:
I just went ahead and fixed the conflicts.

PR #2777, opened 2018-03-26
syncdoc: skip undefined codemirrors in close

haraldschilly:
This is hopefully harmless until this is fixed properly


williamstein:
This code will get entirely deleted soon (replaced by the new frame-tree stuff), so this is definitely fine to hack temporarily.

Issue #2778, opened 2018-03-26
TypeError: this._save_local_view_state is not a function


haraldschilly:
for an `html` file (although, very similar code is also in the tasks editor!)

chrome: `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/65.0.3325.181 Safari/537.36`

smc_git_rev: 7540469b969a9a460bab71590265385193ec3824

Date: 2018-03-26 10:10:47.554486

`this._save_local_view_state is not a function`

```
TypeError: this._save_local_view_state is not a function
    at t.value (https://cocalc.com/static/smc-f6370dd7e08c9fce487f.cacheme.js?f6370dd7e08c9fce487f:6:1038692)
    at Object.save_editor_state (https://cocalc.com/static/smc-f6370dd7e08c9fce487f.cacheme.js?f6370dd7e08c9fce487f:6:1020959)
    at l (https://cocalc.com/static/smc-f6370dd7e08c9fce487f.cacheme.js?f6370dd7e08c9fce487f:6:289761)
    at function.e._wrapper.e._wrapper (https://cocalc.com/static/lib-f6370dd7e08c9fce487f.cacheme.js?f6370dd7e08c9fce487f:6:2103426)
    at https://cocalc.com/static/lib-f6370dd7e08c9fce487f.cacheme.js?f6370dd7e08c9fce487f:6:2104936
```

This `save_editor_state` is

```
value:function(e,t){var n,r,i;if(y(this,h),null!=(r=this.store.get("local_view_state"))){if(n=null!=(i=r.get("editor_state"))?i:c.Map(),null==t){if(!n.has(e))return;n=n.delete(e)}else n=n.set(e,c.fromJS(t));return this.setState({local_view_state:r.set("editor_state",n)}),this._save_local_view_state()}}},
```

I can see that `_save_local_view_state` is deleted in the end, maybe it is still being called? I.e. a delayed call via that debouncing?

I just checked, and I think the "l" function mentioned in the stacktrace is this l inside debounce:

```
w.debounce=function(e,t,n){var r,i,o,s,a,l=function(){var c=w.now()-s;c<t&&c>=0?r=setTimeout(l,t-c):(r=null,n||(a=e.apply(o,i),r||(o=i=null)))};
```

haraldschilly:
I'm a bit puzzled about this.

[...]

PR #2779, opened 2018-03-26
code-editor: safeguard save_editor_state against called after close

haraldschilly:
idea for #2778

but I'm not sure if I understand this correctly ... and there is similar code for the tasks editor, which might need a similar check.


williamstein:
You do understand it -- once the state enters closed it will never leave, and any last minute actions should be ignore...

PR #2786, opened 2018-03-26
Wait until success to update coupon use

johnjeng:
Fixes case when failed subscriptions would count against the user's coupon count.

@DrXyzzy can you test this?

### DIFF
--- a/src/smc-hub/client.coffee
+++ b/src/smc-hub/client.coffee
@@ -2180,6 +2180,9 @@ class exports.Client extends EventEmitter
                     dbg("Successfully added subscription; now save info in our database about subscriptions....")
                     @database.stripe_update_customer(account_id : @account_id, stripe : @_stripe, customer_id : customer_id, cb: cb)
                 (cb) =>
+                    if not options.coupon
+                        cb()
+
                     if options.coupon
                         dbg("add coupon to customer history")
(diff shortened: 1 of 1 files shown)


src/smc-hub/client.coffee
```diff
@@ -2239,6 +2239,15 @@ class exports.Client extends EventEmitter
...
+                    dbg("Update the subscription.")
+                    changes =
+                        quantity : mesg.quantity
+                        plan     : mesg.plan
+                        coupon   : mesg.coupon_id
+                    @_stripe.customers.updateSubscription(customer_id, subscription_id, changes, cb)
+                (cb) =>
+                    @database.stripe_update_customer(account_id : @account_id, stripe : @_stripe, customer_id : customer_id, cb: cb)
                 (cb) =>
                     if mesg.coupon_id
```
williamstein: There is no cb called at all if mesg.coupon_id isn't true!  Oops.  Make the code

```
if not mesg.coupon_id
    cb()
... then the rest not indented
```

johnjeng: Fixed. Weird how this change is being recorded.

williamstein: You forgot the return after the cb(), which is even worse :-)

williamstein:
i'll just merge and fix myself.

PR #2792, opened 2018-03-28
db: add projects.compute_image field to specify which one to run in kucalc

haraldschilly:
I think this is harmless, ...


williamstein:
I will wait until low load to restart all hubs, which merging this requires (so tonight or tomorrow).

PR #2794, opened 2018-03-28
static page dedicated to latex editor

haraldschilly:
nothing to see yet …


williamstein:
Sad to see new code being written in pug.  I'm **definitely** going to rewrite anything in pug to be done in React instead.

PR #2803, opened 2018-03-30
make "something bad" message more useful

DrXyzzy:
Replace `Something bad must have happened.` message with text that
- identifies CoCalc as source of the message
- offers information on the nature of the error

@haraldschilly  please review for correctness. Also, can you suggest a quick way to simulate the error in testing?

If merged, also update [this wiki entry](https://github.com/sagemathinc/cocalc/wiki/Troubleshooting#load-timeout)

### DIFF
--- a/src/webapp-lib/app.pug
+++ b/src/webapp-lib/app.pug
@@ -16,7 +16,7 @@ html(lang="en")
         div.message
           | Timeout while loading CoCalc.
           br/
-          | If the problem persists, click <a href="https://github.com/sagemathinc/cocalc/wiki/Troubleshooting">Troubleshooting</a> for more information.
+          | If the problem persists, click <a href="https://github.com/sagemathinc/cocalc/wiki/Connection-Timeout">Connection Timeout</a> for more information.
     div#smc-startup-banner-status
       | Initializing ...
 
--- a/src/webapp-lib/app.pug
+++ b/src/webapp-lib/app.pug
@@ -14,9 +14,11 @@ html(lang="en")
         img(src=require('!url-loader?mimetype=image/svg+xml!cocalc-icon-white-transparent.svg'))
       div.banner-error
         div.message
-          | Something bad must have happened.
+          | Timeout while loading CoCalc.
           br/
           | Try <a href="javascript:location.reload(true);">reloading the page</a> while holding your shift-key.
+          br/
+          | Or disable browser extensions blocking external sites, or try a different browser.
     div#smc-startup-banner-status
       | Initializing ...
 
--- a/src/webapp-lib/app.pug
+++ b/src/webapp-lib/app.pug
@@ -16,9 +16,7 @@ html(lang="en")
         div.message
           | Timeout while loading CoCalc.
           br/
-          | Try <a href="javascript:location.reload(true);">reloading the page</a> while holding your shift-key.
-          br/
-          | Or disable browser extensions blocking external sites, or try a different browser.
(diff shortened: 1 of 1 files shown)


src/webapp-lib/app.pug
```diff
@@ -14,9 +14,11 @@ html(lang="en")
         img(src=require('!url-loader?mimetype=image/svg+xml!cocalc-icon-white-transparent.svg'))
       div.banner-error
         div.message
-          | Something bad must have happened.
+          | Timeout while loading CoCalc.
           br/
           | Try <a href="javascript:location.reload(true);">reloading the page</a> while holding your shift-key.
+          br/
+          | Or disable browser extensions blocking external sites, or try a different browser.
```
williamstein: I wonder if a link to a wiki page of troubleshooting techniques for the initial page load would be way more useful?

E.g., it could remark that IE11 is known broken *today*.

And it could be more clear, e.g., "whitelist cocalc in any ad blocking extensions..."

src/webapp-lib/app.pug
```diff
@@ -14,9 +14,11 @@ html(lang="en")
         img(src=require('!url-loader?mimetype=image/svg+xml!cocalc-icon-white-transparent.svg'))
       div.banner-error
         div.message
-          | Something bad must have happened.
+          | Timeout while loading CoCalc.
```
williamstein: One funny thing about this "timeout" is that if your network is slow and it takes 65 seconds to load the page, then it'll show this error, wait 5s and still load fine.  So it's only a timeout in the sense that it's "suspicious".  I've seen this a lot when doing dev (via cell phone network), where the page load size went to 15MB+...  That's why it's 60s and not 30s.

src/webapp-lib/app.pug
williamstein: In short, maybe replace all this with "It is taking a long time to load CoCalc, which is suspicious.  Please see these [troubleshooting ideas](...) or refresh your browser to try again."

DrXyzzy:
Link added to wiki page on [Troubleshooting](https://github.com/sagemathinc/cocalc/wiki/Troubleshooting). The wiki page includes instruction to click Shift Enter, which was in the code.

williamstein:
No - I don't like this.  You can't link to the completely general troubleshooting or even "Debugging connectivity issues in CoCalc" page when the user sees that box.  Why would they see that box? 

  - using a very old unsupported browser.
  - very slow network, so loading is going to work fine, but it takes more than 60s
  - I pushed out a comletely broken version of cocalc
  - Temporary issue serving cocalc.

More of the page you link to now is about what to do assuming you can connect at all.

DrXyzzy:
How about linking to a new wiki page with the following:

link: .../Connection-Timeout

## Timeout When Loading CoCalc Main Page

If you see the error message "Timeout while loading CoCalc." when you try to connect to CoCalc, there are several possible causes:

- **You are using a very old unsupported browser.** Check for updates available to the browser you are using, or switch to the latest version of a different browser.

- **You are on a very slow network.** Loading is going to work fine, but it takes more than 60 seconds. If you are on a slow network, loading may succeed after the timeout warning is displayed. 

- **There is a temporary issue serving CoCalc, or an internal problem in CoCalc itself.** If the reasons above don't apply, let us know by sending email to help@sagemath.com.

williamstein:
Yes, that's perfect.

DrXyzzy:
Wiki page [Connection Timeout](https://github.com/sagemathinc/cocalc/wiki/Connection-Timeout) added and link in app.pug updated.

PR #2804, opened 2018-03-30
Use babel-loader for .js and .jsx

johnjeng:
I remembered to polyfill es2015+ but not transpile it. Oops.


haraldschilly:
Ahh! That explains everything!

Issue #2725, opened 2018-03-03
process video-src links to point to "raw" files


haraldschilly:
Currently, the file `smc-webapp/process-links.coffee` only deals with `a`, `img` and `object` tags.

This ticket's goal is to also deal with `video` tags, which need a very similar replacement of their `src` attribute. However, it's not as easy as img tags, because these `video` tags can have nested children like that:

```
<video>
  <source src=" ... " />
  <source src=" ... " />
</video>
```

PR #2809, opened 2018-04-04
webapp: process links should also fix <video> tags

haraldschilly:
issue #2725

Testing: upload a small webm video and do this in a chat or a jupyter markdown textarea:

```
<video controls src="video.webm" style="width:50%">
</video>

<video controls style="width:50%">
    <source src="video.webm" >
</video>
```

Of course, other image tags shouldn't break.

There is also similar code buried inside the `sagews/sagews.coffee` file. I tried to fix this in a similar way but then it all broke. That should better be fixed as part of the upcoming rewrite.


williamstein:
> That should better be fixed as part of the upcoming rewrite.

Yes, that's definitely right.

PR #2813, opened 2018-04-05
webapp install: strategically calling git to initialize submodules

haraldschilly:
I found 4 places, where it makes sense to strategically call git to initialize and update submodules.


williamstein:
Thanks!

PR #2814, opened 2018-04-05
Issue/2810

### LINKED ISSUE #2810: markdown link formatting is broken when the target link has a space in it.
1. Put this in a markdown file or task list:
```
[file with space](./a b.txt) and [file without space](./a.txt)
```

2. Notice that the first does NOT get properly converted!

screenshot 2018-04-04 at 1 28 23 pm

REPORTED BY: Aaron Tresham

johnjeng:
Fixes #2810 by using marked.js

Options may need adjusting.


williamstein:
@haraldschilly any comments about this? Changing markdown processes is... interesting and dangerous.

haraldschilly:
uhm, I can try to test this tomorrow. there are many changes, yes.

haraldschilly:
Hmm, I am not sure what's going on, but when I started to test this, I switched to the branch and with a clean working tree in git, I did `npm run clean; npm run make` 

The result is:

```
[...]
# static react pages update must come *before* webpack
cd $SMC_ROOT
update_react_static
/home/user/smc/src/smc-webapp/node_modules/katex/contrib/auto-render/auto-render.js:3
import katex from "katex";
^^^^^^
 
SyntaxError: Unexpected token import
    at createScript (vm.js:68:6)
    at Object.runInThisContext (vm.js:76:34)
[...]
```

haraldschilly:
I had to merge with master, see commit above, because the very first md file I opened was a test for the issue from yesterday. :boom:

haraldschilly:
Well, I continued testing, I have to think about what all to check for, though.

What I saw and I don't know if even related, are pre tags in sagews markdown cells. i.e. 

```
%md
<pre>
...
</pre>
```

renders as normal font. I think there is a CSS class missing, `.sagews-output-md pre { font-family: monospace; }`

haraldschilly:
Another detail: in a markdown cell in a jupyter notebook (cocalc's), enter this:

```
<code>$x</code> $ bar
```

which ends up as `$x</code> $ bar` and monospace font formatted.

haraldschilly:
Hmm, what's also missing are the environments. I.e. at least this one in a markdown cell in an ipynb is no longer detected:

```
\begin{aligned}
x &=& y \\
y &=& z
\end{aligned}
```

The full list is in smc-util/misc.coffee: `mathjax_environments = ...`

haraldschilly:
I also found a maybe completely unrelated issue, in ipynb files. It works in production, so maybe this is ticket related...

Do the following sequence of keystrokes: [Esc] to be in command mode. Now: "b", "m", "Return", "x", "Shift+Return".

It should do: insert new cell below, switch to markdown mode, return to edit it, enter the letter "x" as it's content, and shift-return to evaluate it. Point is, it's not evaluating to markdown, but stays formatted as a cell. In production, you can see a "text"-"x" and it's not a cell any more.

williamstein:
Changing to needs work, since it seems like @haraldschilly points out we can't just switch to this...

johnjeng:
>Hmm, looks like this react static rendering is broken.

Urgh. It's cause it needs to transpile/polyfill stuff from es6/7 to es5. Eventually (soon) we'll need to integrate this better with Webpack. For the time being I guess I'm going to hack it and just check in r_misc if we're making the static pages.

johnjeng:
Moot.

PR #2817, opened 2018-04-06
add new workshop subscription/plan, which is 1 week

williamstein:
screenshot 2018-04-06 at 11 03 54 am


williamstein:
Thoughts?

slel:
For workshops, maybe use 'people' or 'participants' instead of 'students'.

PR #2827, opened 2018-04-11
webapp/landingpage: small changes to the trial info box

haraldschilly:
here is how it looks on my screen

cocalc-landing-tweak

### DIFF
--- a/src/smc-webapp/landing_page.cjsx
+++ b/src/smc-webapp/landing_page.cjsx
@@ -639,6 +639,8 @@ Connecting = () ->
     </div>
 
 exports.LandingPage = rclass
+    displayName: 'LandingPage'
+
     propTypes:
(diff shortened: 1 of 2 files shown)


src/smc-webapp/landing_page.cjsx
```diff
@@ -676,27 +680,59 @@ exports.LandingPage = rclass
             forgot_password_success = {@props.forgot_password_success}
         />
 
+    # this is an info blob on the landing page, clarifying to the user that "free" is a perpetual trial
+    render_trial_info: ->
+        return null if not (@props.customize?.is_commercial ? true)
```
williamstein: This is wrong; this is not how redux props works.  It should be @props.is_commercial.  I'll fix this.

Issue #2868, opened 2018-05-07
Enable support for "from __future__ import ..." in sage worksheets


dv-extrarius:
In a previous ticket, I mentioned that I wanted support for the python 3-style print function (typically enabled via "`from __future__ import print_function`" in python 2 code) in sage worksheets, and a specific utility function was added to support that single use case.

However, I also want support for 'true division', which is typically enabled with something like "`from __future__ import division`" in python 2 code, and I'd really like to be able to stick to the common python 2 syntax for such things.

williamstein:
This is now live.

Just out of curiosity, how would anybody ever find out about this functionality?

Is there any way to make it so typing
```
from __future__ import print_function
```
in a worksheet shows an error, and says `please use python_future_feature` instead...??

DrXyzzy:
Note: this issue is related to #2862.

@williamstein - we could make the import `from __future__ import print_function` show an error - using logic similar to implicit multiplication detection in the commits for #2513. But with the approach in the present PR, the import simply works, as if the user were running a .py file. The latter seems preferable to me.

williamstein:
> simply works.

oh -- nice!!

PR #2869, opened 2018-05-07
Enable support for "from __future__ import ..." in sage worksheets

dv-extrarius:
Ref: #2868

Add support for python's __future__ statements along with a utility function python_future_feature(feature=None, enable=None) to enable, disable, and query the state of 'future features' like "division" and "print_function".

### DIFF
--- a/src/smc_sagews/smc_sagews/sage_parsing.py
+++ b/src/smc_sagews/smc_sagews/sage_parsing.py
@@ -14,6 +14,38 @@ Code for parsing Sage code blocks sensibly.
 
 import string
 import traceback
+import __future__ as future
+import ast
+
+def get_future_features(code, mode):
+    if '__future__' not in code:
+        return {}
+    features = {}
+    node = ast.parse(code, mode = mode)
+    #Make it work for all outer-container node types (module, interactive, expression)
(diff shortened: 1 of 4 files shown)


DrXyzzy:
Thanks for the PR! Reviewing now.

DrXyzzy:
Looks good, and thanks for including pytest cases. All pytests passed (except the usual 5 skipped) in my test run. One comment: the function in `test_sagews.py::TestPythonFutureImport` that clears future feature settings:
```
class TestPythonFutureImport1:
    def test_pyfutimp_0(self,exec2): ...
```
is not needed, because a new sage_server client is spawned each time a test class is run.

DrXyzzy:
Also note that the following code works, although [maybe it shouldn't](https://docs.python.org/3/reference/simple_stmts.html#future), because statements at the same level of indentation in a sagews code cell are executed as separate code blocks:
```
print("hello")
from __future__ import division
17r/6r
from __future__ import print_function
sin(sqrt(2)).n()
print("foo", end=" X\n")
python_future_feature()
  ->
hello
2.8333333333333335
0.987765945992736
foo X
['division', 'print_function']
```

Issue #2870, opened 2018-05-08
make sagews2pdf robust against special characters in date


haraldschilly:
There is a case where special characters ended up in the date field by default. The string encoding should take care of it.

Browser+OS is IE/Edge on Win 10

date: `'\u200e5\u200e/\u200e7\u200e/\u200e2018'`

```
"command 'smc-sagews2pdf' (args=./Sage Demo Module 1.sagews --outfile ./Sage Demo Module 1.pdf --title Sage Demo Module 1 --author xxx --date ‎5‎/‎7‎/‎2018 --subdir false --contents false) exited with nonzero code 1 -- stderr='Traceback (most recent call last):
 File \"/cocalc/bin/smc-sagews2pdf\", line 12, in 
 sys.exit(main())
 File \"/cocalc/lib/python2.7/site-packages/smc_pyutil/sagews2pdf.py\", line 785, in main
 style = args.style
 File \"/cocalc/lib/python2.7/site-packages/smc_pyutil/sagews2pdf.py\", line 711, in sagews_to_pdf
 style=style)
 File \"/cocalc/lib/python2.7/site-packages/smc_pyutil/sagews2pdf.py\", line 684, in latex
 + FOOTER
 File \"/cocalc/lib/python2.7/site-packages/smc_pyutil/sagews2pdf.py\", line 655, in latex_preamble
 s += r\"\\date{%s}\"%tex_escape(date) + \"\
\"
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 6: ordinal not in range(128)
'"
```

williamstein:
sagews is deprecated

PR #2871, opened 2018-05-08
sagews2pdf: changing default encoding of strings to address special chars in date and author strings

haraldschilly:
issue #2870

I've never seen this default encoding trick for strings before. It seems to work, though. 

Setting @DrXyzzy as a reviewer, because maybe you have some additional insights.


DrXyzzy:
There are [warnings](https://stackoverflow.com/questions/3828723/why-should-we-not-use-sys-setdefaultencodingutf-8-in-a-py-script) about the `sys.setdefaultencoding("utf-8")` trick, but it seems to work here. I tested by doing the following in a sage worksheet:
```
s = u'\u0420\u043e\u0441\u0441\u0438\u044f\u200e5\u200e/\u200e7\u200e/\u200e2018'
print(s)
```
Then copying and pasting the output into a .term and running the conversion utility on some .sagews file:
```
z="...pasted output..."
smc-sagews2pdf foo.sagews --date $z >/dev/null 2>/dev/null
open foo.pdf
```
Without the patch from this PR, trying to view `foo.pdf` in CoCalc causes the error `Failed to load PDF document.` With the patch from this PR, a pdf with the correct date is written and can be viewed.

PR #2882, opened 2018-05-14
change 'invite' to 'add' in text

### LINKED ISSUE #2873: change buttons from "Invite" to "Add" etc. when adding collaborators
In Project Settings / Add people to project, the button to add someone should be changed:
* from "Invite user" to "Add user" if one user is found on search
* from "Invite selected user" to "Add selected user" if more than one user is found in search and only 1 is selected
* from "Invite n users" to "Add n users" if more than one user is found in search and more than 1 is selected

Needed because we don't have an invite/accept process, but instead just an add process.

tscholl2:
This should fix #2873 unless I am missing an "invite" somewhere, but my very-quick grep search found one other instance in billing here: https://github.com/sagemathinc/cocalc/blob/invite_to_add/src/smc-webapp/billing.cjsx#L1212

```
You can also invite collaborators to work with you inside a project
```

I thought this was an ok use for the word "invite" so I left it, but maybe for consistency it might be better to switch it also.

### DIFF
--- a/src/smc-webapp/collaborators/add-to-project.cjsx
+++ b/src/smc-webapp/collaborators/add-to-project.cjsx
@@ -88,7 +88,7 @@ exports.AddCollaborators = rclass
         replyto      = redux.getStore('account').get_email_address()
         replyto_name = redux.getStore('account').get_fullname()
         if replyto_name?
-            subject = "#{replyto_name} invites you to project #{@props.project.get('title')}"
+            subject = "#{replyto_name} added you to project #{@props.project.get('title')}"
         else
             subject = "CoCalc Invitation to project #{@props.project.get('title')}"
         @actions('projects').invite_collaborator(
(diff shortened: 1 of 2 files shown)


DrXyzzy:
You might also change the wording of the Subject of the collaborator-added email in src/smc-webapp/collaborators/add-to-project.cjsx to not use "invitation". At present it's

`CoCalc Invitation to project #{@props.project.get('title')}`

tscholl2:
Good point. I saw that but couldn't think of a nice way to rephrase it. My initial idea was "CoCalc Addition to Project", but that doesn't sound right. What about "You've been added to project..."?

DrXyzzy:
Looks good. I would put the word `CoCalc` in the Subject.

tscholl2:
Added!

williamstein:
@DrXyzzy -- do you sign off on this? :-)

src/smc-webapp/collaborators/add-to-project.cjsx
```diff
@@ -88,9 +88,9 @@ exports.AddCollaborators = rclass
         replyto      = redux.getStore('account').get_email_address()
         replyto_name = redux.getStore('account').get_fullname()
         if replyto_name?
-            subject = "#{replyto_name} invites you to project #{@props.project.get('title')}"
+            subject = "#{replyto_name} added you to CoCalc project #{@props.project.get('title')}"
```
williamstein: Technically you shouldn't explicitly use "CoCalc" is messages to users.  There's a variable somewhere that gets set.  People can make their own install and call it something else...

williamstein:
I think I'll merge this anyways, but can somebody fix it sometime?

PR #2883, opened 2018-05-16
use COCALC_PROJECT_ID env variable instead of SMC info file

tscholl2:


src/dev/project/util.py
```diff
@@ -14,14 +14,15 @@ def chdir():
 
 def base_url(port=None, write=True):
     print("base_url(port=%s)"%port)
-    info_file = join(os.environ['SMC'], 'info.json')
-    info = json.loads(open(info_file).read())
+    project_id = os.environ['COCALC_PROJECT_ID']
+    if project_id is None or project_id == "":
```
williamstein: It can't be None.  If the var isn't defined os.environ will throw an error.  You're thinking `os.environ.get('COCALC_PROJECT_ID', None)`.   

Also, if this isn't set the error should really be "COCALC_PROJECT_ID environment variable not found. You can only use dev/project to run CoCalc from *inside* of a CoCalc project."

PR #2887, opened 2018-05-16
matplotlib backend selection in salvus server

haraldschilly:
This is work in progress, but right now it is at least in state where sagews worksheets produce some output in ubuntu 18.04 again.

The background is that matplotlib config files or env variables might set the backend of matplotlib to one of such a type, where it tries to open a graphical window. e.g. configs like that

```
$ grep ^backend /etc/matplotlibrc
backend      : TkAgg
```

This change here makes this robust against that and selects "Agg" regardless of what the configuration says.

Loading speed? Since in that file pylab is getting imported globally, and also some other parts of matplotlib, I can't imagine how this can have any impact.

Testing: ~~what's missing is to double check with the current ubuntu 16.04 environment.~~ ok, this works in the current software environment, too (no surprise)

### DIFF
--- a/src/smc_sagews/smc_sagews/sage_salvus.py
+++ b/src/smc_sagews/smc_sagews/sage_salvus.py
@@ -12,6 +12,9 @@
 #                  http://www.gnu.org/licenses/                                         #
 #########################################################################################
 
+# set backend of matplot lib before any other module is loaded
+import matplotlib
+matplotlib.use('Agg')
 
 import copy, os, sys, types, re
 
--- a/src/smc_sagews/smc_sagews/sage_server.py
+++ b/src/smc_sagews/smc_sagews/sage_server.py
@@ -4,8 +4,9 @@ sage_server.py -- unencrypted forking TCP server.
 
 Note: I wrote functionality so this can run as root, create accounts on the fly,
 and serve sage as those accounts.  Doing this is horrendous from a security point of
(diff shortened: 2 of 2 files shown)


DrXyzzy:
Review: Looks good.

Testing:

1. All pytests in smc_sagews/smc_sagews/tests pass in branch for this PR.

    ```
    python -m pytest -x
    ...
    == 167 passed, 5 skipped in 352.01 seconds ==
    ```

1. No adverse effect on startup time.

    Excerpt from sage_server logs from 3 runs **without** this PR:
    ``` 
    sage_server: seconds to import sage_server: 10.7005848885
    imported sage library and other components in 0.331412792206 seconds
    sage_server: seconds to import sage_server: 8.26413607597
    imported sage library and other components in 0.217766046524 seconds
    sage_server: seconds to import sage_server: 7.25702404976
    imported sage library and other components in 0.319604873657 seconds
    ``` 

    Excerpt from sage_server logs from 3 runs **with** this PR:
    ```
    sage_server: seconds to import sage_server: 8.03657007217
    imported sage library and other components in 0.3473341465 seconds
    sage_server: seconds to import sage_server: 8.55658817291
    imported sage library and other components in 0.198640108109 seconds
    sage_server: seconds to import sage_server: 4.06400489807
    imported sage library and other components in 0.184757947922 seconds
    ```

williamstein:
Thanks @DrXyzzy !

PR #2901, opened 2018-05-25
Hotfix course invites

haraldschilly:
cover your eyes, an ugly hotfix

**DON'T MERGE THIS** it's just for reference for where the problems are. I also pushed the version (forgot this initially)

hence more importantly: **UPDATE VERSION** in master, when this is properly fixed.


johnjeng:
What are the reproduction steps for this? I'm guessing that I broke this...

haraldschilly:
yes, sorry, it's the situation where a student is invited to a course, or a new one is created.  make a course, invite yourself as `j3+studentNNN@cocalc.com` and then it breaks when it tries to make the log entry. (student projects aren't created, no emails sent, etc)

williamstein:
Fixed by reverting the code here: https://github.com/sagemathinc/cocalc/commit/1ab693a67cfacbd3becce89123c6d992418eb01c

The typescript version of smc-react is still broken, but it probably doesn't use the broken aspects of this.

PR #2902, opened 2018-05-25
Ts stopwatch

johnjeng:
Convert Stopwatch to TS.


williamstein:
This breaks the stop icon.  
screenshot 2018-05-26 at 11 19 58 am

Issue #2903, opened 2018-05-26
REDUX/typescript -- address the smc-react comments in this commit


williamstein:
This commit: https://github.com/sagemathinc/cocalc/commit/1ab693a67cfacbd3becce89123c6d992418eb01c

johnjeng:
This is fixed.

PR #2905, opened 2018-05-29
Issue/2903

johnjeng:
Fixing #2903 Revert getProjectStore/Actions/Table to old semantics. 

Also enforce that stores and actions are defined when calling getStore and getActions when using Typescript.

Note: This creates a semantic difference between the typescript and coffeescript interfaces for our global redux object! 

@williamstein I don't want to change the coffeescript behavior because I'm guessing a lot of code relies on checking if `redux.getStore` is undefined and refactoring all of that sounds unnecessarily dangerous. You mentioned that you'd like things to be guaranteed to be defined when possible for typescript so this is my solution.

### DIFF
--- a/src/smc-webapp/project_store.coffee
+++ b/src/smc-webapp/project_store.coffee
@@ -37,7 +37,7 @@ misc      = require('smc-util/misc')
 
 misc_page = require('./misc_page')
 
-{Actions, rtypes, computed, depends, project_redux_name, Table, register_project_store, redux}  = require('./smc-react')
+{Actions, rtypes, computed, depends, project_redux_name, Table, redux}  = require('./smc-react')
 
 exports.file_actions = file_actions =
     compress  :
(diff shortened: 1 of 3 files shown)


williamstein:
> @williamstein I don't want to change the coffeescript behavior because I'm guessing a lot of code relies on checking if redux.getStore is undefined 

Maybe.  I don't really know.  It's such a pain to even tell with coffeescript...

> and refactoring all of that sounds unnecessarily dangerous. 

+1   -- let's just make it so we can delete it :-).

> You mentioned that you'd like things to be guaranteed to be defined when possible for typescript so this is my solution.

That does seem like a reasonable approach.

williamstein:
Did you test this yet?  I pulled it, tried to open a project, and got a big traceback immediately...
screenshot 2018-05-29 at 4 46 54 pm

Maybe I did something wrong.

johnjeng:
Didn't test yet. Meant to comment that but forgot while rushing out of cafe.

williamstein:
Ok. I only spent a minute on running it and reading the code was helpful...

johnjeng:
@haraldschilly What are the reproduction steps for that course bug related to opening project_stores?

williamstein:
> What are the reproduction steps for that course bug related to opening project_stores?

1. create a course
2. add a student

It fails when trying to add the student to their project, since it tries to add them to the project log for that project... and that project is not opened.  The code does getProjectActions(' student project id').log(...), which obviously fails if getProjectActions returns undefined for closed projects.

johnjeng:
This should be ready for testing by someone else now then.

williamstein:
Works!

Issue #2906, opened 2018-05-30
jupyter: showing "?"-help leads to exception


haraldschilly:
stacktrace from prod, I'll try to reproduce in dev

```
 Uncaught TypeError: this.props.message.get(...).forEach is not a function
    at Object.render (cell-output-message.cjsx?fa2b:206)
    at finishClassComponent (react-dom.development.js?2357:7873)
    at updateClassComponent (react-dom.development.js?2357:7850)
    at beginWork (react-dom.development.js?2357:8225)
    at performUnitOfWork (react-dom.development.js?2357:10224)
    at workLoop (react-dom.development.js?2357:10288)
    at HTMLUnknownElement.callCallback (react-dom.development.js?2357:542)
    at Object.invokeGuardedCallbackDev (react-dom.development.js?2357:581)
    at invokeGuardedCallback (react-dom.development.js?2357:438)
    at renderRoot (react-dom.development.js?2357:10366)
```

where the top frame fails in `cell-output-message.cjsx`

```
    render: ->
        type  = undefined
        value = undefined
        @props.message.get('data').forEach (v, k) ->   // <---- BANG
            type  = k
            value = v
            return false
        if type
            [a, b] = type.split('/')
```

Reproduce it via selecting the "gap" kernel and evaluate `FreeGroup?`. 

My guess is, the kernel replies, but doesn't send any payload data back.

PR #2907, opened 2018-05-30
Issue/2906

haraldschilly:
see #2906 

With that fix, the "data" is a string, a weird one: `Unsupported message: {"found":"true","data":"Spass hassen spass"}`


williamstein:
Is this supposed to be a quick hotfix that gets removed soon, and for which there is a high priority ticket to do something better. Just silently throwing away an error seems bad...  It's not clear to me.

haraldschilly:
for reference, after this fix it looks like that (the help shows the weird message)

screenshot from 2018-05-30 19-56-53

williamstein:
OK, seems safe now that you explained it on the video chat.

PR #2921, opened 2018-06-15
Remove smc-react.coffee and Type Project_store

johnjeng:
Finally seems to run correctly. Code needs some cleaning up still.

@DrXyzzy could you help test this?

# Known Issues
- [x] Creating files from the `+ New` Page doesn't work
- [x] Creating files from project listings doesn't work
- [x] Project Actions is just totally broken. Okay


DrXyzzy:
@j3, "test everything" is a lot. :-) What things do you think most need testing? I'll log the results as I go along.

johnjeng:
@DrXyzzy Right. Haha. Specifically I had in mind Sagews and Jupyter stuff.

johnjeng:
@DrXyzzy Hold off on testing. Lots of stuff broken.

williamstein:
>  I'll log the results as I go along.

Maybe some of what you end up doing can later be automated with mocha...

johnjeng:
@DrXyzzy This is ready to undergo some testing.

DrXyzzy:
@johnjeng, are the 3 "known issues" at the top of this PR supposed to be fixed now? They were not when I checked a few hours ago, before the last few commits. For example, when I checked out the branch for this PR and built cc-in-cc, I could not create files from (+) New or Files tab.

johnjeng:
@DrXyzzy They should be fixed. Though if it returns with an error that isn't `failed` then that's not an issue I'm aware of.

PR #2923, opened 2018-06-15
rewrite Sage Worksheets

williamstein:


src/smc-webapp/frame-editors/sagews-editor/editor.ts
```diff
@@ -0,0 +1,60 @@
+/*
+Top-level react component for editing Sage Worksheets.
+
+Ultimately we'll have several views:
+
+ - a single document view: one big editor, with output widgets (using either codemirror+hacks or prosemirror.)
```
jasongrout: I experimented a bit with monaco recently (https://github.com/jupyterlab/jupyterlab-monaco - a *very* initial rough proof of concept embedding monaco into JuptyerLab). I was impressed, though I haven't had time to dig very deeply into it. It might be something to consider.

It seemed pretty heavy-weight for an editor (e.g., using megabytes of memory for even small bits of text), whereas CodeMirror used more like tens or hundreds of kilobytes, so it may not be very suitable for a cell view where every cell is a separate live editor.

PR #2925, opened 2018-06-17
hub/test: fixing one postgres test

haraldschilly:
tiny fix for one postgres test. not entirely sure about the additional fields. I bet they were added during the last year, that's all.


haraldschilly:
merging it, because harmless. just made the PR to communicate this.

Issue #2836, opened 2018-04-14
it is difficult to know what user I'm signed in as, so change the gear icon next to "Account" to be the user's avatar


williamstein:
This is bog standard and what every other website does (show the avatar).

williamstein:
Here's the relevant commented out code: https://github.com/sagemathinc/cocalc/blob/master/src/smc-webapp/desktop_app.cjsx#L116

tscholl2:
This was merged in 2935.

PR #2935, opened 2018-06-21
center labels on navtabs, use avatar instead of gear when possible

tscholl2:
fixes for #2836 and also center the labels on the navtabs on the top


DrXyzzy:
I'm testing in cc-in-cc.

src/smc-webapp/app_shared.cjsx
```diff
@@ -65,7 +65,7 @@ exports.NavTab = rclass
     propTypes :
         label           : rtypes.string
         label_class     : rtypes.string
-        icon            : rtypes.string
+        icon            : rtypes.oneOfType([rtypes.string,rtypes.object])
```
johnjeng: Shouldn't this be `rtypes.element`?

tscholl2: Probably. I need to look at the list of `rtypes` to get an idea of which ones there are. I can change that.

DrXyzzy:
First project tab doesn't look right.

button-ctr

PR #2938, opened 2018-06-22
make sagews pytest suite work with python 3.5

DrXyzzy:
- explicit use of encode()/decode()/chr() when going between bytes and strings
- dict.items() instead of dict.iteritems()
- allow more time for some tests to complete

to test, apply this patch, then
```
python -m pytest
==>
== 167 passed, 5 skipped in 332.97 seconds ==
```

### DIFF
--- a/src/smc_sagews/smc_sagews/tests/conftest.py
+++ b/src/smc_sagews/smc_sagews/tests/conftest.py
@@ -59,7 +59,7 @@ class ConnectionJSON(object):
 
     def _send(self, s):
         length_header = struct.pack(">L", len(s))
-        self._conn.send(length_header + s)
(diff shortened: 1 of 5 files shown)


haraldschilly:
all fine, I made `python3 -m pytest` explicit in the readme and had one timeout.

besides that, a future adventure for you might be to update from anaconda3 to anaconda5. I saw this in the tests (grep for `TestAnaconda3Mode`) but then remembered that this is also in the sage worksheets. however, maybe we should switch to use the python3 kernel in sage worksheets and not use anaconda. (is `%python3` anaconda? or make this choice explicitly? such a change might also touch some front-end  aspects and documentation)

PR #2945, opened 2018-06-27
Rnw in latex v2

haraldschilly:
well, the only detail I don't understand is forward-search. patching the synctexfile and inverse-search works, but in forward it always jumps to the first page. either a bug of patching, or some weird detail being off.

screenshot from 2018-06-27 18-33-45


haraldschilly:
uhm, well. I might be chasing a ghost. is this mentioned forward-search issue for all tex files? Is it known? I'll try to fix it ... otherwise this patch is actually ok merge.

haraldschilly:
sorry, nevermind, I'm tired. I overlooked a comment in the old code, and thought something is broken, that's essentially all.

williamstein:
Tomorrow include instructions for review and if this is ready.  :-)

haraldschilly:
ok, time for testing.

1. download: [test.rnw](https://cocalc.com/share/20e4a191-73ea-4921-80e9-0a5d792fc511/latex-tests/test.rnw)
2. check if new editor opens up
3. change something in the `c(1,2...)` array, rebuild, histogram and sum should change
4. inverse search
5. forward search
6. clicking the cleanup waste bin button also cleans the generated `*.tex` file
7. changing the build command inserts the `*.tex` file, not `.rnw`. the latex build command doesn't know it is building from a generated file.
8. the syncdb aux shadow file is tied to the `*.rnw` file, not the autogenerated `*.tex`. That makes sense to me, since that's the "main file", but I don't know if my changes are ok. I think I figured it out, though.
9. .. and related to that, when opening up the file more than once, building and changing the build command is synced.

williamstein:
Everything works as far as I can tell. However, I'm going to not merge your PR as is since it is **too brittle**, and nobody could possibly use this as is, since "to error is human".  I think you should parse the output of the knitr build, and if it fails, do NOT try to then run latex.  Instead, show a meaningful error to the user.  E.g., in the above example, in "build control" the knitr log does show this:
```
...
processing file: 2018-06-29-093714.rnw
Quitting from lines 30-35 (2018-06-29-093714.rnw) 
Error in paste(x, collapse = "+") : object 'x' not found
Calls: knit ... inline_exec -> hook_eval -> withVisible -> eval -> eval -> paste
Execution halted
```

williamstein:
I changed my mind and will merge this, since the current version in prod is just as bad in the same way.  I'll make a followup ticket.

williamstein:
Follow up issue: https://github.com/sagemathinc/cocalc/issues/2954

PR #2950, opened 2018-06-28
webapp: some support for formatting py files

haraldschilly:
this is just me playing around for now, to see how this could work. it currently fails with errors in the project, so not sure what's going on.


haraldschilly:
The point where I'm stuck is this stacktrace in the project's local hub:

```
(node:163637) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'length' of undefined
    at diff_match_patch.patch_make (/home/user/smc/src/smc-util/dmp.js:1769:32)
    at make_patch (/home/user/smc/src/smc-util/syncstring.coffee:83:28)
    at StringDocument.make_patch (/home/user/smc/src/smc-util/syncstring.coffee:2134:16)
    at SyncString._sync_remote_and_doc (/home/user/smc/src/smc-util/syncstring.coffee:2091:28)
    at SyncString._save (/home/user/smc/src/smc-util/syncstring.coffee:1313:10)
    at /home/user/smc/src/smc-project/prettier.ts:118:36
    at step (/home/user/smc/src/smc-project/prettier.ts:45:23)
    at Object.next (/home/user/smc/src/smc-project/prettier.ts:26:53)
    at fulfilled (/home/user/smc/src/smc-project/prettier.ts:17:58)
    at <anonymous>:null:null
 
(node:163637) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled
 with .catch(). (rejection id: 1)
```

I do know that `pretty` in `smc-project/prettier.ts` is a string, containing the formatted code, but it fails like the above in line `syncstring._save(() => res.json({ status: "ok" }));`. It works for latex, though. So, what must be missing is some additional setup somewhere?

PR #2951, opened 2018-06-29
webpack: failed attempt to exclude ts and tsx files in ROOT/data

haraldschilly:
webpack breaks for cocalc-in-cocalc dev projects when there are problematic .ts files  inside of them. I mean, webpack shouldn't care about any files in `./data`. I tried to exclude it by an absolute path, or via some pattern, but it is still there. this commit is my failed attempt, maybe someone has an idea what's going on. in the comment is a ticket I found about this.

### DIFF
--- a/src/webpack.config.coffee
+++ b/src/webpack.config.coffee
@@ -512,6 +512,12 @@ if MEASURE
     bundleAnalyzerPlugin = new BundleAnalyzerPlugin({analyzerMode: 'static'})
     plugins = plugins.concat([bundleAnalyzerPlugin])
 
+# we exclude the ROOT/data directory (where the dev projects are, etc.)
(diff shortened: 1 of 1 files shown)


williamstein:
WORKAROUND: set the env variable:
```
export COCALC_PROJECT_PATH="/home/user/projects/"
```

and move data/projects there, then restart the hub.  Then the hub will use that directory instead.   I've been doing this for a while.

Obviously, disabling autobuilding of ts files would be way better, but I don't know how either.

haraldschilly:
oh yes, of course, moving the files also works. thx.

johnjeng:
This is my fix. https://github.com/sagemathinc/cocalc/commit/6d19aabf774ebb5f807b6e71c744ef836049951c

Basically ts-loader [isn't a very good webpack citizen](https://github.com/TypeStrong/ts-loader/issues/552).

PR #2956, opened 2018-07-02
Dealing with Knitr errors

haraldschilly:
This patch accomplishes a few details regarding Knitr processing of Rnw files.

* do not continue to compile any latex if there is a knitr problem
* parse the knitr output and try to generate appropriate error and warning messages (there isn't much information to get, but it's still helpful)
* if there are no knitr errors but latex errors, these latex errors are on top
* pure latex processing is not affected.

below are screenshots for various scenarios (R syntax, missing R variable, latex error)

screenshot-cocalc com-2018 07 02-16-05-57

screenshot-cocalc com-2018 07 02-16-07-26

screenshot-cocalc com-2018 07 02-16-10-05

screenshot-cocalc com-2018 07 02-16-12-52

### DIFF
--- a/src/smc-webapp/frame-editors/latex-editor/actions.ts
+++ b/src/smc-webapp/frame-editors/latex-editor/actions.ts
@@ -56,6 +56,7 @@ interface LatexEditorState extends CodeEditorState {
   zoom_page_height: string;
   build_command: string | List<string>;
   knitr: boolean;
+  knitr_error: boolean; // true, if there is a knitr problem
 }
 
 export class Actions extends BaseActions<LatexEditorState> {
@@ -90,7 +91,7 @@ export class Actions extends BaseActions<LatexEditorState> {
         // changing the path to the (to be generated) tex file makes everyting else
         // here compatible with the latex commands
         this.path = change_filename_extension(this.path, "tex");
-        this.setState({ knitr: this.knitr });
+        this.setState({ knitr: this.knitr, knitr_error: false });
       }
     }
   }
@@ -239,8 +240,8 @@ export class Actions extends BaseActions<LatexEditorState> {
   async run_build(time: number, force: boolean): Promise<void> {
     this.setState({ build_logs: Map() });
     if (this.knitr) {
-      let knitr_errors = await this.run_knitr(time, force);
-      if (knitr_errors) return;
+      await this.run_knitr(time, force);
+      if (this.store.get("knitr_error")) return;
     }
     await this.run_latex(time, force);
     if (this.knitr) {
@@ -252,7 +253,7 @@ export class Actions extends BaseActions<LatexEditorState> {
     }
   }
 
-  async run_knitr(time: number, force: boolean): Promise<boolean> {
+  async run_knitr(time: number, force: boolean): Promise<void> {
     let output: BuildLog;
     const status = s => this.set_status(`Running Knitr... ${s}`);
     status("");
@@ -266,12 +267,12 @@ export class Actions extends BaseActions<LatexEditorState> {
       );
     } catch (err) {
       this.set_error(err);
-      return true;
+      this.setState({ knitr_error: true});
+      return;
     } finally {
       this.set_status("");
     }
     output.parse = knitr_errors(output);
-    console.log(output.parse);
     this.set_build_logs({ knitr: output });
     this.clear_gutter("Codemirror-latex-errors");
     update_gutters({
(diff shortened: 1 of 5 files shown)


williamstein:
WOW, this is awesome!

src/smc-webapp/frame-editors/latex-editor/actions.ts
```diff
@@ -260,16 +261,31 @@ export class Actions extends BaseActions<LatexEditorState> {
...
       );
     } catch (err) {
       this.set_error(err);
-      return;
+      return true;
     } finally {
       this.set_status("");
     }
+    output.parse = knitr_errors(output);
+    console.log(output.parse);
```
williamstein: delete this console.log

src/smc-webapp/frame-editors/latex-editor/actions.ts
```diff
@@ -279,7 +295,7 @@ export class Actions extends BaseActions<LatexEditorState> {
       await patch_synctex(
         this.project_id,
         this.path,
-        force ? undefined : (time || this._last_save_time),
+        force ? undefined : time || this._last_save_time,
```
williamstein: Your change is technically right, because conditional has lower precedence than ||.  But I think it's very important to make this code *clear*, and conditional is literally [**just below** || in precedence... i.e., just barely](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence).   Also, the rest of the code in this file leaves in the parens to make the precedence clear.    It's like write `2 + (3 * 5)` instead of `2 + 3 * 5`.   Do you really find it more readable without the parens?

src/smc-webapp/frame-editors/latex-editor/actions.ts
```diff
@@ -322,6 +338,7 @@ export class Actions extends BaseActions<LatexEditorState> {
     output.parse = new LatexParser(output.stdout, {
       ignoreDuplicates: true
     }).parse();
+    console.log(output.parse);
```
williamstein: delete console logs....

src/smc-webapp/frame-editors/latex-editor/knitr.ts
```diff
@@ -18,17 +20,98 @@ export async function knitr(
...
+  const warnings: Error[] = [];
+  const typesetting: Error[] = [];
+
+  let file: string = "";
+  let err: Error | undefined = undefined;
+
+  const warnmsg = "Warning message:";
+
+  for (let line of output.stderr.split("\n")) {
+    console.log(`line ${i}: ${line}`);
```
williamstein: console logs

williamstein:
I added some **very easy to address** comments.

src/smc-webapp/frame-editors/latex-editor/actions.ts
haraldschilly: well, it's the prettier magic. Personally, I'm also much more happy with rather more than less parens in logical expressions.

williamstein: Well I'm OK with just letting prettier do what it does in this particular case...

src/smc-webapp/frame-editors/latex-editor/latex-log-parser.ts
```diff
@@ -93,12 +93,12 @@ interface File {
   files: string[];
 }
 
-export interface ProcessedLatexLog {
-  errors: Error[];
-  warnings: Error[];
-  typesetting: Error[];
-  all: Error[];
-  files: string[];
+export class ProcessedLatexLog {
```
williamstein: why is this suddenly a class and not an interface?  How does this even work?

williamstein:
You did something weird with class versus interface...

haraldschilly:
I made a change, which makes all this work again, but I'm not as happy as I wished about it. Maybe I'm missing a detail. Anyways. I leave this as some sort of philosophical question for @johnjeng ... In a nutshell, there is now an interface (signaled by the "I") `ProcessedLatexLog` and an associated class. An instance of that class sets all the fields in the constructor. 

### upside

instead of two times writing code like

```
const all: Error[] = [];
const errors: Error[] = [];
const warnings: Error[] = [];
const typesetting: Error[] = [];
```
and a convoluted `return {all, errors, ...}`, it's just a more logically-encapsulated

```
const pll = new ProcessedLatexLog();
pll.errors.push(...)
return pll;
```

### drawback

tsx has this line:

```
const content = this.props.build_logs.getIn([tool, "parse", group]);
```
which peeks right into the JS/immutable object. That doesn't work, if the object in "parse" is a class.

Hence the reason why this is split apart into interface and class, and has a trivial `toJS()` method.

williamstein:
Merged!

PR #2959, opened 2018-07-03
db: add "addons" row to project table

haraldschilly:
This is just for adding a row like `compute_image` to the projects table. It's a general purpose configuration for "academic" aspects, e.g. additional software, special product keys (for env variables), and whatever we come up in the future. First step is to have a place to store this in a project specific datastructure.

I was also thinking about using the quota system for this, but I fear that's getting too complex.

The associated branch for kucalc is `proj-edu`.

### DIFF
--- a/src/smc-util/db-schema.coffee
+++ b/src/smc-util/db-schema.coffee
@@ -792,9 +792,9 @@ schema.projects =
         compute_image :
             type : 'string'
             desc : 'Specify the name of the underlying (kucalc) compute image (default: "latest")'
-        academic :
+        addons :
             type : 'map'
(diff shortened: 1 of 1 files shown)


src/smc-util/db-schema.coffee
```diff
@@ -821,6 +824,7 @@ schema.projects =
                 action_request : null   # last requested action -- {action:?, time:?, started:?, finished:?, err:?}
                 course         : null
                 compute_image  : 'latest'
+                academic       : {}
```
williamstein: Note that action_request is also an object above, as is course, but the default returned value is null, not {}.  I think that would be better for consistency.

williamstein:
+1 to addons.

PR #2961, opened 2018-07-03
cookie banner

haraldschilly:


src/smc-hub/hub_http_server.coffee
```diff
@@ -326,6 +330,15 @@ exports.init_express_http_server = (opts) ->
                         res.type(filename)
                     res.send(data)
 
+    router.get '/cookies/banner', (req, res) ->
```
williamstein: Please

1. Change the endpoint to something like /location
2. Move all the rest of the code to the frontend.

Why?    Because we might do all kinds of things with this geolocation info, e.g., something involving GDPR, or multidatacenter projects or something else entirely.   Also, updating the list of country codes shouldn't require restarting the hubs.

williamstein: @haraldschilly responds: NOPE.  Don't see the relevant info in the frontend... So leave as is.

PR #2975, opened 2018-07-13
webapp/global banner: reactivate for K2 announcement

haraldschilly:
I'm trying to make this work again. There is a comment I don't understand, though

> We need to delete this function and instead do like in the _init of actions, i.e.,
>    set derived state based on changes to account store in AccountActions.  -- william

When I add a `_init` function to the account actions, it isn't called at all. Is such a call supposed to happen by the framework? Is this documented anywhere? If so, I haven't found it in the app-framework examples. 

So, what I ended up doing is just calling it right where it is created, in order to make it do anything.

### DIFF
--- a/src/smc-webapp/r_account.cjsx
+++ b/src/smc-webapp/r_account.cjsx
@@ -1118,7 +1118,7 @@ OtherSettings = rclass
 
     toggle_global_banner: (val) ->
         if val
-            # this must be null, not undefined. no idea why...
+            # this must be "null", not "undefined" – otherwise the data isn't stored in the DB.
             @on_change('show_global_info2', null)
         else
             @on_change('show_global_info2', webapp_client.server_time())
--- a/src/smc-webapp/r_account.cjsx
+++ b/src/smc-webapp/r_account.cjsx
@@ -1116,19 +1116,37 @@ OtherSettings = rclass
     on_change: (name, value) ->
         @props.redux.getTable('account').set(other_settings:{"#{name}":value})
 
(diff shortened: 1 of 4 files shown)


haraldschilly:
Someone should look over this. took me hours to figure out how to set it up again. My main struggle was that `redux.getTable('table name').set(key:{key2:undefined})` is different from `redux.getTable('table name').set(key:{key2:null})`. I only discovered this by accident when trying all sorts of changes. Aren't we using undefined everywhere? Or is this a bug?

williamstein:
> Aren't we using undefined everywhere? Or is this a bug?

The current behavior is absolutely right and NOT a bug, and used in many places.  Here's why:

```
> JSON.stringify({a:undefined, b:null})
"{"b":null}"
```

src/smc-webapp/r_account.cjsx
```diff
@@ -1116,27 +1116,46 @@ OtherSettings = rclass
     on_change: (name, value) ->
         @props.redux.getTable('account').set(other_settings:{"#{name}":value})
 
+    toggle_global_banner: (val) ->
+        if val
+            # this must be null, not undefined. no idea why...
```
williamstein: Please update this with a useful comment.

haraldschilly:
Today, I wrote some text for the [linked wiki page](https://github.com/sagemathinc/cocalc/wiki/Maintenance2018)

DrXyzzy:
Corrected link to wiki page: [Maintenance 2018](../wiki/Maintenance-2018)

PR #2983, opened 2018-07-20
Typescript courses

johnjeng:
Convert coffee courses to typescript courses.
Initial pass, still a lot of cruft to remove in the next pass.

For quickly reverting, change this line. https://github.com/sagemathinc/cocalc/pull/2983/commits/2e99f68cfaa3888eda753cf988774940d454a55c


johnjeng:
@DrXyzzy or @tscholl2 Can you test this PR?

DrXyzzy:
Also both times when I created a test account in cc-in-cc for this PR branch *after* adding the email address as student to a .course, a phantom "No Title" project appeared on first login by the test account in addition to the student course project.

notitle

johnjeng:
Distributing assignments seems to be broken in my project though it was working before. Hopefully it's my project.

johnjeng:
Seems like it was a slow copy issue.

DrXyzzy:
After above commits:
- no white screen after adding student
- no phantom project (but I will still open sep ticket )

New findings:
- create shared project: only 1 of 3 students was added to shared project in addition to the instructor; it was the one student who was not logged in when the shared project was created
- require students to pay - warning about payment flashes when student project is refreshed in browser, then disappears (possibly because `commercial` was not enabled in dev project admin settings)

Test report in [private file](https://cocalc.com/projects/ccd4d4a4-29a8-4c39-85c2-a630cb1e9b6c/files/TESTPLANS/pr2983.md?session=default).

williamstein:
@tscholl2  can you look this over???   I know @DrXyzzy tested it, but it would be nice to have somebody who knows typescript at least look at it before I do (and merge).

johnjeng:
@DrXyzzy Shared project behavior is expected based on the code. If the student doesn't have an account id yet, they don't get added. They won't get added until the course file configured. Behavior should be that they're added automatically upon account creation but this patch isn't about that.

I do not see the second issue when commercial is enabled.

DrXyzzy:
@johnjeng all three students had accounts and had been added to the course when I tried creating a shared project. But only one of them was logged in at the time.

johnjeng:
Huh, it worked for me even if the person wasn't logged in. I'll try to replicate fiddling with other variables.

Issue #2991, opened 2018-07-22
(high priority) user search bug


williamstein:
1. Search for "james davenport" to add as a collaborator to a project.
2. Get no results.

Do the same search as an admin; get a result!

![screenshot 2018-07-22 at 3 01 07 pm](https://user-images.githubusercontent.com/1276278/43050573-38294094-8dc0-11e8-9b15-3443d0ae1280.png)

Searching for James by his email address also works.  So something is messed up with the query.

haraldschilly:
Some code is merged, but I'm not sure if we can close it. Maybe there are additional restrictions by time encoded in the webapp. So, until this is confirmed to work, we leave it open.

haraldschilly:
this was fixed by me via another issue and two PRs.

williamstein:
Heh, this is not fixed now :-(

williamstein:
I'll fix this -- it's very easy.

PR #2993, opened 2018-07-23
db user search: search all users, not just active ones -- #2991

haraldschilly:
did three test searches directly on the db in prod, which all were below 3 ms.

### DIFF
--- a/src/smc-hub/postgres-server-queries.coffee
+++ b/src/smc-hub/postgres-server-queries.coffee
@@ -906,12 +906,13 @@ exports.extend_PostgreSQL = (ext) -> class PostgreSQL extends ext
     #     OR
     #     (created >= NOW() - $3::INTERVAL)
     #   )
+    #   ORDER BY last_active DESC NULLS LAST
     #   LIMIT $4::INTEGER
     user_search: (opts) =>
         opts = defaults opts,
             query  : required     # comma separated list of email addresses or strings such as 'foo bar' (find everything where foo and bar are in the name)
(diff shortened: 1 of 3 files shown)


williamstein:
I think I made a huge mistake having an active search window at all, honestly!!  It's really confusing.
We should remove it, or have two options to the search api, and let the user search for "possibly inactive users".  It's really good to have events that trigger activation of inactive users, after all.  Actively discouraging it is stupid.    So I don't like this PR....

haraldschilly:
another proposal: what I would do is to order by last_active, descending, and the null values should be at the end (null means they're very old, right?), without a constraint, but still have a limit.  that way, dormant accounts automatically fall off the radar if there are too many matches, or otherwise -- if unique enough or a more precise search -- old accounts are still found. 

example query for myself

```sql
SELECT first_name, last_name, last_active, created
FROM accounts
WHERE lower(last_name) = 'schilly'
ORDER BY last_active DESC NULLS LAST
LIMIT 10;
```

```md
 first_name | last_name |        last_active         |         created
 ------------+-----------+----------------------------+-------------------------
 Harald     | Schilly   | 2018-07-23 19:28:39.46267  | 2015-04-10 20:06:29.921
 Harald     | Schilly   | 2017-10-26 10:40:38.740904 |
 Harald     | Schilly   |                            |
 Harald     | Schilly   |                            |
```

query takes longer, I checked a few, 80 to 100ms, but that's still fine I guess.

williamstein:
Sounds good to me!

PR #3004, opened 2018-07-26
Lean Support

williamstein:


williamstein:
[My private todo list](https://cocalc.com/projects/10f0e544-313c-4efe-8718-2142ac97ad11/files/lean.tasks?session=default)

Issue #3018, opened 2018-07-31
latex force rebuild does not force reload the PDF preview


haraldschilly:
this is a bit tricky to explain, but here is an attempt:

1. build .tex file, which contains a custom .sty file. at the bottom is an example just for testing.
2. building again doesn't change anything.
3. modify the .sty file
4. building again doesn't change (expected, and related to #3017 )
5. clicking "force rebuild" in the build dialog does cause a rebuild. the bottom of the output clearly says
   ```
   Rule 'pdflatex': File changes, etc:
   Changed files, or newly in use since previous run(s):
      'myownstyle4.sty'
   ```
6. but the PDF preview doesn't change! switching to PDF native shows the changes, though.

expected: after a force rebuild, also the PDF is reloaded without any additional conditions... (I haven't checked the code, though)

---

tex file

```
[...]
\usepackage{myownstyle4}
\begin{document}
hi: \hi{}
\end{document}
```

`myownstyle4.sty` in the same directory

```
\NeedsTeXFormat{LaTeX2e}[1994/06/01]
\ProvidesPackage{mypackage}
  [2011/01/11 v0.01 LaTeX package for testing]
\def\hi{FOO BAR}  % change FOO BAR
\endinput
```

haraldschilly:
hmm, I experimented a bit around, thought something like adding `this._forget_pdf_document();` to the `force_build` function helps, but it doesn't. I also saw that forget-function is called in `run_latex`. However, maybe it's just an oversight to not forcefully refresh the view, i.e. that `set_reload` method? Can we improve the documentation of  `set_reload`?

haraldschilly:
PR #3019 solely implements updating the PDF when there is a forced rebuild  … hence I think this ticket can be closed

PR #3019, opened 2018-07-31
latex: force build triggers a force reload of the pdf preview 

haraldschilly:
see comments in #3018


haraldschilly:
thank you for looking into this.

Issue #2997, opened 2018-07-25
*.Rtex files should be handled similarly to *.Rnw


haraldschilly:
this is a simple compatibility fix for these files, e.g. coming from overleaf. 

besides registering the editor with the files, there are maybe some adjustments necessary for properly replacing the file extension in the knitr specific commands.

~~it's also a bit weird, because technically the Rnw and Rtex syntax is different, but I have Rtex files which have the same syntax as Rnw. Maybe there is just no real distinction?~~ no, there is, the knitr command seems to behave differently based on the file extension.

reference file: [005-latex.Rtex](https://github.com/yihui/knitr-examples/blob/master/005-latex.Rtex) for testing.

haraldschilly:
done. a straightforward generalization of Rnw to also include Rtex. In the end, Knitr just looks at the extension to know what to do.

PR #3022, opened 2018-07-31
support for .Rtex files, generalizing what was done for .Rnw

haraldschilly:
see #2997

this generalizes .Rnw extra logic to also work for .Rtex files.

testing:webapp: 

1. create a .Rtex file (smc_pyutil contains a suitable example)
2. check that the pdf preview eventually shows up (this here patchs a utilitiy function, which derives the pdf filename)
3. forward/inverse search should work
4. R code syntax errors should show up in the error box


haraldschilly:
note for review: install the updated pyutil package in your cc-in-cc project via `cd smc_pyutil; /usr/bin/pip install --user --upgrade ./`. Then restart the project and/or the project hub. That way, when you click on the entry in the file dropdown button (in + add) for "R Knitr Rtex", that `default.rtex` template will be picked up instead of an empty file. Afterwards, `/usr/bin/pip uninstall smc-pyutil psutil pyyaml` ...

williamstein:
Can you fix this:

1. Open any new .rtex file
2. See this error for a few seconds: image

Otherwise, it totally works.

This is so good we should merge it now, even with this bug.

PR #3025, opened 2018-08-01
webapp/latex: attempt to fix changing build command

haraldschilly:
I spent some time trying to fix editing the latex editor command, but I completely failed and I am giving up. The react doc says `componentWillReceiveProps` is deprecated and shouldn't be used. From the docs I think `getDerivedStateFromProps` is what should work.

What's right now also an issue is to do this: First, ctrl-a to select the whole command, then hit delete + return. If that's quick enough, the command is the empty string and you're stuck with an infinite "Loading …".

That's why I also added lines to set the state of the command to default. But thinking about that, maybe it's ok to have an empty string and then do nothing on the backend (equivalent to disabling building at all), but in that case, there should still be a textbox and the dropdown button.

### DIFF
--- a/src/smc-webapp/frame-editors/latex-editor/build-command.tsx
+++ b/src/smc-webapp/frame-editors/latex-editor/build-command.tsx
@@ -32,25 +32,47 @@ interface Props {
 
 interface State {
   build_command: string;
+  prev_build_command?: string;
   focus: boolean;
 }
 
 export class BuildCommand extends Component<Props, State> {
   constructor(props) {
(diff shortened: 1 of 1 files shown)


williamstein:
Since this is clearly confusing, we should just change the UI to be way more straightforward and less clever.  I had a bad feeling when I wrote this in the first place.

williamstein:
@tscholl2 and I just did this properly, and pushed it to master, and made it live. (Or will in a moment)

PR #3033, opened 2018-08-02
UI selector to choose underlying project image

haraldschilly:
the whole gist of this is to make DB changes like that, where the allowed values are validated and users also have some idea about what they select.

```
smc=# select title, project_id, compute_image from projects where project_id = 'bc6f81b3-25ad-4d58-ae4a-65649fae4fa5';
  title  |              project_id              | compute_image
---------+--------------------------------------+---------------
 test123 | bc6f81b3-25ad-4d58-ae4a-65649fae4fa5 | default
(1 row)
 ```

click, select, confirm, restart, ... and then

```
smc=# select title, project_id, compute_image from projects where project_id = 'bc6f81b3-25ad-4d58-ae4a-65649fae4fa5';
  title  |              project_id              | compute_image
---------+--------------------------------------+---------------
 test123 | bc6f81b3-25ad-4d58-ae4a-65649fae4fa5 | old
```

what's missing is to only show this in the UI when the project runs in the kucalc system. what's the canonical way to do this? I saw some lines about a `@props.kucalc`, but I'm not sure.

### CODE REVIEW

**Comment by williamstein** [comment]
Please delete all the changes you made to client.coffee, messages.coffee, etc.  The point of our entire "new" (as of 2016?) db-schema/postgresql/user query functionality is that users can modify a field in a table in the database by simply making a user query.   If you need to put guaranteed constraints on that change, that code goes as a hook in db-schema.coffee.   I thought it was really dumb in the old code adding functions to client.coffee, new messages, new code to the database, etc... all redundant.  So there is a much better way and I took great joy in deleting all that old code.

**Comment by haraldschilly** [comment]
well, I don't know how anything about that "query" functionality, is there a documentation somewhere? how do these hooks work?

There is also nothing backed by a database, because I am not sure where to set or get the data.

**Comment by williamstein** [comment]
> well, I don't know how anything about that "query" functionality, is there a documentation somewhere? how do these hooks work?

It's all in db-schema.coffee.   (docs, code, examples where this is used).

> There is also nothing backed by a database, because I am not sure where to set or get the data.

This is ALL backed by a database, not nothing.  This PR is full of code like [here](https://github.com/sagemathinc/cocalc/pull/3033/files#diff-9361843c2e6938b6e05eacdd2b644f98R2294) that is interacting with the database.

If this is really confusing, I can probably rewrite it myself sometime this week...

Maybe you can look at the code for things like starting a project, etc.,?  It *all* works by "user sets something in database via a query as defined in db-schema.coffee, and something else responds to this change".   This is kind of the *core architectural idea* of CoCalc's backend, and is the reason it is possible to run the same code in cc-in-cc and Kubernetes...

File: src/smc-webapp/project_actions.ts [hunk 1]
```diff
@@ -2647,6 +2649,17 @@ export class ProjectActions extends Actions<ProjectStoreState> {
   close_free_warning(): void {
     this.setState({ free_warning_closed: true });
   }
+
+  async set_compute_image(new_image: string): Promise<void> {
```
williamstein: Harald -- this one little function *replaces most of the code* you added to smc-util/client.coffee, message.coffee, smc-hub/client.cofee, postgres-server-queries.coffee.   I spent a huge amount of time figuring out how to do this sensibly, rather than having to edit 5 files to do anything.

src/smc-util/compute-images.coffee
```diff
@@ -0,0 +1,19 @@
+{defaults, required} = require('smc-util/misc')
```
williamstein: I just moved this to smc-util, so COMPUTE_IMAGES is already compiled into the client.  This is better since:

- there isn't an async race condition when querying for the images to put the menu entries in project settings,
- less load on the hub servers.
- it's far *easier* to update the client frontends than the hubs, which we basically rarely want to restart, since that destabilizes things for all users, and increased db load a lot. 
- I bet the list of compute image tags won't change that often, but if it does, we'll have a better solution.

haraldschilly: well, I wanted these entries to be either in the database (that's what I meant with that they aren't there yet) and there might be new ones. The list here is just very generic, and e.g. "old" makes no sense right now. What I wasn't sure is if it is a good idea to put it into the database at all. A much more direct way would be to query the image server.

src/smc-webapp/project_actions.ts
haraldschilly: looks nice, what I don't understand is how this checks that the image name is valid.

Issue #3043, opened 2018-08-03
markdown: emojis should be ignored in code blocks (and other places?)


williamstein:
1. Put this in a chat message (or anywhere that does markdown):
<pre>
```
:-)
```
</pre>

2. See this mess:

![image](https://user-images.githubusercontent.com/1276278/43663011-26d8e218-9725-11e8-8dab-32368957afc5.png)

---

What should happen is what happens above with github...

haraldschilly:
we'll might want to deprecate what we have and also introduce using an emoji picker like http://blog.tommoor.com/emojione-picker/

williamstein:
We should switch to using unicode for emoji, which just makes this problem go away.

williamstein:
There's a similar bug in math mode:

![image](https://user-images.githubusercontent.com/1276278/103972348-ab9e9700-5121-11eb-9353-646f1235b951.png)

williamstein:
Fixed in 0d766b118215257c66ce4a7bfadcdd72be7bbe7a, which isn't merged into master yet, but will be soon-ish.

williamstein:
That's a commit in this branch: https://github.com/sagemathinc/cocalc/pull/5147

PR #3045, opened 2018-08-03
use markdown it emoji for smilies

haraldschilly:
this fixes the situations outlined in #3043 … but it can't fix all our emojis, because some of our current ones contain markdown text indicators (underlines). I'm rather for removing what we have and switch to that library.

screenshot from 2018-08-03 23-11-57


haraldschilly:
what's left is to do

- [ ] decide, if we run it for all markdown, or only the ones for chatting (which means to add a boolean parameter)
- [ ] enable more smilies [upstream shortcut.js file](https://github.com/markdown-it/markdown-it-emoji/blob/master/lib/data/shortcuts.js)
- [ ] probably change `:-\` to something else, because of the comment in shortcuts.js … although not sure what this could be
- [ ] we wrap them in some extra ´<span ...>` tags, maybe for padding and font size (?) ... [that's described here](https://github.com/markdown-it/markdown-it-emoji#change-output)
- [ ] testing

williamstein:
- [ ] decide, if we run it for all markdown, or only the ones for chatting (which means to add a boolean parameter)

ALL markdown.  I don't want to have confusion due to inconsistency :-)

- [ ] enable more smilies [upstream shortcut.js file](https://github.com/markdown-it/markdown-it-emoji/blob/master/lib/data/shortcuts.js)

Do that in *another* PR.

- [ ] probably change `:-\` to something else, because of the comment in shortcuts.js … although not sure what this could be

Just remove it.  I don't even know what it is. :-\

- [ ] we wrap them in some extra ´<span ...>` tags, maybe for padding and font size (?) ... [that's described here](https://github.com/markdown-it/markdown-it-emoji#change-output)

Do this in a followup PR.

- [ ] testing

Done by the reviewer.

williamstein:
And another todo for another PR: delete misc.smiley.

williamstein:
Three problems so far in testing:

Doesn't work when preview enabled:
screenshot 2018-08-04 at 8 23 33 am

My fav emoji -- :shrug: is gone:
screenshot 2018-08-04 at 8 22 53 am

Side chat doesn't work yet:
screenshot 2018-08-04 at 8 21 01 am


---

Github solves this problem via unicode and a completions menu... 🤷‍♂️

tscholl2:
I searched for "smiley" in the code, this shows a few other places where the old smiley code is being used:

image

williamstein:
This whole approach to emojis is very bad compared to the solution that Github uses, which is "just make it easy to insert unicode".  That's what we should definitely do... at some point.  It's vastly better, since it requires no markdown extension at all, so it will work with vanilla Jupyter, github, etc.,   Whereas what we are doing here produces docments that only work in cocalc, but won't work when exported to most any other renderer.  The fact that we have so many questions about which emoji's to support, just proves that most markdown renderers will fail on this.  

If we support unicode, you can write emoji's in an .md file when making a blog post (say), and it will work!

So don't put much more effort into this, IMHO, except to fix the bug I reported.

williamstein:
To be clear: I'm fine with this markdown-it approach for now, but we should aim for unicode.

williamstein:
I said:
> This whole approach to emojis is very bad

but @haraldschilly set me straight in video chat.  I'm maybe wrong.

Issue #369, opened 2016-01-14
support init.sage in sage worksheets sagews


williamstein:
It seems  

      /<project>/.sage/init.sage

works only for command-line sage.   It should also work in Sage worksheets.

Reported by: Pedro Cruz

PR #3052, opened 2018-08-06
Issue/369

DrXyzzy:
Ref: #369 

Add `SAGE_STARTUP_FILE` (defaults to `~/.sage/init.sage`) processing to sage_server.

Commits with "baseline" in comment were to get pytest suite working with current CoCalc before `sage_server` changes in this PR.

Pytest cases added in `tests/test_env.py`. To test, apply patch for this PR, then:
```
python -m pytest
==>
= 172 passed, 5 skipped in 431.19 seconds =
```

### DIFF
--- a/src/smc_sagews/smc_sagews/sage_server.py
+++ b/src/smc_sagews/smc_sagews/sage_server.py
@@ -1038,6 +1038,8 @@ class Salvus(object):
                             # BUGFIX: be careful to *NOT* assign to _!!  see https://github.com/sagemathinc/cocalc/issues/1107
                             block2 = "sage.misc.session.state_at_init = dict(globals());sage.misc.session._dummy=sage.misc.session.show_identifiers();\n"
                             exec compile(block2, '', 'single') in namespace, locals
+                            b2a = "try: load(os.environ['SAGE_STARTUP_FILE'])\nexcept: pass\n"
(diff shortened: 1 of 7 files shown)


src/smc_sagews/smc_sagews/sage_server.py
```diff
@@ -1038,6 +1038,8 @@ def execute(self, code, namespace=None, preparse=True, locals=None):
                             # BUGFIX: be careful to *NOT* assign to _!!  see https://github.com/sagemathinc/cocalc/issues/1107
                             block2 = "sage.misc.session.state_at_init = dict(globals());sage.misc.session._dummy=sage.misc.session.show_identifiers();\n"
                             exec compile(block2, '', 'single') in namespace, locals
+                            b2a = "try: load(os.environ['SAGE_STARTUP_FILE'])\nexcept: pass\n"
```
williamstein: If init.sage has an exception for any reason, e.g., something goes wrong when it is loading, will this code you wrote make that error be silently hidden?   E.g., if init.sage contains `1/0`, then Sage should show an error while starting up.

I will merge this as is, assuming it works, but please open a followup issue about silently ignoring errors in init.sage.

williamstein:
Follow up for you: https://github.com/sagemathinc/cocalc/issues/3056

Issue #3050, opened 2018-08-05
Tab completion in Jupyter does not put cursor at end of line when refining the search


kevinywlui:
To reproduce:
1. Start a new Jupyter notebook.
2. In the first cell, type Ab[Tab]e[Enter]
3. The cursor is now after the b instead of the end of line.

haraldschilly:
I tested this. First, I wasn't able to reproduce. Then, it became clear that it's crucial to do some refinement of the search by entering another character. It looks like entering one or more additional characters for the completion search sets the offset for advancing the cursor position too early (or sets it to 0). Once that was clear it's easy to reproduce.

haraldschilly:
@kevinywlui this should be fixed (maybe you have to refresh your cocalc tab in the browser to get the update).

PR #3053, opened 2018-08-06
webapp/jupyter: fix completion after pressing a key

haraldschilly:
ref #3050, which also explains how to test


haraldschilly:
I worked with the debugger to see this. What I find really odd is that the `id` is `"__!!!!!!this is a required property!!!!!!__"`. That doesn't look ok. Closing/opening again and it works fine. So, I don't think that patch introduces this anomaly.

screenshot from 2018-08-06 17-13-53

PR #3063, opened 2018-08-08
Persist grading

### LINKED ISSUE #3049: when grade is entered it sometimes vanishes from the form
Reported by Gabriela Dorfman Furman:

In a course, after distributing an assignment:
1. Click the "Collect .." button (task)
1. Try to write in the window under the caption "Enter Grade" (where it says "Grade (any text) ...")
1. After a few seconds the grade is deleted (like nothing is written) or the save button does not work

Sometimes, if the bug does not appear at first, you can hit tab to enter a comment, then shift-tab to go back and enter new text for the grade, and _that_ grade vanishes after a second or so.

Seen with browsers: 
- Chrome 67 on Windows
- Safari 11.0.3 on macOS

May be related to #2928 - grade & comments are not retained on tabbing away and back.

johnjeng:
Persist grading assignments and fix #3049.

### DIFF
--- a/src/smc-webapp/course/actions.ts
+++ b/src/smc-webapp/course/actions.ts
@@ -1508,7 +1508,7 @@ export class CourseActions extends Actions<CourseState> {
 
     const open_grades = store.get("active_feedback_edits");
     const new_open_grades = open_grades.delete(
-      assignment.get("assignment_id") + student.get("student_id")
+      assignment_identifier(assignment, student)
     );
     this.setState({ active_feedback_edits: new_open_grades });
   };
@@ -1535,7 +1535,7 @@ export class CourseActions extends Actions<CourseState> {
       current_edited_comments = current_edited_feedback.get("edited_comments");
     }
 
-    const grade: string;
+    let grade: string;
     if (new_edited_grade != undefined) {
       grade = new_edited_grade;
     } else if (current_edited_grade != undefined) {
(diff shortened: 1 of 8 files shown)


src/smc-webapp/course/actions.ts
```diff
@@ -1492,42 +1497,104 @@ export class CourseActions extends Actions<CourseState> {
...
-    grades[student.get("student_id")] = grade;
-    (obj as any).grades = grades;
-    this._set(obj);
-  }
 
-  set_comments(assignment, student, comments) {
-    let left;
+    const open_grades = store.get("active_feedback_edits");
+    const new_open_grades = open_grades.delete(
+      assignment.get("assignment_id") + student.get("student_id")
```
haraldschilly: I'm curious. Should this line maybe call `utils::assignment_identifier` to semantically express that this is the common key used everywhere?

johnjeng: Oh, yup. Missed that one.

src/smc-webapp/course/actions.ts
```diff
@@ -1544,14 +1544,14 @@ export class CourseActions extends Actions<CourseState> {
...
+    let comments: string;
     if (new_edited_comments != undefined) {
       comments = new_edited_comments;
     } else if (current_edited_comments != undefined) {
       comments = current_edited_comments;
     } else {
-      comments = store.get_comments(assignment, student) || ""
-
+      comments = store.get_comments(assignment, student) || "";
+    }
```
johnjeng: This got deleted somehow. Oops.

PR #3070, opened 2018-08-10
webapp/project files tab: add slight border around tabs

haraldschilly:
a little experiment how a slight border around file tabs looks like

screenshot from 2018-08-10 13-39-56


williamstein:
@johnjeng thoughts?  I'm not convinced this is a good idea.  There is a string trend away from borders that aren't really needed...  E.g., in new chrome there is just a little divider between tabs:

image

And in this very giithub box I'm typing in now, the preview tab has no border around it at all:

image

williamstein:
I'm closing this.  Somebody can re-open it if they have thoughts or endorse it.

PR #3079, opened 2018-08-14
improve key for caching connected components

haraldschilly:
In the `rclass` of the app framework, there is a cache for connected components, which is based on the keys of passed in redux props. imagine, the keys are `barbar` + `foo` vs. `bar` + `barfoo`:  they are the same `barbarfoo` key.

Here, I played with using a common [32 bit hash function](https://en.wikipedia.org/wiki/MurmurHash), but well, by default this here uses a json serialization. I haven't done any timings.


williamstein:
A 32-bit hash function has a very high probability of collision.  It's trivial to find hundreds of collisions in a few minutes of code running.    This is nothing like sha1 (what git uses) or uuidv4s, where the odds of a collision are astronomically small.

Basically, murmor3 = "definitely a hash collision if you do a million hashes", whereas uuid's and sha1 -- the sort of thing I can live with -- is "likely a hash collision if you do 10**20 hashes".

So can you delete the code using murmor3?   (Or save it in a different branch.)

src/smc-webapp/app-framework.ts
```diff
@@ -457,24 +456,9 @@ x.actions must not be defined.
...
 
-type cache_key_methods = "slow" | "fast";
-
-const compute_cache_key = function(
-  data: immutable.Map<string, any>,
-  hash_method?: cache_key_methods
-): number | string {
-  const method: cache_key_methods = hash_method || "slow";
+const compute_cache_key = function(data: immutable.Map<string, any>): string {
   const keys = misc.keys(data).sort();
```
williamstein: misc.keys does not work correctly on immutable Map:

image


The problem though here is that data is not of type `immutable.Map<string, any>`.  It's a normal object.  There's a different way to declare a normal mutable map.

williamstein: Also, I wrote a typescript version of misc.keys in frame-editors/generic/misc.ts... though it's just this one liner:
```
export const keys = Object.keys;
```

williamstein:
I'll just merge this and fix the mistake myself.  I hope you saw the comment.

PR #3081, opened 2018-08-14
Update project collab settings

tscholl2:


williamstein:
NOTE TO SELF: Remember to restart the hubs (hub-user-query) before making this live, since the query depends on a new vritual table!

PR #3084, opened 2018-08-15
Fix double directory fetch

johnjeng:
Prevent double directory fetching inside certain actions


src/smc-webapp/project_actions.ts
```diff
@@ -1130,7 +1133,7 @@ export class ProjectActions extends Actions<ProjectStoreState> {
         if (store.get("active_project_tab") === "files") {
           this.set_url_to_path(path);
         } else {
-          this.set_active_tab("files");
+          this.set_active_tab("files", {update_file_listing: false});
```
williamstein: Why do you NOT want to update the file listing here?

src/smc-webapp/project_actions.ts
```diff
@@ -352,7 +352,7 @@ export class ProjectActions extends Actions<ProjectStoreState> {
   //            or a file_redux_name
   // Pushes to browser history
   // Updates the URL
-  set_active_tab(key: string): void {
+  set_active_tab(key: string, opts: {update_file_listing: boolean} = {update_file_listing: true}): void {
```
williamstein: 100% of other users of opts in CoCalc have opts as the *only option* to the function.  Don't missing positional args and opts.  Use one or the other.    This is also a weird use of opts, where this is only one option?

williamstein:
It doesn't work:

screenshot 2018-08-31 at 7 48 17 am

src/smc-webapp/project_actions.ts
johnjeng: Because `set_current_path` also updates the file listing.

PR #3085, opened 2018-08-15
Issue/3083

tscholl2:
Fixes 3083

### DIFF
--- a/src/smc-webapp/_examples.sass
+++ b/src/smc-webapp/_examples.sass
@@ -48,7 +48,7 @@ $code-height: 230px
           border-radius : 0
 
     div.webapp-examples-descr,
-    pre.code
+    .webapp-examples-code>div>pre
       height            : $code-height
       overflow          : auto
       border-color      : $COL_GRAY_DDD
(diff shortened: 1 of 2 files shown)


williamstein:
But I can't test the code, since the assistant is broken in my master dev/review project... ?

haraldschilly:
I'll test this. I'm wondering about adding a language and forgetting to update that dictionary. Is codemirror robust against an undefined language?

williamstein:
> I'm wondering about adding a language and forgetting to update that dictionary.

I wonder too -- at least it should fail gracefully.  I'm OK with a fix for doing things right (we should just directly use file_associations right, rather than copy from it!) being in a follow-up PR, since having syntax highlighting only some of the time is OK.

tscholl2:
You can comment out the line where it passes the language and find out. I believe because of:

https://github.com/sagemathinc/cocalc/blob/master/src/smc-webapp/jupyter/codemirror-static.tsx#L83

It will render any new language with python syntax highlighting. This matches what I observed while developing.

williamstein:
> with python syntax highlighting

OK, probably better than nothing.

Issue #3092, opened 2018-08-17
jupyter2 bug supporting clear_output


haraldschilly:
The `clear_output` command doesn't work correctly in jupyter2. This is maybe a bit similar to #2520? Here is some code:

```
from IPython.display import clear_output
print(1)
clear_output()
print(2)
clear_output(wait=True)
```

and that's how it looks in python 3

![screenshot from 2018-08-17 12-06-09](https://user-images.githubusercontent.com/207405/44260885-2778ab80-a216-11e8-828e-6e79af526342.png)

also note, that these red boxed do not pass validation when opened in classical jupyter.

the expected behavior is 

![screenshot from 2018-08-17 12-09-10](https://user-images.githubusercontent.com/207405/44260962-673f9300-a216-11e8-9f38-2fb7c620ee7a.png)

haraldschilly:
this is fixed via #3093 which also properly sets all other message types. i.e. maybe this fix did also clean up some flakiness around run status information of cell execution.

PR #3093, opened 2018-08-17
jupyter2/project:

haraldschilly:
ref:  #3092

root cause: look closely at the changed lines in execute-code: `msg_type` was always undefined, because earlier the `mesg` object was already stripped of the `header` dict.

I also added types in another file I sifted through via debugging, I hope that's fine.

test:

```
from IPython.display import clear_output
from time import sleep

for i in range(10):
    clear_output(wait=True)
    print('.' * i)
    sleep(.1)
```

which should just show 10 dots in the end.


williamstein:
Very good solution.  This was my fault, I think.

PR #3106, opened 2018-08-20
highlight and select collaborators by clicking

tscholl2:


williamstein:
Damn, this is nice :-)

PR #3107, opened 2018-08-20
websockets jupyter fast

williamstein:


williamstein:
NOTE -- I will probably delete this branch and start over from scratch on this problem.

PR #3108, opened 2018-08-21
running preflight checks when cocalc /app starts

haraldschilly:
WIP: this patch is "doing something", but it's not pretty. Sometimes, the browser still works, although it's not ES6 compatible. I think we could let that slip through and still load the page? Hmm, but I pushed a small change to stop loading the page altogether. So, the first few screenshots below still show the loading banner. After ad462408db7c2c0a87a78730d50974d0e86945df this is not always the case.

---

firefox 45 on win 8:

screenshot from 2018-08-21 15-45-35

---

Safari 8 on OS X Yosemite

screenshot from 2018-08-21 15-49-10

---

Chrome 52 on Windows 8.1

screenshot from 2018-08-21 15-52-19

---

Opera 40 on macOS Sierra

screenshot from 2018-08-21 15-54-30

---

Internet Explorer 11 on Windows 7

screenshot from 2018-08-21 16-08-25

---

Opera 46 on OS X Yosemite

screenshot from 2018-08-21 16-11-23


haraldschilly:
did some design and linking to a wiki page. right now in my chrome it looks like the screenshot below.

you can trigger it by editing the preflight check file and requiring an absurdly high version number, e.g. `const oldChrome = spec.name === "Chrome" && spec.version < 162`

what's missing for me is a  link to "continue on your own caution". I would append a URL query param, which bypasses that check.

screenshot from 2018-08-27 20-56-11

williamstein:
I don't know how to re-open this, but I force removed this from master.  It's very broken.

PR #3113, opened 2018-08-22
jupyter: format code cells using existing prettier infrastructure

haraldschilly:
this is a proof of concept how this could work. I was also curious how the API is implemented. To explain what's going on look at the before & after screenshots below. I did add the "needs review" label, which is actually "Please give me some early feedback about what I'm doing here". Should we add such a label in orange? I'll add it.

What's missing:

- [ ] am I using the API correctly?
- [ ] yes, `_format_cell` returns something, I'm aware of that ... but the limit function doesn't want a void promise.
- [ ] button is a stub, obviously
- [ ] add a menu entry to format all cells
- [ ] deal with non-python code (e.g. show a warning message to the user)

before: all 4 cells are selected

screenshot from 2018-08-22 13-50-36

after: 

screenshot from 2018-08-22 13-50-50

### DIFF
--- a/src/smc-project/browser-websocket/api.ts
+++ b/src/smc-project/browser-websocket/api.ts
@@ -42,12 +42,19 @@ export function init_websocket_api(
   });
 }
 
-async function handle_api_call(client: any, data: any, primus:any, logger:any): Promise<any> {
+async function handle_api_call(
+  client: any,
+  data: any,
+  primus: any,
+  logger: any
+): Promise<any> {
   switch (data.cmd) {
     case "listing":
       return await listing(data.path, data.hidden);
     case "prettier":
       return await prettier(client, data.path, data.options);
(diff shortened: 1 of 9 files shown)


haraldschilly:
sage code in jupyter is close to python, but still a syntax error when fed through yapf. a half-half trick could be to attempt formatting it and suppressing an eventual error (and do nothing) otherwise. that's how it looks:

screenshot from 2018-08-28 17-46-42

apart from that, formatting markdown always works, regardless of the kernel language.

---

the new menu entries are following my gut feeling

screenshot from 2018-08-28 17-48-27

---

python syntax errors aren't too informative, but more than nothing

screenshot from 2018-08-28 17-49-11

---

R syntax errors also show up

screenshot from 2018-08-28 17-51-53

PR #3116, opened 2018-08-22
format html files using tidy

haraldschilly:
WIP


williamstein:
> WIP

Can't wait until you finish this, since I'll definitely use it, as will my wife!

PR #3117, opened 2018-08-22
latex: fix #3111 edge case

### LINKED ISSUE #3111: latex: build command is empty, causing build process being stuck
syncdb is `{"key":"build_command","value":"","time":1533135594745}` ... which might be a leftover from when I tested this a while ago...

exception in console:

```
smc-4011c55f6b6d7f825412.cacheme.js?4011c55f6b6d7f825412:5 ERROR  TypeError: Cannot read property 'indexOf' of undefined
    at c (smc-4011c55f6b6d7f825412.cacheme.js?4011c55f6b6d7f825412:5)
    at Object.<anonymous> (smc-4011c55f6b6d7f825412.cacheme.js?4011c55f6b6d7f825412:5)
    at smc-4011c55f6b6d7f825412.cacheme.js?4011c55f6b6d7f825412:5
    at Object.next (smc-4011c55f6b6d7f825412.cacheme.js?4011c55f6b6d7f825412:5)
    at i (smc-4011c55f6b6d7f825412.cacheme.js?4011c55f6b6d7f825412:5)
e.(anonymous function) @ smc-4011c55f6b6d7f825412.cacheme.js?4011c55f6b6d7f825412:5
21:10:07.889 
```

and it stays like that forever:

screenshot from 2018-08-21 22-06-01

I think the problem must be in the `build_command_string` … probably a quick fix.

haraldschilly:
see #3111 

what I did to test this:

* scenario A (sanitization)
  1. vim-edit the `.foo.tex.syncdb` file to have `""` as the build command
  2. open the tex file
  3. check that it builds and the build command in syncdb is the default

* scenario B (changes)
  1. either vim-edit the syncdb file or open up the page twice
  2. changes on one instance properly show up on the other side and vice versa


src/smc-webapp/frame-editors/latex-editor/actions.ts
```diff
@@ -142,23 +142,32 @@ export class Actions extends BaseActions<LatexEditorState> {
...
+    const set_cmd = (cmd: string | undefined): void => {
+      if (cmd === undefined || cmd.length === 0) {
+        const default_cmd = build_command(
+          "PDFLaTeX",
+          path_split(this.path).tail,
+          this.knitr
+        );
+        this.set_build_command(default_cmd);
+      } else {
+        this.setState({ build_command: fromJS(cmd) });
```
williamstein: The fromJS doesn't make sense if cmd is a string, as the type suggestions in line 146.  So... did you misdeclare the type or what happened?

williamstein:
I tested this:
1. Make a new document.
2. Total fail. 

It doesn't even build it at all. This PR completely breaks latex.

PR #3122, opened 2018-08-25
Rewrite Terminals in React

williamstein:


williamstein:
[My internal todo list for this](https://cocalc.com/projects/45f4aab5-7698-4ac8-9f63-9fd307401ad7/files/terminal-react.tasks?session=default)

PR #3129, opened 2018-08-28
Preflight checks v2

haraldschilly:
take two of https://github.com/sagemathinc/cocalc/pull/3108

the problem was no errors were reported for such a file -- only upon the first webpack run.  I just restarted it again and I swear there are no errors again. I'll test on k2 later today …


haraldschilly:
ok, and here in its final form

old opera, showing the skip link

screenshot from 2018-08-28 15-31-38

and here is how it looks after skipping the check in "Firefox 49 on macOS Sierra" -- if you want to shoot yourself in your foot, be our guest :gun: 

screenshot from 2018-08-28 15-33-24

PR #3133, opened 2018-08-29
Control course compute image

johnjeng:
Needs testing on Kucalc I think? 

I tested the UI and changing the compute image seemed to work.


DrXyzzy:
1. When I try changing the compute image from .course file Configuration tab, a strange red error box shows up at upper right:

screenshot 2018-08-29 21 43 50

2. I was logged in as student from a separate device. As instructor, I changed image from `Default` to 
`Old image` by using the dropdown under .course: `Student projects control` > `Selected image`. The instructor project restarted with the Old image, but no change appeared in the student project, which remained on the Default image. I tried restarting the student project from the project settings tab of the student session, and the student course project came back in the Default image.

3. After the student image was changed in .course to `Old image` as above, I added a new student. I created the student account and had student logged in on second device, with no projects. I added the second student in the .course. The second student had a course project created, but it used the Default image, not the Old image as specified in the .course.

4. In the .course under `Student projects control`, I also tried `Stop all` to stop all student projects. The status displayed was `2 of 2 student projects currently running.` Clicking Stop all and then clicking it again in the confirmation dialog had no effect. The confirmation dialog disappeared immediately. There was no status update indicating anything was happening, i.e. it still showed both student projects running. And I had one of the students logged in and viewing the student course project. It did not restart.

williamstein:
@hal, I assume all of the above was on https://test.cocalc.com

DrXyzzy:
yes all tests were on test.cocalc.com

williamstein:
And 4. Please put a sequence of steps that are exactly what you did to test this before you set it to "needs review", so I can walk through those same steps.    Normally it is obvious with PR's how to test, but I'm really puzzled about how basically this doesn't work at all for a second time -- since surely you are doing a sequence of tests before setting it to "needs review" and it passes them, right? (otherwise, you are wasting our time.)   If you said you did (a), (b) and (c), and they don't work at all for me, then maybe I would realize that I'm testing it wrong.

johnjeng:
In this case I had only tested that the UI displayed correctly (since there was a refactor and conversion to typescript) and that changing the image didn't break everything. I wasn't sure if I had to test on Kucalc hence the question. 

>The database for projects has one field compute_image that should get set. That's all that matters, and means this PR can be fully and safely tested in either kucalc or cc in cc. Do this in a terminal to check:

This is the information I was looking for. In retrospect I suppose the correct label is "Needs-help". I'll keep that in mind for the future!

>the act of restarting the project containing the .course will definitely break this image stuff having any impact at all, and probably even make it so the .course file isn't saved (?). It's a sort of catch 22.

Oh yup, very good point.

>Instead, if they are out of sync, show a message about this.

I'll give them a button to sync it inside the error message.

>Move this to the very bottom

👍

haraldschilly:
I've merged this and well, there were several conflicts. I tried to resolve this, but it's not trivial.

opening the project settings, I get

`: WARNING: redux spec is invalid because it contains 'undefined' as a key. {"undefined":{},"customize":{}}`

```
he above error occurred in the <Connect(ProjectControlPanel)> component:
    in Connect(ProjectControlPanel)
    in Unknown (created by ProjectSettings-ProjectSettingsBody)
    in div (created by Col)
    in Col (created by ProjectSettings-ProjectSettingsBody)
    in div (created by Row)
    in Row (created by ProjectSettings-ProjectSettingsBody)
    in div (created by ProjectSettings-ProjectSettingsBody)
    in ProjectSettings-ProjectSettingsBody (created by Connect(ProjectSettings-ProjectSettingsBody))
```

Upside: the project image control component shows up in the course config. It's only for the officially provided images, though.

PR #3143, opened 2018-08-31
enhance rmarkdown

haraldschilly:
the goal of this is to offer proper "rmarkdown" support (which has some extra features) instead of just low-level "knitr" support

changes

- code format button in editor buttonbar for Rmd (it's markdown after all!)
- support for frontmatter in markdown (e.g. useful for our blog) and rmd files
- building rmd files slightly different, by using a bit of frontmatter info to set a parameter. that's just to help users having a better experience.
- building it itself uses the last time of file saved as a timestamp (I copied this from latex) and a bit of debouncing
- more panels for Rmd, in case someone opts into generating a PDF instead of html.
  * of course, one could tell rmd to produce a PDF but look at the html view and vice versa. there is a heuristic to scan the current directory for expected output file names and suggests to switch the panel to the other view. see [1] below. I'm aware that this could be better, but that's how far I came. We could add a follow up ticket.
- the markdown panel does show the unprocessed but rendered markdown (I think that's what hal once talked about) for increased readability. Besides that, processing markdown at all isn't a good idea in the first place, because there is a [know bug](https://stackoverflow.com/questions/39183406/do-not-escape-backslashes-in-formulas-with-rmarkdown-md-document) which mangles backslashes in latex formulas.
- enhancing the "masked file" data-structure used by the files listing to gray out Rmd-derived files [2]
- codemirror multiplexing to highlight code accordingly: support for a few selected ones, i.e. R, python, ruby, go, etc. should be good enough. [3]

testing:

* check if formatting and frontmatter syntax highlighting in markdown and rmd files works. yaml fontmatter is formatted nicely, too.
* I'm working on various Rmd test files to see what we can support and where we lack some libraries. 
probably a simple example to get started is to just try this (remove the 4 space indent to make the file work ... I added it to disable rendering it here on github)
→ **[collecting rmd test files in our cocalc examples](https://github.com/sagemathinc/cocalc-example-files/tree/master/RMarkdown)**

        # Test
        
        ```
        just code formatted
        ```
        
        and some R code
        
        ```{r}
        xx <- rnorm(2000)
        paste("length of xx:", length(xx))
        plot(xx)
        ```
        
        **ok** _italic_ and $\tfrac{1}{997} = `r 1/997`$.
        
        <div align='right'>right hand side</div>
        
        link to [google](https://google.com).
        
        ## and a table
        
        ```{r echo = FALSE, results = 'asis'}
        library(knitr)
        kable(mtcars[1:15, ], caption = "this is a knitr kable.")
        ```
        

### [1]

screenshot from 2018-09-13 16-15-50

screenshot from 2018-09-13 16-16-22

### [2]

screenshot from 2018-09-13 16-21-47

### [3]

screenshot from 2018-09-13 16-28-15


haraldschilly:
with the latest commit here, codemirror multiplexing for R, R inline, python and ruby looks like that. Produced output in html mode visible on the right hand side. 

screenshot from 2018-09-11 17-46-36

src/smc-webapp/codemirror/custom-modes.coffee
```diff
@@ -89,6 +90,35 @@ CodeMirror.defineMode "sagews", (config) ->
...
+    modes['octave'] = 'octave'
+    modes['bash'] = modes['sh']
+
+    options = []
+
+    # blocks (ATTN ruby before r!)
+    # all engine modes: names(knitr::knit_engines$get())
+    for name in  ['ruby', 'r', 'python', 'octave', 'fortran95', 'fortran',  'octave', 'bash', 'go', 'julia', 'perl']
+        mode = modes[name]
+        open = new RegExp("```\\s*{#{name}[^}]*?}")
```
haraldschilly: the block to highlight is a markdown triple backticks codeblock, which has `{r [, options ...]}` next to it in the first line. it could also be indented when used in a list, which this also matches, but that's sort of not a really good idea to do in general.

src/smc-webapp/codemirror/modes.coffee
```diff
@@ -65,6 +65,7 @@ require('codemirror/mode/verilog/verilog.js')
 require('codemirror/mode/xml/xml.js')
 require('codemirror/mode/xquery/xquery.js')
 require('codemirror/mode/yaml/yaml.js')
+require('codemirror/mode/yaml-frontmatter/yaml-frontmatter.js')
```
haraldschilly: yaml frontmatter enhances normal markdown to support the yaml introduction. that's used in the cocalc blog, so, that's why it is also enabled for normal markdown. github itself also supports it for the wiki pages. etc.

src/smc-webapp/frame-editors/code-editor/actions.ts
```diff
@@ -81,6 +81,7 @@ export interface CodeEditorState {
   read_only: boolean;
   settings: Map<string, any>; // settings specific to this file (but **not** this user or browser), e.g., spell check language.
   complete: Map<string, any>;
+  derived_file_types: Set<string>;
```
haraldschilly: this is new: down below is code to check which expected compiled files do really exist. this set will contain nothing, just `"html"`, or `"html", "pdf"`. it's only used for Rmd.

src/smc-webapp/frame-editors/code-editor/actions.ts
```diff
@@ -1520,6 +1521,7 @@ export class Actions<T = CodeEditorState> extends BaseActions<
         parser = "typescript";
         break;
       case "md":
+      case "rmd":
```
haraldschilly: rmd is markdown, hence we can support markdown source formatting

src/smc-webapp/frame-editors/frame-tree/frame-tree.tsx
```diff
@@ -214,6 +217,7 @@ export class FrameTree extends Component<FrameTreeProps, FrameTreeState> {
           id={id}
           name={this.props.name}
           actions={this.props.actions}
+          mode={spec.mode}
```
haraldschilly: mode is just used to signal that this is for `"rmd"` or something else.

src/smc-webapp/frame-editors/frame-tree/util.ts
```diff
@@ -39,6 +39,7 @@ export const PRETTIER_SUPPORT = {
   js: true,
   jsx: true,
   md: true,
+  rmd: true,
```
haraldschilly: prettier again

src/smc-webapp/frame-editors/generic/client.ts
```diff
@@ -38,7 +38,11 @@ export interface ExecOutput {
 
 // async version of the webapp_client exec -- let's you run any code in a project!
 export async function exec(opts: ExecOpts): Promise<ExecOutput> {
-  return callback_opts(webapp_client.exec)(opts);
+  let msg = await callback_opts(webapp_client.exec)(opts);
+  if (msg.status && msg.status == "error") {
+    throw new Error(msg.error);
```
haraldschilly: there could be an error sitting inside the returned msg object

src/smc-webapp/frame-editors/html-editor/iframe-html.tsx
```diff
@@ -98,13 +101,24 @@ export class IFrameHTML extends Component<PropTypes, {}> {
   }
 
   render_iframe() {
+    if (
+      this.props.mode == "rmd" &&
+      this.props.derived_file_types != undefined
+    ) {
+      if (!this.props.derived_file_types.contains("html")) {
+        return this.render_no_html();
```
haraldschilly: just in "rmd" mode, and when we do know the situation about derived file types, and we've learned that the "html" version doesn't exist, this iframe won't be able to show the html file → then give the user a hint what is going on

src/smc-webapp/frame-editors/latex-editor/pdfjs.tsx
```diff
@@ -464,7 +467,34 @@ class PDFJS extends Component<PDFJSProps, PDFJSState> {
...
+      </div>
+    );
+  }
+
   render() {
+    if (
+      this.props.mode == "rmd" &&
+      this.props.derived_file_types != undefined
+    ) {
+      if (!this.props.derived_file_types.contains("pdf")) {
```
haraldschilly: like in the html iframe case, rmd mode and we know about alternatives but there is no pdf → give user a hint about what's going on

src/smc-webapp/frame-editors/rmd-editor/actions.ts
```diff
@@ -2,30 +2,98 @@
...
+import { markdown_to_html_frontmatter } from "../../markdown";
 import { FrameTree } from "../frame-tree/types";
+import { redux } from "../../app-framework";
+import { change_filename_extension } from "../generic/misc";
 
 export class RmdActions extends Actions {
+  private _last_save_time: number = 0;
+
   _init2(): void {
+    super._init2(); // that's the one in markdown-editor/actions.ts
```
haraldschilly: This is highly confusing for me and maybe it's a bug: the line below suggests, that there is something running after "markdown". The `Actions` class is derived from the one for Markdown, and it also has a `_init2` method. So, the parent `_init2` initializes spellchecking and hooks up on change events for synchronizing. So, I think this is an oversight, and therefore I am calling the parent `_init2`. That's why spellchecking is active.

src/smc-webapp/frame-editors/rmd-editor/actions.ts
```diff
@@ -2,30 +2,98 @@
...
+      this._init_rmd_converter();
+      this._check_produced_files();
     }
   }
 
-  _init_rmd2md(): void {
-    this._syncstring.on("save-to-disk", () => this._run_rmd2md());
-    this._run_rmd2md();
+  _init_rmd_converter(): void {
+    const run_debounced = debounce(() => this._run_rmd_converter(), 5 * 1000, {
```
haraldschilly: that's the lodash debounce, and well, it felt better that way because the whole mechanism is triggered by a save and it was behaving odd when I manually saved frequently. long term, I think we should have a build button and run-state indicator, similar to latex. Maybe worth a follow up ticket?

src/smc-webapp/frame-editors/rmd-editor/actions.ts
```diff
@@ -2,30 +2,98 @@
...
+      return;
+    }
+    const listing = dir_listings.get(project_store.get("current_path"));
+    if (listing == undefined) {
+      return;
+    }
+
+    let existing = Set();
+    for (let ext of ["pdf", "html"]) {
+      const expected_fn = change_filename_extension(this.path, ext);
```
haraldschilly: the core of this: checking if an expected compiled file exists

src/smc-webapp/frame-editors/rmd-editor/actions.ts
```diff
@@ -2,30 +2,98 @@
...
-    let markdown: string;
+    this.set_status("Running RMarkdown...");
+    this.set_error("");
+    let markdown = "";
     try {
-      markdown = await convert(this.project_id, this.path, time);
+      const md: string = this._syncstring.to_str();
+      let frontmatter = "";
+      if (md !== undefined) {
+        const md2html = markdown_to_html_frontmatter(md);
```
haraldschilly: there are two markdown functions, and this is the one which gives us the frontmatter. it's used in the `convert` function to decide a little bit what to do.

src/smc-webapp/markdown.ts
```diff
@@ -39,7 +41,15 @@ See https://github.com/sagemathinc/cocalc/issues/2863
...
 
-export function markdown_to_html(markdown_string: string): string {
+export interface MD2html {
+  html: string;
+  frontmatter: string;
+}
+
+function process(
+  markdown_string: string,
+  mode: "default" | "frontmatter"
```
haraldschilly: `process` is a little enhancement of the markdown processing we already have, where the `mode` switches to the one with frontmatter processing. I didn't want to do it unconditionally, because it needs to  instantiate the parser again and  maybe there is some overhead. tl;dr, all existing markdown processing calls behave exactly the same as before.

src/smc-webapp/markdown.ts
```diff
@@ -48,14 +58,36 @@ export function markdown_to_html(markdown_string: string): string {
...
+
+  let html: string;
+  let frontmatter = "";
+
+  // avoid instantiating a new markdown object for normal md processing
+  if (mode == "frontmatter") {
+    const md_frontmatter = new MarkdownIt(OPTIONS).use(
+      MarkdownItFrontMatter,
+      fm => {
+        frontmatter = fm;
```
haraldschilly: this is weird code, but it's synchronous and that frontmatter processor still has a function as an argument. using a module-level variable `frontmatter` is likely not a safe thing to do...

src/smc-webapp/markdown.ts
```diff
@@ -48,14 +58,36 @@ export function markdown_to_html(markdown_string: string): string {
...
+      }
+    );
+    html = md_frontmatter.render(text);
+  } else {
+    html = markdown_it.render(text);
```
haraldschilly: that's how it is done before in the default case

Issue #2875, opened 2018-05-10
firefox (newer versions) does not stay connected to Cocalc


DrXyzzy:
I have seen this repeatedly over past 3 weeks. It may be related to recent user issues connecting to CoCalc with Firefox on macOS. Seen with macOS 10.13.1, Firefox 59.0.2 (64-bit). No anti-virus or ad-blocker software installed. No browser extensions are enabled.

Sample timeline [added: see below for quicker way to replicate the problem]:
- 8:52 login to CoCalc
- 8:54 fans go on but no process showing over 10% CPU
- switch browser tabs so CoCalc session is not displayed
- 9:19 return to cocalc tab **CoCalc is no longer connected**. Other browser tabs and windows can be opened and refreshed normally.
  - see orange "connecting..." notification upper right, alternating with
"disconnected"
  - get message "Your connection is unstable or CoCalc is temporarily not available."
  - attempts to login by hitting refresh or closing and opening a new tab and entering CoCalc URL stall with blank CoCalc tab
  - login is possible after restarting the browser

Browser console log shows:
```
initializing prometheus client smc-6cc8c1ade66041fbb28e.cacheme.js:5:713630
websocket - connect smc-6cc8c1ade66041fbb28e.cacheme.js:5:1085992
websocket -- delete cookie smc-6cc8c1ade66041fbb28e.cacheme.js:5:1087744
websocket - connecting smc-6cc8c1ade66041fbb28e.cacheme.js:5:1085992
attempt: 1 and num_recent_disconnects: 0 smc-6cc8c1ade66041fbb28e.cacheme.js:5:3210873
unreachable code after return statement[Learn More] js:69:1217
[11.1s Δ 11.1s] websocket -- delete cookie smc-6cc8c1ade66041fbb28e.cacheme.js:5:53179
[11.1s Δ 0s] websocket - reconnect scheduled (attempt 1 out of 100000) smc-6cc8c1ade66041fbb28e.cacheme.js:5:53179
Firefox can’t establish a connection to the server at wss://cocalc.com/hub/?_primuscb=MDARKlu. 1-6cc8c1ade66041fbb28e.cacheme.js:2:917
[12.1s Δ 1s] websocket - connecting smc-6cc8c1ade66041fbb28e.cacheme.js:5:53179
[12.1s Δ 0.01s] attempt: 1 and num_recent_disconnects: 1 smc-6cc8c1ade66041fbb28e.cacheme.js:5:53179
[...]

PR #3144, opened 2018-09-01
static: check for buggy firefox 

haraldschilly:
amends preflight check, relates to issue #2875 ... I should have added that right from the start, but sort of forgot about it.

the only thing left to do is to test it. compile on k2 and run through browserstack...


haraldschilly:
FF 61 on some OSX

screenshot from 2018-09-01 11-48-31

old chrome on most recent osx (still the usual message)

screenshot from 2018-09-01 11-49-40

Firefox 60 on Windows 10

screenshot from 2018-09-01 11-51-02


---

and it still starts fine in FF 62 and recent chrome

PR #3162, opened 2018-09-10
webapp/project: format c, c++ files

haraldschilly:
This should be straight forward to review. For testing, paste a c, cc, h or c++ file and use the format button. 

before

screenshot from 2018-09-10 13-53-11

after

screenshot from 2018-09-10 13-53-24


williamstein:
It doesn't seem like "Indent size" has any impact.  This may be some unrelated thing though.

PR #3167, opened 2018-09-10
go code editor: starting with gofmt

haraldschilly:
* merged with master (formatting branch from yesterday), hence there is no conflict to expect
* calls gofmt like those other cmd line utilities
* fixes some codemirror options, because go is using tabs and not spaces
* post-processing of error messages to get rid of showing the user the temporary file name. below are before/after screenshots of that
* there are plenty of examples online, e.g. https://play.golang.org/p/I63ge2ISDs should be ok for a quick spin

## unprocessed

screenshot from 2018-09-11 12-14-54

## processed

screenshot from 2018-09-11 12-22-01


src/smc-project/gofmt.ts
```diff
@@ -0,0 +1,78 @@
...
+  return ret.join("\n");
+}
+
+export async function gofmt(
+  input: string,
+  options: ParserOptions,
+  logger: any
+): Promise<string> {
+  // create input temp file
+  const input_path: string = await callback(tmp.file);
```
williamstein: Put the body of this function in a try/finally block, and delete input_path in the finally part.  As far as I can tell reading this code, input_path is never cleaned up at all.      If you don't want to indent everything in a try/finally, move the body of code to a helper function.

williamstein: Alternatively, since gofmt takes input from stdin by default, why not just write to stdin?  If it works, shouldn't it be more efficient than creating a temporary file?    E.g., there's an example here https://stackoverflow.com/questions/13230370/nodejs-child-process-write-to-stdin-from-an-already-initialised-process

williamstein: (If you do use stdin, be sure to test on a file that's big-ish e.g., at least 40K or so, to make sure there is no weird locking/buffering messing things up.)

williamstein:
I fixed the issues myself, since they were all my fault!

PR #3174, opened 2018-09-12
preflight check: test explicitly for FF 60.2 ESR

haraldschilly:
because that's the only one in version 60 that works, and remove that little alert box from the browser client

rationale behind this is explained here https://github.com/sagemathinc/cocalc/issues/2875#issuecomment-420682930

testing: so far I haven't …  and compiling it suddenly causes

```
2018-09-12T15:39:55.992Z - debug: Uncaught exception: TSError: ⨯ Unable to compile TypeScript:
../smc-webapp/jupyter/complete.tsx(32,5): error TS2304: Cannot find name '$'.
```

well, added another commit, now it compiles, and works locally for me. we should check if existing warnings for FF 60 and 61 still show up.

plan to finish: check if my dev project loads find in my chrome and ff 60.2 esr, while the warning appears for an older firefox.

### DIFF
--- a/src/smc-webapp/client_browser.coffee
+++ b/src/smc-webapp/client_browser.coffee
@@ -101,14 +101,6 @@ class Connection extends client.Connection
         if prom_client.enabled
             @on('start_metrics', prom_client.start_metrics)
 
-        # Wait until load totally done.
-        setTimeout(@_firefox60_bug, 5000)
-
-    _firefox60_bug: =>
-        {name, version} = require('misc/browser').get_browser()
-        if name == 'Firefox' and (version == '59' or version == '60' or version == '61')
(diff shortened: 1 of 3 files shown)


williamstein:
> ../smc-webapp/jupyter/complete.tsx

I touched that code yesterday...

haraldschilly:
that error showed up when starting the hub

williamstein:
Put this at top of the file:
```
declare const $: any;
```

haraldschilly:
yes, I already did that in that other commit. I compiled this branch on k2 and went through all recent browser version on win 10, two older FF (showing the warning), and the 60.2ESR on my local machine. Everything works as expected.

Issue #2757, opened 2018-03-16
list all your existing collaborators  when adding a collaborators 


haraldschilly:
Right now, the "add collaborator" dialog only shows a search box.

The goal of this ticket is to add a dropdown list for adding an existing collaborator of yours from another project you're part of. That way, you can select someone from your "known circle" of cocalc users instead of having to search for.

williamstein:
I don't agree that we should implement this.  It's clutter.  Facebook doesn't need it.  By just showing all collabs first, we accomplish the same thing better.

counterargument?

williamstein:
![screenshot 2018-03-16 at 11 09 42 am](https://user-images.githubusercontent.com/1276278/37537364-8bc99746-290a-11e8-9647-16592fecef6a.png)

PR #3186, opened 2018-09-14
webapp/sagews: hotfix to bypass #2757

haraldschilly:
Ref: #2757

not the prettiest thing, and it's also not a "fix", but existing worksheets do at least load up without showing control characters and a user learns about the problem

screenshot from 2018-09-14 21-33-04


haraldschilly:
and why on earth is there a package-lock file in the commit? it doesn't even show up in my git repo. mehh .. I'll force push something to get rid of it

haraldschilly:
got rid of package-lock diff ...  :roll_eyes:

PR #3190, opened 2018-09-19
hub: update prom client and some monitoring aspects

haraldschilly:
this is mainly about monitoring of hub and project

* main change is the aggregation of client information per hub, i.e. there are no longer lines specific to a client with a concrete `client_id`. in a cc-in-cc dev project, that can be seen at such an url: `https://cocalc.com/... your dev project ID ... /port/ ... your port number ... /metrics`

* there is an update of prometheus client, and it is also added to projects. its only purpose there is to capture nodejs related memory and process info. I don't know how that can be made visible in a test project, though. It should be the same general data that's also visible for hubs.

* it removes client specific `query_queue_info` from the hub. that needs to be done in a different (better) way anyways

* exposes OOM stats for projects

## testing

I should probably just deploy this in the test namespace. If it doesn't blow up, all this should be fine.


williamstein:
I looked over the code with my eyes, and nothing looks suspicious.  Test it *somewhere*, then merge it to master whenever you are sure it won't break everything.

Issue #2845, opened 2018-04-25
the email is valid check when creating an account is clearly not sufficient


williamstein:
Every time I send emails to new users there are numerous clearly invalid email addresses, which gmail catches.  We should not let people enter a clearly invalid email address when creating an account.

williamstein:
Here is an example of an email address that somebody gave us, but which gmail rejects:
```
foo@ntu.edu.s
```

We should change our validator to be stronger, e.g., include all valid TLD's.

The above person *clearly* meant to type "foo@ntu.edu.sg", but made a typo and will henceforth get very confused and locked out of their account when they try to sign in.  They will quit using cocalc or have to make a desperate support attempt. We need to make this situation less likely.

williamstein:
Here is the current TLD list: http://data.iana.org/TLD/tlds-alpha-by-domain.txt

williamstein:
In my experience, people typo their TLD about 1% of the time...

williamstein:
This is properly solved via email verification, which is implemented.

PR #3196, opened 2018-09-21
Check for validity of TLD in email verification

johnjeng:
Threw this together in a few min. Fix for #2845.

If someone wants to pick it up as warm up, please do.

Issues:
- Tests didn't run for some reason. I think it's unrelated.
- List of valid TLDs should probably be its own file or it should query the website directly.


williamstein:
Ping @johnjeng

PR #3198, opened 2018-09-21
Rseo mods

DrXyzzy:
fix pricing link in base at bottom of [r seo page](https://cocalc.com/doc/r-statistical-software.html), plus a couple tiny things


haraldschilly:
yes, thank you, that missing "PREFIX" was 100% an oversight.

PR #3199, opened 2018-09-21
add "start free" boilerplate text at the bottom

haraldschilly:
trivial addition, just like on the R page


williamstein:
seems safe and a good idea.

Issue #3171, opened 2018-09-12
searching or relatively recent collaborators is still broken


williamstein:
I just tried to search for my nephew, who has multiple accounts under "Luca Begay", and was active 10 months ago. But look:

![screenshot 2018-09-11 at 4 59 04 pm](https://user-images.githubusercontent.com/1276278/45394162-2dc73100-b5e4-11e8-8b39-b4ab908808b8.png)

(no results!)

But see, on another project, he has accounts and at least one was used in the last 10 months!

![screenshot 2018-09-11 at 4 59 15 pm](https://user-images.githubusercontent.com/1276278/45394156-27d15000-b5e4-11e8-8f10-c290723ebc0f.png)

PR #3210, opened 2018-09-26
webapp/hub: increase default for user_search to 13 months and ignore "last active" for admins

haraldschilly:
this looks trivial, but I have to add that I haven't tested it so do not merge it without a check

* I indirectly noticed #3171 today because searching for a 2 year dormant user via email didn't give any results in the admin interface
* and well, then noticing  how the default is set, this makes it clear why #3171 is still an issue
* setting `{active:undefined or null}` for admin doesn't work, because the string as a default overwrites it ... hence I added the clause to the if  in the hub. (which requires a restart)

### DIFF
--- a/src/smc-hub/postgres-server-queries.coffee
+++ b/src/smc-hub/postgres-server-queries.coffee
@@ -982,7 +982,7 @@ exports.extend_PostgreSQL = (ext) -> class PostgreSQL extends ext
                     query += ', email_address'
                 query += ' FROM accounts'
                 query += " WHERE deleted IS NOT TRUE AND (#{where.join(' OR ')})"
-                if opts.active
+                if opts.active and not opts.admin
                     params.push(opts.active)
                     # name search only includes active users
                     query += " AND ((last_active >= NOW() - $#{i}::INTERVAL) OR (created >= NOW() - $#{i}::INTERVAL)) "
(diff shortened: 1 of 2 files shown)


williamstein:
Looks good to me.  Please test it!!!  You should test it, right?

williamstein:
OK, that makes perfect sense.  I can't imagine this doesn't work perfectly. 

... famous last words  🤷‍♂️

haraldschilly:
for testing, I added a log line to see the query params. 

admin query from the logs:

```
"query params=%test%,100"

SELECT account_id, first_name, last_name, last_active, created, email_address FROM accounts WHERE deleted IS NOT TRUE AND (((lower(first_name) LIKE $1::TEXT OR lower(last_name) LIKE $1::TEXT OR lower(email_address) LIKE $1::TEXT))) ORDER BY last_active DESC NULLS LAST LIMIT $2::INTEGER
```

user query

```
"query params=test1%,13 months,25"

SELECT account_id, first_name, last_name, last_active, created FROM accounts WHERE deleted IS NOT TRUE AND (((lower(first_name) LIKE $1::TEXT OR lower(last_name) LIKE $1::TEXT))) AND ((last_active >= NOW() - $2::INTERVAL) OR (created >= NOW() - $2::INTERVAL))  ORDER BY last_active DESC NULLS LAST LIMIT $3::INTEGER
```

so, that looks good, I'll merge it

PR #3213, opened 2018-09-27
webapp/lean: restart lean server instance via a button

haraldschilly:
main complication so far: is that lean api somewhere documented?

### DIFF
--- a/src/smc-project/lean/lean.ts
+++ b/src/smc-project/lean/lean.ts
@@ -250,6 +250,16 @@ export class Lean extends EventEmitter {
     }
   }
 
+  restart(): void {
+    this.dbg("restart");
+    if (this._server != undefined) {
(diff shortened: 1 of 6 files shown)


williamstein:
>  is that lean api somewhere documented?

which lean api?  The CoCalc one or lean itself?

haraldschilly:
ok, this really works!

1. add this to a lean file at the top
```
import data.nat.prime -- prime number stuff 
open nat
#eval factors (2^32+1)
```
→ which throws errors

2. add this `leanpkg.path` to the home dir:
```
builtin_path
path /ext/lean/lean-3.4.1-linux/mathlib/
```

3. click the restart button → then it works.
4. you can add `-- ` in the second line of the path file, and then restart again. it no longer knows about mathlib and fails again like in step 1.

PR #3220, opened 2018-09-28
webapp/frame editor: catch unhandled exception when scrolling in codemirror to a line that does not exist

### LINKED ISSUE #3219: latex → check_for_fatal_error → Cannot read property 'getIn' of undefined
I'll fix this, looks easy. It seems to happen when a latex build was running but the user tabbed away to the files listing or whatever.

chrome: `Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/69.0.3497.100 Safari/537.36`

smc_git_rev: 9dde507f765bbe78bafc6d00b22375271ebd1c16

Date: 2018-09-28 14:01:28.233185

`Cannot read property 'getIn' of undefined`

```
TypeError: Cannot read property 'getIn' of undefined
    at n.check_for_fatal_error (https://cocalc.com/static/smc-2bee88b0c9d7755adaa4.cacheme.js?2bee88b0c9d7755adaa4:5:882171)
    at n.<anonymous> (https://cocalc.com/static/smc-2bee88b0c9d7755adaa4.cacheme.js?2bee88b0c9d7755adaa4:5:886262)
    at https://cocalc.com/static/smc-2bee88b0c9d7755adaa4.cacheme.js?2bee88b0c9d7755adaa4:5:879552
    at Object.next (https://cocalc.com/static/smc-2bee88b0c9d7755adaa4.cacheme.js?2bee88b0c9d7755adaa4:5:879657)
    at i (https://cocalc.com/static/smc-2bee88b0c9d7755adaa4.cacheme.js?2bee88b0c9d7755adaa4:5:878403)
```

haraldschilly:
* this should fix #3219, although I wasn't able to reproduce it
* I actually noticed another codemirror exception (unhandled async), when jumping to a line which doesn't exist. message: `There is no line NNN in the document.`

however, I was able to trigger that for an rtex file and  it seems like reconciling the tex errors with the original file doesn't work any more. I'll investigate and make another issue if true → #3221. However, fixing this little codemirror problem should be something we do anyways.

### DIFF
--- a/src/smc-webapp/frame-editors/code-editor/actions.ts
+++ b/src/smc-webapp/frame-editors/code-editor/actions.ts
@@ -1056,7 +1056,7 @@ export class Actions<T = CodeEditorState> extends BaseActions<
 
   _get_most_recent_cm_id(): string | undefined {
     return this._get_most_recent_active_frame_id(
-      node => node.get("type").slice(0,2) == "cm"
+      node => node.get("type").slice(0, 2) == "cm"
     );
   }
 
@@ -1135,9 +1135,9 @@ export class Actions<T = CodeEditorState> extends BaseActions<
   set_syncstring(value: string): void {
     if (this._state === "closed") return;
     const cur = this._syncstring.to_str();
-    if (cur === value) { // did not actually change.
-      return;
-    }
+    // did not actually change.
+    if (cur === value) return;
+
     this._syncstring.from_str(value);
     // NOTE: above is the only place where syncstring is changed, and when *we* change syncstring,
     // no change event is fired.  However, derived classes may want to update some preview when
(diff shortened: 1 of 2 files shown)


williamstein:
To reproduce/test this, make a file a.txt, then  (with your project id)
```
a = smc.redux.getEditorActions('878f95de-9465-4627-a26c-dead9b73c65c','a.txt')
await a.programmatical_goto_line(10, true, true)
await a.programmatical_goto_line(-1, true, true)
await a.programmatical_goto_line(0, true, true)
```

The right fix should be to not call CodeMirror with invalid input, rather than using try/catch.  It turns out line number input is valid if and only if the line is a positive integer.

williamstein:
OK, I pushed a different fix.

haraldschilly:
but wouldn't that jump to the end of the file in a case like #3221? I would rather like my cursor to stay where it is.

williamstein:
> but wouldn't that jump to the end of the file in a case like #3221? I would rather like my cursor to stay where it is.

Yes.  This is the right behavior for a function called `progammatically_goto_line`, e.g., it agrees with what vim, emacs, and Codemirror themselves all do when you call their goto line functions with a number larger than the buffer.

Issue #3222, opened 2018-09-28
latex/codemirror: render_gutters → Cannot read property 'line' of null


haraldschilly:
In the source code for the stacktrace's top frame below, I see two `<obj>.line` occurrences. I think it's `lineInfo.line`, but I can't narrow it down any further. The part "of `null`" is a pretty good hint, that a firm `!==` check for not `undefined` is not good enough. A fix should maybe change the test to `obj != undefined` ?

chrome: `Mozilla/5.0 (Windows NT 6.3; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/67.0.3396.99 Safari/537.36`

smc_git_rev: 9dde507f765bbe78bafc6d00b22375271ebd1c16

Date: 2018-09-28 15:04:58.856267

`Cannot read property 'line' of null`

```
TypeError: Cannot read property 'line' of null
    at t.render_gutter (https://cocalc.com/static/smc-2bee88b0c9d7755adaa4.cacheme.js?2bee88b0c9d7755adaa4:5:3983279)
    at https://cocalc.com/static/smc-2bee88b0c9d7755adaa4.cacheme.js?2bee88b0c9d7755adaa4:5:3983627
    at https://cocalc.com/static/1-2bee88b0c9d7755adaa4.cacheme.js?2bee88b0c9d7755adaa4:6:285070
    at Ue.iterate.Ge.iterate (https://cocalc.com/static/1-2bee88b0c9d7755adaa4.cacheme.js?2bee88b0c9d7755adaa4:6:290821)
    at Fe.__iterate (https://cocalc.com/static/1-2bee88b0c9d7755adaa4.cacheme.js?2bee88b0c9d7755adaa4:6:285039)
    at Fe.forEach (https://cocalc.com/static/1-2bee88b0c9d7755adaa4.cacheme.js?2bee88b0c9d7755adaa4:6:317925)
    at t.render_gutters (https://cocalc.com/static/smc-2bee88b0c9d7755adaa4.cacheme.js?2bee88b0c9d7755adaa4:5:3983596)
    at t.render (https://cocalc.com/static/smc-2bee88b0c9d7755adaa4.cacheme.js?2bee88b0c9d7755adaa4:5:3983735)
    at s (https://cocalc.com/static/1-2bee88b0c9d7755adaa4.cacheme.js?2bee88b0c9d7755adaa4:6:4378437)
[...]

PR #3224, opened 2018-09-29
webapp/frame: codemirror lineInfo for an unkown line gives null

haraldschilly:
Ref #3222


src/smc-webapp/frame-editors/code-editor/codemirror-gutter-markers.tsx
```diff
@@ -29,7 +29,9 @@ export class GutterMarkers extends Component<Props, {}> {
     const handle = info.handle;
     if (handle != null) {
       let lineInfo = this.props.codemirror.lineInfo(handle);
-      if (lineInfo !== undefined) {
+      // lineInfo for unknown line gives null
+      // https://github.com/sagemathinc/cocalc/issues/3222
+      if (lineInfo !== undefined && lineInfo !== null) {
```
williamstein: In Javascript, this whole line is equivalent to `if (lineInfo != null)`.

PR #3239, opened 2018-10-03
Fix browser forward/back navigation

johnjeng:
# Description
Forward/back browser navigation is currently broken. You can go back one action but can't go back anymore. Additionally forward is not possible.

The issue is that every `set_active_tab` call at the page level pushes onto the browser history stack.

This change adds an option for `load_target` and page level `set_active_tab` to not push onto browser history.
Changed `window.onpopstate` to pass `change_history=false` to `load_target` so that forward and back don't alter the history stack.

# Testing Steps
1. Go to account settings
1. Go to Projects 
1. Open a project
1. Use forward and back on your browser to navigate the previous 3 actions.

# Relevant Issues
Couldn't find an issue.

### DIFF
--- a/src/smc-webapp/history.coffee
+++ b/src/smc-webapp/history.coffee
@@ -110,7 +110,7 @@ exports.set_url = (url) ->
     analytics_pageview(window.location.pathname)
 
 # Now load any specific page/project/previous state
-exports.load_target = load_target = (target, ignore_kiosk=false) ->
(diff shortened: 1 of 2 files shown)


williamstein:
Wait, it's that easy?

williamstein:
This works as far as it can.  However, there are sadly definitely some bugs -- probably this one https://github.com/sagemathinc/cocalc/issues/3039 -- that become more obvious with this.

PR #3247, opened 2018-10-05
webapp: support adb ada files

haraldschilly:
well, when we install the compiler, we should also support file formatting. this is closely modeled after the lean file. good enough for the very basics.

screenshot from 2018-10-05 16-45-06


williamstein:
Nice!  I wrote a surprisingly large amount of ADA code myself once... back in the 1990s...

PR #3259, opened 2018-10-08
Don't remove anyone if student id unknown

johnjeng:
My hypothesis is that `student_account_id` becomes undefined for some reason (which is why a non-user invite is sent) and then the project checks if account ids on the student project are valid. Since `student_account_id` is undefined, all course-collabs (teacher, TAs, etc) will remain but anyone else, including the student, will be removed from the student course.


williamstein:
It's not that the id becomes undefined -- it's that it wasn't defined yet!

PR #3266, opened 2018-10-08
Cleanup privacy policy

haraldschilly:
# Description

This is sanitized HTML, which is derived from an internal RMarkdown file. In paricular, the section numberings are fine now. for example:

## Testing
* someone has to double-check that all content matches with the original. 

## before

screenshot from 2018-10-08 15-00-36

## after

screenshot from 2018-10-08 15-01-09


haraldschilly:
updated with @johnjeng changes

PR #3268, opened 2018-10-08
webapp: remove bootstrap, in favor of just using react-bootstrap

haraldschilly:
# Description

see issue #1872, the referenced upstream ticket is closed and someone reported that bootstrap removal as the solution.

basically this is a double-bootstrapping problem.

# Testing Steps
I did very little testing, just confirmed that #1872 is fixed. My fear is there could be unrelated problems. However, I'm pretty sure importing `react-bootstrap` does also introduce some code and css in such a way, that it might be fine just the way it is.

### DIFF
--- a/src/smc-webapp/package.json
+++ b/src/smc-webapp/package.json
@@ -27,7 +27,6 @@
     "async-await-utils": "^2.0.4",
     "awaiting": "^3.0.0",
     "bootbox": "^4.4.0",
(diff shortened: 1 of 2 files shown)


williamstein:
I just 100% reverted this code.  It breaks Sage worksheets -- the menus are gone.  It also makes popover tracebacks appear whenever you click on certain buttons.   Revisit next year when the react rewrite is done.

Issue #3216, opened 2018-09-28
jupyter2: api endpoint signal: no kernel with path 'undefined'


haraldschilly:
IMHO, whatever is this actually about, the webapp code shouldn't issue an API call with "undefined" payload. I see this as a sign that some code is asserting some variable is defined and this stacktrace  is a symptom of that.

smc_git_rev: 9dde507f765bbe78bafc6d00b22375271ebd1c16

Date: 2018-09-28 12:50:40.892748

Error frequency: several times per hour

`Error: api endpoint signal: no kernel with path 'undefined'`

```
Error: Error: api endpoint signal: no kernel with path 'undefined'
    at e.<anonymous> (https://cocalc.com/static/smc-2bee88b0c9d7755adaa4.cacheme.js?2bee88b0c9d7755adaa4:5:5014814)
    at https://cocalc.com/static/smc-2bee88b0c9d7755adaa4.cacheme.js?2bee88b0c9d7755adaa4:5:5014352
    at Object.next (https://cocalc.com/static/smc-2bee88b0c9d7755adaa4.cacheme.js?2bee88b0c9d7755adaa4:5:5014457)
    at i (https://cocalc.com/static/smc-2bee88b0c9d7755adaa4.cacheme.js?2bee88b0c9d7755adaa4:5:5013203)
```

---

this is triggered by an unset `data.path` in `handle_api_call` in `smc-project/browser-websocket/api.ts`

```
case "jupyter":
      return await jupyter(data.path, data.endpoint, data.query);
```

---

aha, and without having any idea how to replicate, is this caused by path in store not being defined? Is there any other usage of jupyter api?

in jupyter actions:

```
return await (await this.init_project_conn()).api.jupyter(
      this.store.get("path"),
      ....
```

haraldschilly:
chrome: `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/69.0.3497.100 Safari/537.36`

smc_git_rev: d1100d5931c2080039388778115ab8ae062aaa1e

Date: 2018-10-08 09:44:06.546045

`unhandledrejection: Error: Error: api endpoint signal: no kernel with path 'undefined'`

```
<generated>
Error
    at s (https://cocalc.com/static/smc-2dc89e44f1881c034548.cacheme.js?2dc89e44f1881c034548:5:862198)
[...]

PR #3273, opened 2018-10-09
webapp/jupyter2: do not call api when path is not known -- issue #3216

haraldschilly:
# Description
Fixing an uncaught exception

# Testing Steps
I don't know how to reproduce. Maybe by opening a jupyter file and closing it quickly?

# Relevant Issues
That whole fix is based on some thoughts outlined in #3216 


williamstein:
This makes no sense.  Instead just use this.path instead of this.store.get('path').  I'll push something very similar, but instead using this.path.

Issue #2215, opened 2017-07-21
latex: word count


haraldschilly:
add a button (or another UI element) to compute the word count of the document. report should show stats for the number of words, lines, formulas, etc.

we also need a utility to actually do this. the first one I found is [texcount](http://app.uio.no/ifi/texcount/download.html) which seems to be still active despite the old school looking webstie. I've no experience, maybe there are better ones out there, ...

williamstein:
There’s no reason to restrict this to LaTeX.

haraldschilly:
this was fixed a long time ago

PR #3284, opened 2018-10-11
webapp/latex: count words

haraldschilly:
# Description
latex word count panel

## Testing Steps
1. open a tex file and select the word count panel, you should see its output
2. similarly, this also works for knitr documents, because it still parses the generated tex file
3. to test if auto updating works:
    1. when hidden and shown again: switch the pdf preview panel to word count, remember the word count you see, switch back to pdf preview (i.e. in such a way, that the word count panel isn't shown anywhere). Then edit the document and save it. Then switch back to the word count panel and see the updated count
    2. when building: edit the tex file and save it, the word count should update.
4. clicking the button is a forced build

It's triggered when the component mounts and during the run_build step.

## Relevant Issues
* #2215 


src/smc-webapp/frame-editors/latex-editor/actions.ts
```diff
@@ -302,6 +306,9 @@ export class Actions extends BaseActions<LatexEditorState> {
     if (update_pdf) {
       this.update_pdf(time, force);
     }
+
+    // and finally, work count finsihes -- to make clear the whole operation is done
```
williamstein: typos in "work count finsihes".

src/smc-webapp/frame-editors/latex-editor/actions.ts
```diff
@@ -277,6 +279,8 @@ export class Actions extends BaseActions<LatexEditorState> {
       await this.run_knitr(time, force);
       if (this.store.get("knitr_error")) return;
     }
+    // update word count asynchroneously
```
williamstein: typo in "asynchroneously"

src/smc-webapp/frame-editors/latex-editor/actions.ts
```diff
@@ -674,6 +681,30 @@ export class Actions extends BaseActions<LatexEditorState> {
     }
   }
 
+  make_timestamp(time: number, force: boolean): number | undefined {
+    return force ? undefined : time || this._last_save_time;
+  }
+
+  async word_count(time: number, force: boolean): Promise<void> {
+    // only run word count if at least one panel is active
```
williamstein: active unfortunately means "focused", not "exists".   Your code will work fine, bu the comment should be changed from "active" to "exists".  

It would be better to define a new method `_has_frame_id_of_type` with similar code to `_get_most_recent_active_frame_id_of_type` (even the same with == null for now would be fine) in the base class and call that instead, to make the intention clearer.

src/smc-webapp/frame-editors/latex-editor/count_words.ts
```diff
@@ -0,0 +1,23 @@
+import { exec } from "../generic/client";
+import { path_split } from "../generic/misc";
+
+// and enhancement might be to generate html via $ texcount -htmlcore
```
williamstein: two typos in your comments here. "and" and "substantically".

src/smc-webapp/frame-editors/latex-editor/latex-word-count.tsx
```diff
@@ -0,0 +1,65 @@
+import { Map } from "immutable";
+import { is_different /* , capitalize, path_split*/ } from "../generic/misc";
+import {
+  Component,
+  React,
+  rclass,
+  rtypes
+  //,  Rendered
+} from "../../app-framework";
+//import { TypedMap } from "../../app-framework/TypedMap";
```
williamstein: delete?

src/smc-webapp/frame-editors/latex-editor/actions.ts
haraldschilly: ok, I can add that function

Issue #3127, opened 2018-08-27
html formatter error reporting


williamstein:
First time I tried this in a project that hasn't been restarted recently and got this:

![screenshot 2018-08-27 at 1 38 00 pm](https://user-images.githubusercontent.com/1276278/44684656-7c7aa580-a9fe-11e8-9bed-72caa66bf9f1.png)

So, obviously, the conversion of errors to text is wrong.

PR #3285, opened 2018-10-11
webapp & project: html formatter error handling

haraldschilly:
# Description
When I checked upon #3127 I thought it is a simple fix. Turns out, that bug is fine and errors show up properly. However, when I changed the name of the binary to one that doesn't exist, I noticed that it doesn't work at all. Hence I added some error handling in case it is not possible to spawn `tidy` and telling the user about this and some note about installing it. I also changed the name in a backwards compatible way.

# Testing Steps
1. well, run tidy, maybe with `<e>` as an unknown tag
2. not sure how to test the case where tidy doesn't exist.

# Relevant Issues
* #3127

## Screenshots showing errors (in a positive way)

screenshot from 2018-10-11 18-58-21

screenshot from 2018-10-11 18-54-07

screenshot from 2018-10-11 18-52-51

screenshot from 2018-10-11 16-56-33


williamstein:
> I also changed the name in a backwards compatible way.

Correct if I'm wrong, but for no good reason you've made it so after I push this, html formatting *breaks* for users unless they restart their project?   

For example, my wife has a project for editing her website.  It has a long idle timeout.   She will open her browser after I merge the above changes.  She will click the HTML format button and be surprised that it is broken suddenly, and won't think to restart her project.

Yes, requiring project restart is important for some major changes (e.g., so rewriting terminals). But in this small case for a rename?  

Or am I confused?

haraldschilly:
yes, you're right, I supposed I was only thinking the other way around

williamstein:
Now that I look at this more closely, I see the whole problem "spawns" from you basically reproducing what is done in
```
src/smc-util-node/execute-code.coffee
```
Instead of repeating that pattern all over, just use that code.   Basically it's `misc_node.execute_code`, and it's used almost everywhere by the backend (and frontend) for running subprocesses, and has a lot of things we generally want (e.g., a default timeout parameter, error detection as you just added in this patch, accumulating all the stdout and stderr).  Can you just rewrite your tidy function to use it?

haraldschilly:
ok, I never used it and I can give it try. Is `execute_code` documented? I can only find the code. Well, I'll try and see what it does.

williamstein:
> ok, I never used it 

It's basically just `webapp_client.exec`, so you've used it indirectly a lot.  It's just the backend side of that frontend function.

haraldschilly:
right, I see. I read it and the only detail that's weird with tidy is that an exit code of `1` is fine. But that can be ignored.

haraldschilly:
I changed it to use node-util's execute_code function and the "optimistic" case word. What's off is the error handling, because it doesn't work. Maybe it's because of that callback_opts wrapper? I looked at it, but it looks fine.

So, to see the issue I am seeing, change the command itself to `tiddy` or mangle an argument to be wrong, e.g. `--sort-attrutes`. I would expect that it throws an error, but it does nothing. I can see a suspicious line `(node:432889) [DEP0013] DeprecationWarning: Calling an asynchronous function without callback is deprecated.` in the project's log, though. (right after execute_code shows the error)

PR #3289, opened 2018-10-15
webapp/project: format xml documents

haraldschilly:
# Description
This is very similar to  HTML formatting. It also tweaks some error code handling, where tidy is a special case with exit code "1" still being ok. There is no issue, I think, but I was checking up upon our DMARC info files and they're just one-liners, which are harder to read than when they're formatted :-)

## Testing Steps
A neat test is some made up xml like that:
```
<?xml version='1.0' encoding='utf-8'?>
<data>
<row id="1">  <height error=".01">8</height>  <value>12</value></row><row id="2"> 
<height>8</height>  <value>12</value></row><row id="3">  <height>8</height> 
<value>12</value></row><row id="4"> 
<height>8</height>  <value>12</value></row></data>
```

- [x] A list of exact steps on how you tested.

## Before/After

screenshot from 2018-10-15 10-34-27

screenshot from 2018-10-15 10-34-41

## Pointing out problems

screenshot from 2018-10-15 10-37-22


src/smc-project/formatters/xml-format.ts
```diff
@@ -0,0 +1,97 @@
...
+          options.parser
+        }' installed?`
+      );
+    }
+
+    // stdout/err capture
+    let stdout: string = "";
+    let stderr: string = "";
+    // read data as it is produced.
+    xml_formatter.stdout.on("data", data => (stdout += data.toString()));
```
williamstein: Please rewrite to use `misc_node.execute_code`. Same comment as with html tidy.

PR #3301, opened 2018-10-18
project/websocket init: check what gets globally executed 

haraldschilly:
# Description

The goal of this here is improve the primus websocket initialization. It adds a check to what is downloaded, before it is globally executed. Another detail I figured out is that in my cc-in-cc dev project the situation of when the project did actually start up until the listing appears does improve. I think that's because of properly throwing errors (the ones in the global evaluation before aren't caught at all) and it also waits for the explicit global evaluation.

# Testing Steps
1. open at least two projects, to check if the listing are different -- bad timing could cause a reuse of the same primus socket across projects
2. do things like: stop project, refresh browser, start it and check what the console says. in the best case szenario, the attempt to connect works quickly and the listing appears right after the connection is established.
3. deliberately break the initialization with a wrong url like `url: url + "xxx",`, to see how the error handling works out.
4. we should try it in the "test" namespace

# Relevant Issues
none

  - there is more logging, but it's in comments

### DIFF
--- a/src/smc-webapp/project/websocket/connect.ts
+++ b/src/smc-webapp/project/websocket/connect.ts
@@ -5,8 +5,8 @@ Create a singleton websocket connection directly to a particular project.
 import { reuseInFlight } from "async-await-utils/hof";
 import { API } from "./api";
 import { retry_until_success } from "../../frame-editors/generic/async-utils";
-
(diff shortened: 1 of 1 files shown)


williamstein:
YEAH!

PR #3308, opened 2018-10-25
Static landing pages improvements

haraldschilly:
# Description
The main aspect of this patch is to introduce a "Features" page pointing to the special latex, R and jupyter notebook pages. I think the wording needs to be fixed a little bit.

# Testing Steps
The pages should still compile. It did work fine for me...

screenshot from 2018-10-25 21-25-17

### DIFF
--- a/src/webapp-lib/_base.pug
+++ b/src/webapp-lib/_base.pug
@@ -117,8 +117,8 @@ html.no-js(lang="en")
             //- li
             //-   a(href=PREFIX + "#a-explore") Explore
 
-            li(class=navbar_highlight("policies"))
-              a(href=PREFIX + "policies/index.html") Policies
+            li(class=navbar_highlight('features'))
+              a(href=PREFIX + "doc/features.html") Features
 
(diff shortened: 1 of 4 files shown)


johnjeng:
This feels too minimal if we expect people to actually look at this page.

johnjeng:
I think 4 tiles with a logo/icon on each one would be better. My point is just something more visual other than text bullet points.

haraldschilly:
@johnjeng right now that page is just empty: https://cocalc.com/doc/features.html

with that PR I just want to tie everything together to be coherent and "connected". I agree that this page (and the one for software) could be much better. I don't want to do this right now, but rather in a PR next week or so.

williamstein:
Looks better than before, and useful for SEO.  Obviously, it'll be great when something stylish like John wants happens.  I plan to merge this tomorrow when I'm fresh.

slel:
Maybe end with a "See also" section with a link to
- https://cocalc.com/doc/software.html

PR #3313, opened 2018-10-29
katex 0.10.1 and fixing CSS files via CDN

haraldschilly:
# Description
ok, this should update katex to 0.10.1 (I didn't commit the lockfiles) and also fixes a couple of referenced CSS files. I wasn't aware that katex is loaded via a CDN, and well, that's obviously not working across versions. There were also other version mismatches and missing integrity checks.

# Testing Steps
1. make sure to update katex in the smc root dir and smc-webapp (npm install or run make)
2. `\copyright` is new and should work (indicator that you really have 0.10 or higher)
3. this also touches print.ts in the frame editor. is this used for printing the html preview in a markdown file? if so, that works
4. markdown cells in ipynb also render those katex formulas
5. formulas in chat
6. upon failure, fall back to mathjax ... I tested this via `\unknown{}` in a formula, then it's red and rendered by mathjax.
7. sharing such a markdown file also renders the formulas fine

I used this snippet in a markdown document:

    $\copyright$

    fraction $y = \frac{2}{3 - x^2}$

    integral fine $\int_{x=0}^{x=1} \sin{x} \cdot \frac{x}{\cos{x}^2} \mathrm{d}x$

    ```python
    def foo(bar):
      bar + baz
    ```

    $\unknown{1+1}$ formula

# Relevant Issues


haraldschilly:
10.1 released: https://github.com/KaTeX/KaTeX/releases/tag/v0.10.1

williamstein:
I'm changing the package.json for katex from `^` to `=` since it seems critical that we keep the version totally under control, and I'll also put some sort of comment in pointing to this PR, so we know all the spots to increase 0.10.1.  I agree that all this loading from CDN's is bad to maintain (and my fault).

haraldschilly:
I never thought about CSS via CDN, that only came to my mind later and indirectly. Good point about using `=` to force the version!

Issue #3315, opened 2018-10-30
add a format button fir .bib files


williamstein:
See this for how to format bib files using a program called biber:

https://tex.stackexchange.com/questions/186733/a-utility-for-auto-formatting-bib-files

PR #3316, opened 2018-10-30
format bib files via biber

haraldschilly:
# Description
this is modeled after the xml formatter, which we streamlined to use cocalc's execute function.

# Testing Steps
1. this only concerns a `.bib` file, which should have a format button.

# Relevant Issues
issue #3315


haraldschilly:
errors are also possible, although lengthy. but here, somewhere it really says where the block is still open. nice.

screenshot from 2018-10-30 21-57-50

PR #3333, opened 2018-11-05
Typescript a Few Things

johnjeng:
# Description
Typescript Tip, Loading, and Space
Typescript feature.coffee

# Testing Steps
1. Use Browserstack to or Chome dev tools to confirm platform detection works
1. Checked that Tips, Loading, and Space work in Project Files and Courses.

# Relevant Issues

### [Checklist](https://github.com/sagemathinc/cocalc/wiki/PR-Checklist):
- [x] No debugging console.log messages.
- [x] All new code is actually used.
- [x] Non-obvious code has some sort of comments.
- [x] A list of exact steps on how you tested.
- [x] Screenshots if relevant.

PR #3334, opened 2018-11-05
Typescript more r misc

johnjeng:
# Description
Typescript basic parts of r_misc.
Move `r_misc` and tests into its own folder.
Use module resolution (`r_misc.cjsx` -> `r_misc/index.cjsx`)

# Testing Steps
1. Run `jest`
1. 
1.

# Relevant Issues
Follow up on #3333 

Front end:


src/smc-webapp/r_misc/selector-input.tsx
```diff
@@ -0,0 +1,82 @@
...
+    | string[]
+    | { value: string; display: JSX.Element }[]
+    | { [keys: string]: JSX.Element };
+  disabled?: boolean;
+  selected?: string;
+  on_change?: (selected: string) => void;
+}
+
+// If the first element is a string, we assume the rest to be a string
+function isStringArrayHeuristic(a: any): a is string[] {
```
williamstein: What does `a is string[]` mean for the return type?  Shouldn't the return type be `boolean`???

I'm not doubting this is valid typescript; it's just that I don't understand it....

src/smc-webapp/r_misc/close-x.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from "react";
+import { Icon } from "./icon";
+
+const closex_style: React.CSSProperties = {
```
williamstein: Oh, so *this* (use `React.CSSProperties`) is how to declare objects and using them for css.  I've been confused about this.    I should have asked you in chat.

src/smc-webapp/r_misc/selector-input.tsx
johnjeng: This is the syntax to tell Typescript that is function is a typechecker. eg
```
// foo: string[] | number;
let length: number = 0;
if (isStringArrayHeuristic(foo)) { // Behaves as if we could do typeof foo == string[]
  length: number = foo.length;  // <-- TS knows foo is a string array so has property length 
} else {
  length = foo; // <-- TS knows foo is a number
}
```

williamstein: (and this code is all merged into master and live on prod already, so this is purely for my education.)

johnjeng: In the documentation is called a "User-Defined Type Guards".

It's on this page https://www.typescriptlang.org/docs/handbook/advanced-types.html

PR #3353, opened 2018-11-20
Fix project upgrades

haraldschilly:
# Description
the overall goal is to limit the contributed upgrades for a project by the configured total amount of maximum upgrades. in particular, already applied upgrades by other users and the free quota were sort of ignored. there is also some consolidation happening, where duplicated code is collected in the same function.

this visualization helped me (it's in the source)

```
0                                total          maximum
|<-------------------------------->|                |
|<----->|<------------------------>|<-------------->|
| admin |  all upgrades by users   | proj remainder |
| +     |<------------>|<--------->|<--------->|    |
| free  |  other users | this user | remaining |    |
|       |              |           | this user |    |
|       |              |<--------------------->|    |
|       |              |  limit for this user  | <= | max
```

# Testing Steps
uff ... many special cases
1. checkbox-upgrades (binary 0/1): 
   * no upgrades → adding 1 is fine
   * user A has upgraded → switch to user B who also wants to upgrade → hitting limit
   * back to user A → remove the same upgrade  → should work
  1.1 like above, but instead of user B, do this for the admin (free quota)
1. non-negative number upgrades:
   * show how much available, and how much left to upgrade
   * also, we have to check the case where the total upgrade is already beyond the limit per project → this does immediately show an error. removing the upgrade removes that error (equivalent to clicking "max", because the max to apply is smaller than the current upgrade value by the user)
1. another small but important detail is to always check the enabled/disabled state of the save button. I don't remember exactly, but there was a situation where it didn't switch to being enabled although it should (probably because the limits are calculated correctly now). It should be enabled if there is indeed a change and none of the errors are active.
1. Here is a way to trigger "too many upgrades" errors: e.g. first (almost) max out mem or disk upgrades, then as admin, add even more upgrades beyond the max, and then edit as the user again. now, the user is told to reduce the upgrade to stay beyond the limit. e.g.

above limit:

screenshot from 2018-11-22 16-41-02

error (save button disabled) -- the screenshot is older, error is different, see below:

screenshot from 2018-11-22 16-41-10

and after clicking max, at the limit and save button enabled

screenshot from 2018-11-22 16-41-18

5. another interesting edge case is if the "admin" upgrades way beyond the upper limit. in that case, the user should be told to reduce upgrades to 0 (their limit), which violates the maximum of the project, but the save button should still be active (otherwise, the user wouldn't even be allowed to remove their quota).

6. also test improvements towards giving users feedback about why it is a problem. Below is a progressing sequence of error messages while being at the allowed limit (where some quotas are already applied -- similarly this should work if there are currently no quotas applied by the user), then exceeding the limit up until the amount of remaining quotas of the user, and then beyond the quotas of the user.

screenshot from 2018-11-23 16-56-19

screenshot from 2018-11-23 16-56-26

screenshot from 2018-11-23 16-56-38

screenshot from 2018-11-23 16-56-47

6.1 similary, in the case where all user upgrades would bring the quota beyond the overall quota limit of the project, the messages report properly:

at the limit
screenshot from 2018-11-23 16-26-43

beyond
screenshot from 2018-11-23 16-26-51

at the maximum of how much quota the user has
screenshot from 2018-11-23 16-27-11

and beyond that
screenshot from 2018-11-23 16-27-18

7. the component also appears in account settings → upgrades

Front end:


williamstein:
I skimmed through the code, fixed some trivial typo, and then tried it.  It **broke** instantly.  

Please convert src/smc-webapp/r_upgrade_adjustor.cjsx to **typescript**, and then I'll look at this again.   I see no point in even considering merging substantial code changes like this with an entire new file's worth of code in coffeescript, when it's just going to trivially be broken in probably dozens of ways that Typescript would automatically catch.    For example, in about 5 seconds I hit this:

screenshot 2018-11-25 at 1 29 28 pm

The typescript compiler would have caught stuff being undefined because you forgot to import it.  I don't want to ever review code again that isn't in Typescript.  It's just too tedious and hard.   If @johnjeng wants to review this he can, but I don't want to.

haraldschilly:
well, since I don't know how to convert this, I moved it back to r_misc. that makes the diff smaller, obviously. I also hope it doesn't overlap with what @johnjeng  is doing.

and btw, the change you made in misc.js was at old code that was already there (not touched by me). however, in 
6b20d2e I combined both function via a single generator function and added tests. that includes your changes.

williamstein:
> since I don't know how to convert this, I moved it back to r_misc.

Good.  As you've found, refactoring code (e.g., moving big chunks to new files) is a basically impossible to get right using CoffeeScript, due to the lack of compile time checks.  It has to be converted to javascript/typescript, or you're just begging for pain.

johnjeng:
If I have 3000 and I want to change it to 4000, deleting the 3 results in all my 0's disappearing.
screen shot 2018-11-28 at 5 01 08 pm
becomes

screen shot 2018-11-28 at 5 01 14 pm

haraldschilly:
@johnjeng I addressed your second comment in c63970f, I hope, but your first one is out of scope. I've started to work on that whole quota calculation code on the backend here #3376 and I want to look into this over there. In particular, as of right now in production (in my test project, but try it yourself), admin: dedicated cpu 2 → save → switches to 0.5. admin: dedicated ram (say) 1234 → save → 0. I really want to keep this PR front-end only and address only the end-user perspective. I'll look into these admin settings, and once I have the whole picture, I'll understand much better what's going on.

src/smc-webapp/r_misc.cjsx
```diff
@@ -1442,18 +1442,52 @@ exports.UpgradeAdjustor = rclass
...
+        #   |<----->|<------------------------>|<-------------->|
+        #   | admin |  all upgrades by users   | proj remainder |
+        #   | +     |<------------>|<--------->|<--------->|    |
+        #   | free  |  other users | this user | remaining |    |
+        #   |       |              |           | this user |    |
+        #   |       |              |<--------------------->|    |
+        #   |       |              |  limit for this user  | <= | max
+        #
+        #   admin/free: could be 0
+        #   all upgrades by users is total_project_quotas
```
johnjeng: This definition and line 1479 `proj_remainder = misc.map_diff(proj_maximum, total_upgrades)` are conflicting. Based on the correct behavior, it seems `total_project_quotas` is actually `total` in this diagram above.

@haraldschilly Please confirm if this is true.

haraldschilly: I pushed a slight change, such that the variable names show up in the diagram. I hope I haven't confused myself again when looking at it :face_with_head_bandage:

williamstein:
I fixed the merge conflict.

PR #3374, opened 2018-11-28
testing: tweaking travis

haraldschilly:
testing only


haraldschilly:
@johnjeng this combines your PR and my tweaks to also do the full "production" build of the webapp in a similar config parameter space as it is done for kucalc

PR #3359, opened 2018-11-23
webapp/project files: library and +new

haraldschilly:
# Description
Reconcile +New contols in Files and fix some layout issues...

# Testing Steps
1. no files
    * create button → dialog
    * searching → create, also some buttons
1. there are files
    *  no search text, +create button
    * some other text in search → create file mode
1. editing the filename works, creating a file should show up with the indicated filetype
1. attempt to create a file without extension → error shows up. 
   *  similarly an error for  a stopped project. ← **that's no longer the case**. looks like this is also broken in production, though. this should be a follow up issue.
1. uploading 
   * by dropping onto existing files
   *  by dropping onto drop area in +create panel
   * in both cases the files listing updates 
1. actions that previously opened up +new should no longer do so (e.g. in jupyter menu, file → new)
1. library shows up and hides when clicking on close
1. grabbing a copy from the library works as usual, and the files listing switches to the target directory
1. settings panel boxes in project settings show up fine

# Relevant Issues

### [Checklist](https://github.com/sagemathinc/cocalc/wiki/PR-Checklist):
- [ ] No debugging console.log messages.
- [ ] All new code is actually used.
- [ ] Non-obvious code has some sort of comments.

Front end:
- [ ] Restart at least one project and check its `~/.smc/local_hub/local_hub.log`
- [ ] Completely restart Webpack with `./w` in `/src`
- [x] Screenshots if relevant.

### basic layout 

screenshot from 2018-12-13 21-09-45

### create panel

screenshot from 2018-12-13 21-07-55

### no files, typing in search text

screenshot from 2018-12-13 21-11-14

### file action delete opened

screenshot from 2018-12-13 21-09-10

### library

screenshot from 2018-12-13 21-08-16

### terminal output
* also still scrollable, aligns on the right hand side, changed the clear button

screenshot from 2018-12-13 21-09-58

### placement of files button and dropdown changed slightly

screenshot-cocalc com-2018 12 13-21-11-53

### upload done + showing the uploaded file

screenshot from 2018-12-13 21-18-53

PR #3376, opened 2018-11-29
hub/quota: test, maximum, refactoring, ...

haraldschilly:
# Questions
* why is that cpu request divided by 1024? I don't understand that at all. Is somewhere an additional factor at play which I don't see? in any case, I'm confused, and at least initially I'm preserving the previous behavior.
   * hsy: I think this is from code a long time ago. The factor is divided away anyways.

# Description
* this properly caps user contributed upgrades
* this depends on #3359 ... not technically, but semantically
* related to that, settings-quotas for the cpu shares of 256 should be set to zero. then the part in the code removed that subtracts 256 out of nowhere. then it's more consistent.
  * I've not changed that part, but instead let's merge this first and then run the query to fix the `cpu_shares`. Then as a follow up we remove that part of the code.

# Testing Steps
1. this calls for k2

# Relevant Issues

Front end:


johnjeng:
What does `PR-pending` mean?

haraldschilly:
they're listed here: https://github.com/sagemathinc/cocalc/labels?utf8=%E2%9C%93&q=pr-

>  "finished, but depends on another branch"

together with work in progress I try to signal that this is not yet ready but close.

haraldschilly:
question for you, maybe, what does "PR-RFC" do? there is one with `[RFC]` in the title, too.  my first guess is "request for comments" ?

johnjeng:
Huh, I don't remember creating the RFC label. But yeah, it's request for comments. Maybe I should just use `needs-help`.

johnjeng:
Oh, wow there are descriptions now. That's great.

src/smc-hub/kucalc/quota.coffee
```diff
@@ -59,63 +88,64 @@ exports.quota = (settings, users) ->
     #            quota.privileged = true
     #            break
 
+    # little helper to caluclate the quotas, contributions, and limits
```
williamstein: typo: "caluclate"

PR #3385, opened 2018-11-30
Fix hash-bar bottom-margin

johnjeng:
# Description
On production, notice the whitespace below the hashtags:
screen shot 2018-11-30 at 4 11 06 pm

Fixed with this PR:
screen shot 2018-11-30 at 4 11 19 pm

Example of many hash tags:
screen shot 2018-11-30 at 4 15 40 pm

# Testing Steps
1. Open a tasks file
1. Make some hashtags
1. Click on some
1. It should not produce extra whitespace below the bar
1. Add a lot of hashtags
1. They should all render

# Relevant Issues

Front end:

### DIFF
--- a/src/smc-webapp/tasks/hashtag-bar.cjsx
+++ b/src/smc-webapp/tasks/hashtag-bar.cjsx
@@ -2,7 +2,7 @@
 Hashtag bar for selecting which tasks are shown by tags
 ###
 
-{Button, ButtonGroup} = require('react-bootstrap')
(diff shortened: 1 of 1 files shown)


williamstein:
Incidentally, the point of the change that made things ugly in the first place was to make it so if there are a ton of hashtags they don't take up the whole screen (hence the overflowY), which is pretty bad on mobile (e.g., phones).   For some reason your fix here breaks that.  I've put in an additional max-height, which fixes things.

PR #3388, opened 2018-12-03
Add signup question

johnjeng:
# Description
Add a simple one line question to sign up to help us understand what a user wants when they sign up.

# Testing Steps
1. Create a new account with some data in the field
1. Inside the hosting project, use `psql smc` and the query `select account_id, tracking_info from accounts;`
1. See the data there.
1. Create an account without the field filled out
1. It should still work.

screen shot 2018-12-03 at 6 54 38 pm
# Relevant Issues

Front end:


haraldschilly:
I've two thoughts:

* the dialog: I would be a bit confused about that first question. Maybe it looks better if there is a bit of a space between the question and the first name box? Instinctively, I imagine myself looking at that box for the first time, and I would enter my name as the answer.
* the entry in the database. I would make this a bit more general, such that it can be used for more than just tracking. maybe two levels? e.g. ` '{info : {  tracking   : { sign_up_usage_intent : .... }}}'`. looks like overengineering, but I think you understand my point

williamstein:
@johnjeng -- I have to agree with @haraldschilly , that it's critical not to confuse new users with this.  Just move the question to the end, and nobody is going to accidentally put their name their.    Put "(optional)" in the label too.

williamstein:
> the entry in the database. I would make this a bit more general, such that it can be used for more than just tracking. maybe two levels? e.g. '{info : { tracking : { sign_up_usage_intent : .... }}}'. looks like overengineering, but I think you understand my point

I *disagree*.  Making a field called "info" is not useful, since every single field is **info**rmation.  I would go the opposite direction and just have a string field called "sign_up_intent" and that's it.   I'm not nearly as good at writing complicated SQL queries as you are @haraldschilly, which is why I would change this to be way simpler, not more complicated.   Also, it's way, way more consistent with how things are already to be simple.  E.g., right now we have a field called `created` with the time when the account was created.  Having sign_up_intent be a top level field is consistent with that.  Having it nested multiple layers deep is not - if we really need such a complicated structure, make a whole new table.

johnjeng:
>Instinctively, I imagine myself looking at that box for the first time, and I would enter my name as the answer.

Really? Is it possible for you to imagine that? What you're saying is that on every sign up form, you type your name into the first box regardless of any text. I have a lot of mental short cuts but I have a hard time believing this.

My argument for putting it first is that they're more likely to write something here.

I'll put "(optional)" though.

williamstein:
You are adding friction for the purpose of feedback.  There was a big
discussion on HN yesterday about adding friction for feedback being a
general trend...

johnjeng:
On the database end, I agree with William. `info` is too generic. Everything on there is info. I guess we might want tracking info to be semantically something else entirely (timeseries data for example). 

I'll pull `sign_up_usage_intent` to the top and put "(optional)".

johnjeng:
Yeah, I agree, I am adding friction for the purpose of feedback. I think with a one line box, the balance is good here. Was [this](https://news.ycombinator.com/item?id=18597960) the discussion you're talking about?

williamstein:
OK, I'm fine with keeping the question for for now.  If we receive any complaints from users I'll switch to the bottom though.  Often changes like this result in angry emails from users, so we'll see.  OK?

johnjeng:
Updated
screen shot 2018-12-06 at 5 21 41 pm

williamstein:
I plan to merge this as it is now.  Due to the DB change, I want to wait until the evening (PST time) or possibly Friday, since it will be slightly disruptive rolling this out.

Issue #3389, opened 2018-12-03
can't scroll terminal


haraldschilly:
there is a vertical scrollbar next to each terminal. I can't use it, because there is somehow another div in front of it (it's the one which displays the terminal content, because when I try to scroll, it selects the text).

Below is a screenshot, where I changed the selected `div`'s z-index to 1 and background to transparent. with these changes I can use the scrollbar again, while I can still mouse-wheel scroll and select terminal text. I don't know if these changes actually make sense, though, or if there is a better solution. That's just what did work for me by editing the DOM directly.

![screenshot from 2018-12-03 13-59-18](https://user-images.githubusercontent.com/207405/49375211-f8724100-f703-11e8-83c5-7c94709a0da8.png)

my browser is chrome 71

williamstein:
I see the same thing in Chrome 70 on ChromeOS Stable.  This was likely caused by switching to the DOM renderer (from canvas).  It's an upstream issue, but I'm fine with us trying to fix it somehow.

PR #3392, opened 2018-12-04
webapp/terminal: fixing hidden scrollbar

haraldschilly:
# Description
This brings the div with the scrollbar in front of the other one: #3389

There is also some artifact left from exposing the render type … but it's no good to have that context menu even in dom mode.

# Testing Steps
I tested this with a bright (black on white) and a dark theme, in chrome and firefox, and also when I've both opened at the same time. Didn't notice any Problems.

# Relevant Issues
 #3389

Front end:

### DIFF
--- a/src/smc-webapp/frame-editors/terminal-editor/connected-terminal.ts
+++ b/src/smc-webapp/frame-editors/terminal-editor/connected-terminal.ts
@@ -118,8 +118,12 @@ export class Terminal {
     this.set_connection_status("disconnected");
   }
 
+  static get rendererType() {
+    return "dom";
+  }
+
   private get_xtermjs_options(): any {
-    const rendererType = "dom";
+    const rendererType = Terminal.rendererType;
     const settings = this.account.get("terminal");
     if (settings == null) {
       // not fully loaded yet.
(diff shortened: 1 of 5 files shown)


src/smc-webapp/frame-editors/terminal-editor/connected-terminal.ts
```diff
@@ -118,8 +118,12 @@ export class Terminal {
     this.set_connection_status("disconnected");
   }
 
+  static get rendererType() {
+    return "dom";
+  }
+
   private get_xtermjs_options(): any {
-    const rendererType = "dom";
+    const rendererType = Terminal.rendererType;
```
williamstein: Why did you change this to a computed getter?   I can't see any gain, except to make things more complicated and harder to understand.  Also, the code is now inconsistent in style to the code for `const fontFamily = "monospace";` directly below.   

The only other renderer is canvas, and it is unusably slow to create new Terminals, so we will likely never use it.

williamstein: Can you address whether this fix should really be upstream (in your opinion)?

haraldschilly: well, I just wanted this to be read-only attribute... there is no `const this.attr = ...`

williamstein:
Please read about readonly and private here then change this code...

https://www.typescriptlang.org/docs/handbook/classes.html

haraldschilly:
I changed it to a readonly attribute.

PR #3393, opened 2018-12-04
testing: kucalc hub mode

haraldschilly:
- the main change here is to have a `--test` flag for hubs, which just means they exit with `0` when they were able to register without an error at the database. this is the only part that actually needs review.
- artificial `err = ...` error, introduced in hub_register leads to failed build, e.g. https://travis-ci.org/sagemathinc/cocalc/jobs/463348367#L1887
- there are two non-trivial changes with those python support scripts: instead of just `os.system` and always exiting with `0`, there are now `if os.WEXITSTATUS(os.system(s)):` checks.
- and well, beyond what's being run here, I think it would also fit to start a project. I don't know how, though. Would it make more sense to mimic how the docker-version of cocalc is being run?

## testing

the (obvious) main aspect to test is if starting the hub like usual works exactly like before. it worked fine in my project.


haraldschilly:
I did build this in namespace "test" and after 15 minutes all hubs work fine. Hence I'm merging this.

PR #3431, opened 2018-12-23
update prettier to be in sync with system-wide version

haraldschilly:
# Description
Compare with

```
$ /usr/bin/prettier --version
1.15.3
```

... and the entry in the global package.json is because as some dependency, a version 1.11 would be installed. this makes sure all of them are the same.

# Testing Steps
1. I tested formatting an Rmd file and a cell in a python3 jupyter notebook.

# Relevant Issues

Front end:

### DIFF
--- a/src/package.json
+++ b/src/package.json
@@ -46,6 +46,7 @@
     "node-cjsx": "^1.0.0",
     "node-glob": "^1.2.0",
     "node-sass": "^4.8.2",
+    "prettier": "^1.15.3",
     "pug": "^2.0.1",
     "pug-loader": "^2.3.0",
     "react-test-renderer": "^16.5.2",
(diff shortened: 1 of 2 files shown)


johnjeng:
Does this make any difference anywhere except when someone is editing outside CoCalc? ie. VScode.

haraldschilly:
well, I don't remember exactly, and it's related to the software updates (the global ones). I think the main problem is cc-in-cc development. Imagine there is a subtle difference between the global version (which runs for everyone) and the (in this case here) outdated older version in the dev environment. For example, a command-line switch or support for more languages. When you want to test this in your dev project (cc-in-cc), you are using a different executable:

```
~$ which prettier
/home/user/smc/src/node_modules/.bin/prettier
```

testing works, but deploying in prod might fail.

PR #3437, opened 2019-01-08
desktop notifications POC

tscholl2:
# Description
I have started using webpages in full screen because I have a small laptop. I would like to know wheb people chat at me, even if I'm not currently looking at CoCalc (which happens sometimes). Desktop notifications seem like they would solve this.

# Testing Steps
1. Open a project with account A
1. Open the same project with a different account B
1. Account B does something to increase the notification count for account A (work on a shared file)
1. Account A should get a desktop notification

# Relevant Issues
Some people find desktop notifications extremely intrusive. This code aims to be polite, but I'm pretty sure it could still lead to a blow-up with lots of notifications. There is a lot of ways you could lessen the amount of notifications: show up to 1 per minute, show one per project, show one per person, etc. I can probably edit the PR if the official CoCalc team (a) actually wants or cares abut this feature, and (b) has a preferred plan to limit notifications.

Front end:


haraldschilly:
Hmm, by just looking at the code, I like the overall idea, but I would request at least two improvements

1. disabled by default (i.e. no question the first time). this is rather easy to do, with a toggle switch in the account settings. I could make a commit and push this here for you, and then what's left is to figure out how to use that value.
2. Just "new file activity" is a bit boring. I would expect the filename, at least :-)
3. Stretch goal would be throttling. My first thought is to add file names to a global queue and call the notification function, which only triggers once a minute tops. It then empties the queue of file names.

tscholl2:
Great! I will continue with this then. I will definitely need some help, especially with wiring up an account setting.

But first, in order to do anything fancier, I think will require a bigger change because the `NavigationBell`  component does not look like it has any extra information to display. Is there a way to add a "listener" for when this information changes in the store? Just browsing, it looks like maybe there is somewhere in the [`FileUseStore`](https://github.com/sagemathinc/cocalc/blob/e529cce3310b12042816f56eff80f41396fb801f/src/smc-webapp/file_use.cjsx#L137) that I can add a listener. This way the notifications won't be tied to a specific react component. It feels better to have the notifications fired closer to the source, but it's much less obvious how to do that.

The actual `new_desktop_notification` function is essentially a utility function. Should it could be moved to some file with `misc` in the name.

Also, I realized when CoCalc is in its "suspended" mode (which happens if it's not an active tab for a while?), does anything still run? Or will notifications also be muted?

williamstein:
> Also, I realized when CoCalc is in its "suspended" mode (which happens if it's not an active tab for a while?), does anything still run? Or will notifications also be muted?

Good point -- Notifications would be muted.  That makes this approach potentially limited.

haraldschilly:
@tscholl2 I did push a change that introduces a setting. when enabled, it immediately asks the user about it, because that way it is clear which action causes which result. there is also an alert in case it is disabled. what I mean by "left of URL bar" is this (knowing about this dropdown was useful while testing this, too!)

screenshot from 2019-01-09 14-14-38

tscholl2:
The latest change adds throttling. However, it's not quite as intuitive as I first imagined, but it seems to solve my use case.

As far as I can tell, "notification count" is done by file. So multiple chats in the same file do not re-trigger the notifications. Putting this logic in the `NotificationBell` component still feels off. But I'm not sure how to add a listener to the store. I was trying to figure out if I could put this stuff [here](https://github.com/sagemathinc/cocalc/blob/master/src/smc-webapp/file_use.cjsx#L253), but I wasn't certain.

Also, I changed it to get the actual notifications from the store. This should allow displaying more detailed information such as filenames, chat message previews, etc. But when I was introspecting the `file_use` objects returned by [this function](https://github.com/sagemathinc/cocalc/blob/master/src/smc-webapp/file_use.cjsx#L300) I didn't understand how to use them properly to look up the extra details.

williamstein:
> As far as I can tell, "notification count" is done by file.   So multiple chats in the same file do not re-trigger the notifications.  

True.

> [...]  I didn't understand how to use them properly

Note that I've completely rewritten sync in a different branch (syncts) in typescript, etc....

williamstein:
> src/smc-util/db-schema.coffee

Is now javascript...

tscholl2:
I will try to update this branch.

johnjeng:
This code is surprisingly short. I'll look at it next week.

Issue #3438, opened 2019-01-09
webapp/tex editor: e[Symbol.iterator] is not a function


haraldschilly:
This happens only for `.tex` files:

chrome: `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36`

smc_git_rev: a1abbd8bcbc2ad60ea0d3b75e8cbf70006969512

Date: 2019-01-08 12:44:12.099518

`e[Symbol.iterator] is not a function`

```
TypeError: e[Symbol.iterator] is not a function
    at https://cocalc.com/static/smc-c15451d38526f55c6480.cacheme.js?c15451d38526f55c6480:5:98344
    at r (https://cocalc.com/static/smc-c15451d38526f55c6480.cacheme.js?c15451d38526f55c6480:5:98504)
    at Array.<anonymous> (https://cocalc.com/static/smc-c15451d38526f55c6480.cacheme.js?c15451d38526f55c6480:5:108263)
    at https://cocalc.com/static/1-c15451d38526f55c6480.cacheme.js?c15451d38526f55c6480:6:2012892
    at https://cocalc.com/static/1-c15451d38526f55c6480.cacheme.js?c15451d38526f55c6480:6:4580334
    at mo.iterN (https://cocalc.com/static/1-c15451d38526f55c6480.cacheme.js?c15451d38526f55c6480:6:186922)
    at Mo.iterN (https://cocalc.com/static/1-c15451d38526f55c6480.cacheme.js?c15451d38526f55c6480:6:188460)
    at Mo.iter (https://cocalc.com/static/1-c15451d38526f55c6480.cacheme.js?c15451d38526f55c6480:6:194330)
    at Ca.eachLine (https://cocalc.com/static/1-c15451d38526f55c6480.cacheme.js?c15451d38526f55c6480:6:264548)
    at i (https://cocalc.com/static/1-c15451d38526f55c6480.cacheme.js?c15451d38526f55c6480:6:4580242)
```

---

tracking down top frame:

the first part is webapp client (and ok), the second part right after it is not.  hmm. I found `arr[Symbol.iterator]()` in a `try{for ...` block in `smc-webapp/node_modules/@babel/helpers/lib/helpers.js`. Not quite 100% this code, though, but it's close.

[...]

PR #3439, opened 2019-01-09
webapp: update react-sortable-hoc due to #3438

haraldschilly:
# Description
This is motivated by https://github.com/sagemathinc/cocalc/issues/3438 ← which happens quite frequently.
After updating it, I noticed that `smc-webapp/node_modules/react-sortable-hoc/dist/` already contains compiled files, which suggests to me that this incompatibility could be taken care of.

Maybe what I did update is nice, but doesn't solve the issue.

# Testing Steps
1. I don't really know what to test ... it must be related to `.tex` files (according to the stacktraces), but the code doesn't suggest so (it's for draggable navigation and tasks). Hmm. Dragging editor tabs, projects and tasks around works for me, though.
2. just compiling it in my cc-in-cc project and opening a `.tex` file works

# Relevant Issues
https://github.com/sagemathinc/cocalc/issues/3438

Front end:

### DIFF
--- a/src/smc-webapp/package.json
+++ b/src/smc-webapp/package.json
@@ -88,7 +88,7 @@
     "react-json-inspector": "^7.0.1",
     "react-redux": "^5.0.7",
     "react-select": "^1.3.0",
-    "react-sortable-hoc": "^0.6.8",
+    "react-sortable-hoc": "^1.3.1",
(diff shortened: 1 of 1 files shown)


williamstein:
@johnjeng  review this!

Issue #3440, opened 2019-01-09
webapp: Cannot read property 'charAt' of undefined


haraldschilly:
Happens with sagews, ipynb and other files

chrome: `Mozilla/5.0 (Windows NT 6.1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36`

smc_git_rev: a1abbd8bcbc2ad60ea0d3b75e8cbf70006969512

Date: 2019-01-08 08:29:48.102757

`Cannot read property 'charAt' of undefined`

```
TypeError: Cannot read property 'charAt' of undefined
    at Object.t.isAbsolute (https://cocalc.com/static/1-c15451d38526f55c6480.cacheme.js?c15451d38526f55c6480:6:588105)
    at Object.t.normalize (https://cocalc.com/static/1-c15451d38526f55c6480.cacheme.js?c15451d38526f55c6480:6:587930)
    at y (https://cocalc.com/static/smc-c15451d38526f55c6480.cacheme.js?c15451d38526f55c6480:5:3970497)
    at r.close_file (https://cocalc.com/static/smc-c15451d38526f55c6480.cacheme.js?c15451d38526f55c6480:5:3987091)
    at r.close_tab (https://cocalc.com/static/smc-c15451d38526f55c6480.cacheme.js?c15451d38526f55c6480:5:3977693)
    at t.close_file (https://cocalc.com/static/smc-c15451d38526f55c6480.cacheme.js?c15451d38526f55c6480:5:1338674)
    at t.onMouseDown (https://cocalc.com/static/smc-c15451d38526f55c6480.cacheme.js?c15451d38526f55c6480:5:1338986)
    at Object.<anonymous> (https://cocalc.com/static/1-c15451d38526f55c6480.cacheme.js?c15451d38526f55c6480:6:4863290)
    at f (https://cocalc.com/static/1-c15451d38526f55c6480.cacheme.js?c15451d38526f55c6480:6:4863328)
    at https://cocalc.com/static/1-c15451d38526f55c6480.cacheme.js?c15451d38526f55c6480:6:4863966
```

---

tracking down function `y`:

**→ this is the `normalize` function at the top of project_actions.ts, used in `close_file`.**

```
function y(e){return"."===(e=m.normalize
(e))?"":e}
"undefined"!=typeof window&&
null!==window&&
(c=n(71),p=n(327));
var b=n(8),
v=n(148).MARKERS,
[...]

PR #3441, opened 2019-01-09
webapp/project actions: path in close_tab is not always defined -- #3440

haraldschilly:
# Description
please see #3440. in a nutshell, a function in misc.js might return null or undefined, but the typescripted code isn't aware of that. I also think there are no negative side effects to be concerned about …

# Testing Steps
1. I don't know why this happens in the first place, just interpreting the stacktrace.

# Relevant Issues

#3440 

Front end:


williamstein:
It would be better to figure out and flag everything that calls close_path incorrectly.  If our code were 100% typescript and typed that would be easy.  But it isn't, so let's go with this for now.  Later we can change the typing back and see what pops out.

haraldschilly:
Yes, I agree. In #3440 I found the root cause for this one:

`misc.tab_to_path` says:
```
// assumes a valid editor tab name...
// If invalid or undefined, returns undefined
```

but there is no (implicit) TS info that it could return undefined, and hence no bells go off.

PR #3523, opened 2019-01-30
Track various user interactions via Google Analytics

johnjeng:
# Description
Add more Google analytics events so that the events view becomes more useful.

screen shot 2019-01-30 at 1 59 22 pm

- [x] Clicking on any top nav button (including a project tab)
- [x] Creating a project from projects page
- [x] Opening a project from the projects page
- [x] Clicking on any navigation tabs inside a project
- [x] Creating a file or folder from the big button in projects listing
- [x] Creating a file or folder from the Create button next to file listing search
- [x] Opening or closing side chat

# Testing Steps
1. Not sure. Is there a way to get a dev Google analytics in a test env?

# Relevant Issues

Front end:

### DIFF
--- a/src/smc-webapp/app_shared.cjsx
+++ b/src/smc-webapp/app_shared.cjsx
@@ -33,6 +33,7 @@ misc = require('smc-util/misc')
 {FileUsePage} = require('./file_use')
 {AdminPage} = require('./admin')
 {show_announce_end} = require('./redux_account')
+{analytics_event} = require('./tracker')
 
 ACTIVE_BG_COLOR = COLORS.TOP_BAR.ACTIVE
 feature = require('./feature')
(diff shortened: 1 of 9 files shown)


haraldschilly:
Hmm, I'm curious about the data, but I would rather record this in our own DB and avoid an external service. I'm aware that we can't do all these visualizations out of the box, but we would have the potential of more insight and "depth". I'm also much happier to not siphon out more data, and well, I would even change how we record these events right now.

williamstein:
@haraldschilly your comment is a little vague to me.    

I'm tempted to merge this PR as is (assuming it works), and collect data for a week or two and let @johnjeng (and others) stare at the results.  If he can use the data to do (or even suggest) something useful to improve cocalc, then it is working.  If he can't, we disable collection and do something else.   

To me the main criteria for whether we should collect data is:  will it result in improvements to cocalc or not?   

It seems like your criteria is different...?

johnjeng:
>we would have the potential of more insight and "depth".

Google analytics lets you query the underlying data so it doesn't seem any different regarding "depth".

Issue #3549, opened 2019-02-06
project log: cursor keys scrolling breaks easily


haraldschilly:
**updated bug description:**

up/down keys should scroll the log entries, but clicking once on the log page or clicking on the previous/next page buttons unfocuses and those keys just move the page. This means the listener to keystrokes should listen on the whole log page. Additionally -- stretch goal -- make page up/down keys flip through the log pages.

![screenshot from 2019-02-06 23-37-59](https://user-images.githubusercontent.com/207405/52378567-5145b900-2a68-11e9-952b-aec196cfc4a5.png)

williamstein:
It's a cursor.   Try hitting the up and down arrow keys on your keyboard and you'll understand. You can select any log entry with your keyboard, then hit return to open the corresponding file.

Our file listing page has a very similar behavior with the cursor, except that it looks visually **completely different** and works slightly differently...  That's inconsistent.

In any case, for the log, should we add a little it of grey text somewhere, like "Use cursor keys to select a log entry. If it is opening a file, you can hit return and open that same file."   ?

haraldschilly:
Ahh! That's what I thought, but I tried the cursor keys, but they didn't work. I even speculated if it scrolls via j/k (vim style).

The bug is this: clicking once on the page (e.g. in my case, I can replicate it by clicking on the area on the right hand side of the listing, maybe row 10 or so) the focus is lost and key commands are ignored.

Or even more striking: clicking once on "next page/previous page" removes the focus and keystrokes are ignored as well.

williamstein:
I'm closing this, since those buttons are gone, and I think the behavior isn't that confusing...

PR #3551, opened 2019-02-07
webapp/project log: focus search box when clicks bubble up -- 

haraldschilly:
# Description
This makes sure the search box is focused, such that arrow keys and so on do work.

# Testing Steps
0. open log with more than one page of entries
1. click somewhere, and type in
2. similarly, up/down keys still work
3. clicking on a link opens the file as usual
4. also works after clicking on "next page"

# Relevant Issues
#3549

Front end:

### DIFF
--- a/src/smc-util/client.coffee
+++ b/src/smc-util/client.coffee
@@ -282,11 +282,14 @@ class exports.Connection extends EventEmitter
                 @emit("data", channel, data)
         @_connected = false
 
-        # start pinging -- not used/needed for primus, but *is* needed for getting information about server_time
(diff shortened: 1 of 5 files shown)


williamstein:
Unfortunately, this PR introduces exactly the sort of bug you reported for a long time:  https://github.com/sagemathinc/cocalc/issues/3069

With this PR merged, if you click anywhere in the log, the page jumps.  The correct fix is to change the code so that the box doesn't have to be focused, but up/down/enter are still captured, by installing (and then removing) a global key handler.

williamstein:
Well I've decided to merge this -- it's better than nothing and can be improved later.

PR #3553, opened 2019-02-07
More analytics tracking

johnjeng:
# Description
Tracking Changes:
- signed_in now semantically matches the user choosing to sign in instead of reconnecting to Cocalc
- Change `topNav` category to `top_nav`, matching formatting in other categories
- Change `project_file` category to `project_file_listing` to be more descriptive

Add tracking to the following actions

## Projects Listing Page
- Filter projects by hashtag 
- Filter projects by Hidden
- Filter projects by Deleted

### File listing
- Download
- Delete
- Rename
- Duplicate
- Move
- Copy
- Share
- Library
- Using the mini-terminal
- Using the in-search-terminal

### Project Settings Page
- Add Collaborator
- Hide the project
- Delete the project
- Add an SSH Key to the project
- Reset the project
- Stop the project
- Change the project image

## Opened File
- Opening the Assistant
- Sending side chat messages
- Toggling video chat

# Testing Steps
1. Test that each of the actions above haven't been broken

# Relevant Issues

Front end:


src/smc-webapp/project_settings.cjsx
```diff
@@ -434,9 +435,18 @@ HideDeletePanel = rclass
     toggle_delete_project: ->
         @actions('projects').toggle_delete_project(@props.project.get('project_id'))
```
williamstein: Please move all the analytics events that are here to actions.  If you put them here, it's a bunch of extra code that has nothing to do with the component **in the component**, which will make eventual unit testing we need to do on components more complicated.  Also, and more importantly, there could be *many* different ways that the toggle_delete_project action (say) gets called.  They should all generate an analytics event, and would if this code were in the project actions class.   Optimally, there would be only one line added to each action (not 4 lines of code), and the actual explicit calls to analytics_event would all be in a new typescript file (thus much more cleanly separating the code).

williamstein: Actually, maybe I'm wrong.  We would lose some information about what is happening with what I'm suggesting, and it's easy enough to mock analytics_event to be a no-op for the purposes of unit testing.  So never mind.

PR #3555, opened 2019-02-08
project/jupyter: set environment only for sage kernels

haraldschilly:
# Description
Only set the environment of a kernel, if it is a sage kernel. (starting with `sage...`).

The result of this is, that a kernel config file for a python kernel containing

```
{
 [...]
  "env": {    "CUSTOM_VAR": "123"  }
}
```

will end up as 

screenshot from 2019-02-08 21-52-06

# Testing Steps
There isn't much to test, since it doesn't change anything for sage kernels and fixes all other cases.

# Relevant Issues

Front end:

### DIFF
--- a/src/smc-project/jupyter/jupyter.ts
+++ b/src/smc-project/jupyter/jupyter.ts
@@ -237,13 +237,16 @@ export class JupyterKernel extends EventEmitter
     dbg("spawning kernel...");
 
     const opts: any = { detached: true, stdio: "ignore" };
-    if (this.name.indexOf("sage")) {
-      // special environment for sage-based kernels
+
+    if (this.name.indexOf("sage") == 0) {
+      dbg("setting special environment for sage.* kernels");
       opts.env = SAGE_JUPYTER_ENV;
     }
+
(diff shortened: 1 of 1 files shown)


src/smc-project/jupyter/jupyter.ts
```diff
@@ -237,13 +237,16 @@ export class JupyterKernel extends EventEmitter
     dbg("spawning kernel...");
 
     const opts: any = { detached: true, stdio: "ignore" };
-    if (this.name.indexOf("sage")) {
-      // special environment for sage-based kernels
+
+    if (this.name.indexOf("sage") == 0) {
```
williamstein: Before I merge this, is this needed?  It seems like this was set before **if and only if** it was NOT a sage kernel, so it was never being set in the case of any sage kernel.... hence it can't be needed, and a better fix is to delete the code?

haraldschilly: now I understand, well, I think it's good to set the variable for R. the one for python is already set, and well, I'll refine the patch after a few tests.

Issue #3601, opened 2019-02-18
jupyter2: this.props.message.get is not a function


haraldschilly:
and ipynb file:

firefox: `Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:65.0) Gecko/20100101 Firefox/65.0`

smc_git_rev: 89520c19d43b306cb60134985e72a516ef5a5bc5

Date: 2019-02-18 15:53:41.861852

`this.props.message.get is not a function`

```
N</t.prototype.render@https://cocalc.com/static/smc-6a37935cf47f274bceff.cacheme.js?6a37935cf47f274bceff:5:992408
zo@https://cocalc.com/static/0-6a37935cf47f274bceff.cacheme.js?6a37935cf47f274bceff:6:4243902
Eo@https://cocalc.com/static/0-6a37935cf47f274bceff.cacheme.js?6a37935cf47f274bceff:6:4243699
Oo@https://cocalc.com/static/0-6a37935cf47f274bceff.cacheme.js?6a37935cf47f274bceff:6:4247546
Wa@https://cocalc.com/static/0-6a37935cf47f274bceff.cacheme.js?6a37935cf47f274bceff:6:4271615
Ga@https://cocalc.com/static/0-6a37935cf47f274bceff.cacheme.js?6a37935cf47f274bceff:6:4271999
zs@https://cocalc.com/static/0-6a37935cf47f274bceff.cacheme.js?6a37935cf47f274bceff:6:4278886
As@https://cocalc.com/static/0-6a37935cf47f274bceff.cacheme.js?6a37935cf47f274bceff:6:4278266
Rs@https://cocalc.com/static/0-6a37935cf47f274bceff.cacheme.js?6a37935cf47f274bceff:6:4279641
Mr@https://cocalc.com/static/0-6a37935cf47f274bceff.cacheme.js?6a37935cf47f274bceff:6:4211487
_/e._wrapper@https://cocalc.com/static/smc-6a37935cf47f274bceff.cacheme.js?6a37935cf47f274bceff:5:685356

```

---

tracking down the top frame

```
t.prototype.render=function(){
return c.React.createElement("pre",{style:j},
JSON.stringify(this.props.message.toJS()))},t}(c.Component),
N=function(e){function t(){
return null!==e&&e.apply(this,arguments)||this}
return o(t,e),t.prototype.render=function(){
var e,t=null!=(e=this.props.message).get("more_output")? <<<<<<<<<<<<<<<
[...]

PR #3602, opened 2019-02-18
webapp/jupyter2: output "message" can be null

haraldschilly:
#  Description

evidence provided by #3601, that's all. I've no idea "why"...

# Testing Steps
I don't know how this could happen, so, no idea.

# Relevant Issues
* #3601

Front end:


williamstein:
Unfortunately, this "fix" is completely wrong.

1. Message is already checked for null immediately before in the code that creates that component. It's definitely impossible for props.message to be null, ensured by both reading the code and the compiler.

2. The error message is `this.props.message.get is not a function `, which is not what you get if props.message is null.  You would get a completely different error.  This problem might happen if message is supposed to be immutable.js, but for some reason it isn't.

Issue #3603, opened 2019-02-18
jupyter2: clear_all_outputs → Unable to get property 'forEach' of undefined or null reference


haraldschilly:
for many ipynb files:

chrome: `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/52.0.2743.116 Safari/537.36 Edge/15.15063`

smc_git_rev: 89520c19d43b306cb60134985e72a516ef5a5bc5

Date: 2019-02-18 15:24:23.753874

`Unable to get property 'forEach' of undefined or null reference`

```
TypeError: Unable to get property 'forEach' of undefined or null reference
   at t.clear_all_outputs (https://cocalc.com/static/smc-6a37935cf47f274bceff.cacheme.js?6a37935cf47f274bceff:5:1816480)
   at clear all cells output.f (https://cocalc.com/static/smc-6a37935cf47f274bceff.cacheme.js?6a37935cf47f274bceff:5:748272)
   at t.command (https://cocalc.com/static/smc-6a37935cf47f274bceff.cacheme.js?6a37935cf47f274bceff:5:1849568)
   at Anonymous function (https://cocalc.com/static/smc-6a37935cf47f274bceff.cacheme.js?6a37935cf47f274bceff:5:1705270)
   at Anonymous function (https://cocalc.com/static/0-6a37935cf47f274bceff.cacheme.js?6a37935cf47f274bceff:6:5063)
   at t.prototype.handleClick (https://cocalc.com/static/0-6a37935cf47f274bceff.cacheme.js?6a37935cf47f274bceff:6:47668)
   at t.prototype.handleClick (https://cocalc.com/static/0-6a37935cf47f274bceff.cacheme.js?6a37935cf47f274bceff:6:9639)
   at Anonymous function (https://cocalc.com/static/0-6a37935cf47f274bceff.cacheme.js?6a37935cf47f274bceff:6:4181785)
   at f (https://cocalc.com/static/0-6a37935cf47f274bceff.cacheme.js?6a37935cf47f274bceff:6:4181695)
   at Anonymous function (https://cocalc.com/static/0-6a37935cf47f274bceff.cacheme.js?6a37935cf47f274bceff:6:4182458)
```

---

top frame track down

```
t.clear_all_outputs=function(){
var e=0;t.store.get("cells").forEach(function(n,r){ <<<<<<<<<<<<<<<<<<<<<<<<<<
[...]

PR #3604, opened 2019-02-18
webapp/jupyter2: undefined store in clear_all_outputs

haraldschilly:
# Description
No idea why this is happening, but it does: #3603 ... I wouldn't be surprised if there is something more fundamental to fix than just this simple test.

# Testing Steps
no idea

# Relevant Issues
* #3603 

Front end:

### DIFF
--- a/src/smc-webapp/jupyter/actions.ts
+++ b/src/smc-webapp/jupyter/actions.ts
@@ -489,6 +489,7 @@ export class JupyterActions extends Actions<JupyterStoreState> {
   };
 
   clear_all_outputs = (): void => {
+    if (this.store == null) return;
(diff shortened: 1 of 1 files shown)


williamstein:
Strangely, again your code definitely would not fix the reported problem.   It is only handling the possibility that `store` is null, which is not what is happening here.   The actual error is `Unable to get property 'forEach' of undefined or null reference`.  forEach is called on `store.get("cells")`, NOT on the store.  So there's absolutely no possibility that store is undefined in this stack trace, and it's 100% that store.get("cells") is not an immutable.map (it is probably undefined or null, maybe due to not being loaded yet, or being closed).

PR #3616, opened 2019-02-21
webapp build: cleanup all of "our" package-lock files

haraldschilly:
# Description
This removes all of "our" (in the cocalc codebase) package-lock files, upon running a full installation or cleaning up.  That way, we can eliminate certain side effects of  these files lying around after a previous build.

My motivation is my previous PR about updating pdfjs, because even after changing back to my master branch (with an older pdfjs), npm installed the newer version.

and well, also install.py, that's what kucalc uses

# Testing Steps
There isn't much to test, because the overall install behavior just getting closer to a "fresh" install

# Relevant Issues

Front end:

### DIFF
--- a/src/install.py
+++ b/src/install.py
@@ -96,7 +96,7 @@ def install_webapp(*args):
         cmd("git submodule update --init")
         cmd("cd examples && env OUTDIR=../webapp-lib/examples make")
         # clean up all package-lock files in cocalc's codebase (before running npm install again)
-        cmd("git ls-files '*/package-lock.json' | xargs rm -f")
+        cmd("git ls-files '../*/package-lock.json' | xargs rm -f")
         for path in [
                 '.', 'smc-util', 'smc-util-node', 'smc-webapp',
                 'smc-webapp/jupyter'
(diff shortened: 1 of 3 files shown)


williamstein:
Cool -- I'm glad you were able to track down what was causing confusion before.  What's the issue with pdfjs, btw?

williamstein:
There is a script in src/scripts to ignore all package-lock files, which I wrote.  This approach `git ls-files '*/package-lock.json' | xargs rm -f` would be way better than what I did.

haraldschilly:
There is no specific issue with pdfjs. It's just one of those packages which has a newer stable release, [list of fixes and improvements](https://github.com/mozilla/pdf.js/releases/tag/v2.0.943)

my only issue was after going back to master, to properly downgrade that library to where it was before

haraldschilly:
> There is a script in src/scripts to ignore all package-lock files ...

Yes, I'm ignoring them as well, and don't notice them. That removal bash code still deletes them, i.e. what I want to say is that masking doesn't impact that git file search.

PR #3632, opened 2019-02-26
webapp/project listing: toggle button to hide masked files

haraldschilly:
# Description
Hide masked files, and well, since hidden files were also masked, they have overlapping functionality. I decided to not mask hidden files -- they're by default just hidden, that's all.

# Testing Steps
1. toggle the buttons at the top right of the listing

## Default (nothing changed) -- there are many masked latex files

screenshot from 2019-02-26 16-52-28

## Showing hidden files (they aren't masked)

screenshot from 2019-02-26 16-52-36

## Hiding hidden files and also masked files (in particular, the latex artefacts are gone)

screenshot from 2019-02-26 16-52-50

Front end:


williamstein:
Is this done for a first pass?

Are you planning to add any sort of persistence, e.g., in localstorage?  Also a default option in account prefs?  In "Other Settings" in account prefs now, there is:
```
[ ] Mask files: grey-out files in the files viewer that you probably do not want to open
```
There could be another option:
```
[ ] Hide files: by default, completely hide files in the files viewer that you probably do not want to see, e.g., autogenerated latex file
```

haraldschilly:
I don't plan anything else. I'm just wondering about how to show hidden files. 

I'm also not aware of that masked files setting. I thought graying out is the default.

haraldschilly:
Well,  disabling it in those others setting  makes this button feel useless. It shouldn't be there.

williamstein:
Greying out is the default.   However it is configurable, as you now see.  Does anyone ever disable it?  I wonder.  If not maybe we should remove that option and replace it by *your* option...

haraldschilly:
I'll check the database.

haraldschilly:
~0.03% have that setting set to false, which would suggest to remove it.

williamstein:
> Well, disabling it in those others setting makes this button feel useless. It shouldn't be there.

Not to me.  That setting is just the DEFAULT.  I might still want to click the button to see all files in a given directory...?

haraldschilly:
> > Well, disabling it in those others setting makes this button feel useless. It shouldn't be there.
> 
> Not to me. That setting is just the DEFAULT. I might still want to click the button to see all files in a given directory...?

well, to get this to a point where we can finalize this, I don't understand your question.

1. masking files enabled (99.97%) of users: they see grayed files for latex build artefacts & co. ok.
1.1 clicking on the hide masked files button and those grayed files no longer show up (that's what this ticket should accomplish)
2. masking files disabled (0.03%), and there are no grayed files
2.1 clicking on the hide masked files button does nothing, because there are no masked files. that feels awkward to me, because without knowing about this setting, the button seems to be broken.

my latest commit here computes the file mask in case 2.1, such that files are indeed no longer visible.

but in general I'm for removing that option from the account settings.

PR #3642, opened 2019-02-28
Change initial page view to match history.coffee

johnjeng:
Send the pathname instead of the default full url.

If this doesn't fix it, I'll dig deeper.


williamstein:
@haraldschilly  already did this (in the exact same way with the same reservations!) before you https://github.com/sagemathinc/cocalc/pull/3636

PR #3654, opened 2019-03-04
doc: little extractor to get a serialized representation of the api doc

haraldschilly:
no need to review, this is just a helper to get the data from point A to B without having to depend on the entire repository


williamstein:
two comments before I merge:

1. You're not actually using any features at all of typescript, right?   Maybe just make it .js?
2. If we installed ts-node systemwide, you could just do ts-node file.ts, I think....

PR #3705, opened 2019-03-21
Performance

johnjeng:
# Description
Uses `react-virtualized` to make file listing much more performant. Try it with a 1000 files! 10000 files even!
https://youtu.be/-vSZnKwhu5A

Implements a few performance improvements mostly involving lists in project listing and file listing.

# Testing Steps
Testing file listing changes
1. Make a lot of files
1. Make a lot more files
1. Set your file per page to a big number
1. Test checking files
1. Shift check multiple files
1. Use file actions
1. Use up and down arrow keys to change the "selected file"
1. Use down arrow keys until it should scroll. Then watch it scroll with your selection.

Testing scroll save
1. Scroll down
1. Navigate away
1. Come back to the same spot
1. Use the arrow keys while your cursor is in the search box
1. Navigating away and back should result in the selection being visible still
1. If you use the arrow keys again but scroll away from the selection, revisiting the file listing should result in your scroll position, NOT your selection position.

Testing `FileNew` changes
1. Create files using the create button
1. Try to use the create button without anything in search
1. It should still work even though `shouldComponentUpdate` is always `false`

Testing `componentDidMount` change
1. Create a course
1. Set students must pay
1. Log onto the student account
1. Add a payment method
1. Log off
1. Without going to the billing page, go to the students file listing
1. Their payment method should be fetched and not have to reenter it.

# Relevant Issues

Front end:

### DIFF
--- a/src/smc-webapp/project_files.cjsx
+++ b/src/smc-webapp/project_files.cjsx
@@ -1524,6 +1524,10 @@ exports.ProjectFiles = rclass ({name}) ->
         shift_is_down : false
 
     componentDidMount: ->
+        # Update AFTER react draws everything
+        # Should probably be moved elsewhere
+        # Prevents cascading changes which impact responsiveness
+        # https://github.com/sagemathinc/cocalc/pull/3705#discussion_r268263750
         setTimeout(@props.redux.getActions('billing')?.update_customer, 200)
         $(window).on("keydown", @handle_files_key_down)
         $(window).on("keyup", @handle_files_key_up)
--- a/src/smc-webapp/project_files.cjsx
+++ b/src/smc-webapp/project_files.cjsx
@@ -1524,7 +1524,7 @@ exports.ProjectFiles = rclass ({name}) ->
         shift_is_down : false
 
     componentDidMount: ->
-        window.setTimeout(@props.redux.getActions('billing')?.update_customer, 200)
+        setTimeout(@props.redux.getActions('billing')?.update_customer, 200)
         $(window).on("keydown", @handle_files_key_down)
         $(window).on("keyup", @handle_files_key_up)
 
--- a/src/smc-webapp/project_files.cjsx
+++ b/src/smc-webapp/project_files.cjsx
@@ -1506,6 +1506,7 @@ exports.ProjectFiles = rclass ({name}) ->
             show_library          : rtypes.bool
             show_new              : rtypes.bool
             public_paths          : rtypes.immutable  # used only to trigger table init
+            file_listing_scroll_top: rtypes.number
 
     propTypes :
         project_id             : rtypes.string
@@ -1787,6 +1788,7 @@ exports.ProjectFiles = rclass ({name}) ->
                     library                = {@props.library}
                     redux                  = {@props.redux}
                     show_new               = {@props.show_new}
+                    last_scroll_top        = {@props.file_listing_scroll_top}
                 />
             </SMC_Dropwrapper>
         else
--- a/src/smc-webapp/project_files.cjsx
+++ b/src/smc-webapp/project_files.cjsx
@@ -1409,7 +1409,7 @@ ProjectFilesNew = rclass
     getDefaultProps: ->
         file_search : ''
 
-    // Rendering doesnt rely on props...
+    # Rendering doesnt rely on props...
     shouldComponentUpdate: ->
         false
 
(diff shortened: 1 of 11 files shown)


src/smc-webapp/project_files.cjsx
```diff
@@ -1520,7 +1524,7 @@ exports.ProjectFiles = rclass ({name}) ->
         shift_is_down : false
 
     componentDidMount: ->
-        @props.redux.getActions('billing')?.update_customer()
+        window.setTimeout(@props.redux.getActions('billing')?.update_customer, 200)
```
williamstein: COMMENTS About this line:  I always question setTimeout when I see it, and always feel happy when I can eliminate it.  Please at least explain why this is needed.   

Also, I usually prefer using `await delay(200)` if possible, where `{delay} = require('awaiting')`.  Generally, the code is more readable, and it also avoids changing `this` (whereas setTimeout does).

I also don't explicitly use `window.setTimeout` anywhere -- why put `window.` explicitly?   E.g., node also has setTimeout but no window.  By putting `window.setTimeout` in code you're needlessly making it so it can't run in node without defining window globally, right?

DrXyzzy:
I ran some tests at the link provided by @williamstein in today's chat.

1. Lots of files. This seemed slow to me, and feedback while typing into the filter text didn't appear until after file list was updated. In Account > Preferences, I set "Number of files per page" to 200 and generated 600 files beginning with 'aa', 'ab', and 'bb'. When I typed 'a' in the filter the first time, it was about 8 seconds for the filtered list of files to appear. After that, starting from an empty filter, I can type 'a', hit backspace to clear the filter, and type 'b', and in each case the new filtered list appears instantly. But if I type 'ab' or 'abcd' etc. into an empty filter, it takes 7 seconds or so for the filtered list to update, even if there are no matching file names. If there is a nonempty list and I click the "X" to clear the filter, it's 7 seconds to get back to the full list. For any filter that takes a long time, there is no echo of my changes to the filter string until the file list is updated. That makes it hard to see whether I made a typo or didn't hit the touchpad just right, etc. until after the wait.

2. Create a file. Works, with latency for large file list as above. I typed a non-matching string into the file search filter so that the "Create" button appears. Then I clicked either it, or one of the options below it offering other filename extensions. In each case, the file is created, but there is again a 7-10 second delay after the click until the screen changes. Regarding "Try to use the create button without anything in search."... without anything in search, I don't see the create button. Or am I misunderstanding the test?

3. Create a course. Works, but note that setting require students to pay checkbox is not enough to make \$14 option available. I created a student login, then as instructor created a course, added the student, and under course config checked "Students will pay for this course". Logging in as student, I could not see the $14 option until I went back to the instructor account in course config and checked "Require that students upgrade by 7 days from now". After that, I was able to log in as student, open the course project, and enter a payment method. I logged out and back in as student and was able to see that the payment method had been saved and did not need to be entered again.

src/smc-webapp/project_files.cjsx
johnjeng: Oh, no reason to use window. Changed to just `setTimeout`.

I didn't want to look too deeply in this case, but it's just to defer cascading changes that resulted from this action. It should probably be moved elsewhere but at least this way it doesn't affect file listing responsiveness as much.

Before it would do this after mount but before drawing anything
Screen Shot 2019-03-21 at 2 04 25 PM

Now it draws and does the update later.
Screen Shot 2019-03-21 at 2 09 05 PM

williamstein: That's a good explanation.  Can you add a comment saying basically that?

DrXyzzy:
I just tested in cc-in-cc with the patch branch. Wow, very fast. Negligible latency clearing nonmatching filter string for file list and create.

I did notice that existing project and files left over from previous run took a long time (a minute or two) to show up in the cc-in-cc instance. Once displayed, file list rendering is as fast as in a newly-created project. New project with 900 test files, set for 300 files per page, started in reasonable time.

johnjeng:
>I did notice that existing project and files left over from previous run took a long time (a minute or two) to show up in the cc-in-cc instance. Once displayed, file list rendering is as fast as in a newly-created project. New project with 900 test files, set for 300 files per page, started in reasonable time.

@DrXyzzy Can you give reproduction steps for what you're describing here?

williamstein:
>  did notice that existing project and files left over from previous run took a long time (a minute or two) to show up in the cc-in-cc instance. 

That sounds completely unrelated to @johnjeng's PR, and probably the project just starting up or the project websocket connecting.

Issue #2919, opened 2018-06-13
file creation: fix up the mess and revamp file creation.


williamstein:
We need to change our File creation system so there is only **one** way to make a file, and deferr when you have to name a thing. Naming things is hard!

> "Imagine if I had to name my puppy before I could even meet and buy her!"

What we need to do:

1. Replace the +New page entirely by a page called Library, and delete everything from there except the library.   (Later this could also have a new section which is a library of examples/templates of "Sage Worksheets", etc....)
2. Move the button in the files page currently labeled "Create" to the far left, rename it "+New", and ensure that it has options that are the most important file types in a group at the top, then a divide line, then *everything* in the current File menu of the *current* +New page.
3. If the user clicks the new +"New" button, which is ONLY available in the Files page, we generated a sequential filename such as "Untitled [n]".
4. We do the hard (but much needed anyways!) work to make file and directory renaming way easier. 

--- 

It would be nice if our new +New dropdown had 3 sections:

- the last five types you've used first, then a bar
- the popular well supported types (e.g., Jupyter, Sage, etc.),
- then a long list of everything else.

williamstein:
The above is what to do.  Here's some discussion about why...

## I watched a user try to make a task list, and here's what happened

> The different ways messed her up? 0.o

It was absurd. 

1. She made something through +New, then deleted it.
2. She clicked on the create button, selected "Task list", then got the error about having to put a filename in the "box above".
3. She put a name in the box above, which changes the main file listing to show that name too with a HUGE button to make a sage worksheet, so she clicked that HUGE button (which in her mind clearly knew she had just tried to make a task list).  She got a sage worksheet.
4. Confusion about what the hell a sage worksheet is.  I stepped in and explained, so she deleted it.
5. She went back to files and clicked "create", then "Task list", and got the error again.
6. She typed the filename, hit return, and got a sage worksheet.
7. Delete sage worksheet.
8. Literally same problems again and *another* damned sage worksheet.
9. Click on +New, saw that page, and I guess thought "oops" and muttered, "definitely not this", and went back to the file listing.
10. Maybe made yet another sage worksheet... and another?
11. Finally, I told her what to do and she got a task list.
12. I think she just saw all the buttons or whatever, and thought "no way I can use this task list", and gave up.
13. I said "click +New to make a new task", so she... clicked the "+New" above and was back in the create files (and library and upload) page.  
14. I got her back to tasks, but at this point is was too late.

williamstein:
[...]

PR #3713, opened 2019-03-26
ask explicitly about filename, after selecting the type

haraldschilly:
# Description
that's an implementation for my idea in #2919, in particular to overcome step 2&3 in the user story. basically instead of an error, show an explicit request for a filename in a panel. it's also less friction than before, because a timestamped-filename is inserted by default and return or one additional click is enough to get the file.

# Testing Steps
in "files" for each, do click "+new" dropdown → e.g. latex or jupyter
   1. press return → new file
   1. press esc → back to square one
   1. type some chars → uses that filename
   1. no filename: create not active, and return shows a specific error

# Relevant Issues
#2919

Front end:

### DIFF
--- a/src/smc-webapp/project/ask-filename.tsx
+++ b/src/smc-webapp/project/ask-filename.tsx
@@ -1,0 +1,106 @@
+import { Component, React, Rendered } from "../app-framework";
+const { default_filename } = require("../account.coffee");
+const { file_options } = require("../editor");
+const {
+  Col,
+  Row,
+  ButtonToolbar,
+  ControlLabel,
+  Button,
+  Form
+} = require("react-bootstrap");
+const { SearchInput } = require("../r_misc");
+const { IS_TOUCH } = require("../feature");
+
+interface Props {
+  actions: any;
+  ext_selection: string;
+  current_path: string;
+}
+interface State {
+  new_name: string;
+}
+
+export class AskNewFilename extends Component<Props, State> {
+  displayName = "ProjectFiles-AskNewFilename";
+  constructor(props) {
+    super(props);
+    this.state = {
+      new_name: default_filename()
+    };
+  }
+
+  componentDidMount() {
+    this.setState({ new_name: default_filename() });
+  }
+
+  cancel = (): void => {
+    this.props.actions.setState({ ext_selection: undefined });
+  };
+
+  create = (name, focus): void => {
+    this.props.actions.setState({ ext_selection: undefined });
+    this.props.actions.create_file({
+      name: name,
+      ext: this.props.ext_selection,
+      current_path: this.props.current_path,
+      switch_over: focus
+    });
+  };
+
+  submit = (val: string, opts: any): void => {
+    this.create(val, !opts.ctrl_down);
+  };
+
+  create_click = (): void => {
+    this.create(this.state.new_name, true);
+  };
+
+  change = (val: string): void => {
+    this.setState({ new_name: val });
+  };
+
+  render_filename() {
+    const data: any = file_options(`foo.${this.props.ext_selection}`);
+    return data.name;
+  }
+
+  render(): Rendered {
+    return (
+      <Row>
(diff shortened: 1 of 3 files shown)


src/smc-webapp/project_files.cjsx
```diff
@@ -1216,6 +1216,85 @@ ProjectFilesActionBox = rclass
...
+        new_name : account.default_filename()
+
+    componentDidMount: ->
+        @setState(new_name : account.default_filename())
+
+    new_filename_cancel: ->
+        @props.actions.setState(filename_selection:undefined)
+
+    new_filename_create: (name, focus) ->
+        console.log("create file #{name}.#{@props.filename_selection}")
```
williamstein: delete or comment out

src/smc-webapp/project/ask-filename.tsx
```diff
@@ -0,0 +1,106 @@
...
+  actions: any;
+  ext_selection: string;
+  current_path: string;
+}
+interface State {
+  new_name: string;
+}
+
+export class AskNewFilename extends Component<Props, State> {
+  displayName = "ProjectFiles-AskNewFilename";
```
johnjeng: displayName isn't necessary in normal react Classes.

src/smc-webapp/project/ask-filename.tsx
```diff
@@ -0,0 +1,106 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      new_name: default_filename()
+    };
+  }
+
+  componentDidMount() {
+    this.setState({ new_name: default_filename() });
```
johnjeng: Why is this called in the constructor and in `didMount`?

src/smc-webapp/project/ask-filename.tsx
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  cancel = (): void => {
+    this.props.actions.setState({ ext_selection: undefined });
```
johnjeng: Make a new action instead of directly setting state.

src/smc-webapp/project/ask-filename.tsx
```diff
@@ -0,0 +1,106 @@
+import { Component, React, Rendered } from "../app-framework";
```
johnjeng: I'm moving to importing `React` from "react" using React.Component. Also the `Rendered` type is usually unnecessary since TS can infer it.

williamstein: > Also the Rendered type is usually unnecessary since TS can infer it.

For me the point of putting types in explicitly is it ensures that I don't screw up the code. I explicitly put return types for my functions because it ensures that I write them properly (without implicitly relying on how they are used), and also makes my intention very clear to readers.   I find this extremely useful.  It immediately ensures that all return paths return a rendered component (or undefined), etc.   I am going to use Rendered explicitly in all my code, unless you can give me a really good argument to not do so.   I consider this similar to using strict null checks, which Typescript of course does not have on by default, but prevents a lot of issues if you do it from the beginning.

src/smc-webapp/project/ask-filename.tsx
williamstein: It should probably only be called in componentDidMount -- it needs to get updated since it is time based.

src/smc-webapp/project/ask-filename.tsx
williamstein: > Make a new action instead of directly setting state.

Based on my experience writing a lot of code at this point, I definitely don't necessarily agree...  I've overall found it much better to just have a good definition of the data in the store, and found putting lots of actions in as wrappers around this to be not necessary, annoying, dangerous (since they can do other things and change, so you're never really sure what state is changing).   I ended up using setState directly a lot more in the newer frame-tree code I wrote, and was way happier with it.     I'm just speaking from experience, and am surprised.

williamstein:
I'm going to merge this despite all the stylistic commentary above, since it *does* work fine and the actual user experience is REALLY good and a major improvement.

PR #3729, opened 2019-04-02
webapp: some enhancements for new filenames

haraldschilly:
# Description
This includes some changes to the dialog asking for a filename, i.e. the +new dropdown **without** entering any filename in the search box. It would be consequential to also add this to the larger "+new" dialog, but that's for later. 

# Testing Steps
1. dropdown select a filename, random filename appears
2. when you select python or sage, there are underlines in the name
3. there is an account setting for the type of random filename, which is also exposed in a dropdown.

# Relevant Issues

Front end:


williamstein:
WOW

johnjeng:
I get this error when trying to click on a file type

Screen Shot 2019-04-03 at 4 33 19 PM

Can you change it to start with `Untitled-`? It felt really odd having that filled in with just `slick-shake`. 
Screen Shot 2019-04-03 at 4 31 48 PM

williamstein:
> Can you change it to start with Untitled-? It felt really odd having that filled in with just slick-shake.

Please make it start by default with the current iso timestamp (so users will see little change initially, except for python files will be valid, say). Then *we* can use this for a few weeks with other name choice options, let the problems shake out, and only then change the default.  There's really no need to make everybody switch immediately.

haraldschilly:
well, I've worked on this a bit more, and I'm reasonable happy with it. The error above happens when the hub code is the old one (and no default is set). I've taken care of this, which means there is always a default even when the hub isn't updated. 

The default is `"iso"` which is the same as right now. Hence when you do not change the filename generation type, you wouldn't notice it. For testing, it's also useful to clear your settings, i.e. in `psql`: 

    update accounts set other_settings = NULL where account_id = '...';

haraldschilly:
well, I've worked on this a bit more, and I'm reasonable happy with it. The error above happens when the hub code is the old one (and no default is set). I've taken care of this, which means there is always a default even when the hub isn't updated. 

The default is `"iso"` which is the same as right now. Hence when you do not change the filename generation type, you wouldn't notice it. For testing, it's also useful to clear your settings, i.e. in `psql`: 

    update accounts set other_settings = NULL where account_id = '...';

haraldschilly:
well, I've worked on this a bit more, and I'm reasonable happy with it. The error above happens when the hub code is the old one (and no default is set). I've taken care of this, which means there is always a default even when the hub isn't updated. 

The default is `"iso"` which is the same as right now. Hence when you do not change the filename generation type, you wouldn't notice it. For testing, it's also useful to clear your settings, i.e. in `psql`: 

    update accounts set other_settings = NULL where account_id = '...';

haraldschilly:
well, I've worked on this a bit more, and I'm reasonable happy with it. The error above happens when the hub code is the old one (and no default is set). I've taken care of this, which means there is always a default even when the hub isn't updated. 

The default is `"iso"` which is the same as right now. Hence when you do not change the filename generation type, you wouldn't notice it. For testing, it's also useful to clear your settings, i.e. in `psql`: 

    update accounts set other_settings = NULL where account_id = '...';

williamstein:
- This shouldn't be called "Random filenames" in account settings.  This is not about random files, it's about what the filename of a *new file* is.  It should be called "New filenames". There are many options.  One, which we have now, is the current time (which is not random).  Another is what major OS's do, e.g., "Untitled 1", "New File 2" (or something like that).

- Change the dice icon to a regenerate or something, since again it isn't necessarily random (time is not random). 

- I don't think I should be seeing this:

image

namely, that the generated filename is different than the algorithm for generating filenames.  This happens if you change the algorithm in account settings, then switch back to the file listing.  I think you need to use reduxProps to listen on the account store...

williamstein:
ping

haraldschilly:
yes, I know it's still open, I'll fix this during the week

haraldschilly:
I think I addressed all 3 points. I hope I didn't introduce a bug by renaming everything. I also don't know why suddenly travis fails. They must have made an update... (something for another branch)

PR #3751, opened 2019-04-12
first attempt at noImplicitAny

williamstein:
; involved tons of changes to app-framework.ts and surely this is going to be really hard.


williamstein:
NOTE: this doesn't work yet!

PR #3768, opened 2019-04-23
enhance user-facing error msg when there is a sign in issue

haraldschilly:
# Description
Since this comes up in support at least once a week, I've enhanced the error message if there is a connection issue. It points to the connection debugging page and suggests to refresh the tab. There is also some style tweaking, because those shorter messages would be squeezed to the far right with line breaks.

# Testing Steps
1. To test this, I've made my email address wrong (by adding  `+'x'`) in the redux account `webapp_client.sign_in` call, and explicitly set `error = 'no conn...'` in the callback. Below are two screenshots.

Screenshot from 2019-04-23 11-52-31

Screenshot from 2019-04-23 11-58-00

# Relevant Issues

Front end:

### DIFF
--- a/src/smc-webapp/landing_page.cjsx
+++ b/src/smc-webapp/landing_page.cjsx
@@ -24,7 +24,7 @@ The Landing Page
 ###
 {rclass, React, ReactDOM, redux, rtypes} = require('./app-framework')
 {Alert, Button, ButtonToolbar, Col, Modal, Grid, Row, FormControl, FormGroup, Well, ClearFix, Checkbox} = require('react-bootstrap')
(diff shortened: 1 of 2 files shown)


williamstein:
I merged this after fixing a grammar issue.  That said, this probably usually happens when the user clicks the "Sign in" button, upon entering details, but their websocket connection has not been established.  In such a situation, we already know that sign in is going to fail, and we know why... so why even try?  We could do something better, make the sign in (or create account) buttons be disabled, and have a message right below them that says "Establishing a connection to the server...".  

For now this quick fix in your PR is clearly the best thing to do.

PR #3769, opened 2019-04-23
webapp/files: avoid unnecessary renderings

haraldschilly:
# Description
upon pressing a key in chat somewhere, a file listing rendering is triggered in the background of another project, which is showing a file listing. this patch avoids this, but it is just a defensive measurement. maybe there is something smarter to do?

# Testing Steps
1. add a debug line to the `render_rows` method to see when it is active. it shouldn't be drawing if you type in chat or somewhere else. also, I deliberately excluded this `shift key` check, because otherwise it would render if I press the shift key anywhere.

# Relevant Issues

Front end:

### DIFF
--- a/src/smc-webapp/project/file-listing/file-listing.tsx
+++ b/src/smc-webapp/project/file-listing/file-listing.tsx
@@ -7,7 +7,7 @@ import {
   CellMeasurerCache
 } from "react-virtualized";
 
-import { debounce } from "lodash";
+import { debounce, isEqual } from "lodash";
 
 const misc = require("smc-util/misc");
 const { Col, Row } = require("react-bootstrap");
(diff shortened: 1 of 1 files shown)


src/smc-webapp/project/file-listing/file-listing.tsx
```diff
@@ -69,9 +69,32 @@ export class FileListing extends React.Component<Props> {
     this.list_ref = React.createRef();
   }
 
+  shouldComponentUpdate(next) {
+    return (
+      !isEqual(this.props.listing, next.listing) ||
```
williamstein: I think props.listing could easily have 500 entries in it. If this is getting called on every single keystroke in a different, then project even this test will waste significant CPU.  We should fix this problem right instead, e.g., by only updating the FileListing if the actual listing changes, which can surely be controlled at a higher level or by passing in a counter instead.

haraldschilly:
who is working on this? it drives me crazy that each time I refresh cocalc I have to go through all open projects and make sure to tab away from the file listing to avoid this lag when typing...

williamstein:
> who is working on this?

OK, I'll fix this properly today.

https://github.com/sagemathinc/cocalc/issues/3799

PR #3777, opened 2019-04-24
NPM package update april 2019

haraldschilly:
# Description
* this contains a fix for invoking LRU cache (needs "new")
* couple of other necessary updates, mostly in those lock files

# Testing Steps
since this is touching a little bit all over the place, one should be careful. on the other hand there aren't that many changes. to test this:
1. `npm run clean` to get rid of your modules
2. `npm run restore-package-lock` to align yours with the ones in the patch
3. `npm run make`

Also, this should be done for the updated node js version 10, which is what it is used in production.

~~I guess the best way to test this is to deploy it in the test or dev namespace~~

**edit** ok, that's what I did. in the test namespace, I first built static, checked it builds & loads fine, then all the hubs, ok, and then the project, also ok. **Each build was one from scratch without caching!**

# Relevant Issues

Front end:


williamstein:
I'm really glad you're doing this!

PR #3788, opened 2019-04-26
python3 print

fchapoton:
# Description

# Testing Steps
1.
1.
1.

# Relevant Issues

Front end:

### DIFF
--- a/src/scripts/sshvm
+++ b/src/scripts/sshvm
@@ -10,7 +10,7 @@ if '@' in n:
 else:
     user = ''
     host = n
-_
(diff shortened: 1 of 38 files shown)


src/scripts/sshvm
```diff
@@ -9,11 +10,11 @@ if '@' in n:
 else:
     user = ''
     host = n
-
+_
```
haraldschilly: I'm reading through the code. I think the underline above is ok, but it's not useful. Or do I miss something?

PR #3793, opened 2019-05-01
framework for cocalc internal user tracking to know where user's come…

williamstein:
You can try something like:
```
https://cocalc.com/[dev project id]/port/[dev port]/app?utm_source=google&utm_campaign=test2
```

and hit the sign in page.  Then even answer the "where did you find out about cocalc" question.  And finally hit a single sign on button... and when you go away and later come back, entries are put in the user_tracking table in.

This is meant to be a foundation that 100% works, unlike https://github.com/sagemathinc/cocalc/pull/2446


haraldschilly:
I looked at the code, and I have a few comments and existing details to share. 

* #2446 tracks all landing pages, and this here only the loading page.   The code that does this right now is here: https://github.com/sagemathinc/cocalc/blob/master/src/webapp-lib/_inc_analytics.pug#L50 ... it reads the information from the url path and the referer, stores it temporarily in a cookie, and that cookie is consumed when a user signs in. that works quite well I think, and we can switch this to use local storage if you like. The missing aspect for #2446 is to extend the utm and referrer information also on the blog and (since recently) also the doc pages.
* AFAIK, the possible UTM fields are 'source', 'medium', 'campaign', 'term', and 'content'. How does this map to "value" ?
* I don't see in the code where the referer is stored. Without that this isn't even half of the fun :-) I suggest to have a look at our real world data right now, e.g. run

      select * from central_log
      where event = 'successful_sign_in' AND value ->> 'referrer' IS NOT NULL order by time desc limit 10;

or also interesting

     select * from central_log
     where event = 'create_account' AND value ->> 'referrer' IS NOT NULL order by time desc limit 10;

or even summaries like

     SELECT count(*) as cnt, value ->> 'referrer' as ref from central_log
     where event = 'successful_sign_in' AND value ->> 'referrer' IS NOT NULL
     AND time >= NOW() - '1 month'::interval
     GROUP BY ref
     ORDER BY cnt DESC LIMIT 100;

* you can also look at currently recorded UTM data

       select * from central_log
       where event = 'create_account' AND value ->> 'utm' IS NOT NULL order by time desc limit 10;

  or look at a stats-plot for currently recorded source/medium pairs per day during the last 3 month.  → [non-public link to plot](https://cocalc.com/369491f1-9b8a-431c-8cd0-150dd15f7b11/raw/scratch/Screenshot%20from%202019-05-01%2012-03-24.png)

all I want to point out with the stats above is, that this branch should at least allow to do the same sort of analytics.

haraldschilly:
ok ,I am able to answer one of my questions, the utm values are stored as a json string. wouldn't it be better to use a jsonb? 

for local storage, there is also [this wrapper](https://github.com/sagemathinc/cocalc/blob/master/src/smc-webapp/misc/local-storage.ts)

williamstein:
Thanks.  I am removing the utm code from this code, changing the data type from string to map, and adding more code related to event tracking.  This was really supposed to be about events (like GA but not anonymous) rather than about utm stuff.

PR #3828, opened 2019-05-15
Block URLs in names

haraldschilly:
# Description
This is admittedly a bit simple, but I think it is sufficient. This
* adds a test for a valid username, with a test
* adds this as a precondition for creating an account
* fixes error reporting in the sign up form (otherwise, the screenshot below wouldn't have a red error message)

Screenshot from 2019-05-15 15-39-18

# Testing Steps
1. creating an account with a normal first and last name still works
1. trying to create an account where the first or last name contains `https://` or so throws an error

# Relevant Issues

Front end:

### DIFF
--- a/src/smc-hub/client/create-account.ts
+++ b/src/smc-hub/client/create-account.ts
@@ -252,7 +252,7 @@ export async function create_account(
       if (reason == null) {
         // generic reason
         reason = {
-          other: `Unable to create account.  Please try later.  ${err}`
+          other: `Unable to create account. Error: "${err}"`
         };
       }
       opts.client.push_to_client(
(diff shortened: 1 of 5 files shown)


williamstein:
Change it to say why the username is not valid.  This is very frustrating and annoying to users.  Imagine if our password checker said "password is not valid" instead of `Password must be between 6 and 64 characters in length.`.

In particular make is_valid_username return a string if the username is not valid, where the string explains why (and maybe also say what we do allow).

williamstein:
Isn't it trivial to defeat this by using a capital letter, e.g., `William Http://wstein.org`

williamstein:
I merged this anyways, but know that it is trivial to get around by using a capital letter.

williamstein:
@haraldschilly  -- 

See how the first one gets through:

image

Also, the error should appear next to the name, like this:

image

Issue #3832, opened 2019-05-15
prettier (or whatever) code formatting for C code never works


williamstein:
1. Make a .c file with something valid in it.
2. Click the '"Format" button. 
3. It always give the same error, and doesn't format. 

See:

![image](https://user-images.githubusercontent.com/1276278/57797416-418cdf00-76ff-11e9-997e-1a34d969ad41.png)

PR #3833, opened 2019-05-15
project/formatter: node 10 fallout regarding fs.unlink

haraldschilly:
# Description
The `fs.unlink` api changed in node 10. Instead of using unlinkSync, I just added a pointless function. That's all. This patch also adds some missing temp file cleanup calls!

# Testing Steps
1. well, all those formatters which were changed should work now

# Relevant Issues
#3832

Front end:


williamstein:
I will merge this, but I don't really agree with what you did.  They changed the api to make it much harder for people to silently ignore errors, which is always a bad thing.  It would have at least been better to make a **wrapper function** around unlink that takes only a file as input, and which reports any error to the log, rather than silently ignoring errors.   

But life is short, so whatever.

PR #3888, opened 2019-06-11
multi_response query not supported by api

DrXyzzy:
# Description

No change to executable code.

Adds 1 line to API docs pointing out that multi_response API query is not supported, as per [src/smc-hub/api/handler.coffee](https://github.com/sagemathinc/cocalc/blob/3f4b15e6e7b9588f9297f493861352e6ead3cf8f/src/smc-hub/api/handler.coffee#L56)

# Testing Steps
1.
1.
1.

# Relevant Issues

Front end:


williamstein:
Maybe also say *why* they are not supported?  The reason is because it makes no sense to support them, since such queries require a longterm persistent connection, and the API simply does not provide such a thing.

Issue #3893, opened 2019-06-13
ts file → shell button → crash


haraldschilly:
I've a `.ts` file with the simple content:

```
if (true) {
  for (let i in [1, 2, 3]) {
    console.log(`${i}`);
  }
}
```

When I click the shell button, webapp crashes hard

```
smc-d90f0523548dfb4cd37b.cacheme.js:5 TypeError: Cannot read property 'join' of undefined
    at t.render_command (smc-d90f0523548dfb4cd37b.cacheme.js:5)
    at t.render (smc-d90f0523548dfb4cd37b.cacheme.js:5)
    at Eo (0-d90f0523548dfb4cd37b.cacheme.js:6)
    at Ao (0-d90f0523548dfb4cd37b.cacheme.js:6)
    at Io (0-d90f0523548dfb4cd37b.cacheme.js:6)
    at qa (0-d90f0523548dfb4cd37b.cacheme.js:6)
    at Wa (0-d90f0523548dfb4cd37b.cacheme.js:6)
```

PR #3895, opened 2019-06-13
webapp/terminal: desc.get('args') could exist but is still undefined …

haraldschilly:
# Description
simple fix for #3893

# Testing Steps
1. see ticket

# Relevant Issues

Front end:


williamstein:
Well that is annoying!  Thanks for the better fix and pattern.   I should be better at typing stuff like this (@johnjeng is way, way better at it).

PR #3918, opened 2019-06-25
Refactor all backend stripe api code out of smc-hub/client.coffee; rewrite in Typescript with async/await

williamstein:
Some initial testing suggests this works.  I need to test it more carefully from scratch with a new user,  **test coupons**, test the api, and test various edge cases.  This code is motivated by https://github.com/sagemathinc/cocalc/pull/3917 -- namely how hearing about and seeing this old code I wrote disgusted me (!).   Also, this rewrite probably fixes many subtle bugs that we might not have noticed, since it's just far easier to write async/await typescript code that is correct, than long callback-hell chains in scope-hell coffeescript.

### DIFF
--- a/src/smc-hub/client.coffee
+++ b/src/smc-hub/client.coffee
@@ -14,7 +14,6 @@ misc                 = require('smc-util/misc')
 {defaults, required, to_safe_str} = misc
 {JSON_CHANNEL}       = require('smc-util/client')
 message              = require('smc-util/message')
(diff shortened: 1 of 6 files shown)


williamstein:
@haraldschilly -- I plan to test this and merge it tomorrow.  However, you might have a look in the meantime, since this is entirely motivated by your complaints in chat, etc., about my ancient horrible code.

Issue #3904, opened 2019-06-17
+new dialog overlaps with file controls


haraldschilly:
well, this is something I noticed today, it must be rather new. before this bug here, it inserted into a row below the buttons.

![Screenshot from 2019-06-17 13-52-13](https://user-images.githubusercontent.com/207405/59602563-4c280300-9107-11e9-8f47-aafa0f49262f.png)

haraldschilly:
this is not easily reproducible. maybe this was just a fluke in my browser?

haraldschilly:
fixed

PR #3936, opened 2019-07-05
files listing: split flex divs to avoid #3904

haraldschilly:
# Description
this is a way how #3904 is fixed for me, tested in chrome and firefox. from trial/error I discovered, that the 100% height style in conjunction with the vertical flex directives are causing the problem. my idea is to split this up into two divs, where the bottom one expands and holds the files.

I marked this as careful, because if this fix works for me it doesn't mean I'm breaking it for others.

# Testing Steps
I tested this for chrome 76.0.3809.36 and Firefox  67.0.3 
1. +new button shows the dialog
1. library shows up
1. uploading a file creates that area showing the upload process
1. file listing shows up, and if there are more files than your page limit, you have these buttons at the bottom

# Screenshots

## Before

Screenshot from 2019-07-05 16-14-13

## With this patch

Screenshot from 2019-07-05 16-14-32

## With this patch, default view (as a reference)

Screenshot from 2019-07-05 16-14-50

Front end:

### DIFF
--- a/src/smc-webapp/project_files.cjsx
+++ b/src/smc-webapp/project_files.cjsx
@@ -60,6 +60,7 @@ ROW_INFO_STYLE = Object.freeze
     height     : '22px'
     margin     : '5px 3px'
 
+
 {FileListing, TERM_MODE_CHAR} = require("./project/file-listing")
 
 feature = require('./feature')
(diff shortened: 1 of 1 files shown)


williamstein:
I've reverted this.

PR #3964, opened 2019-07-23
Draft for startup actions

haraldschilly:
# Description
As discussed yesterday, this is a draft for specific actions happening on startup. This is far from finished, but I think this is a first step.

What it does.

There are two phases: first it reads the parameter and stores it in local storage. After signing in, checks local storage and dispatches a function. In detail:

1. it reads the `launch` URL query parameter, parses it, and dispatches actions depending on it. There are three types: `share` for a specific share ID, `csi` for custom software image (if it knows the ID, it creates and opens a new project), and `binder`, which is not implemented at all. For now, let's focus on the share part.

2. I wrote a bit of code to save the information in a redux store, such that it can be surfaced on the sign up/in page. Below is the POC screenshot for an URL like `port/[ ... ]/app?launch=share/1234-uuid-xyz` when you're not signed in.

Screenshot from 2019-07-23 15-05-36

3. After signing in (i.e. that sign in hook happens), it dispatches the action. In the case above, the console log ends up noting: 

`launching share with ID = 1234-uuid-xyz`

Finally, regarding launching a share based on a custom software image, I think this shouldn't be part of this scheme. Rather, there should be a new field `compute_image` in `public_paths`, which defaults to `DEFAULT_COMPUTE_IMAGE` (i.e. "default"). My thoughts are that a share is tied to a specific image, and hence it makes sense to record this in its table at the point in time when it is create and updated. Beyond that I would defer any combinations like that until later.

# Testing Steps
1. what's actually worth to test is signin in as usual, because this touches the landing page!

# Relevant Issues

Front end:


src/smc-webapp/landing-actions.ts
```diff
@@ -0,0 +1,212 @@
+/*
+ * Landing actions are certain "intentions" for specific actions,
+ * which are triggerd upon starting the webapp.
```
williamstein: "triggerd" is a typo.  Also, what's with the *'s at the beginning of each line?

PR #3971, opened 2019-07-29
doc index entry for api user_auth

DrXyzzy:
# Description

Add an index entry for sphinx docs to message.js. This is a test with one change before adding others.
- only change is inside a string parameter
- cc-in-cc builds & runs ok (projects start, ipynb widgets run)
- test build of docs shows the two new index entries

# Testing Steps
1.
1.
1.

# Relevant Issues

Front end:

### DIFF
--- a/src/smc-util/message.js
+++ b/src/smc-util/message.js
@@ -2786,6 +2786,7 @@ API(
       }
     },
     desc: `\
+.. index:: pair: Token; Authentication
 Example:
 
 Obtain a temporary authentication token for an account, which


haraldschilly:
Hmm, did you try this by just editing the exported `api.json` file? My bet is this won't work, because this is treated as imported markdown and only processed for formatting. I'm also wondering if you need an additional newline to separate `.. [command]` from content.

haraldschilly:
ok, nope, I edited the json file manually, and it does work. good to know! (maybe when I tried it first it didn't work, but after some upgrades it is fine now.)

Issue #3453, opened 2019-01-11
changing account email in CoCalc does not update email address in Stripe


DrXyzzy:
to replicate:
1. create test account using email address
1. sync with stripe using admin tab
1. verify CoCalc account id's match in stripe and admin tab
1. change account email address using Cocalc `Account` > `Preferences`
1. sign out and back in with new email
1. verify new email address matches CoCalc account id in admin tab
1. sync with stripe again
1. search by CoCalc account id in stripe
1. observe that old address is still shown in stripe customer record under both "Account Information:Email" and "Billing Information:Send emails to"

williamstein:
This doesn't surprise me, since we never implemented anything to do this.  They unique identifier between stripe and cocalc is the customer id, which we also have in our database (and also the  account_id metadata, but that's not used), not the email address.  Many users don't even have email addresses (e.g., maybe they use single sign on and don't provide one), or they have multiple addresses (one for github, one in cocalc, etc.).  

So to me this isn't a bug, but I can see how it could be confusing to you.   We could have a feature to add a link to the stripe page for a user to the admin page in cocalc, since the stripe url is just `https://dashboard.stripe.com/customers/customer_id`.

williamstein:
Fixed, then made no longer relevant...

PR #3972, opened 2019-07-29
hub/stripe: propagate email changes from our DB to stripe 

haraldschilly:
# Description
* Updates to a user's email address should propagate to stripe. See #3453
* `sync_stripe`
    - syncs the name to the `name` field in stripe (sounds like a good idea), if name or description changed. it also updates the description field.
    - of course, it also syncs over the email address
    - I saw the comment about rate limiting, that's why I added a short 10ms delay
* while testing and fiddling around with stripe directly, I ended up with exceptions about missing data on  fields. That's why I also added some purely defensive code to `billing.cjsx`.

# Testing Steps
everything I tested is for cc-in-cc against my own stripe API test token

1. running `hub --stripe_sync`
2. changing email address, checking stripe
3. changing name, running `stripe_sync`, checking if name changed
4. new user, creating stripe account, subscribing to standard
5. finally, also check if there is a valid entry in *our* DB, in the `stripe_customer` column

# Relevant Issues
#3453

Front end:

### DIFF
--- a/src/smc-hub/postgres-server-queries.coffee
+++ b/src/smc-hub/postgres-server-queries.coffee
@@ -782,8 +782,11 @@ exports.extend_PostgreSQL = (ext) -> class PostgreSQL extends ext
             stripe      : undefined  # api connection to stripe
             customer_id : undefined  # will be looked up if not known
             cb          : undefined
-        customer = undefined
-        email_address = undefined
+        locals =
+            customer : undefined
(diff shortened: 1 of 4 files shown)


williamstein:
**RANT:** It would be vastly better to rewrite these functions in typescript (in another file) then add new code than adding a bunch of new coffeescript.  I don't trust coffeescript at all.

PR #3985, opened 2019-08-01
webapp/kiosk: remotely control an embedded cocalc instance

haraldschilly:
# Description

just a POC, we'll discuss this later. auxiliary files are here: https://github.com/haraldschilly/cocalc-embedding-test/ (hosted on my personal domain)

* extra work:
   - [x] merge conflict
   - [x] opening a file with a direct path no longer works

# Testing Steps
* probably the most important aspect is that this doesn't impact normal usage
* it typescripts these server settings, hence they should still work
* opening a link for fullscreen=kiosk mode should open that file and nothing else (can be tested by shift clicking on a file tab)
* fullscreen kiosk mode has a couple of modifications.
  *  compared to what's in master, this adds a bit of logic to deal with switching projects by merging what's in the store, etc. I think I got it, and from my tests it all works, but this definitely needs a second pair of eyes. i.e. projects.cjsx → projects table → _change.
  * there is also a `load_single_project` function, which goes a step further from what's there before. in turn, `load_recent_projects()` is no longer called at all.
  * no session management, no point in that
  * the app_shared.cjsx logic for the active app content is also more restrictive. i.e. you can only see a file or a cocalc banner (no file listing, etc.)
* the actual meat is this iframe-communication file. it basically receives and checks json objects from a parent page. then it dispatches commands. the security aspect is that it only listens to https and a configured list of domains. that's explained in that new server setting config. there is also a test file for that. so, for now I'm for being restrictive. once we know this is really safe, we can open this up to anyone, I guess.
* I've added you to two of my test projects in "test". so, as long as you're signed in, you can open [this page](https://dev.harald.schil.ly/cocalc-embedding-test/embedding-test.html) and click the buttons. e.g.
  *  you can try right after refreshing the page to quickly click on "status". first, no response, then it tells you it is "connecting" and then "connected".
  * another neat thing to try is clicking "halt" in the jupyter notebook

… and I noticed there is a new test failure after merging with master. So, that's something i'll take care of tomorrow.

2019-08-01-cross-domain-comm

Front end:


haraldschilly:
Update: in the screenshot below, the parent website has an additional button called `"status"` to query what's going on inside of cocalc. The repsonse is (for now) just some static object, so, that's not useful -- but my idea is to report back which projects are opened, files opened/loading, etc. Therefore, it should be possible to come up with a full control loop, consisting of a limited set of commands (`open` a file, `close` a file, ...) and query the result.

Below, the js-console shows the output from cocalc, and the status button is at the top. Next to it, it renders the json representation of the object coming back from cocalc.

Screenshot from 2019-08-02 19-47-41

williamstein:
Hmmm, this reminds me of how we have the "smc" (and redux) objects avialable in cc-in-cc dev mode, but NOT in production, in order to make cocalc more secure and much less amenable to cross site scripting attacks.    

If we had a query param for loading cocalc that made that smc object available again, e.g., for users who feel they don't need the extra xss protection, then it would be very easy to script/control anything in the cocalc iframe from outside, right?  A page that contains an iframe can run any javascript commands inside that iframe -- at least that's how my wrapping of classical jupyter with sync works.  This would make it easy to close/open projects, files, or do absolutely anything else.

I don't understand what you're doing though, so I don't know if it is the above or something else.

williamstein:
Testing steps?  What have you done to test this?  What are the concerns?  Security issues?

williamstein:
**The blocker for this is that you need to provide reviewer steps, etc.**

haraldschilly:
yes, I already started to edit it to come up with some points  :-)

williamstein:
OK -- good -- I just wanted to make sure it was clear why I wasn't reviewing/merging this.

williamstein:
I can say that skimming over the **code**, it looks good.

haraldschilly:
I updated the description with some points, but I noticed there is a new test failure after merging with master. So, that's something i'll take care of tomorrow.

williamstein:
I tried https://dev.harald.schil.ly/cocalc-embedding-test/embedding-test.html but none of the buttons at the top did anything.  I'm not sure what I'm supposed to do or expect.

williamstein:
When I did [this](https://github.com/sagemathinc/cocalc/issues/4145) I noticed that I could send/receive events from the page itself, and didn't have to use an iframe.  Is there any similar mode for developing or testing this?

src/smc-hub/hub.coffee
```diff
@@ -506,6 +498,14 @@ exports.start_server = start_server = (cb) ->
         (cb) ->
             # This must happen *AFTER* update_schema above.
             init_smc_version(database, cb)
+        (cb) ->
```
williamstein: Why?  This change has nothing to do with kiosk mode, etc., so why did you make it? Is it fixing a bug that you just happened to notice while doing something else?

src/smc-webapp/projects.cjsx
```diff
@@ -981,8 +998,27 @@ load_recent_projects = =>
...
-load_recent_projects()
+
+if not COCALC_MINIMAL
+    load_recent_projects()
+
+
+_project_tables = {}
+_previous_project_id = undefined
+
+load_single_project = (project_id) =>
```
williamstein: Document what this function does.  I don't know.  It's name is "load_single_project" and in your notes you talk about it "merging what's in the store".  But the code you have here is just deleting the Table and making a new one for a different project.   It seems to me that this will cause subtle problems.  E.g., what if the original project stops?  In the store you'll think it is still running, since you just deleted the Table that would tell the store that the project stopped.

It seems to me that was is needed is a way to **actually** add a single project by having a way to make a changefeed to track what happens with a given list of projects.  That doesn't exist yet, and you didn't write it, and it's maybe not easy.  But I don't see how you can possibly solve this problem without doing that... unless you explicitly and purposely choose to make the information in the projects table be wrong.  Maybe that is OK for some use case.

haraldschilly:
> I tried https://dev.harald.schil.ly/cocalc-embedding-test/embedding-test.html but none of the buttons at the top did anything.

It's pointing at the setup for the test namespace. Did you build the static pages and the hub? There ar e recent images to run as well.

src/smc-webapp/projects.cjsx
haraldschilly: Well, unfortunately I don't know anything about this. No, my goal is not to make the table wrong. My only goal was to be able to switch between projects.

haraldschilly:
> Have you run stress/loading testing against this? My biggest worry is that something in here breaks minerva's use of cocalc, and they get pissed... At least let's run that once after merging this.

yes, good idea. The usual use case of pointing to a file should obviously still work fine.

>  there any docs anywhere, or are you planning to write those later?

yes, later, this will be a page like the api docs, or even part of that. That test page uses [this script](https://github.com/haraldschilly/cocalc-embedding-test/blob/master/script-test.js) with a couple of hardcoded projects and paths.

>  how would any person besides you use this?

for a cc-in-cc dev project, roughly this should work:

1. host static files on a webserver on your own domain – this must be behind `https://`
2. as admin user in your cc-in-cc dev project, whitelist your domain in the site settings
3. restart the hub to make that setting effective (I think that's necessary, I'm not sure)
4. embed your dev project in an iframe, pointing to `.../app?fullscreen=kiosk` is good enough
5. write code or manually enter commands to get the back and forth communication going. For my own dev project use [this code](https://github.com/haraldschilly/cocalc-embedding-test/blob/master/script.js) for testing.

src/smc-webapp/projects.cjsx
williamstein: Can you rename the function then?   E.g., shouldn't it be called "switch_between_projects"?

haraldschilly: yes, `switch_to_project` sounds good to me

src/smc-webapp/session.ts
```diff
@@ -8,11 +8,19 @@ const { throttle } = require("underscore");
...
-const async = require("async");
-
-exports.session_manager = (name, redux) => new SessionManager(name, redux);
+exports.session_manager = (name, redux) => {
+  const sm = new SessionManager(name, redux);
+  if (COCALC_MINIMAL) {
+    // we only need the session manager to open a target URL
+    return undefined;
+  } else {
+    sm;
```
williamstein: This is a MAJOR BUG!  It's because of fat arrow functions and implicit returns!

PR #3998, opened 2019-08-06
initialize less when in kiosk mode

haraldschilly:
# Description

simple idea: when started in kiosk mode, the variable `COCALC_MINIMAL`  causes some aspects to not initialize. the effect is less network traffic and overhead, but I'm not sure if this really helps significantly or is just a drop on a hot stone. This is maybe just a first step.

# Testing Steps
1. I refreshed various pages, in particular a jupyter notebook. All seem to work fine.
1. normal pages and `fullscreen=default` do not trigger this mode, hence they work just like before.

# Relevant Issues

Front end:


williamstein:
This look great!  Even if this itself doesn't help at all (hot stone), it's a nice start.

Issue #4000, opened 2019-08-06
dropdown menu for Source, etc is not visible in default tex frame editor in ~~future versions of~~ Chrome


DrXyzzy:
Seen on macOS 10.14.3.
Seen with:
* Google Chrome (Canary) 78.0.3875.0 (Official Build) canary (64-bit) 
* HeadlessChrome/77.0.3844.0 (comes with current puppeteer-1.19.0.tgz)

**NOT seen** with:
* Google Chrome (stable) Version 76.0.3809.100 (Official Build) (64-bit)

This may be of interest so that we can prevent breakage when Chrome is updated.

Current view, Chrome 76 (ok):

<img width="639" alt="tex-menu-chrome-76" src="https://user-images.githubusercontent.com/528072/62583148-978f9f80-b874-11e9-99bb-7330ba634859.png">

Future view, Chrome 77 and 78 (can't see pulldown menu without resizing):

<img width="640" alt="tex-menu-chrome-78" src="https://user-images.githubusercontent.com/528072/62583157-9fe7da80-b874-11e9-80b1-d2ba7cdbb5a6.png">

williamstein:
The relevant css/code that manages this is pretty lame/brittle, and I always wanted to rewrite it to be better.    I'm not too surprised it breaks.

haraldschilly:
I think I saw this as well, I'll take a look.

DrXyzzy:
See also #3718.

williamstein:
I reverted @haraldschilly 's PR, so this is still open.

DrXyzzy:
I am seeing this now with Chrome Version 76.0.3809.132 (Official Build) (64-bit) macOS 10.14.3 (18D109)

![Screen Shot 2019-09-09 at 6 45 17 PM](https://user-images.githubusercontent.com/528072/64573970-09a34a80-d333-11e9-8b69-5d53f063c1db.png)

I don't see it in current Safari due to another issue, #3718.

<img width="593" alt="Screen Shot 2019-09-09 at 6 47 41 PM" src="https://user-images.githubusercontent.com/528072/64573986-23dd2880-d333-11e9-8c44-ab8c1a0b7ac3.png">

haraldschilly:
weird. can you remember if this worked at least once since that august 8th patch? in any case, I'll look into this.

DrXyzzy:
@haraldschilly 

> can you remember if this worked at least once since that august 8th patch? 

[...]

PR #4002, opened 2019-08-07
webapp/frame editor: fix overlapping buttons 

haraldschilly:
# Description
I thought #4000 is easy, but it's not. With this change it behaves as before in firefox, while it overlaps in chrome. The overlapping is at least in such a way, that the right hand controls are always accessible.

chrome beta:

chrome

firefox

Screenshot from 2019-08-07 14-28-44

# Testing Steps
1.  I tested the latex editor in chromium (stable), chrome beta and firefox

# Relevant Issues
#4000

Front end:


williamstein:
I merged this, then realized it horribly breaks things far far worse.  It makes it so the dropdown on the right to select the type of frame is UNDER the frame, hence not visible.  I'm reverting this.

PR #4014, opened 2019-08-14
Jupyter IFrame heuristic

haraldschilly:
# Description
This is a slight enhancement to the iframe heuristic. I.e. when some html looks like it is a "full" page, it embeds it in an iframe – otherwise it leaves it as it is. With that, html pages for more advanced plotting outputs, like [altair vis](https://altair-viz.github.io/) work out of the box, more or less. (as of right now, its html is inserted as embedded html, which implicitly removes parts, makes the scripts not work properly, etc.)

What's unclear/missing IMHO:

*  "load viewer button", I don't understand its purpose. Improving it should be a follow up issue.
* the height of the iframe is hardcoded. this makes no sense. the output could be a thin bar or very high. solving this needs more insight into all this. (colab does it properly)

# Testing Steps

What I did is `from IPython.display import HTML` and then
1. simple div, comes up embedded
2. complex `<html>...</html>`, which is an iframe. 

Screenshot

Screenshot from 2019-08-14 10-43-06

actual use case

Screenshot from 2019-08-14 10-47-58

# Relevant Issues

Front end:


williamstein:
Very good idea!

PR #4015, opened 2019-08-14
Plaintext invite emails

haraldschilly:
# Description

# Testing Steps
1.
1.
1.

# Relevant Issues

Front end:


williamstein:
This makes me happy (e.g., increased security, rewriting in typescript)!

src/smc-hub/client.coffee
```diff
@@ -1407,25 +1408,11 @@ class exports.Client extends EventEmitter
...
-                        <a href='#{base_url}'>#{base_url}</a>
-                        and sign in using your email address '#{locals.email_address}'.
-                        #{direct_link}</b><br/>
-                        """
-
-                    # The following is only for backwards compatibility with outdated webapp clients during the transition period
-                    if not mesg.title?
-                        subject = "Invitation to CoCalc for collaborating on a project"
+                    try
+                        email_body = makeEmailBody(subject, mesg.email,locals.email_address,  mesg.title, mesg.link2proj)
```
williamstein: It would be nice to use Javascript's standard camelCase variable/function names.  However, we do that basically nowhere in our entire codebase, and not even elsewhere in this same line (e.g., locals.email_address)... I feel like randomly introducing it will just be confusing/annoying, and make us have to think about stuff we don't really have time to think about...  We could discuss this tomorrow morning.

williamstein:
I don't think "Plaintext invite emails" is really the right description for this -- what you're doing is much better than that.  Maybe "Safe invite emails"?

PR #4024, opened 2019-08-20
Type db schema

johnjeng:
# Description
Adds some type checking to db-schema to ensure that get/set fields are actually declared. Remaining rules options are a work in progress.

Uses a `create` function so that you don't have to type fields and the write the fields for the object again. There could be a better way to type it but no [real luck yet](https://stackoverflow.com/questions/57517135/can-i-dynamically-type-an-object-literals-keys-to-be-based-on-a-siblings-keys).

@williamstein Maybe pull and make some typos to check how the errors look to you?

# Testing Steps
1. No compile errors
1. Run the main site
1. Check that accounts work

# Relevant Issues

Front end:


williamstein:
It definitely feels useful.  I'm very impressed that it is possible to declare what you did in Typescript so elegantly.

williamstein:
If this works, I'm happy to merge it.

johnjeng:
It works as far as I can tell.

haraldschilly:
did you see the travis errors? e.g.

> `2019-08-20T07:08:17.815Z - debug: PostgreSQL.update_schema: "error creating tables -- TypeError: Cannot read property 'virtual' of undefined"`

and in a test suite for webapp are 34 failing.  master is fine, hence I'm quite sure this is caused by the code here

johnjeng:
Oh, yup, this is broken.

johnjeng:
API tests seem to fail on master and Travis checks so I assume those are known to be broken.

PR #4026, opened 2019-08-21
static: reinstate fixed navbar via flex

haraldschilly:
# Description

# Testing Steps
I tested this in chrome and FF

# Relevant Issues

Front end:

### DIFF
--- a/src/webapp-lib/_base.pug
+++ b/src/webapp-lib/_base.pug
@@ -115,170 +115,7 @@ html.no-js(lang="en")
     style
       :sass
         @import "smc-webapp/_colors.sass"
-        img[data-src]
-          min-height: 50px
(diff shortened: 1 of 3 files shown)


williamstein:
What did you change?  It's a LOT of work to tell from this patch, because you moved all the relevant code to another file, then changed it there...

williamstein:
Well, I don't understand this, but it seems harmless, and worse case I can revert.

haraldschilly:
yes, I also cleaned this up a bit, makes it hard to follow. The key idea is to split the top part in two divs and then

```
body
  display: flex
  flex-direction: column
  height: 100vh
  overflow: hidden

div#content
  flex: 1 1 auto
  overflow: auto

nav#nav-main
  flex: 0 0 auto
```

PR #4040, opened 2019-08-29
rewrite billing.cjsx in typescript

williamstein:
@haraldschilly -- heads up; you need to be aware of this if you try to finish your "move the FAQ" branch...


williamstein:
@haraldschilly -- please be AWARE that the FAQ component you were working on is now moved and in typescript!!

haraldschilly:
ok, I see, very good

PR #4045, opened 2019-08-30
build: enable incremental typescript builds

haraldschilly:
# Description

This patch enables it, still builds, but I don't know if it is doing anything at all …

I also updated ts-node, because maybe that's the reason why it isn't doing anything.

It's probably fine to merge this, and we'll experience something some day in the future when this actually works?

---

that's below is no longer true, but worth a check during review

status: As it is, this is broken. `update_react_static` gives

```
error TS5023: Unknown compiler option 'incremental'.
error TS5023: Unknown compiler option 'tsBuildInfoFile'.
```

and I'm not looking into this for now. 

Documentation: https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#faster-subsequent-builds-with-the---incremental-flag

# Testing Steps
1.
1.
1.

# Relevant Issues

Front end:


williamstein:
I'm confused -- you set it to needs review, but also said it is broken?

haraldschilly:
@johnjeng  brought this up, that's why I set him to be the reviewer and take a look. Maybe he has heard of a detail I'm not aware of.

williamstein:
> I also updated ts-node, because maybe that's the reason why it isn't doing anything.

I could be wrong, but my impression is that this incremental option is **irrelevant when using webpack**.  This is only something that might speed up starting projects and the hub in cc-in-cc dev mode.

williamstein:
I'm changing this to needs work, since it is not known to solve any known problem and doesn't seem to do anything.

williamstein:
I'm closing this, since I feel like our current typescript speed situation is sufficiently fast, given the changes I made to our use of webpack.  Re-open it if you feel differently.

PR #4060, opened 2019-09-04
Clearly exclude package-lock

johnjeng:
I never got around to using @williamstein's solution to assume no change so I just see garbage all the time.

In any case, ignoring them is what we actually do so there's no reason to have the lock files in our repo. If anything it's entirely misleading.


williamstein:
> I never got around to using @williamstein's solution to assume no change so I just see garbage all the time.

It's built in -- just do this:

```
~/cocalc/src$ . smc-env
~/cocalc/src$ ignore-package-lock 
```

haraldschilly:
well, besides that, it's hard to review any PR where the tests are broken. So, what's with the first one: `Error: Expected 'URLs are not allowed. Found is.test.ok' to be undefined` ?? I don't think this PR here is the culprit.

haraldschilly:
I just realized you missed two files.No problem, I'll make a follow up in a branch. Just for the record, you can use git to search for files. In this case here:

```
~/cocalc/src$ git ls-files '../*/package-lock.json'
```

PR #4069, opened 2019-09-10
Fix frame menu 4000

### LINKED ISSUE #4000: dropdown menu for Source, etc is not visible in default tex frame editor in ~~future versions of~~ Chrome

### LINKED ISSUE #4064: file dropdown button in code editor does a line wrap
via CSS, prevent that triangle from doing a newline.

Screenshot from 2019-09-06 13-54-05

haraldschilly:
# Description
different approach to fix #4000 … there are some contradicting problems I can't solve across browsers. So, this here isn't perfect either, just a bit better.

bonus: fix #4064

# Testing Steps
I tested this on chrome 76 and 77 and firefox 69. 

# Relevant Issues

Front end:


williamstein:
1. Can you make a PR that doesn't have thousands of lines of package-lock.json changes?  This PR doesn't even involve packages.
2. Any summary of the contradicting problems, and what this PR breaks?   It's easier to review/merge something with known problems if you tell me what those problems are...

williamstein:
(I just realized that this is not needs-review yet)

haraldschilly:
I've removed the lockfile removal.

The contradictions are using `float:"right"` for the control panel and causing the latest problem in #4000 vs. the dropdown menus do not show up (either in the control bar or in the buttonbar)  if `z-index` is used to fix the overlap. Another issue was a missing background in the title and the connection information. Otherwise one can see through one div to the other one. A third issue is using the flex layout for the top bar, as `direction: column`. This becomes apparent when some of the flex related styles slightly change. That's why the title bar style got a `flex: 0 0 auto` to avoid expanding to 50% height.

As it is here right now, it addresses most of the issues, but for very narrow frames the control buttons start to disappear to the right. flex seems to have a limit how far it likes to shrink and my attempts tuning the settings didn't help. The alternative would be to make the control always show up, by aligning the content to the right (and not the left). but then the opposite happens and the button bar disappears to the left.  Well, technically this is still an improvement (given it doesn't break to hefty on that newer safari, which I haven't tried yet).

williamstein:
> I've removed the lockfile removal.

I see what's going on too - i've just merged that other PR about that.

haraldschilly:
I've made further changes, and well, this solves a couple of problems but it's not perfect. I'm for merging this and then making new issues to brainstorm ideas. The browsers I tested are chrome 77, firefox 69 and safari 12.1 via browserstack.

williamstein:
Wow, this makes the most important thing way worse.  I have to rever this PR somehow. Look my fullscreen toggle button in the upper right of the left frame is just gone:

image

williamstein:
OK, I've fixed things so the upper right control is always visible.  You might consider this breaking things though... 🤷‍♂

PR #4079, opened 2019-09-12
Add store getIn types

johnjeng:
# Description
- Fix deep typing for `getIn` when the entire path was always defined
- Add deep typing for `getIn` when some part of the path MAY be defined
- Add type `DeepImmutable<T>` for converting from plain JS objects to our use of immutable js.
    - Does NOT support custom classes. So *the typing* (not runtime, so use `as any` if you must) will break if you stored a custom class inside a store. So far we don't do this it seems.

# Testing Steps
```
cd src/smc-webapp
npm test
```

- Spin up a dev server and open cocalc.
- Open account
- Go to billing
# Relevant Issues

Front end:

### DIFF
--- a/src/jest.config.js
+++ b/src/jest.config.js
@@ -2,20 +2,27 @@ const { defaults: tsjPreset } = require("ts-jest/presets");
 const path = require("path");
 
 module.exports = {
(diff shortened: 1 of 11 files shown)


williamstein:
Looking forward to more typing with this.

PR #4111, opened 2019-09-27
install: fix executor to make cocalc-docker build

haraldschilly:
# Description

This should fix docker-build.

# Testing Steps
Let's just wait until travis-ci finishes … well, travis is :green_heart: 

# Relevant Issues

Front end:

### DIFF
--- a/src/install.py
+++ b/src/install.py
@@ -2,8 +2,8 @@
 from __future__ import print_function
 import argparse, os, sys, time
 from concurrent.futures import ThreadPoolExecutor
-executor = ThreadPoolExecutor(max_workers=3)
 
+WORKERS = 3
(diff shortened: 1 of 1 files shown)


williamstein:
> This should fix docker-build.

Just out of curiosity... why?

haraldschilly:
This uses a ThreadPool to parallelize these npm build commands. What I didn't test was reusing the same one more than once. That's where it broke. That doesn't happen with the usual build, because there is just one usage. Instantiating new ones each time fixes this. I don't remember for sure, but this is somehow different from when I used this the last time – probably too long ago, or from the now ancient python3 times.

williamstein:
Thanks for the excellent explanation!

PR #4132, opened 2019-10-05
sagews: fix compatibility issue with 8.9

haraldschilly:
# Description
The goal is to make `smc_sagews` and hence sage worksheets work with sage 8.9.
What changed is the [sage.plot.multigraphics](https://doc.sagemath.org/html/en/reference/plotting/sage/plot/multigraphics.html) module.
I also made a couple of small cleanups and well, the bulk of linechanges is formatting …

Beyond that, I've made the changes by mimicking what I see in the code, that's all. Hence "careful".

# Testing Steps
I'm not sure what to watch out for. What I did is toggle versions via

```
~/bin$ ln -s /ext/bin/sage-8.9 sage  # or removing it
~/cocalc/src/smc_sagews$ smc-sage-server restart
```

and tested if multi graphics and other plots show up. 

# Relevant Issues

Front end:


williamstein:
This is a nightmare to review "carefully" given that you put the formatting in the exact same commit as the actual changes.  Please don't do that in the future if you can help it.    Sure, mix the two steps up for something that doesn't need "careful" review... but for something that does, please don't.  It makes it much harder.

src/smc_sagews/smc_sagews/sage_salvus.py
```diff
@@ -2763,11 +2748,34 @@ def plot3d_using_matplotlib(expr,
     plt.show()
 
 
-from sage.plot.graphics import Graphics, GraphicsArray
+# 8.9 introduced https://doc.sagemath.org/html/en/reference/plotting/sage/plot/multigraphics.html#sage.plot.multigraphics.MultiGraphics
+try:
```
williamstein: You should do the logic in the opposite order or actually check the version of Sage.  The way you are doing it now you are assuming ` from sage.plot.graphics import GraphicsArray` fails implies "pre 8.9".  However, in two years for sage-9.4 somebody could happily create a module called graphics that exports GraphicsArray, and suddenly this code totally breaks!

In fact, that's not such a crazy scenario.  Imagine this happening:
1. People complain that sage.plot.graphics.GraphicsArray went away and lots of code broke as a result (the public api of sage changed without a deprecation warning),
2. Somebody decides to add back a shim so that sage.plot.graphics.GraphicsArray exists again.
3. Boom -- your code just broke.

williamstein:
> This is a nightmare to review "carefully" given that you put the formatting in the exact same commit as the actual changes.

I just had a clever idea!  I just formatted sage_salvus.py using yapf as a single commit and pushed that to master.  Then I rebased your commit against *that*.  As a result, now diff with master should show only what you actually did that isn't formatting.  🤷‍♂

Issue #2549, opened 2017-12-06
hub: remove ttl_dt for computing stats


haraldschilly:
getting rid of the no longer needed

```
ttl_dt : 15       # 15 secs subtracted from ttl to compensate for computation duration when called via a cronjob
```

because now we have only  one update job

haraldschilly:
https://github.com/sagemathinc/cocalc/commit/871457e920c7e9f1b2a5d4a592b221ff2896ba43

PR #4134, opened 2019-10-07
hub/stats: get rid of ttl offset, since we are not using cronjob

haraldschilly:
# Description
see https://github.com/sagemathinc/cocalc/issues/2549

fallout of migrating stats to run as a microservice

# Testing Steps
1.
1.
1.

# Relevant Issues

Front end:


williamstein:
Stats are still computed by the docker image so I don't understand the justification for this.   I also don't understand #2549.  I don't even understand why the code you want to remove this there...

haraldschilly:
I guess this was added when this was running in a loop? I don't know. I don't see the point as well. It's much more intuitive to tune that TTL time and that's it -- i.e. no additional compensation for the time of the query. The `time` field is set when the queries start, which means there is no need to compensate for their duration.

williamstein:
I still don't really know what this is about (despite having written the code, probably), but merging seems harmless.

Issue #4136, opened 2019-10-07
ban specific sub release versions of chrome 77


haraldschilly:
due to a bug with websocket in chrome 77 (which will be fixed by a new subrelease today or tomorrow), we have to ban 77 prior to a subrevision.

- [ ] which subrelease actually ?
- [ ] parse `navigator.appVersion` to get it as  an integer 

upstream: https://bugs.chromium.org/p/chromium/issues/detail?id=1006243

williamstein:
When we decide to really ban this, we'll remove
```
(window as any).buggyCh77 = buggyCh77;
```
and  the `false &&` from `src/webapp-lib/preflight-checks.ts`.

williamstein:
I'm just going to delete this code. Chrome 77 is ancient history and worst case, people will figure out to refresh their browser.  Also, I greatly reduced what would cause this in the first place.

PR #4139, opened 2019-10-08
preflight: ban specific chrome 77 

haraldschilly:
# Description
* extracting exact chrome version as a quadruple of integers
* testing them against what's in the chromium bugreport

# Testing Steps
I tweaked the numbers in the test to ban myself. Numbers are also parsed correctly.

I should check with browserstack edge, FF, etc. ... but I don't see how this could break.

Did the above, looks good. Money shot:

Screenshot from 2019-10-08 15-00-56

# Relevant Issues

 #4136

Front end:


williamstein:
This is good, but I think we should wait a significant amount of time (a couple of weeks) after this is fixed upstream before rolling it out.  Reasons:

1. We do have a message with a reliable workaround (open a new tab) that does get triggered when the problem happens.
2. I used cocalc with broken chrome 77 all day long actively... without any trouble (since I didn't hit the 256 reconnect limit).

williamstein:
Also, I have a problem with the wording of the message, since it says "You cannot use cocalc with your current browser because..." but that is **simply not true**.   What happens in this instance is you can't use cocalc after 255 failed websocket connection attempts.   In some situations things could work perfectly for many days before hitting that, and for others it could be quicker.

williamstein:
I wonder how quickly chromium really gets updated in the wild?   I may have to do more to try to workaround this bug, since it's pretty severe.   E.g., looking at today on Google analytics we have nearly 70% of users are using chrome, and of those well over 70% are using the version that has this bug.  So _right now_ the vast majority of users are using a version of Chrome that has this websocket bug.

image

williamstein:
I'm setting this to blocked, since there is no way we're banning the browser 60% of users are using to connect to cocalc.com right now.

haraldschilly:
today we seemed to have passed >50% for 77-subreleases >= 114

williamstein:
> today we seemed to have passed >50% for 77-subreleases >= 114

Nice. And yet I'm still stuck on 105 because I'm using ChromeOS... on Google's flagship Pixel Slate nonetheless.  The fact that Chrome can't be updated without a full OS update, and those are rolled out much more slowly, is a major con of this platform.

haraldschilly:
yes, that's not so good. I bet they patch it a little bit for chrome OS and hence it's lagging behind.

williamstein:
Interestingly, this code breaks when doing the build on Kubernetes/docker:
```
[2604] multi ./webapp-lib/primus/primus-engine.min.js 28 bytes {0} [built]
    + 2591 hidden modules

ERROR in /cocalc/src/webapp-lib/preflight-checks.ts
[tsl] ERROR in /cocalc/src/webapp-lib/preflight-checks.ts(148,16)
      TS2532: Object is possibly 'undefined'.

ERROR in /cocalc/src/webapp-lib/preflight-checks.ts
[tsl] ERROR in /cocalc/src/webapp-lib/preflight-checks.ts(148,29)
      TS2339: Property 'join' does not exist on type 'string | number[]'.
  Property 'join' does not exist on type 'string'.
Child html-webpack-plugin for "app.html":
```

I'll fix...

PR #4141, opened 2019-10-08
try to better workaround a major chrome 77 websocket bug?

williamstein:
See https://bugs.chromium.org/p/chromium/issues/detail?id=1006243


williamstein:
Not a good idea as is.

PR #4142, opened 2019-10-09
webapp/jupyter2: resize too large images to container width

haraldschilly:
# Description
this is something that bugs me and I today just saw it in a notebook again. if an image has no height/width, it's show as it is. for large ones, they're off screen and it's hard to scroll around, etc. this here limits the width to 100% and if you need to see it full size, you can click. this doesn't impact any of the other plots, because they all have height and width set.

# Testing Steps
I used dask to plot a graph, which is in this case larger than my screen and hence hard to understand.

```
import dask.array as da
x = da.random.random((1000, 1000), chunks=(300, 300))
y = x + x.T
z = y[::75, 100:].mean(axis=1)
rz = (z[:100].sum() / z[:100].shape[0])
rz.visualize()
```

# Relevant Issues

Front end:

### DIFF
--- a/src/smc-webapp/jupyter/output-messages/image.tsx
+++ b/src/smc-webapp/jupyter/output-messages/image.tsx
@@ -13,6 +13,7 @@ interface ImageProps {
 
 interface ImageState {
   attempts: number;
+  zoomed: boolean;
 }
 
 export class Image extends Component<ImageProps, ImageState> {
@@ -20,7 +21,8 @@ export class Image extends Component<ImageProps, ImageState> {
 
   constructor(props: ImageProps, context: any) {
     super(props, context);
-    this.state = { attempts: 0 };
+    this.img_click = this.img_click.bind(this);
+    this.state = { attempts: 0, zoomed: false };
   }
 
   load_error = async (): Promise<void> => {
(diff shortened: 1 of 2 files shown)


src/smc-webapp/jupyter/output-messages/image.tsx
```diff
@@ -43,18 +45,37 @@ export class Image extends Component<ImageProps, ImageState> {
...
+  render_image(src, on_error?): Rendered {
+    const props = {
+      src,
+      width: this.props.width,
+      height: this.props.height,
+      onClick: this.img_click
+    };
+    if (this.props.width == null && this.props.height == null) {
+      const cursor = this.state.zoomed ? "zoom-out" : "zoom-in";
+      props["style"] = { cursor } as React.CSSProperties;
```
williamstein: Why are you writing props["style"] instead of props.style, and props['onError"] instead of props.onError.  It looks like you're writing **PYTHON** code....

williamstein:
Wow, this is cool!

src/smc-webapp/jupyter/output-messages/image.tsx
haraldschilly: no idea, at least temporary my brain did output it that way

haraldschilly:
a follow up of this could be to deal with images that are smaller than the view width, or well, all images even those with a specific height and width. it's a bit more tricky, because it involves measuring the image after it was mounted...

PR #4163, opened 2019-10-17
npm install: delete old node_modules by moving them out of the way

haraldschilly:
# Description
this is just for speeding up `npm run make`. deletes old node modules in parallel of npm ci. also,  building project first, because there is more C++ code being built.

# Testing Steps
well ... if it causes issues, it should be easy to fix. I've not noticed any problems.

# Relevant Issues

Front end:


williamstein:
Seems safe.

Issue #4170, opened 2019-10-17
jupyter2: new_cell_pos → Cannot read property 'getIn' of undefined


haraldschilly:
chrome: `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/77.0.3865.90 Safari/537.36`

smc_git_rev: b612379cd2218f31d34a2eb1eab1255c436ffbb2

Date: 2019-10-14 13:05:27.880916

`Cannot read property 'getIn' of undefined`

```
TypeError: Cannot read property 'getIn' of undefined
    at Object.t.new_cell_pos (https://cocalc.com/static/smc-75b1a24eaac3878d8530.cacheme.js?75b1a24eaac3878d8530:5:601020)
    at t.insert_cell_adjacent (https://cocalc.com/static/smc-75b1a24eaac3878d8530.cacheme.js?75b1a24eaac3878d8530:5:2739054)
    at e.insert_cell (https://cocalc.com/static/smc-75b1a24eaac3878d8530.cacheme.js?75b1a24eaac3878d8530:5:2827583)
    at Object.f (https://cocalc.com/static/smc-75b1a24eaac3878d8530.cacheme.js?75b1a24eaac3878d8530:5:834727)
    at https://cocalc.com/static/smc-75b1a24eaac3878d8530.cacheme.js?75b1a24eaac3878d8530:5:1409832
    at dispatch (https://cocalc.com/static/0-75b1a24eaac3878d8530.cacheme.js?75b1a24eaac3878d8530:6:445876)
    at v.handle (https://cocalc.com/static/0-75b1a24eaac3878d8530.cacheme.js?75b1a24eaac3878d8530:6:443888)
    at e._wrapper.e._wrapper (https://cocalc.com/static/smc-75b1a24eaac3878d8530.cacheme.js?75b1a24eaac3878d8530:5:698235)
```

PR #4171, opened 2019-10-17
jupyter: fix cells undefined in uils/new_cell_pos

haraldschilly:
# Description
well, this looks easy, but once I adhere to the text "Returned undefined whenever don't really know what to do" and encode that in typescript, I had to come up with a fallback. I don't know if that's reasonable or not. 

# Testing Steps
no idea how to reproduce, must be an odd corner-case.

# Relevant Issues
#4170

Front end:

### DIFF
--- a/src/smc-webapp/jupyter/actions.ts
+++ b/src/smc-webapp/jupyter/actions.ts
@@ -865,7 +865,7 @@ export class JupyterActions extends Actions<JupyterStoreState> {
       id,
       delta
     );
-    return this.insert_cell_at(pos, save);
+    return this.insert_cell_at(pos != null ? pos : 0, save);
   }
 
   delete_selected_cells = (sync = true): void => {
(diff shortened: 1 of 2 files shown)


williamstein:
I agree that this PR is going in the wrong direction to hide one problem by making things maybe worse.  E.g., if a user were to ever hit these code paths they would end up with a cell they wouldn't even see at the top of the document -- that doesn't seem useful to do on purpose. 

Of course the right solution is to use typescript to ensure that anything that calls this new_cell_pos function has to give a parameter that is defined.   In practice this is of course mostly likely happening due to calling it before something is fully initialized, or after things are being closed.

I would change this PR to either throw an exception so we can find out what causes the trouble (when it gets reported again), or at least put in a big comment saying that this is a quick temporary hack until the typescripting of this module is improved.

williamstein:
OK, the root cause of this is that most of the types in jupyter/store.ts are `any`, since the initial conversion from CoffeeScript to TypeScript is far from done...

PR #4179, opened 2019-10-21
Improve typing for Store.getIn and TypedMap.getIn

johnjeng:
# Description
Simplifies `getIn` typing from `Store` and `TypedMap`. 
Puts the shared `getIn` and `get` types in the same location.

Fixed edge cases involving nesting of different types of immutable structures that using Immutable.js and TypedMap would involve.

Eg. The following used to break (typing-wise) from being unable to infer what was inside `immutable.Map`. 
```typescript
declare store: Store<{
  accounts: immutable.Map<string, {first_name: string;}>
}>;

let account_id = "some-uuid";
store.getIn(["accounts",  account_id, "first_name"]); // <- Now correctly returns type: string
```

Added support well typing `5` layers deep, up from `3` before.

# Testing Steps
1. Run up a test project.
1. Open a course
1. In `smc-webapp`, run `npm test`

# Relevant Issues

Front end:


johnjeng:
Hmm, passing on Travis CI but running OOM on my dev project... Suspicious.

williamstein:
This is so important for making it so we can type our stores and get much more value out of Typescript.

williamstein:
Unfortunately, this introduced a bug.  

1. Open Account --> Subscriptions and course packages.
2. Expand one of the "Invoices and Receipts" entries.
3. See $NaN.   

On the master branch in 3 you correctly see the price.

johnjeng:
Fixed the NaN issue. NoImplicitAny would have found the bug in the first place. Would be good to turn that for parts of our code soon probably.

williamstein:
> Would be good to turn that for parts of our code soon probably.

Unfortunately, I think it's impossible to turn no implicit any on only for certain files:

https://github.com/microsoft/TypeScript/issues/8855   

Otherwise, we would be using it.   Turning it on everywhere is also difficult for us since it generates thousands of errors throughout our codebase that are daunting to fix everywhere (I did try once, but it's ridiculous).  I (maybe?) wish I had turned it on on day 1...

The only possible workaround is to patch typescript itself to ignore errors from certain files, but that sounds like a mess.

PR #4187, opened 2019-10-22
do not use rtypes.object for Invoices in billing (#4179)

williamstein:
temp


johnjeng:
Incorporated into #4179

PR #4199, opened 2019-10-30
code editor in a frame -- take three...

williamstein:


src/smc-webapp/frame-editors/code-editor/code-editor-manager.ts
```diff
@@ -0,0 +1,80 @@
...
+import { Actions, CodeEditorState } from "../code-editor/actions";
+import { get_file_editor } from "../frame-tree/register";
+import { redux } from "../../app-framework";
+
+export class CodeEditor {
+  private project_id: string;
+  private path: string;
+  private actions: Actions;
+
+  constructor(project_id: string, path: string): void {
```
haraldschilly: FYI, I checked travis, the `void` causes

```
 ERROR in /home/travis/build/sagemathinc/cocalc/src/smc-webapp/frame-editors/code-editor/code-editor-manager.ts(14,50)
      TS1093: Type annotation cannot appear on a constructor declaration.
```

williamstein: I wonder why we don't see that error when using typescript + webpack.

haraldschilly: I really don't know.

PR #4200, opened 2019-10-30
Run prettier and eslint

johnjeng:
# Description
Runs prettier on all our `.ts` and `.tsx` files.
Runs `eslint --fix` on all our `.ts` and `.tsx` files.
Installs Eslint with Typescript and React bindings.
Changes the `lint` script in package.json to use Eslint instead of the unused coffeelint.
Turns off the following linting rules from the defaults:
 - camelcase
 - no-inferrable-types
 - no-explicit-any

THIS PR DOES NOT FIX ALL ISSUES FOUND BY ESLINT.
✖ 7418 problems (2612 errors, 4806 warnings)
There are many even after `--fix`...
But we will be able to run eslint on new code.

# Testing
Run `npm run lint` from `src/`.

# Relevant Issues

Front end:


johnjeng:
@williamstein Thoughts? 

Examples of rules I'd be into:
Must use import if available [@typescript-eslint/no-var-requires](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-var-requires.md) 
Must explicitly write return types [@typescript-eslint/explicit-function-return-type](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-function-return-type.md)

johnjeng:
Probably dangerous by sheer diff size? Semantically I think it's pretty safe...

williamstein:
This is a nightmare to review.   It's a mix of thousands of let --> const changes combined with an occasional manual change, combined with mystery changes.    Using github I have to click "show diff" for most files because there are so many, which just makes it even more difficult.  I hope I'm not going to regret this.

But I definitely like the consistency and code quality improvements!!

PR #4201, opened 2019-10-30
Run eslint in smc-webapp with --fix

johnjeng:
# Description
Example diff of running Eslint with `--fix`


johnjeng:
--fix seems to mostly run Prettier and change `let` to `const` when safe.

williamstein:
> Yay or nay on having #4200 also run eslint with--fix on all our ts/tsx code?

Is it SAFE?  What's the probability of introducing bugs?

johnjeng:
It appears very safe. Looking through this diff, I couldn't find any change that wasn't prettier or simply a change from `let` to `const`.

I will add this to #4200 then.

PR #4233, opened 2019-11-14
npm package updates

haraldschilly:
# Description

this reduces the number of different package versions, although it's more conservative than in the first round. It's also just a subset of packages, because I want to keep this small. The next PR will tackle the next patch.

 this also updates prettier, which I'm also updating in the software environment (i.e. both should be the same as well)

# Testing Steps
well, there shouldn't be any changes.

# Relevant Issues

Front end:


williamstein:
I'm curious what, if any, testing you've done?

PR #4239, opened 2019-11-20
webapp/jupyter: change cell type selector to be an antd dropdown

haraldschilly:
# Description
This isn't doing much, just changing the cell type dropdown. That way I get at least a little bit done, more to come :roll_eyes: 

Screenshot from 2019-11-20 16-11-37

# Testing Steps
dropdown changes cell type

# Relevant Issues

### [Checklist](https://github.com/sagemathinc/cocalc/wiki/PR-Checklist):
- [ ] No debugging console.log messages.
- [ ] All new code is actually used.
- [ ] Non-obvious code has some sort of comments.

Front end:
- [ ] Restart at least one project and check its `~/.smc/local_hub/local_hub.log`
- [ ] Completely restart Webpack with `./w` in `/src`
- [ ] Completely restart the hub by killing and restarting `./start_hub.py` in `/src/dev/project`
- [ ] Screenshots if relevant.

PR #4245, opened 2019-11-22
webapp: converting many r_misc const imports

haraldschilly:
#4239  Description

changing "const" imports to "import/from". various TS errors appeared, hence there are also subtle changes. I repeatedly used

    git grep r_misc | grep -v from | grep -v cjsx | grep -v coffee | cut -d: -f1  | sort -u | xargs open

to find the files, hence this isn't the whole set, but a lot.

# Testing Steps
this should be low risk, also webapp only

# Relevant Issues

Front end:


src/smc-webapp/project/utils.ts
```diff
@@ -22,7 +22,7 @@ export type NewFilenameTypes =
   | "ymd_semantic";
 
 export const NewFilenameFamilies = Object.freeze<
-  Readonly<{ [name in NewFilenameTypes]: string }>
+  { [name in NewFilenameTypes]: string }
```
haraldschilly: looks like typescript started to understand that freezing is making it readonly

src/smc-webapp/r_misc/selector-input.tsx
```diff
@@ -7,7 +7,8 @@ interface Props {
   options:
     | string[]
     | { value: string; display: JSX.Element }[]
-    | { [keys: string]: JSX.Element };
+    | { [keys: string]: JSX.Element }
+    | Readonly<{ [keys: string]: string }>;
```
haraldschilly: this is because of `project/ask-filename.tsx`. to check, use the new filename dialog and click on the selector how to generate new names

haraldschilly:
for the sake of sanity, I'll add notes to the non trivial changes

src/smc-webapp/admin/users/projects.tsx
```diff
@@ -60,21 +60,23 @@ export class Projects extends Component<Props, State> {
...
+              description: null,
+              users: null,
+              last_active: null,
+              last_edited: null
+            }
+          ]
+        },
+        options: [{ account_id: this.props.account_id }]
+      })
+    ).query.projects;
```
haraldschilly: formatting

src/smc-webapp/frame-editors/frame-tree/editor.tsx
```diff
@@ -43,7 +42,7 @@ interface FrameTreeEditorReduxProps {
   errorstyle: ErrorStyles;
   cursors: Map<string, any>;
   status: string;
-  load_time_estimate?: Map<string, any>;
+  load_time_estimate?: LoadingEstimate;
```
haraldschilly: `Map<string, any>;` is just not what the component gets. There is the type `LoadingEstimate`  now. I hope this is fine here?

src/smc-webapp/r_misc/icon.tsx
```diff
@@ -14,7 +14,7 @@ interface Props {
   inverse?: boolean;
   Component?: JSX.Element | JSX.Element[];
   style?: any;
-  onClick?: () => void;
+  onClick?: (event?: any) => void; // TODO tighten what event could be
```
haraldschilly: optionally adding `event` here is necessary for one usage

src/smc-webapp/r_misc/loading.tsx
```diff
@@ -2,13 +2,15 @@ import * as React from "react";
 import { Icon } from "./icon";
 import { TypedMap } from "../app-framework/TypedMap";
 
+export type Estimate = TypedMap<{
+  time: number; // Time in seconds
+  type: "new" | "ready" | "archived";
+}>;
+
```
haraldschilly: this is the `LoadingEstimate`

src/smc-webapp/admin/users/ban.tsx
```diff
@@ -70,7 +70,7 @@ export class Ban extends Component<Props, State> {
     return (
       <ErrorDisplay
         error={this.state.error}
-        onClick={() => {
+        onClose={() => {
```
haraldschilly: there were a few cases, where `ErrorDisplay` had `onClick` inside of it. I bet this is a bug. I changed all of them like this one here

src/smc-webapp/r_misc/icon.tsx
johnjeng: Use `event?: React.MouseEvent`. https://fettblog.eu/typescript-react/events/

src/smc-webapp/r_misc/loading.tsx
johnjeng: Wow. Drag to highlight comment region. I didn't know this was possible on GitHub.

src/smc-webapp/r_misc/icon.tsx
johnjeng: Use `event?: React.MouseEvent`. https://fettblog.eu/typescript-react/events/

haraldschilly: ok

src/smc-webapp/frame-editors/frame-tree/editor.tsx
williamstein: LoadingEstimate **is a** Map<string, any>, so this is fine, since LoadingEstimate is a TypedMap (which is just a special Map<string, any>).  You just replaced one valid type with a vastly better valid type.

PR #4300, opened 2019-12-23
LTI support code

haraldschilly:
# Description
This is only code for supporting an LTI plugin. It basically deals with adding a few more (very general) fields to the database and some extras for dealing with auth tokens. Regardless how LTI will be implemented, this should be useful in any case. Hence I wonder if this should be merged rather soon or if we leave it until all this settles. In any case, I leave this here as a PR ... which is better than just a branch.

# Testing Steps
1.
1.
1.

# Relevant Issues

Front end:


src/smc-hub/hub.coffee
```diff
@@ -80,6 +80,15 @@ MetricsRecorder = require('./metrics-recorder')
 # express http server -- serves some static/dynamic endpoints
 hub_http_server = require('./hub_http_server')
 
+if fs.existsSync(path_module.join(SMC_ROOT, 'smc-hub', 'lti', 'lti.ts'))
```
williamstein: This seems weird to me.  How is this going to work?  Why not just use try/catch?   Can you put some comments in the code.    Also, what is the intention longterm?  I.e., is it that we copy an lti.ts file in at runtime?

williamstein:
I agree that we should make sure this is safe & stable and merge it soon!   What's the danger if any right now?

src/smc-hub/hub.coffee
haraldschilly: true, I changed that to try/catch.

... and basically yes, I treat the whole `smc-hub/lti` subdirectory as being optional and it's only the main hub file that knows anything about it. Deployment could e.g. mean to copy or git-clone this LTI module somewhere and then set a symlink.

haraldschilly:
> What's the danger if any right now?

I think the only real danger is anything that related to authentication via a token. E.g. at some point it sets the password to an empty string in `sign-in.coffee`. There are also some small changes in the auth-token code.

williamstein:
I merged this after reading it thinking that if the lti option isn't set, then it wouldn't change any behavior.  It seems to seriously break anonymous account creation, so I'm going to go through and try to figure out why.     I'll basically try to 100% ensure that if the lti option is NOT set, then no behavior is changed.

src/smc-util/message.js
```diff
@@ -554,6 +554,7 @@ message({
   hub: required, // ip address (on vpn) of hub user connected to.
   account_id: required, // uuid of user's account
   email_address: undefined, // email address they signed in under
+  lti_id: undefined, // alternatively if email_address isn't set, there must be an lti_id
```
williamstein: I'm trying to figure out how your broke anonymous account creation, but at least this comment is a hint.  An anonymous account definitely doesn't have an email address or lti_id set...

williamstein:
Progress -- OK, the real problem is only that the fallback behavior when a sign in token is required no longer works.  If no sign in token is required, then everything seems to work fine still.

williamstein:
Upon further testing I think this PR is 100% fine, but *my* code for falling back to the login screen is not sufficiently robust, e.g., working in some situations but not others randomly.  So I'm fixing that.  This PR is fine.

PR #4302, opened 2019-12-24
edit wording

DrXyzzy:
# Description

One-word edit under "How it works" when "Share" is clicked after selecting a file in Files tab.

# Testing Steps
1.
1.
1.

# Relevant Issues

Front end:


williamstein:
> (unlisted). Files are automatically made public **within** 30 seconds any

Unfortunately, that is not necessarily true.    There is a queue and if the server were very busy, it could take longer than 30s (in fact, in theory, arbitrarily longer).

DrXyzzy:
Ok. But the original wording reads as if a word is missing. How about:

    Files are automatically made public within about 30 seconds any time you change them.

PR #4319, opened 2020-01-14
Fix computed values

johnjeng:
# Description
The computed values are run every time they're called at the moment. This PR fixes that.

# Testing Steps
1. `cd src/smc-util`
1. `npm test`
1. In the browser console `smc.redux.getProjectStore(project_id).selectors.stripped_public_paths.fn.recomputations();` with `project_id` appropriately filled in.
1.  Check any file
1. Run the same code above.
1. The recomputation count should not go up

# Relevant Issues

- [x] Run eslint on new and edited files

Front end:


src/smc-util/misc.js
```diff
@@ -1595,7 +1595,11 @@ const mathjax_environments = [
   "Vmatrix",
   "vmatrix"
 ];
-const mathjax_delim = [["$$", "$$"], ["\\(", "\\)"], ["\\[", "\\]"]];
+const mathjax_delim = [
+  ["$$", "$$"],
+  ["\\(", "\\)"],
+  ["\\[", "\\]"]
+];
```
johnjeng: Prettier formatting.

src/smc-util/misc.js
```diff
@@ -2647,7 +2651,8 @@ exports.log = function() {
 exports.wrap_log = function() {
   if (
     !exports.RUNNING_IN_NODE &&
-    (typeof window !== "undefined" && window !== null)
+    typeof window !== "undefined" &&
+    window !== null
```
johnjeng: Prettier.

src/smc-util/test/misc-test.coffee
```diff
@@ -1373,6 +1373,14 @@ describe 'top_sort', ->
...
         .toThrow("No sources were detected")
 
+    DAG4 =
+        node1 : ["node0"]
+        node2 : ["node0", "node1"]
+
+    it 'Works with implict sources', ->
+        expect misc.top_sort(DAG4)
+        .toEqual [ 'node0', 'node1', 'node2' ]
+
```
johnjeng: Add test for the broken case where node were found implicitly in the dependency graph.

src/smc-webapp/account/store.ts
```diff
@@ -39,7 +39,7 @@ export class AccountStore extends Store<AccountState> {
     return this.get("account_id");
   }
 
-  selectors: any = {
+  selectors = {
```
johnjeng: Woohoo!  selectors are correctly typed based on the `State` in `Store<State>` now!

src/smc-webapp/account/store.ts
```diff
@@ -49,19 +49,19 @@ export class AccountStore extends Store<AccountState> {
...
         );
       },
-      dependencies: literal([
+      dependencies: [
         "email_address",
         "passports",
         "is_logged_in",
         "lti_id"
-      ])
+      ] as const
```
johnjeng: Change to use the newer `as const` from TS 3.4

src/smc-webapp/app-framework/Store.ts
```diff
@@ -107,7 +111,7 @@ export class Store<State> extends EventEmitter {
     notSetValue?: any
   ): any => {
     if (this.selectors && this.selectors[field] != undefined) {
-      return this.selectors[field].fn();
+      return this.selectors[field].fn(this.getState());
```
johnjeng: `reselect` memoizes the return function based on the args so we have give a changing value here. Upside is that if `getState` doesn't change, no recomputes will happen.

src/smc-webapp/app-framework/Store.ts
```diff
@@ -135,7 +139,7 @@ export class Store<State> extends EventEmitter {
    * happens call the given callback.
    */
   wait<T>(opts: {
-    until: (store: Store<State>) => T; // waits until "until(store)" evaluates to something truthy
+    until: (store: any) => T; // waits until "until(store)" evaluates to something truthy
```
johnjeng: Typescript's inability to write `(store: this) => T` as a function argument where `this` is either a Store or subclass makes this awkward either way. Callers of `wait` should specify the store instance type.

williamstein:
Looking good. Note that this uses something from typescript 3.4 so I am updating our backend kucalc manage-* service to use newer typescript since that's needed before I can merge this.

johnjeng:
@williamstein Any idea how this code might break connection to projects? I checked if it's stale state but it turns out that my dev client actually doesn't connect to the project.

williamstein:
> @williamstein Any idea how this code might break connection to projects? I checked if it's stale state but it turns out that my dev client actually doesn't connect to the project.

I don't understand the question.  Is your code causing something to not work, e.g., projects don't start up?  What did you do?  What should have happened?  What actually happened...?

johnjeng:
> I don't understand the question. Is your code causing something to not work, e.g., projects don't start up? What did you do? What should have happened? What actually happened...?

Fixed it. The connection icon to the project would constantly be orange and I couldn't fetch directory listings. Project settings told me the project was running but the issue turned out to be that the project **wasn't** running. There was a Typescript error in the start up which I caught after looking at the project's `local hub.log`.

src/smc-webapp/app-framework/Store.ts
williamstein: Why isn't this an actual comment in the source code?  It seems pretty useful for people using the code...

PR #4328, opened 2020-01-17
expose the sage version in the UI

haraldschilly:
# Description
This exposes the current version of sage to the webapp. For now, it just shows up in the settings config panel, but later we can use it to show some additional info in the worksheet for 9.0.

# Testing Steps
I tested the combination of new project and new webapp, and also current project code (that version info is absent) and new webapp.

# Relevant Issues

Front end:


williamstein:
As far as I can tell, the Sage version does NOT get updated, even if the user changes it via sage_select.  I think that's a serious problem.  E.g., if I were to type "sage_select 9.0" to switch my default version of Sage, and I was aware of this configuration parameter, I would surely look in settings to confirm that the version changed. However, reading through the code, it looks to me that the only way to update the version would be to close and re-open the project in the frontend.   Even just restarting the project wouldn't change things.  Moreover, if anything in the frontend depends on this version, it'll be potentially wrong if sage_select is used.

In fact, if the user changes the project image does the configuration get updated?  It doesn't seem like it to me.

Some ideas for a solution to this problem:

- Change the project configuration information to be a JSON field of the projects table, so it is just part of the SyncTable, and have the project (1) periodically run tests for conf, and (2) update if things change.  This is kind of bad because it would be info that is set for *all* projects that the user is a collab on, which is a lot of info.  And configuration could get big, right?

- (Probably best approach:) Add a new synctable that is information about as single project.  The browser client opens it when it connects to the project, and it's stored entirely in RAM (not in the database at all).  We now fully support such synctables.  This could store tons of useful information about configuration, directory listings, top output, etc.    Again, for things like the sage version, check periodically (every 30 seconds or minute?) if it has changed, and if so the table gets updated.  

- Add a button labeled "Refresh" to the top of the "Features and configuration" panel in project settings, which makes it possible to recompute the project config when the user explicitly clicks it.  Basically the standard old school approach to this problem...  And it won't help at all if other parts of the UI depend on the sage version.

- Update config every time project settings is displayed.  Same problem as above.

williamstein:
If you just want a quick fix for now to get this in, a refresh button would at least address my current immediate concern, since the version isn't used anywhere else in the UI.  It's setting us up for problems later though.

haraldschilly:
> Even just restarting the project wouldn't change things

Restarting the project causes the config to be fetched again. You can try this by selecting the "old image", which misses a few features.

I agree that option 1 isn't great. I don't know how to do 2. Neither how to create such a table, nor how to use it without running into problems. I can look into adding a refresh button.

williamstein:
Is this ready for review?  I don't think you need to make restarting the sage worksheet server trigger a refresh in this PR.

haraldschilly:
well, that would be the only detail I would add. maybe tomorrow as a quick follow up PR. I've changed the labels…

PR #4337, opened 2020-01-21
bypass project level caching of configuration info

haraldschilly:
# Description

* `no_cache` parameter to bypass project level caching
* add the usual default paths to the PATH variable

# Testing Steps
I'll test this in "test"

# Relevant Issues

Front end:


williamstein:
Done?

haraldschilly:
I hope so, but I didn't test it yet.

williamstein:
Amusingly, I'm waiting a long time right now 100% because of this bug.  I'm trying to figure out the best way to install jupyter lab in my cocalc-kubernetes projects, and have to restart them after doing `pip install` because configuration is cached.  So this will be a welcome improvement.

haraldschilly:
just to spell out where I'm stuck:

when I click the "refresh" button, the request to get the info is supposed to be called with `no_cache=true` … but the project log says `primus-api request object {"cmd":"configuration","aspect":"main","no_cache":false}`

somewhere this flag is getting lost

williamstein:
> just to spell out where I'm stuck:
> 

Thanks!  I'm sure I can very easily fix this, and I'll take over from here.

src/smc-project/configuration.ts
```diff
@@ -63,12 +78,14 @@ async function sage_info(): Promise<{
   if (exists) {
     // We need the version of sage (--version runs quickly)
     try {
-      const info = (await exec("sage --version")).stdout.trim();
+      const info = (
+        await exec(`env PATH="${PATH}" sage --version`)
```
williamstein: Why would you do it this way, rather than using the env option to exec?   https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback

williamstein:
> somewhere this flag is getting lost

Answer -- you forgot to update smc-util/client.coffee.   **If that file were in typescript** you definitely wouldn't have made this mistake (typescript would have immediately given an error).

williamstein:
The relevant function client.coffee is:

```
    # This is async, so do "await smc_webapp.configuration(...project_id...)".
    configuration: (project_id, aspect) =>
        if not misc.is_valid_uuid_string(project_id) or typeof(name) != 'string'
            throw Error("project_id must be a valid uuid")
        return (await @project_websocket(project_id)).api.configuration(aspect)
```

What the heck is `        if not misc.is_valid_uuid_string(project_id) or typeof(name) != 'string'`, given that name isn't even defined here? I'll come up with something meaningful and deleted that name thing.

williamstein:
I made this issue about moving client.coffee to typescript: https://github.com/sagemathinc/cocalc/issues/4350

src/smc-project/configuration.ts
haraldschilly: hmm, I didn't think about that. Setting env var directly is better.

williamstein: I encourage you to do so in a followup PR.  I didn't make the change myself.

PR #4338, opened 2020-01-21
make default and max upgrade quotas configurable

haraldschilly:
# Description
* no (intended) changes for production
* for on-prem setups, this adds default and max quota fields to the site settings
* these settings are fed into the `quota()` function and if they exist, they overrule hardcoded defaults

# Testing Steps
1. running   `smc-hub$ SMC_DB_RESET=true SMC_TEST=true npx mocha test/kucalc/quota.coffee` → `24 passing`
1. I've built hub, manage and static for this branch in "test". 
    * I added some upgrades via my subscription and the upgrades did appear as they should. Hence I checked that there is no negative side effect. – which shouldn't be a surprise, because I didn't modify the tests and all of them still pass
    * similarly, I used one of the site licenses to upgrade internet & hosting of a new project. that worked as expected.

# Relevant Issues

Front end:


src/smc-hub/kucalc/quota.ts
```diff
@@ -85,10 +86,94 @@ interface Settings {
...
+  memory_limit: DEFAULT_QUOTAS.memory, // upper bound on RAM in MB
+  cpu_limit: DEFAULT_QUOTAS.cores, // upper bound on vCPU's
+  idle_timeout: DEFAULT_QUOTAS.mintime // minimum uptime
+} as const;
+
+function calc_default_quotas(site_settings?: SiteSettingsQuotas): Quota {
+  const q: Quota = Object.assign({}, BASE_QUOTAS);
+
+  // overwrite/set extras for any set default quota in the site setting
+  if (site_settings != null && site_settings?.default_quotas != null) {
```
williamstein: Shouldn't this be just:
```
if (site_settings?.default_quotas != null) {
```
-- or --
```
if (site_settings != null && site_settings.default_quotas != null) {
```

src/smc-hub/kucalc/quota.ts
```diff
@@ -85,10 +86,94 @@ interface Settings {
...
+    const dq = site_settings?.default_quotas;
```
williamstein: Aren't you in an if statement where you already checked that site_settings is defined?

williamstein: (If you're working around shortcomings in Typescript, I would say so explicitly in a comment...)

src/smc-hub/kucalc/quota.ts
```diff
@@ -211,9 +303,27 @@ exports.quota = function(
         contribs += factor * parse_num(num);
       }
     }
+    // if we have some overcommit ratio set, increase a request
+    if (site_settings?.default_quotas != null) {
+      const { mem_oc, cpu_oc } = site_settings?.default_quotas;
```
williamstein: The question mark in this line is not necessary.

src/smc-hub/kucalc/quota.ts
```diff
@@ -211,9 +303,27 @@ exports.quota = function(
...
+      if (
+        typeof cpu_oc == "number" &&
+        name == "cpu_request" &&
+        quota.cpu_limit != null
+      ) {
+        const oc_cpu = quota.cpu_limit / cpu_oc;
+        contribs = Math.max(contribs, oc_cpu - base);
+      } else if (
+        typeof mem_oc == "number" &&
```
williamstein: What are you REALLY testing when you write `typeof mem_oc == "number"`?  Are you really testing that at is defined at all?   So if mem_oc is accidentally a string (say), the it just gets silently ignored (not even a warning message)?  

I would probably replace `typeof mem_oc == "number"` with a test for weather `mem_oc` is true-ish, since that implies it is defined **and** nonzero.  If it were zero then you have a divide by 0 in line 321, which makes me nervous.

williamstein:
I have a number of comments.  They are all stylistic, so you can fix them in a follow up PR if you have the time.

src/smc-hub/kucalc/quota.ts
haraldschilly: I replaced this with a little wrapper to sanitize it. The point with being explicit about "number" is mainly typescript. It wants to make sure both sides of these arithmetic ops are numbers.

Issue #4345, opened 2020-01-26
invite email sent twice


haraldschilly:
when I send an email invite (e.g. to a course) to myself, I receive it twice. I suspect this happens for other email types as well. This must be a config problem, bug, or maybe we have to update the node library, I don't know.

haraldschilly:
After merging this PR above, I tested this for a new course file. I got two emails, despite only one invitations. What's weird: there are two hub-user-api hubs and both of them are sending? Or am I reading this wrong?

```
[prod3.test]> get pods | grep hub-user-a
hub-user-api-68f4d568c4-2xrnh                  1/1     Running   0          14h     192.168.219.86    kucalc-prod3-node-pnqj   <none>           <none>
hub-user-api-68f4d568c4-5b29d                  1/1     Running   0          14h     192.168.181.44    kucalc-prod3-node-pvcq   <none>           <none>
```

```
[prod3.test] > c log hub-user-api-68f4d568c4-5b29d | grep student481di2@schil
2020-01-28T11:48:10.220Z - debug: PostgreSQL._user_query(id=181ecc): "[{\"patches\":{\"string_id\":\"9af62934d83e0f2b6145a535c590685303b99cad\",\"time\":\"2020-01-28T11:48:10.041Z\",\"patch\":\"[1,[{\\\"email_address\\\":\\\"harald+student481di2@schil.ly\\\",\\\"table\\\"
:\\\"students\\\",\\\"student_id\\\":\\\"4632f47a-9b82-4a78-9604-537712e59100\\\"}]]\",\"user_id\":1,\"prev\":\"2020-01-28T11:47:59.781Z\",\"format\":1}}]"
2020-01-28T11:48:10.221Z - debug: PostgreSQL._user_query(id=e1e48d): "{\"patches\":{\"string_id\":\"9af62934d83e0f2b6145a535c590685303b99cad\",\"time\":\"2020-01-28T11:48:10.041Z\",\"patch\":\"[1,[{\\\"email_address\\\":\\\"harald+student481di2@schil.ly\\\",\\\"table\\\":
\\\"students\\\",\\\"student_id\\\":\\\"4632f47a-9b82-4a78-9604-537712e59100\\\"}]]\",\"user_id\":1,\"prev\":\"2020-01-28T11:47:59.781Z\",\"format\":1}}"
[...]

PR #4346, opened 2020-01-26
hub/email: update sendgrid api to v3 and stop using helpers

haraldschilly:
# Description
this is a modernization of sending emails and fixing #4345. the main difference is that this is more "low level" and therefore we can do everything the V3 api is talking about. the callback has a more rich response, which is discarded. in the future, it might be a good extension to check what's actually coming back from the service. I also added a silly small test file, which helped me with development.

# Testing Steps
in my cc-in-cc project, I created an account for an invite email, invited myself, added a new student, and did a fake password reset. in all cases I got one email.

# Relevant Issues

Front end:

### DIFF
--- a/src/smc-hub/email.ts
+++ b/src/smc-hub/email.ts
@@ -30,8 +30,8 @@ const os_path = require("path");
 const async = require("async");
 const winston = require("./winston-metrics").get_logger("email");
 
-// sendgrid API: https://sendgrid.com/docs/API_Reference/Web_API/mail.html
-const sendgrid = require("sendgrid");
+// sendgrid API v3: https://sendgrid.com/docs/API_Reference/Web_API/mail.html
+const sendgrid = require("@sendgrid/mail");
 
 const misc = require("smc-util/misc");
 const { defaults, required } = misc;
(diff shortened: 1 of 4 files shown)


williamstein:
I merged this and made it live on test.cocalc.com.   I couldn't get it to send any emails at all.  In all cases it seemed to silently fail. Hmm...

williamstein:
Checking the logs for test.cocalc shows:
```
2020-01-27T21:28:32.088Z - debug: send_email(to:hsy@sagemath.com) -- starting sendgrid client...
2020-01-27T21:28:32.089Z - debug: send_email(to:hsy@sagemath.com) -- email_server: explicitly disabled -- so pretend to always succeed for testing purposes
2020-01-27T21:28:32.089Z - debug: send_email(to:hsy@sagemath.com) -- successfully sent email
```
so I need to enable things there for testing first.

williamstein:
OK, I added sendgrid credentials and restarted the hubs on test.cocalc.com.   I got one password reset email as I should.  I then tried adding a non-account to a course via email and received **two identical emails**, so this change does not fix  #4345.

PR #4366, opened 2020-02-05
project/config: set PATH var via env option

haraldschilly:
# Description

trivial fix for https://github.com/sagemathinc/cocalc/issues/4364

# Testing Steps
1.
1.
1.

# Relevant Issues

Front end:

### DIFF
--- a/src/smc-project/configuration.ts
+++ b/src/smc-project/configuration.ts
@@ -79,7 +79,7 @@ async function sage_info(): Promise<{
     // We need the version of sage (--version runs quickly)
     try {
       const info = (
-        await exec(`env PATH="${PATH}" sage --version`)
+        await exec("sage --version", { env: { PATH } })
       ).stdout.trim();
       const m = info.match(/version ([\d+.]+[\d+])/);
       if (m != null) {


src/smc-project/configuration.ts
```diff
@@ -79,7 +79,7 @@ async function sage_info(): Promise<{
     // We need the version of sage (--version runs quickly)
     try {
       const info = (
-        await exec(`env PATH="${PATH}" sage --version`)
+        await exec("sage --version", { env: { PATH } })
```
williamstein: Oops, that sets the entire environment to have only one variable PATH, and deletes everything else from the environment.  This will cause subtle bugs.  Please rewrite this to take process.env, set the PATH in that, and then set env to that.

```
> require('child_process').exec("export", {env:{PATH:'foo'}}, console.log);0;
0
> null 'export PATH=\'foo\'\nexport PWD=\'/home/cocalc/kucalc/cluster2\'\n' '' 
```

williamstein:
I'm just going to merge this then immediately fix it since it is easy.

haraldschilly:
yes, good catch. ok, thank you!

Issue #4318, opened 2020-01-14
implement email address verification


williamstein:
I disabled email verification for cocalc, since there seemed to be  [tons of bugs](https://github.com/sagemathinc/cocalc/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+verification),  and we didn't really, *really* need it for anything.

This  needs to be fixed since we now have limited site license functionality, which depends on an email address. 

Regarding how it works, I think we should:

- automatically count email addresses that come from single sign on (google, etc.) profiles as valid.
- only have the validation flow for other emails.

Also, a lot of attempts to validate will fail due to the messages being labeled as spam, since we know that happens already with password reset messages.  That said, lots of other sites have working email verification... so how does it work without being broken?

haraldschilly:
>  tons of bugs

one of these is about a jupyter notebook? others apply to sending emails in general, not just email verification. I should probably edit the tickets. What I still don't know is why it was broken in the first place leading you to disable it and well, I also have to figure out how it was actually disabled.

haraldschilly:
>  a lot of attempts to validate will fail due to the messages being labeled as spam, since we know that happens already with password reset messages. 

the main problem is that we do not verify email addresses and therefore send emails to invalid addresses. that harms the overall quality score of emails we send and that's causing our provider to bundle us with lower quality senders and at the side of email services, downgrade us as well. other sites proactively verify emails addresses very early to avoid this.

besides that, I'm curious what data you have to claim that there are a lot of password resets being classified as spam. I don't have that information.

williamstein:
> besides that, I'm curious what data you have to claim that there are a lot of password resets being classified as spam. I don't have that information.

The only data I have is just that when I search zendesk/my email for "password reset", there are about 20 threads per month, sometimes complaining that they never received a reset message.  However, when we directly (using gmail or zendesk) interact with that same user they always get our messages. That's the only data I have; nothing else.

It could be that the entire solution to this problem is that you just re-enable email verification and it "just works".  It's possible.  

Also, I changed my site license code (and whole approach, in fact) to no longer involve the email adddress, so this is less relevant now.

williamstein:
> I'm curious what data you have to claim that there are a lot of password resets being classified as spam.

They tend to just role in.  Here's one from a user today: 

> "I never got the email from the forgot password exercise (I have no spam filter) and so I am sending this email."

[...]

Issue #2707, opened 2018-02-14
email verification in account preferences completely ignores errors from server


williamstein:
See:

![screenshot 2018-02-14 at 1 24 21 pm](https://user-images.githubusercontent.com/1276278/36228922-72660178-118a-11e8-91d9-b4a22fc172f5.png)

I see "email sent", but in the console log it says "TODO: error sending email verification: error sending email -- "unable to read the file '/home/user/cocalc/src/data/secrets/sendgrid', which is needed to send emails.""

williamstein:
Assigning to @haraldschilly

williamstein:
dup of #2505

Issue #2505, opened 2017-11-19
handle error case in email verification properly


williamstein:
The first test I did in my project isn't good (but low priority):

![screenshot 2017-11-19 at 9 21 12 am](https://user-images.githubusercontent.com/1276278/32993313-24b2b71e-cd0b-11e7-98f4-1798c43df5b9.png)

What I expect:
  - to see a clear error message reported to user, rather than a clear statement to the user that it worked (!).

Even better:
  - don't even show the email verification line at all if email isn't setup... I.e., there is a server_settings url that says what's available (e.g., commercial, etc.); add email to it and don't show this at all if it isn't available.   E.g., if/when I next update the personal CoCalc Docker  image, it is going to show this email verification thing, which is silly.

Another thing (lower priority): it seems a little weird that the only way to resend an email verification is refresh the browser.

PR #4368, opened 2020-02-05
re-enabled email verification and extend server settings configuration

haraldschilly:
# Description

so far, this PR does

* re-enable email verification (actually, it was never really disabled, just hidden)
* present a feedback page about what's going on, especially if there are errors
* add a setting to enable/disable email related aspects in the UI (+ make progress towards configuring email in general)

what's left to do is to go through the various issues around this and figure out what they are about and fix them

[this comment](https://github.com/sagemathinc/cocalc/issues/4318#issue-549253885) in #4318 claims these are tons of issues to take care of:

* [x] https://github.com/sagemathinc/cocalc/issues/2507 → already addressed or out of scope, closing
* [x] https://github.com/sagemathinc/cocalc/issues/3148 → I'm not able to reproduce this at all. 
  * [ ] double check this when deployed in "test"
* [x] https://github.com/sagemathinc/cocalc/issues/2943 → this is about a broken jupyter notebook, not in scope of this PR
* [ ] https://github.com/sagemathinc/cocalc/issues/2707 → dup of #2505 ??? In some sense this is addressed, because the verification section in the UI is gone if email sending is disabled. However, what remains is 
     * [ ] check if the reported error happens, it doesn't hurt to show it. 
     * [ ] don't send any (welcome or other) email if the email backend is disabled
* [x] https://github.com/sagemathinc/cocalc/issues/2506 → there is a small page saying if it is successful or if there was an error
* [x] https://github.com/sagemathinc/cocalc/issues/2505 → similar to #2707 … the main take away is that there is no proper email config. a proper solution would be to make it a setting, fully configurable how the hubs send emails. that's a good thing to have, for all contexts.
* [x] https://github.com/sagemathinc/cocalc/issues/2470  → mixes up several things, probably a too large meta ticket
     * creating an annoying banner, IMHO out of scope for re-enabling this
     * without an email address, there is no verification. that's already the case.
     * if you change your email address to one that's not verified, it says so. you have to verify it. changing the address back to a verified one doesn't require re-verification.
     * analyzing bouncing/blocked emails and showing this in the UI sounds interesting, but that's way more than just enabling them again. it would probably require a new service interacting with the api and ingesting the entire stream of email activities. sounds like a ticket for its own as well.
* [ ] https://github.com/sagemathinc/cocalc/issues/886  → this is about invite codes. I don't know what this has to do with verified emails. They are mentioned, but there is no reason why.

# Testing Steps
* main flow:
   1. existing user: click the verification button in the account settings → email should arrive → click the link → see it did work → redirected to cocalc. done.
   2. new user: same as above, except there is a link in the welcome email
* problematic flow: create a new account, but do not click on the verification link. instead, copy/paste it and
  1. modify the email in the url → no such account error
  2. modify the token → shouldn't work, and email is still not verified
* trivial case: already verified, but you click on the email a second time. it tells you that you're already verified.

# Relevant Issues

Front end:


src/smc-webapp/admin/site-settings.tsx
```diff
@@ -182,65 +198,202 @@ export class SiteSettings extends Component<{}, SiteSettingsState> {
...
 
-  render_row(name: string, value: string): Rendered | undefined {
-    if (value == null) {
-      value = site_settings_conf[name].default;
+  private render_default_row(name): Rendered | undefined {
+    const conf: Config = site_settings_conf[name];
+    return this.render_row(name, conf);
+  }
+
+  private render_extras_row(name): Rendered | undefined {
```
williamstein: @haraldschilly you don't need to do `Rendered| undefined` since Rendered **is** undefined | JSX.Element.  I.e., it allows for undefined already...

PR #4376, opened 2020-02-07
various email related enhancements (password reset emails via SMTP server, verification, ...)

haraldschilly:
# Description

this depends on #4368 for all the settings. it only makes sense to merge all of this at once. the hub should be updated first, it's all backwards compatible. besides just the password resets and all its configuration options, it also introduces some additional settings related to email sending. additional changes include:

* send out welcome email to anonymous users who sign up
* make email verification a setting – show/hide UI or tweak behavior accordingly
* if email verification is enabled, send out such an email when email address is changed (related to anonymous user's welcome emails)
* correctly route all types of emails, i.e. welcome, verification and mention are the first (default) category. password resets might be handled via a different route.
* makes some minor fixes to the welcome email
* show an error alert when verification email sending fails
* use more of site settings for sending emails (overriding hardcoded defaults)
* add the base url prefix to password verification emails
* this also cherry picks https://github.com/sagemathinc/cocalc/pull/4396

# Testing Steps
there are basically three interesting cases to test:
1. no settings (I deleted all of them for testing) and the result should be that sendgrid sending works as usual (via the secrets file)
2. default email sending is configured, either as sendgrid or smtp. all 6 types of emails should go through the configured service. I have to add that I only tested this for gmail's smtp server ... but in the end this should support all common types of smtp gateways. I checked these types:
    - welcome
    -  password reset
    -  verify email
    -  mentions
    - invite collaborator (existing user)
    - invite collaborator (not an existing user)
3. if "password reset" is set to be smtp (instead of default), route it differently.

### namespace "test"
* building this now, let's see how this works out
  * there is a weird stacktrace, which I never saw during dev and I also can't reproduce. hmm ... `at Function.mapToProps: TypeError: Cannot read property 'category' of undefined`.
  * It's happening in the line `if (type.category === "IMMUTABLE") {` in `app-framework.ts` …
  * ok, got it, it was `rtypes.bool` vs `rtypes.boolean` (and I'm sure I'll never remember which one is correct)
* today, I made a fresh build after merging. I tested it in "test" right after updating the hub, then also after updating static, and then again some tests after configuring all new settings and specifying a separate smtp sever for password resets. no problems.

# Notes

This is how emails looks like, sent via the smtp relay

Screenshot from 2020-02-07 17-33-47

and the header clearly says this this originates from google itself

```
Return-Path: <password@cocalc.com>
Received: from [127.0.0.1] (107.45.227.35.bc.googleusercontent.com. [35.227.45.107])
        by smtp-relay.gmail.com with ESMTPS id c189sm16181ywa.9.2020.02.07.08.33.11
        for <harald@schil.ly>
        (version=TLS1_2 cipher=ECDHE-RSA-AES128-GCM-SHA256 bits=128/128);
        Fri, 07 Feb 2020 08:33:11 -0800 (PST)
```

Front end:


williamstein:
> ok, got it, it was rtypes.bool vs rtypes.boolean (and I'm sure I'll never remember which one is correct)

rtypes.boolean would of curse make the most sense, because `typeof true` is "boolean" in Javascript.  So *of course* `rtypes.bool` is what is right. Ugh.

```

```

PR #4395, opened 2020-02-20
inconsistent use of "instead_of_change"

haraldschilly:
# Description

As described internally, there are crashes like

```
TypeError: cb is not a function
    at Object.instead_of_query (/cocalc/src/smc-util/db-schema/db-schema.ts:179:11)
    at async.series (/cocalc/src/smc-hub/postgres-user-queries.js:2347:32)
    at /cocalc/src/smc-hub/node_modules/async/lib/async.js:718:13
    at Immediate.iterate (/cocalc/src/smc-hub/node_modules/async/lib/async.js:262:13)
    at runCallback (timers.js:705:18)
    at tryOnImmediate (timers.js:676:5)
    at processImmediate (timers.js:658:5)
```

which I can trace back to `f3bb7ab2f9` which seems to trigger this as a side effect of that change. I don't know the overall reasoning behind this, needs a check. 

I also added more typing info for the callbacks, which seems to be a good idea.

# Testing Steps
* hub and static compiles and runs without errors … that's all I tested, though

# Relevant Issues

Front end:

### DIFF
--- a/src/smc-util/db-schema/db-schema.ts
+++ b/src/smc-util/db-schema/db-schema.ts
@@ -174,7 +174,7 @@ schema.blobs = {
   },
   user_query: {
     get: {
-      instead_of_query(database, obj, _account_id, cb) {
+      async instead_of_query(database, obj, cb): Promise<void> {
         if (obj.id == null) {
           cb("id must be specified");
           return;
@@ -207,7 +207,13 @@ schema.blobs = {
         blob: true,
         project_id: true
       },
-      instead_of_change(database, _old_value, new_val, _account_id, cb) {
+      async instead_of_change(
+        database,
+        _old_value,
+        new_val,
+        _account_id,
+        cb
+      ): Promise<void> {
         database.save_blob({
           uuid: new_val.id,
           blob: new_val.blob,
@@ -885,7 +891,7 @@ schema.projects = {
     "USING GIN (users)", // so get_collaborator_ids is fast
     "USING GIN (host jsonb_path_ops)", // so get_projects_on_compute_server is fast
     "lti_id",
-    "USING GIN (state)",  // so getting all running projects is fast (e.g. for site_license_usage_log... but also manage-state)
+    "USING GIN (state)" // so getting all running projects is fast (e.g. for site_license_usage_log... but also manage-state)
   ],
 
   user_query: {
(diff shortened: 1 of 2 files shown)


src/smc-util/db-schema/db-schema.ts
```diff
@@ -207,7 +207,13 @@ schema.blobs = {
...
         project_id: true
       },
-      instead_of_change(database, _old_value, new_val, _account_id, cb) {
+      async instead_of_change(
+        database,
+        _old_value,
+        new_val,
+        _account_id,
+        cb
+      ): Promise<void> {
```
haraldschilly: formatting happened

src/smc-util/db-schema/db-schema.ts
```diff
@@ -174,7 +174,7 @@ schema.blobs = {
   },
   user_query: {
     get: {
-      instead_of_query(database, obj, _account_id, cb) {
+      async instead_of_query(database, obj, cb): Promise<void> {
```
haraldschilly: removing `_account_id`, which is called as if it is the callback

src/smc-util/db-schema/types.ts
```diff
@@ -49,15 +49,14 @@ interface TableSchema<F extends Fields> {
       instead_of_query?: (
         database,
         obj,
-        instead_of_query,
```
haraldschilly: I have no idea why there is `instead_of_query` inside of `instead_of_query`

src/smc-util/db-schema/db-schema.ts
williamstein: Your PR definitely doesn't fix the bug.   I'm guessing this is totally my fault and was caused by my changing the inputs to instead_of_query, but not realizing it was used here using he older API.   The new api is `instead_of_query(database, opts, cb)`, where opts is 
```
            account_id : undefined
            project_id : undefined
            query      : required
            multi      : required
            options    : required
```
and that's not the same as "obj".   I'll merge your fix, then immediately make the additional changes that are needed.

src/smc-util/db-schema/types.ts
williamstein: Yes, what he had made little sense.  But also that obj should be the typescript version of this (due to my later changes):
```
opts:
            account_id : undefined
            project_id : undefined
            table      : required
            query      : required
            multi      : required
            options    : required
```

src/smc-util/db-schema/db-schema.ts
williamstein: Fortunately, I can't think of anywhere in our codebase that would use this code.  It must be somewhere, but I can't think of it...

PR #4401, opened 2020-02-20
webapp/project log: fix impendance between recording and showing downloaded files

haraldschilly:
# Description
the core bug is showing `event.files[0]` where files is actually a string. 

# Testing Steps
1. download one or more files, before testing this
2. switch to this, the entries should show up
3. do a single/multiple file download and this time, instead of the zip file name for multiple files, the files itself show up.

# Relevant Issues

Front end:

### DIFF
--- a/src/smc-webapp/project_actions.ts
+++ b/src/smc-webapp/project_actions.ts
@@ -2627,13 +2627,15 @@ export class ProjectActions extends Actions<ProjectStoreState> {
       auto: true,
       print: false,
       timeout: 45
-    });
+    } as { path: string; log: boolean | string[]; auto: boolean; print: boolean; timeout: number });
 
+    // log could also be an array of strings to record all the files that were downloaded in a zip file
     if (opts.log) {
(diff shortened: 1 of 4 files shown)


src/smc-webapp/project_actions.ts
```diff
@@ -2627,13 +2627,15 @@ export class ProjectActions extends Actions<ProjectStoreState> {
...
-    });
+    } as { path: string; log: boolean | string[]; auto: boolean; print: boolean; timeout: number });
 
+    // log could also be an array of strings to record all the files that were downloaded in a zip file
     if (opts.log) {
+      const files = Array.isArray(opts.log) ? opts.log : [opts.path];
       this.log({
         event: "file_action",
         action: "downloaded",
-        files: opts.path
```
haraldschilly: about this: the `FileActionEvent` says `path` and here it's `files`. I made this `files` everywhere, backwards compatible, and allows to use the `multi_file_links` function

PR #4414, opened 2020-02-25
mathjax 2.7.4 → 2.7.7 + minor updates

haraldschilly:
# Description

I considered updating to mathjax 3. That would require a few changes in the code, which is fine (promises instead of queues). However, it doesn't support all features, especially line breaks. Hence this is just a minor update + small fixes.

# Testing Steps
* I don't know if there is a way to force it to be used. If I enter a non-sensical formula like `$\foo$` it falls back to it and marks it red. So, I guess it's all fine.

# Relevant Issues

Front end:


williamstein:
> I don't know if there is a way to force it to be used. If I enter a non-sensical formula like $\foo$ it falls back to it and marks it red. So, I guess it's all fine.

I think katex doesn't support \newcommand which you can use to test rendering anything in mathjax.

williamstein:
> I don't know if there is a way to force it to be used. If I enter a non-sensical formula like $\foo$ it falls back to it and marks it red. So, I guess it's all fine.

I think katex doesn't support \newcommand which you can use to test rendering anything in mathjax.

PR #4422, opened 2020-02-26
Reorganize Settings Page

jengjob:
# Description
Reorganize project settings page.
All setting panels are rendered in a single column.
Menu links jump to corresponding settings panel.
newmenu
newmenudarkmode

# Testing Steps
1.
1.
1.

# Relevant Issues
Any CSS tweaks.
The way the menu jump currently works is by just having divs between each panel with an id attached to it and having links to them in the menu. Probably not the best way to do it but it works well and I cannot think of any severe downsides.
- [x] Run eslint on new and edited files

Front end:


haraldschilly:
hmm, there is already a conflict. what's the name of the main cocalc repo? `upstream`?

1. `git remote add upstream https://github.com/sagemathinc/cocalc.git` (or something like that)
2. `git fetch upstream`
3. `git merge upstream/master` → which should urge you to fix the conflicts, check them, and then **after** it works fine again, `git add [the body file]` and `git commit`

---

besides that I'm wondering what `import { Component, OnInit } from "angular/core";` is about.

williamstein:
- On a normal/default resolution display, the left column overlaps the right column.  Fix that.  I bet you run at a non-default resolution with small fonts, you lucky youngster with good eyesight :-).
 - Make it so that when you click on an item on the left, the right hand scrolls the section into view in such a way that it is maybe 30px from the top. Add some blank space to the very bottom so this even works with the last panel.

jengjob:
- Fixed overlapping in different resolutions
- Added small buffer to scroll when item is clicked

jengjob:
Scrolling issue fixed. Currently still breaks when the window height is VERY small(so small that you can only see one menu item) and you click on a menu item.

williamstein:
If you make your screen about 700 pixels wide, then a lot of the functionality of project settings is just not usable with this change applied.   E.g., selecting the image is cut off.   This same functionality is not impaired with the current live design.  With github their settings also work because you can **scroll horizontally**.

src/smc-webapp/project/settings/body.tsx
```diff
@@ -132,122 +132,302 @@ export const Body = rclass<ReactProps>(
...
-                <SSHPanel
-                  key="ssh-keys"
+      const scrollToTargetAdjusted = (id) => {
+        var element: HTMLElement | null = document.getElementById(id);
+        if (!element) {
+          console.log("document.getElementById(id) returned null");
+          return;
+        }
+        element.scrollIntoView(true);
+        //document.getElementById("thisone")!.scrollTop;
```
williamstein, line 142: Debugging code?  Did you mean to delete this?

williamstein:
CoCalc (and all modern -- post 2013? -- websites...) use something called "responsive design" to handle multiple devices and page widths.  You're not using that at all here.  Can you google responsive design, learn something about it, etc.   In particular, code like
```
      if (window.innerWidth > 700) {
        // LARGER SCREENS
        return (
          <div className="smc-vfill" style={{ flexDirection: "row" }}>
```
is a nonstarter for solving this sort of problem, as you can see by loading the page with the browser window wide, then just dragging the window to make it skinnier.  This approach (just render best on browser size) was done once long ago in the dark ages, but in modern times (i.e., during the last decade, but definitely today), CSS has come a LONG ways.   

To see responsive design in action, just load the current project settings page with the browser window wide then make the window skinnier and skinnier.  Like magic, observe that no content is ever cutoff off the side of the screen (content and layout resizes).  Then as you hit a threshold the panels are all reorganized into a single column.  This is all done with CSS (not javascript), and is a basic feature of react bootstrap (and any) grid layout, I think even in Bootstrap v1. 

I'm sorry for mentioning Github as a model, since that may have misled you.  The github website is really, reallly bad regarding working on anything but a fullsize desktop.   It has basically no mobile support at all (instead they have a completely separate hardly-functional mobile site).  This is probably because github is a server side rendered ruby-on-rails application, which was first written long long ago before rich client side javascript web applications were a thing.  Github was already huge and the dominant site at what it does back in 2012, and it hasn't really changed much in basic design since then.   Now that I look, I consider a lot of how it works to not be good enough for our basic cocalc requirements.    That said, github users are almost entirely software developers will relatively big screens, not mobile users.  With cocalc I really worry a lot more about mobile usage, since students often try to use cocalc on ipads or in classes, and at least I want the design of cocalc to be reasonable multi-device friendly, since that's sort of the standard for "modern" websites, at least starting in 2013...

PR #4429, opened 2020-03-04
hub: landing pages server plugin

haraldschilly:
# Description
the net impact of this is that the hub loads some plugin server code with the argument `--landing`. 
without that and any other adjustments, things should work just like before.
It also updates react dom + adds @types for it.

# Testing Steps
1. build the hub and check that it runs.

# Relevant Issues

Front end:


williamstein:
Good idea.

PR #4465, opened 2020-03-23
hub/auth: typescript auth, generalize strategies, extend UI

haraldschilly:
# Description
* typescript and cleanup auth.coffee
* generalize the setup routines, i.e. instead of copy/pasta for each strategy, there is data + a single general function.
* with that, there are also modules for ldap, oauth2, now, which can be configured with a given identifier. e.g. in the DB:
   * `strategy` = `acmecorp` ← must be unique and url friendly
   * `conf` = 
     ```{json}
     {"type": "oauth2next", 
     "clientID": "1234567890",
     "scope": ["email"],
     "clientSecret": ".....",
     "authorizationURL": "https://domain.edu/.../oauth2/authorize",  
     "userinfoURL" :"https://domain.edu/.../oauth2/userinfo",  
     "tokenURL":"https://domain.edu/.../oauth2/...extras.../access_token", 
     "login_info" : {"emails" :"emails[0].value"},
     "display": "Acme Inc",
     "icon": "https://domain.edu/pix/square.svg"}
     ```
* the last two conf entries are siphoned to the front end, that's why there is a version 2 endpoint for `/auth/strategies`! → display is show as the name, and (if set), there is also a square image that shows up. well, actually, both should be set to make this work well. the other settings depend on the strategy … `oauth2next` has many settings :chart_with_upwards_trend: 
* the `login_info` is a mapping from the returned profile to fields for cocalc. the one above is just one example, there are more. the string is standard dot-object notation.
* for github, it also queries public profile data to get the email address.
* there is no global `STRATEGIES` list in webapp – I don't miss it
* emails set via SSOs are automatically verified
* account creation token
  * if any SSO strategy is public, disable it to be set
  * on the signup page, move the account creation token field down to the email fields... since it only applied to them

# Testing Steps
1. in test, I did setup github oauth. I can link my account and also sign in now.
1. fancy configs similar to the `oauth2next` config above is already running in an on-prem setup in production
1. where I really have problems with is the sequence of checks and operations when there is already an account or when to merge. I believe it is the same as before (and e.g. it properly complains if there is already an account with email address X and you are signed out and try to SSO in with that email address) … but what are all the cases and what to check for?
   - what I explicitly tested is this pathological case
       1. I do not have my google account linked, and also my email isn't taken
       1. sign in using google
       1. unlink google
       1. sign in again → complains there is already an account with that email
       1. you have to do a password reset & sign in
       1. link with your google account
- sign in using github, and if it is set/shared, you'll have an email address

# deployment
1. first hub, it's backwards compatible with the strategies object
2. webapp

# Relevant Issues

Front end:


haraldschilly:
ok, I'm pretty happy with this now. I finally tested this in "test" ... the only detail I'm not happy with is are the icons in the account box. the ldap one is custom image, while the other ones are by some library. 

Screenshot from 2020-05-13 21-01-33

williamstein:
> ok, I'm pretty happy with this now

OK -- I'll wait until you change it to "needs review" before reviewing or merging.

haraldschilly:
ok, finally, got it. at least it's consistent … (tested in FF, and Chrome)

Screenshot from 2020-05-14 15-28-04

Screenshot from 2020-05-14 15-27-53

williamstein:
What are instructions and issues related to releasing this into production?

haraldschilly:
in the first comment I added a deployment section. it says that the hub must come first, it is backwards compatible. then the static webapp.

src/smc-util/db-schema/site-defaults.ts
```diff
@@ -319,4 +320,12 @@ export const site_settings_conf: SiteSettings = {
     valid: only_booleans,
     to_val: to_bool,
   },
+  email_signup: {
+    name: "Allow email signup",
+    desc:
+      "Users can sign up via email&password. Could be subject to an 'account creation token'.",
```
williamstein, line 326: put space here, so "email & password", not "email&password".

williamstein:
I'll plan to make live later today (or this evening).  I trivial have a comment about a string that admins see, so can you fix that?

haraldschilly:
yes sure

>

PR #4511, opened 2020-04-14
typescript trial banner and tweak text a little bit

haraldschilly:
# Description
* typescript the trial banner
* right now, it is always small, because we don't load the log just for the banner and it might be truncated as well. instead, this uses the project's creation time! it wasn't available, hence the change to the db schema and the requirement to upgrade hubs (if that happens, #4516 should be merged to make upgrading them worthwhile). the config right now makes the banner red after one month. in practice, I expect many banners to suddenly grow and turn red. 
* the phrasing itself is only mentioning implications to make this more tangible.
* this also gets rid of extra text, which we never displayed anyway. there is more text in the project settings.

# Testing Steps
basically, 4 cases of missing quotas, with screenshots
* internet, host

Screenshot from 2020-04-17 15-29-46

and after 1 month or more

Screenshot from 2020-04-17 15-25-48

* host

Screenshot from 2020-04-17 15-21-30

* internet

Screenshot from 2020-04-17 15-22-23

* none: well, project also renders fine :grin: 

# Relevant Issues

Front end:


williamstein:
Looks good.  I'll likely not merge this until I merge https://github.com/sagemathinc/cocalc/pull/4480 and do both at the same time (since they both require hub restarts).

PR #4514, opened 2020-04-15
fix DNS for "normal" password reset emails

haraldschilly:
# Description
a rather trivial case, where the hardcoded dns is used but it should be the one from the server config (if it is more than an empty string)

# Testing Steps
* I didn't test it yet

# Relevant Issues

Front end:


williamstein:
Seems equivalent to what we already have if not set, so LTGM.

Issue #4534, opened 2020-04-25
make it possible to load cocalc without it attempting to load anything from external network resources


williamstein:
Example applications:

> Using CoCalc in an airgapped environment via cocalc-docker. Without internet access we have noticed a lot of lag with the application due to its inability to reach CDN sources for page content. A similar application might be using cocalc-docker locally on a laptop in an airplane or remote area with no network access. 

Another application is improving predictability of page loads.   By relying on numerous external CDN sources makes the initial load time for CoCalc less predictable and more subject to potential external outages.  

Here are some external URL's CoCalc tries to reach on load:

- https://cdnjs.cloudflare.com/ajax/libs/mocha/5.2.0/mocha.min.css
- https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/3.3.7/css/bootstrap.min.css
- https://cdn.jsdelivr.net/npm/katex@0.11.1/dist/katex.min.css
- https://storage.googleapis.com/cocalc-extra/cocalc-screenshot-20200128-nq8.png
- https://fonts.googleapis.com/css?family=Roboto+Slab&display=swap

Mocha should be just removed -- it was only for some frontend testing in development, and we don't use that anymore.   Bootstrap, katex, the screenshot should be in webpack.  The fonts... I don't know.

williamstein:
REQUESTED BY: Kevin Bruner

williamstein:
This should now be completely fixed for the main cocalc application, due to work of @haraldschilly .  I've updated https://cocalc.com so it uses this, and I always updated the docker image.

Note that the *share* server on the docker image is still not done.  For some reason it grabs these resources from cocalc.com, rather than the docker image:

bootstrap.min.css , codemirror.css, katex.min.css. analytics.js


![image](https://user-images.githubusercontent.com/1276278/81618664-b2131e00-939c-11ea-9c77-5e51c979a2f3.png)

haraldschilly:
did you set the domain name? the reason the server needs it lies in production, where we cross the subdomain from share.cocalc.com to just cocalc.com … the server needs to know that, and that's why it uses a setting. So, in the settings, the domain name should be for all modes, which it currently isn't. 

In this case here, set kucalc mode to be no, and then it shows up.

![Screenshot from 2020-05-12 10-00-08](https://user-images.githubusercontent.com/207405/81657940-9bbaa000-9438-11ea-9b7c-ab5f925de552.png)

williamstein:
[...]

PR #4545, opened 2020-04-29
remove resources pointing to 3rd party servers

haraldschilly:
# Description
* [x] mocha css: dynamically inject it
* [x] infrastructure to serve 3rd party libs via version specific paths and just the plain name
* [x] packages: pin bootstrap to the latest 3.x version (we used it in the lockfile) and slightly update codemirror
* [x] function to create such css resource links, which are used by the share server, sagews printing and frame tree editor printing (printing code has customized CSS, I'm not touching it)
* [x] to make the share server load the files from the correct url, it needs to know about the server settings. with that, I also customized the displayed site name and got rid of reading the GA token.
* [x] rename res → resources as requested
* [x] getting rid of loadz, only makes sense in hub landing, not here
* [x] not changing Rmarkdown (I think that's too much of a detail). The workaround is to just set this in the yaml preamble:
```{markdown}
---
output:
  html_document:
    mathjax: /static/mathjax/MathJax.js?config=TeX-AMS-MML_HTMLorMML
---

$$\int_0^1 \exp(1-x^2)\;\mathrm{d}x$$
```

# Testing Steps
these aspects can be broken
* index page: title font, overall css
* webapp: sagews printing and printing the html preview in a frame editor. in both cases, check if formulas render, bootstrap style, etc.
* share: site name &  title in header, GA token, katex formula layout works, css. all three main types render: listing, a file and the user page

# Relevant Issues
#4534

Front end:


williamstein:
I wouldn't be surprised if this fixes some weird un-debuggable random reports from users that loading the cocalc site is flaky.  Maybe a third party resources is being blocked in their country or by their outgoing firewall, etc...

PR #4566, opened 2020-05-05
start supporting expiring PII info

haraldschilly:
# Description
* a new server setting `pii_retention`
* if not `never`,  a little helper to derive number of seconds from it + display it
* then uses this number to create an expiration date in the future (using that specific misc function)
* and sets an expiration date when inserting into the `central_log` , iff the type of event is of a certain kind or contains `email_address` or `ip_address`
* **bonus** if the server settings are set, the cache is reset

# Testing Steps
* What I did is to just restart the hub (nothing set) and check if it works → no expiration
* then set to 30 days → expires in the future
* I checked if these strings give a meaningful results via a quick test

# Relevant Issues

Front end:

### DIFF
--- a/src/smc-hub/postgres-server-queries.coffee
+++ b/src/smc-hub/postgres-server-queries.coffee
@@ -47,7 +47,8 @@ collab = require('./postgres/collab')
 
 SERVER_SETTINGS_EXTRAS = require("smc-util/db-schema/site-settings-extras").EXTRAS
 SITE_SETTINGS_CONF = require("smc-util/schema").site_settings_conf
-SERVER_SETTINGS_CACHE = require("expiring-lru-cache")({ size: 10, expiry: 10 * 1000 })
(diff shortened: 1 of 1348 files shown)


src/smc-hub/postgres-server-queries.coffee
```diff
@@ -221,6 +227,9 @@ exports.extend_PostgreSQL = (ext) -> class PostgreSQL extends ext
                         'value::TEXT' : (new Date()).toISOString()
                     conflict : 'name'
                     cb     : cb
+            # and clear the cache
+            (cb) ->
+                SERVER_SETTINGS_CACHE.reset()
```
williamstein, line 232: You forgot to call cb here.  And clearing the cache probably isn't async anyways, so this should just be done if there isn't an error right before line 234 below.   

It is WAY too hard to write callback-hell code, and I can't wait until we convert all this to async/await (+typescript), since that would have made **precisely** this bug simply not happen!

williamstein: I'll fix this, but please ack that you saw it.

Issue #4584, opened 2020-05-14
analytics.js not found on pressing the login button


arm2arm:
Hello, I am geeting following error on login page.
> GET https://my.domain.de/analytics.js?fqd=false net::ERR_ABORTED 404 (Not Found)

The server running with non commercial flag on.

Thanks,
Arman.

haraldschilly:
similar to #4583

williamstein:
I'm building a new Docker image with @haraldschilly 's fix right now, which if it all checks out I'll push to Dockerhub later today...

williamstein:
OK, after testing @haraldschilly 's fix, that doesn't fix this problem.  It is actually a mistake with the haproxy configuration on cocalc-docker...  which was easy to fix.  I've fixed that and am building a new image now.    https://github.com/sagemathinc/cocalc-docker/commit/867de7041a78394166c0779f4f18f330e5397ca2

PR #4585, opened 2020-05-14
only query analytics.js in a commerical setup

haraldschilly:
# Description
addresses #4584 

add a "commercial" env variable for the templates and introduce it in webpack and the open cocalc server

# Testing Steps
1.
1.
1.

# Relevant Issues

Front end:


williamstein:
I agree with what this PR will do, but also it is **really confusing*(* that the commercial-ui database setting does NOT control whether or not analytics.js is used.  I can see what that is and how it would be hard to do otherwise.  But still it is confusing, especially since you didn't document this inconsistency.

PR #4599, opened 2020-05-18
webapp/breadcrumb nav: don't forget history when going back home

haraldschilly:
# Description
this is a simple fix to not forget the nav history when going back to "home" and well, the depth calculation must be tweaked a little bit to avoid an off-by-one error.

# Testing Steps
navigate 2 or 3 dirs deep, switch between files and check how the path looks, also switch between files in paths that are not in a parent/child relationship.

# Relevant Issues

Front end:


williamstein:
Wow, super excited to see this!   When I was adding this functionality to some other places, I instantly wanted the home click to NOT forget the path, even though I never noticed this shortcoming before.  I tried for a moment to fix it, but failed.  It's like you read my mind by fixing this.

PR #4614, opened 2020-05-24
webapp: fix xsl syntax highlighting -- fixes #4613

### LINKED ISSUE #4613: XML highlighting for .xsl files
I do a decent amount of work with XSLT stylesheets, which are XML files usually with the extension `.xsl`. It'd be nice to have XML syntax highlighting for files with that extension. E.g. (from https://en.wikipedia.org/wiki/XSLT )

```
<?xml version="1.0" encoding="UTF-8"?>
<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0">
  <xsl:output method="xml" indent="yes"/>

  <xsl:template match="/persons">
    <root>
      <xsl:apply-templates select="person"/>
    </root>
  </xsl:template>

  <xsl:template match="person">
    <name username="{@username}">
      <xsl:value-of select="name" />
    </name>
  </xsl:template>

</xsl:stylesheet>
```

haraldschilly:
# Description
see #4613

# Testing Steps
1.
1.
1.

# Relevant Issues

Front end:

### DIFF
--- a/src/smc-webapp/file-associations.ts
+++ b/src/smc-webapp/file-associations.ts
@@ -93,7 +93,7 @@ const codemirror_associations: { [ext: string]: string } = {
   xml: "xml",
   cml: "xml", // http://www.xml-cml.org/, e.g. used by avogadro
   kml: "xml", // https://de.wikipedia.org/wiki/Keyhole_Markup_Language
-  xsl: "xsl",
+  xsl: "xml",
   v: "verilog",
   vh: "verilog",
   "": "text",


williamstein:
Weird...  I don't understand this fix.  Could you add a comment explaining why the natural solution (that you're deleting) is wrong...?

williamstein:
My guess -there is no codemirror xsl mode.  It simply doesn't exist. Maybe it did at one point and was removed...?

haraldschilly:
maybe, I don't know. it just didn't highlight anything and there is no xsl mode I can find online at the codemirror website.

williamstein:
OK, well this commentary here should be sufficient documentation...

Issue #4618, opened 2020-05-29
latex multifile doesn't correctly detect all files


haraldschilly:
the main problem is a latex file with many parts (included via `\include`), which doesn't build upon pressing the build button. 

to solve this ticket, a change in an included file or subfile and then clicking on build should trigger a build.

I don't have a public example yet.

haraldschilly:
I tried to figure out what's going on, but I have no real clue. The first thing that surprises me that the logfile is parsed for filenames. I bet that can break randomly. Back when we discussed this I was proposing to use latexmk to list the dependencies, since I'm confident it can do this well.

williamstein:
@haraldschilly fixed this -- yeah!  https://github.com/sagemathinc/cocalc/pull/4619

PR #4619, opened 2020-05-29
webapp/latex: make detecting of multifiles more robust 

haraldschilly:
# Description
this uses latexmk to detect all the files making up the project. this is more robust than reading the log. e.g. there is a large example (internally only) which breaks somewhere and only some files show up. this is negative side effects regarding saving and building the project.

# Testing Steps
* for multifile documents, all filename show up in the dropdown list.
* an important case are existing documents. for that,
  *  I did first go back to `master` → reload page → select build command → no `-deps` → closed tex file
  *  then I checked the `.*.syncdb` file that it is saved (as an array)
  * checked out this branch  → build → reload page
  *  opened the tex file and then it inserted `-deps`. saving the file, I confirmed it also changed the syncdb file. (this didn't work the first time around, not sure why)
  * then I manually edited the build command and removed `-deps` (saved as a string in syncdb file)
  * reload page → reopen tex file  → and `-deps` is there again.

for the second case I added an additional sanitization function.

# Relevant Issues
#4618

Front end:

### DIFF
--- a/src/smc-webapp/frame-editors/latex-editor/actions.ts
+++ b/src/smc-webapp/frame-editors/latex-editor/actions.ts
@@ -274,7 +274,7 @@ export class Actions extends BaseActions<LatexEditorState> {
           // set; however, it's possible it isn't in case this is
           // an old document that had the build_command set before
           // we implemented output directory support.
-          const build_command: List<string> = this.ensure_output_directory(cmd);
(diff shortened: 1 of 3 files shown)


haraldschilly:
so, I think I got it. my main problem was that at some point I saved the map {`build_command`: ...}` as the value of the `build_command` in the syncdb file. that should either be an array or string, not a map.  I started over with a new file.

PR #4635, opened 2020-06-04
launching custom software images

haraldschilly:
# Description
* remove the unused binder endpoint. this must be done differently.
* similar to the `share/...` launch url, support custom software images.
* if there is already a project with that underlying image, use the most recent one
* wishlist
  * [x] combine this with copying over files from the share server
  * [x] support anonymous usage

# Testing Steps
* well, regarding custom images in cc-in-cc dev, I  pick one of the random ones for testing. 
* testing what happens if you do the following for being signed in or not signed in:
* in the case of launching a share, it could be one for the default software, or with an image. to simplify this, I just test creating a share as not being signed in and anonymous (i.e. via share server link) in both cases
* overall, there are also safeguards if custom software image names aren't know or if random things are fed into the launch URL 

| signed in | anonymous | launch/csi | launch/share      | tested  |
| ---------- |  ---------- | ----------- | ------------ | ------- |
| yes | [not set] | [not set] | [not set] | :green_circle:   |
| yes | true | a valid id | [not set] |  :green_circle: |
| yes | true | [not set] |  share link, default project | :green_circle:  |
| yes | true | [not set] |   share link, custom software image |  :green_circle:  |
| yes | [not set]  | a valid id | [not set] | :green_circle: |
| no  | [not set] | [not set] | [not set] | :green_circle: |
| no  | true | a valid id | [not set] | :green_circle: |
| no  | true | [not set]|  share link, default project | :green_circle: |
| no | true | [not set] |   share link, custom software image |  :green_circle: |
| no  | [not set]  | a valid id | [not set] | :green_circle: ¹ |
| no  | [not set]  | [not set] | a valid share link | :green_circle: ¹ |

¹ the last two show the landing page. there are 3 cases, where in each of them the launch action happens
* existing account
* create new account
* anonymous

# Relevant Issues

Front end:


williamstein:
> if there is already a project with that underlying image, use the most recent one

I wonder if this was a good choice by me.  Maybe one should always get a new project... at least until there is an option to select the target project (or create a new one).    New projects are cheap and clean.    Anyway, just a thought, and you can ignore it for this PR...

williamstein:
1. Do I *have* to restart all the hubs for that one config change?
2. What's your assessment of the _danger_ of releasing this, assuming normal account creation and anonymous=true do both work?

src/smc-webapp/launch/custom-image.ts
```diff
@@ -3,76 +3,140 @@
...
-    .getStore("projects")
-    .wait_until_project_created(token, 30, (err, project_id) => {
-      if (err != null) {
-        console.error(`Error creating project -- ${err}`);
-      } else {
-        actions.apply_default_upgrades({ project_id });
-        actions.open_project({ project_id, switch_to: true });
-      }
+  // returns project_id or if there is a problem, undefined
+  async launch(): string | undefined {
```
williamstein, line 34: @haraldschilly I want to point out that you totally got bit by the ts-fork checker bug here!!   This is clearly invalid code, since an async function can only return a Promise:

image

It silently failed to be reported due to ts-forker breaking...  There will be many typescript errors in your code, probably.

I think I finally fixed this problem using a different webpack config option.

williamstein: I'll fix all these for you.

PR #4649, opened 2020-06-09
project: support custom environment vars

haraldschilly:
# Description
This is pure backend behind-the-scenes code. All it does is to inject a `{string:string}` map from the database for a specific project into the local hub. It's done in a way that the project itself is not affected (will always start, nothing funny will happen) while all spawned processes will have these env vars. What's obviously missing is some UI to set that field in the DB…

# Testing Steps
1. base case is clear: if nothing is set, projects still behave as usual
2. the fun starts once something like that is in the database:
```
update projects set env = '{"FOO": "THIS IS A TEST", "BAR": "123"}'::JSONB where project_id ='[project id]';
```

for my `cc-in-cc` project it works:
```
~$ echo $FOO 
THIS IS A TEST
```
because there is a special function `dev_env` in `smc_compute.py` passing it on. A similar line is there for the usual usage (docker?) and I also prepared the template file for cocalc kubernetes. All other code is just to tie everything together.

Regarding kubernetes, there is a branch `project-env-4523` in kucalc, which I've tested as well. `manage` must be rebuilt with `--no-cache`. Changes to our other kubernetes setups are very similar (I didn't do them):
1. pass on the env variable in the project `Dockerfile`, at the bottom where project init.sh is called.
2. the project's `init.sh` file must set `export COCALC_EXTRA_ENV="$2"`

Screenshot from 2020-06-10 19-32-43

# Deployment

Hub first (!) to update the database. Then manage and project.

# Relevant Issues

Front end:

### DIFF
--- a/src/smc-hub/compute-client.coffee
+++ b/src/smc-hub/compute-client.coffee
@@ -1333,6 +1333,8 @@ class ProjectClient extends EventEmitter
                     else
                         opts.cb()
             return
+        locals =
+            env: {}
         async.series([
             (cb) =>
                 if opts.set_quotas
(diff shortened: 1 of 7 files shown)


williamstein:
So glad you're doing this!  (since it needs to get done, and I just don't feel like doing it at all...)

Issue #4660, opened 2020-06-15
don't auto-open converted docx files


haraldschilly:
when I open a docx file, some sort of "information editor" opens up, which tries to tell me something about what's going on.

the problem is I can't read it. every time I try to open that tab it opens up a converted text file ....

it would be much  better if the conversion happens optionally when I click a button. then, I also have a chance to read the remainder of the text.

PR #4673, opened 2020-06-17
webapp/generic: do not open docx automatically

haraldschilly:
# Description
this is #4660 + some general cleanup/modernization

# Testing Steps
Screenshot from 2020-06-17 11-42-43

Screenshot from 2020-06-17 11-42-35

# Relevant Issues

Front end:

### DIFF
--- a/src/smc-webapp/editor-data/generic.tsx
+++ b/src/smc-webapp/editor-data/generic.tsx
@@ -7,15 +7,13 @@
 //
 // See https://github.com/sagemathinc/cocalc/issues/2462
 
-import { React, Component, Rendered } from "../app-framework";
+import { React, Rendered, useActions } from "../app-framework";
 const { register_file_editor } = require("../project_file");
-
-import { Well } from "react-bootstrap";
-
 import { Markdown } from "../r_misc";
 import { webapp_client } from "../webapp-client";
-
 import { keys, filename_extension } from "smc-util/misc2";
+import { COLORS } from "../../smc-util/theme";
+import { Button, Well } from "../antd-bootstrap";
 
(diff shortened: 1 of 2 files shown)


src/smc-webapp/editor-data/generic.tsx
```diff
@@ -67,44 +65,58 @@ interface Props {
   path: string;
 }
 
-class DataGeneric extends Component<Props, {}> {
-  render_hint(): Rendered {
-    const ext = filename_extension(this.props.path);
+function DataGeneric(props: Props) {
```
williamstein, line 68: I read a lot about this, and clearly the best way with typescript to make a functional component is:
```
const DataGeneric : React.FC<Props> = (props) => {
```

This correctly checks the types both of the inputs *and* the outputs and makes it very clear that you're defining a React Functional Component (in contract, if I see `function DataGeneric(props: Props) {` I don't even know it is a functional component).  See https://levelup.gitconnected.com/using-react-functional-components-with-hooks-in-typescript-6d61075edddc and also the many components I've converted in the last few weeks.

williamstein:
I also don't like putting the memo here:
```
register_file_editor({
  ext: keys(INFO),
  icon: "question",
  component: React.memo(DataGeneric),
});
```
Why?  Because whether or not to React.memo this component is a feature of the component not how it is being used.  If somebody where to export the component from this file and not notice this memo'ing at the bottom, they would miss that.  Also, when debugging it would be very easy to think the component is not React.memo'd because of neglecting to read the very bottom.  Please don't do this for the sake of future-debugging.  I'll change this.

Issue #4434, opened 2020-03-05
firefox: "not saving" banner in save button broken


haraldschilly:
this is how the red "not saved" banner in the save button looks like

Firefox 73.0.1 on Linux

![Screenshot from 2020-03-05 10-53-11](https://user-images.githubusercontent.com/207405/75969723-a5bba200-5ecf-11ea-8a76-f13260551df1.png)

haraldschilly:
the fix worked, but now it's line-breaking here (in edge cases)

![Screenshot from 2020-03-27 20-27-28](https://user-images.githubusercontent.com/207405/77792938-8bf22280-7069-11ea-97af-d0ba25888d84.png)

PR #4676, opened 2020-06-17
tweak save button

haraldschilly:
# Description
* this started by fixing #4434, but then I looked into this "react-logic" behind it and did some tests. so, I don't like how this is done at all.
* I ended up lifting up the state of showing the uncommitted information to the store. otherwise it resets each time one switches frames.

# Testing Steps
* I tested this by having two frames in a textfile, stopped my dev hub, and then tried to save a change. error shows up, and resolves once I start my dev hub again.
* The button also appears in chat, so, same there. Chat has no changes, because it is backwards compatible.

# Relevant Issues
#4434

Front end:

### DIFF
--- a/src/smc-webapp/frame-editors/frame-tree/save-button.tsx
+++ b/src/smc-webapp/frame-editors/frame-tree/save-button.tsx
@@ -4,11 +4,8 @@
  */
 
 const { Button } = require("react-bootstrap");
-
 import { Icon, VisibleMDLG } from "smc-webapp/r_misc";
-
-import { React, Rendered, Component } from "../../app-framework";
-
+import { React } from "../../app-framework";
 import { UncommittedChanges } from "../../r_misc";
 
 interface Props {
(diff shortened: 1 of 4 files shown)


src/smc-webapp/frame-editors/frame-tree/save-button.tsx
```diff
@@ -20,51 +17,63 @@ interface Props {
...
+  set_show_uncommitted_changes?: Function;
 }
 
-export class SaveButton extends Component<Props, {}> {
-  render(): Rendered {
-    const disabled: boolean =
-      !this.props.has_unsaved_changes ||
-      !!this.props.read_only ||
-      !!this.props.is_public;
+const SaveButtonFC: React.FC<Props> = (props: Props) => {
```
williamstein, line 24: Please don't name functional components with FC at the end.  They are clearly functional components because of React.FC, and this conflicts with what everybody else does.  I think it's also probably better to just do:
```
const SaveButtonFC: React.FC<Props> = React.memo((props: Props) => { ...
```
so it's crystal clear right when you're reading the code that memo is being used.

williamstein:
I merged this and fixed the minor style issue I had.  Then I found a stale-state bug that's pretty bad while testing this, but it's *NOT* caused by this.  Will report and fix soon:  https://github.com/sagemathinc/cocalc/issues/4685

Issue #4597, opened 2020-05-16
Sign out button in Account tab vertically crowds horizontal menu for some widths


DrXyzzy:
Seen with recent Chrome and Firefox

1. Open Account screen.
1. Shrink width of browser window so `Sign out...` button is over `Support` or other buttons.

![signout-too-close](https://user-images.githubusercontent.com/528072/82108588-0861c300-96f5-11ea-8796-95b968934c81.png)

williamstein:
@haraldschilly  #4679 fixes this.  At the same time, it breaks the problem that this button solves if the screen is somewhat narrow...  Maybe a new ticket should be opened about that.

PR #4679, opened 2020-06-18
 move signout button into the row of tabs

haraldschilly:
# Description
this fixes the problem in #4597 but at the same time it's not much better. well, this probably needs another iteration after this one here.

# Testing Steps
1.
1.
1.

# Relevant Issues

Front end:

### DIFF
--- a/src/smc-webapp/account/account-page.tsx
+++ b/src/smc-webapp/account/account-page.tsx
@@ -21,15 +21,10 @@ import {
   Component,
 } from "../app-framework";
 import { Col, Row, Tab, Tabs } from "../antd-bootstrap";
-
 import { LandingPage } from "../landing-page/landing-page";
-
 import { AccountPreferences } from "./account-preferences";
-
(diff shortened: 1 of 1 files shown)


williamstein:
Did you test this on mobile / skinny screens?  Since antd tabs I think turn into some scroll thing on mobile, it might make this **supposed to be always very visible** "sign out" button into a hidden button which defeats the whole point.   I didn't test yet.  I'm just asking.

williamstein:
> Did you test this on mobile / skinny screens? Since antd tabs I think turn into some scroll thing on mobile, it might make this **supposed to be always very visible** "sign out" button into a hidden button which defeats the whole point. I didn't test yet. I'm just asking.

I tested and indeed this completely hides the sign out button on mobile.  It's even hidden on desktop if your window isn't wide enough:

image

I'm merging this anyways because it looks better.  However, you'll probably want to redo this differently...

Issue #4438, opened 2020-03-08
firefox: magn. glass +/- doesn't work in frame editors


haraldschilly:
No matter if I try .md or .Rmd files: the magnification glasses to zoom in or out in the preview frame do not work. 

maybe 

```
This page uses the non standard property “zoom”. Consider using calc() in the relevant property values, or using “transform” along with “transform-origin: 0 0”.
```

williamstein:
works now in firefox.

PR #4681, opened 2020-06-19
html viewer fix +/- zoom button

haraldschilly:
# Description
* fix +/- zoom in iframe viewer. it uses IE's "zoom" which is nonstandard. [reference](https://caniuse.com/#feat=css-zoom)  and see  #4438
* respect the default font size. there was a hardcoded number 16, which makes no sense. i.e. the first time you click + or -, it should be more or less based on your default font size -- not 16.

# Testing Steps
all I did was to make rmd file and +/- zoon in the iframe html viewer

# Relevant Issues
 #4438

Front end:


williamstein:
Looks good. I think I might have made this zoom mistake elsewhere, e.g., maybe in the PDF viewer.  Not sure.

Issue #4247, opened 2019-11-22
add title to sign-in page


haraldschilly:
reported by a user

> Some password managers suggest the sign in credentials in a website based upon the HTML title. The page https://cocalc.com/app?session= has an empty HTML title.

PR #4689, opened 2020-06-22
app landing page: make sure a title is set 

haraldschilly:
# Description

it might be the case that #4247 is already fixed ... in any case, this sets the title of the initial landing page to name/description.

# Testing Steps
1.
1.
1.

# Relevant Issues

Front end:

### DIFF
--- a/src/webapp-lib/app-init.ts
+++ b/src/webapp-lib/app-init.ts
@@ -22,7 +22,8 @@ function script_error() {
 function style() {
   const NAME = CUSTOMIZE.site_name || "CoCalc";
   HELP_EMAIL = CUSTOMIZE.help_email || "help@cocalc.com";
-  document.title = NAME;
+  const SITE_DESCR = CUSTOMIZE.site_description || "";
+  document.title = `${NAME} – ${SITE_DESCR}`;
   const msg = document.getElementById("cc-message");
   if (msg == null) {
     // happens when loading is very quick and message is already removed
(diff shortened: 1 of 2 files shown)


src/webapp-lib/app-init.ts
```diff
@@ -22,7 +22,8 @@ function script_error() {
 function style() {
   const NAME = CUSTOMIZE.site_name || "CoCalc";
   HELP_EMAIL = CUSTOMIZE.help_email || "help@cocalc.com";
-  document.title = NAME;
+  const SITE_DESCR = CUSTOMIZE.site_description || "";
```
williamstein, line 25: I know it is equivalent, but doesn't it make more sense to write:
```
CUSTOMIZE.site_description ?? "";
```

This is much clearer to me since it means "if CUSTOMIZE.site_description is null or undefined evaluate to the empty string".  What you wrote means "if CUSTOMIZE.site_description is falsey evaluate to the empty string".     I prefer not to depend on "falsey" in javascript if I can avoid it since it has weird semantics.   Also it just feels funny because the empty string is falsey.

williamstein: Answer to my question: it messes up the codemirror syntax highlighting!

image

Well that's just depressing.

williamstein: And yes this is broken in the latest version of codemirror.

Issue #4696, opened 2020-06-24
jupyter: can't generate student version


haraldschilly:
it seems all it creates is a subdirectory and then tries to open a file that doesn't exist

PR #4697, opened 2020-06-24
webapp/jupyter/nbgrader: fix broken assignment  

haraldschilly:
# Description
I tried to address #4696 but I have no idea what's really going on. 
* I'm 100% sure the while loop in line 158 never stops, that's what debugging shows clearly
* The remainder is a problem with closing the file and opening it again. E.g. it just hangs at `await once(actions.jupyter_actions.syncdb, "ready");` indefinitiely.

# Testing Steps
* fresh: students dir + file does NOT exist → new notebook should appear
* update: it exists → notebook should get modified cells.

With the fix below it almost works, but updating the notebook seems to be flaky. E.g. I get errors about cells being unmodifiable. However, running it twice seems to work. So, maybe a step towards fixing this, but this change here is not enough.

# Relevant Issues

Front end:


williamstein:
I better merge yours as is since you made this live and I'm about to build some other bugfix I made to master!!!!

PR #4812, opened 2020-08-25
half-baked whitelabeing idea

haraldschilly:
this is like a 50% MVP. via the `whitelabeling` table, someone can customize the webapp theme. what's missing is the actual landing page (my idea would be to not use the special landing pages, but instead the simpler cocalc landing page, like for docker) and also customizations regarding email sending (i.e. what's going on in the hub, server side). hubs must be aware of the front-end theme, which means requests like sending an email have to contain the "dns" value, which is then used to derive the specific configuration. so, there is a path forward.

# Testing Steps
well, the most important aspect is that the usual case doesn't degrade. the key is to check that `smc-hub/hub_http_server.coffee`'s `/configuration` endpoint exposes `server_settings.pub` –  which is `this.data.pub` in `smc-hub/webapp-configuration.ts` – in those cases where the is just the basic domain name. the `.pub` field should also be updated automatically as before (I still have to actually test this, though)

well, I checked it in my cc-in-cc dev project, that's all. I think all this is safe. Since it requires a hub restart and nothing urgent is fixed, there is no need to rush it.

# deployment

update hub-websocket, which will also introduce a new table in the DB


src/smc-util/db-schema/site-whitelabeling.ts
```diff
@@ -0,0 +1,86 @@
+/*
+ *  This file is part of CoCalc: Copyright © 2020 Sagemath, Inc.
+ *  License: AGPLv3 s.t. "Commons Clause" – see LICENSE.md for details
+ */
+
+// This is for the commerical setup only!
```
williamstein, line 6: "commerical" --> "commercial".

src/smc-hub/webapp-configuration.ts
```diff
@@ -0,0 +1,88 @@
+/*
+ *  This file is part of CoCalc: Copyright © 2020 Sagemath, Inc.
+ *  License: AGPLv3 s.t. "Commons Clause" – see LICENSE.md for details
+ */
+
+// this manages the webapp's configuration based on the hostname (allows whitelabeling)
+
+import { parseDomain, ParseResultType } from "parse-domain";
```
williamstein, line 8: really need all these commented out imports?

williamstein:
Changing to needs work, just because I have several unanswered questions above, not because anything is broken.

src/smc-hub/webapp-configuration.ts
haraldschilly: No

haraldschilly:
I addressed your comments and well, it's far from being done. I started a bit about server side settings as well,  but  to finish this I'll make another PR with more refactoring. 

I'm building it now for testing…

PR #4819, opened 2020-08-28
hub/auth: use cookie to transport token to UI

haraldschilly:
# Description
this uses a short living session cookie to store the pw reset token before the UI loads up. also fixes removing the cookie in the client in chrome.

# Testing Steps
* cc-in-cc works
* works fine in "test" … it's not necessary to update static, just the hubs is fine

# Relevant Issues

Front end:


williamstein:
Code looks fine.  What's the motivation?

PR #4831, opened 2020-09-02
webapp/software update: reorder some lines to avoid violating hook rules

haraldschilly:
# Description

well, this should help

# Testing Steps
1.
1.
1.

# Relevant Issues

Front end:

### DIFF
--- a/src/smc-webapp/project/page/software-env-upgrade.tsx
+++ b/src/smc-webapp/project/page/software-env-upgrade.tsx
@@ -52,25 +52,17 @@ function useComputeImage(project_id) {
 export const SoftwareEnvUpgrade: React.FC<{ project_id: string }> = ({
   project_id,
 }) => {
-  // if we're outside cocalc.com, this is not applicable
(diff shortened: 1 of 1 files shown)


williamstein:
This code is still not 100% right because it has a conditional return statement before a bunch of hooks.  Please read this blog post: https://ohseemedia.com/posts/rendered-fewer-hooks-than-expected-reactjs-error/

The rules of hooks think on the react page isn't sufficiently clear.  That post is clear, where it says:

> "It also means that you cannot use hooks after your conditional statement, if your conditional statement returns early, ie:"

If you think about how hooks have to be actually implemented in Javascript, the real "rule of hooks" should be "every hook must get called every time in the exact same order".  Basically this assumption is critical to make it possible to actual implement hooks at all.    

The situation of your component, where you want to exit early (and not involve a bunch of other hooks at all) if some condition is satisfied, is solved by breaking your one component up into two components, where one component renders the other component.    If that doesn't make sense, let me know.

williamstein:
I'll merge this then fix it properly.

PR #4834, opened 2020-09-04
webapp: loading compute_images in minimal mode 

haraldschilly:
# Description

I was testing embedding cocalc. I noticed that when all projects should close, an exception appears. It's somehow happening with those react store hooks and the project listing. I don't know how to really fix this, and well, since this might happen elsewhere as well, I enabled pulling in the compute image data for the embedded mode. (or is there a way to say that a certain store might not always exist?)

The call triggering this below is `page_actions.close_project_tab(project_id);` for the last open project.

Screenshot from 2020-09-04 12-49-06

# Testing Steps
1.
1.
1.

# Relevant Issues

Front end:

### DIFF
--- a/src/smc-webapp/custom-software/init.ts
+++ b/src/smc-webapp/custom-software/init.ts
@@ -6,7 +6,6 @@
 // Manage DB <-> UI integration of available *custom* compute images
 // TODO: also get rid of hardcoded official software images
 
-import { COCALC_MINIMAL } from "../fullscreen";
 const { redux, Store, Actions, Table } = require("../app-framework");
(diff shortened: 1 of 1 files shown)


williamstein:
>  (or is there a way to say that a certain store might not always exist?)

Either (1) Don't access that store with a hook (just use redux.getStore), or (2) write a new hook that gets the store or null and is more painful to use from typescript.  The reason that we check if the store exists when using it via a hook is because of typescript.  If we allowed null, then there would be a massive amount of (mostly) unnecessary null checks all over any code that used the useStore hook.

PR #4836, opened 2020-09-05
upgrading to typescript  4

haraldschilly:
# Description

I started to port our codebase to typescript 4 + updating various packages. there is a lot of work to do … as far as I can tell all due to a stricter check `TS2790` when deleting an attribute. I wonder if there is a better way, but since this is so widespread, there might not be a simple solution. … I also don't want to disable this check, it makes a lot of sense to keep it.

## testing

details that are touched which should  be tested

- [ ] loading full project log
- [ ] x11
- [ ] adding `string_cols=[]` somewhere related to chat, and a bunch of other small changes in sync
- [ ] jupyter
- [ ] project related, smc-project/sync was touched. path watcher, listings, etc.

## problems

- [x] right now, I see there are no project configuration settings coming back. very weird.
  - well, this one was really strange. on the project's side, the dictionary for the configuration had entries like `{foo: await foo(), ...}` where the expected entry should be a configuration map. in reality, it just returned `{foo: true}`. that's why part of the configuration side in the project is refactored to call the promise functions (with a different name) explicitly. so, this is either a change in functionality or a clash with the names.
- [ ] first time opening an md file, preview just says "Loading...". editing the file makes it rendering. first time around there must be something missing …
- [ ] jupyter broken
  - clicking "run" logs `NotebookFrameActions(frame_id='44883cc3').command confirm restart kernel` ?
  - it seems to be stuck properly loading up. maybe it's the same root cause as the problem with MD files.

## deployment

there is nothing to take care of, because no functionality changes


src/smc-hub/postgres/changefeed-query.ts
```diff
@@ -53,11 +53,11 @@ function key(obj: { [key: string]: any }): string {
 type State = "ready" | "closed";
 
 class ThrottledTableQueue extends EventEmitter {
-  private table: string;
-  private queue: { [key: string]: TableQuery } = {};
-  private db: PostgreSQL;
-  private process_timer: any;
-  private interval_ms: number;
+  private table?: string;
```
williamstein, line 56: No!  It would be better not to.delete these on.close than have to
put unnecessary null checks everywhere.
I'm -1 to this.  I only delete things in close for garbage collection
reasons.  There must be a better way than null.checks everywhere.
Maybe one assertion at the top of all functions?

haraldschilly: well, I don't know the best way forward. Regarding garbage collection, I don't understand why they are deleted. Is the database object being recreated all the time? A string like `project_id` or `table` is not even a complex instantiated object. 

So ok, maybe we should just get rid of most `delete ...` lines and keep those which are really in some way meaningful. I can certainly try that instead and see how far I get.

haraldschilly:
pushed another change, with that the hub starts up again. it resets object-fields instead of deleting them, which has them same effect. I left those cases which look like a more complex situation.

williamstein:
Does not deleting the fields cause memory leaks?

haraldschilly:
> Does not deleting the fields cause memory leaks?

well, not on its own. all objects that aren't reachable from the global context will be removed. tricky are situations with callbacks and  closures, and well, passing on a reference. In general it's also hard to know if the instance is about to be deleted or if it is somehow reused. I'm not sure what edge cases can happen, that's why I did those changes where I reset the field to an empty object.

williamstein:
Please do this in two separate steps:

(1) upgrade to typescript v4 and disable this "delete checking",
(2) enable delete checking and make code changes.

I want to make 100% sure that we sort out **all** issues with typescript v4 before worrying about (2).

haraldschilly:
hmm, I found no way to disable a specific test. I only found `tsc-silent` which feels like a lot of work to integrate just to get rid of it afterwards. another option is to comment each line with `@ts-ignore` which is not a pretty solution either. well, I'm all for modifying the delete statements to reset the fields. I haven't seen any other errors showing up.

williamstein:
>hmm, I found no way to disable a specific test.

In that case, +1 to precisely what you're doing here in this PR!

williamstein:
I thought about this more and I don't like  the approach of removing all the delete attributes in the close function.   I'm going to see if I can come up with a better approach... (if not, then we have no choice, but I hope there is a better approach).   I can't imagine how not deleting these references isn't going to lead to lots of horrible memory leaks, given the lack of weakrefs in Javascript right now...   I really hope there is a better way.

src/smc-util/sync/editor/generic/sync-doc.ts
```diff
@@ -260,7 +263,7 @@ export class SyncDoc extends EventEmitter {
     ]);
 
     if (this.change_throttle) {
-      this.emit_change = throttle(this.emit_change, this.change_throttle);
+      this.emit_change = throttle(this._emit_change, this.change_throttle);
```
williamstein, line 266: Don't do this.  The leading underscore convention is something used in Python because there is no actual notion of "private" in Python.  In coffeescript and javascript it is useful too, but in Typescript it is totally unnecessary.   I've been systematically removing use of leading underscores in our typescript code, since it isn't needed...

williamstein:
@haraldschilly I'm going to start over on this in a way that is more iterative (not all or nothing) and fits better with the intention of how this should work, but doesn't change the logic or potential memory leaking of the code.   I'm looking at your code and using it as inspiration, but you don't have to work anymore on this branch.

williamstein:
OK, I ended up doing this.  I copied some of your code.  Regarding delete, in some cases I made the property optional and added a type assertion function (`assertDefined`), in some cases I just added explicit checks, and in some cases I added a new function `close` that just deletes *everything* from the object (to maximally free up any potential memory references!) and skips the type checking.   Which depended very much on context and how the object is used in practice, and how much each approach impacted the code.

After fixing the delete issues and your configuration issue, and a few tiny things you found, everything seemed to work really well!  Welcome to the brave new world of TS 4.   https://medium.com/better-programming/whats-new-in-typescript-4-fe8d50f08e31

PR #4852, opened 2020-09-12
webapp/terminal: make renderer configurable

haraldschilly:
# Description

this is just a quick start, please take over …


williamstein:
I think this is the wrong approach.  The correct approach is an automatic fallback if webgl isn't supported, not a config option in account preferences that many people wouldn't find.

williamstein:
This is what I mean: 57e1b28cccf1ccb07c88da2543f7d2324506cddb

PR #4859, opened 2020-09-15
session: attempt to keep tab ordering when restoring

haraldschilly:
# Description


williamstein:
NOTE: try this again by bypassing the absolute path lookup for session restore.

williamstein:
I used a very different approach just now to fix this.

PR #4865, opened 2020-09-16
combine 4 initial http requests into one and optimize them

haraldschilly:
# Description

well, the idea is simple: the webapp learns everything it initially needs to know via `/customize?type=full`. this requires a bit of plumbing in the hub and all aspects are a little bit cached and the customize request itself is also cached.

deployment: hub first (it learned new tricks in a backwards compatible way) → then webapp. it's backwards compatible with an older webapps, just in case.

testing: (1) newer hub, existing webapp, reloading it should do 4 requests. (2) with the newer webapp, the inital customize request includes more information. things to check are things like a customized name, SSO strategies, signup token.

I should probably build this in test.


williamstein:
This looks very, very good.  The only question is when to merge/release it (to avoid disruption).

PR #4868, opened 2020-09-17
fix issue with stop/start project sagemathinc/cocalc-docker#97

HighwayStar:
# Description
Workaround for stop/start project issue  described in https://github.com/sagemathinc/cocalc-docker/issues/97

This probably not compatible with python 2 if this still matters.

With this change project can be correctly stopped and strarted using current cocalc in docker.


williamstein:
Thanks!

PR #4869, opened 2020-09-17
project: remove console server

haraldschilly:
# Description

this carves out everything related to the console server from the project, but leaves all entry points and a script file. as far as I can see, everything works as before.


williamstein:
This makes sense!  I completely rewrote terminals when I changed the architecture of cocalc so that browser clients talk "directly" to projects, not using hub-websocket to pass messages along.  However, I didn't delete the bloat / clutter left from before. Yikes!

PR #4870, opened 2020-09-18
optimize project memory usage

haraldschilly:
# Description

~~:warning:  this depends on https://github.com/sagemathinc/cocalc/pull/4869~~

changes (for all the below there is a corresponding branch in kucalc)

* fix a `--test` mode, which allows us to do a compile time check like with the hubs 
* do not load the typescript and coffeescript transpilers
* lazy-load prettier
* rename api call to format code to "formatter", but keeping the deprecated name in the webapp, such that it stays backwards compatible for now (with older projects)
* type those api messages

together with some tweaks on kucalc's side, this reduces the average memory consumption from ~140mb to (initially) 75mb. there is room for further optimizations, but this needs more work.

testing: I have already checked this PR + `proj-mem-tune` in kucalc, but first the PR above must be merged. the parts to check are basic project functionalities, like calling formatter for prettier, sync, etc. there is no real fundamental change in functionality, though.

deployment: depends on a branch in kucalc, and well, I'll build and deploy this now. the real critical thing to watch out for is the long term cpu and memory consumption of a project. initially it will be smaller than usual, but there could be unforeseen effects.


williamstein:
Wow!!! "together with some tweaks on kucalc's side, this reduces the
average memory consumption from ~140mb to 85mb. "   just wow!

haraldschilly:
interesting, right after I deployed this for testing and restarted my project (which cleared `/tmp` + some typescript cache) I saw new TS errors. That's certainly something to keep in mind. At least for the project, there could be stale stuff…

In other news, here is the effect on a randomly picked node, ordered by memory: the oder hubs are on top, then comes a mix of nodes and python3 kernels. i.e. now we are at least in the same ballpark as a jupyter notebook. there is room for more, but it's getting harder and harder.

PR #4871, opened 2020-09-19
Pricing info licenses

haraldschilly:
# Description

This adds course pricing info to the landing page. 

It also changes some parts to be functional components and some slight refactoring.

This has no dependencies, except for rebuilding the landing hub. I'll start a build now and if it works, I'll release this.

The only real issue I had was to match the prices I get via the form. I debugged it, and it seems to depend on that time interval. So, what I did is to find start/end pairs that almost match what I see when I use the form.


williamstein:
Excellent!!

PR #4873, opened 2020-09-21
clear terminal

haraldschilly:
# Description

this adds a trash button to clean up everything that is going on right now in that terminal. this includes seeing phantom characters from some previous output or getting out of vim if you happen to be lost there.


williamstein (changes requested):
1. Should we rename it from "clear terminal" to "clear"?  There could be a clear button for other editors.  Naming this "clear terminal" seems to me kind of like naming a copy button "copy text editor".

2. Should there be a popconfirm?  I'm a little worried about a button that deletes everything without confirmation. What if you bump it?  In cocalc, pretty much any button you hit can be undone easily (e.g., closing a terminal -- just open it again and it's back).  It seems that clear is more destructive, so a confirmation would be good. What do you think?

haraldschilly:
well, I'm about to give up. I saw there is a `clear_terminal_command` somewhere and I enhanced it to also do a reset. That's all I attempted to do here. When I try to move this clear method into one of the classes in the terminal-editor subdirectory, I don't see how I can access the terminal or how to call this. Is there somewhere an explanation for all this?

e.g. I tried to use this terminal manager, because it has a `get(id)` method, but it's private and hence `Property 'terminals' is private and only accessible within class 'Actions<T>'` … which tells me that's not how to access anything on a terminal in `terminal-editor/actions`

Besides that, a popconfirm is a good idea, but the text should be specific to the editor type. So, how do I set the title/description of what's shown for a terminal in that title-bar ?

I'm trying to rearrange some things further, but so far there are only dead ends.

haraldschilly:
well, I made some changes and made this field visible. 

but in any case, it's still not working as I want. if I open a terminal in a jupyter notebook, the title-bar's `is_visible` check gets `props.type == "terminal"`, but `props.actions` is an instance of `JupyterEditorActions`. Looks like a bug to me.

haraldschilly:
ok, I moved this back. so, now there is this clear function for the terminal in the actions for the code editor. I wrapped it in a condition to safeguard what it is doing.

Issue #4875, opened 2020-09-22
course/licenses: remove license from deleted projects


haraldschilly:
A deleted student project should get the license removed. That way it will not use up a seat if it is started accidentally.

PR #4876, opened 2020-09-22
course: remove license if student is deleted

haraldschilly:
# Description

* see #4875

ok, I'm not sure about this. there must be somewhere some code that removes a license from a project if the student is deleted, but I don't know where. I also saw in a real course that a deleted student project had a license. So, I don't know, maybe this here helps as an additional safeguard or it makes things even worse.


williamstein:
Your solution is exactly right, and I stupidly forgot to deal with this case.  Thanks!!!

PR #4882, opened 2020-09-23
real-time project info

haraldschilly:
# Description

This consists of two parts:

1. project: which collects information about itself in a pretty efficient way (no subprocesses, etc.) and also has a "status" service inside of it, which boils this down to some essentials (currently, there are just a couple of alerts and the project version). my future ideas here are to also use this for the project configuration.
2. webapp: if there is an alert, it highlights this in the UI (project, info tab). and well, the info tab itself then shows real-time info about the project's cgroup and running processes.

deployment: project should come first, but it is ok if you haven't restarted your project. then webapp.


williamstein:
Holy sh!t, this is amazing!!

PR #4892, opened 2020-09-29
project/jupyter:  rename _identity to identity

haraldschilly:
# Description

well, while working on that project info, I'm also adding TS types here and there. When I do that for the known jupyter kernel instances, this popped out. I'm sure this is a typo, but please check if this is maybe not so good (so far, that `if` was always false)

the only thing I tested is that I can start/halt a kernel.


williamstein:
Very good!

Issue #4893, opened 2020-09-29
lean: update_status_bar is not a function


haraldschilly:
for a lean file: 

chrome: `Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/70.0.3538.77 Safari/537.36`

smc_git_rev: 97dbda84f57f91100ec8495aeead92efa07f09e4

Date: 2020-09-29 11:22:18.233885

`t.update_status_bar is not a function`

```
TypeError: t.update_status_bar is not a function
    at t.eval (eval at <anonymous> (https://cocalc.com/app:1:7557), <anonymous>:2408:3627)
    at i (eval at <anonymous> (https://cocalc.com/app:1:7557), <anonymous>:258:13894)
    at eval (eval at <anonymous> (https://cocalc.com/app:1:7557), <anonymous>:258:13413)
    at function.e._wrapper.e._wrapper (eval at <anonymous> (https://cocalc.com/app:1:7557), <anonymous>:439:1980)
    at eval (eval at <anonymous> (https://cocalc.com/app:1:7557), <anonymous>:439:3482)
```

haraldschilly:
interestingly, also this one happens for lean. I bet there is a common cause. This clean function to delete all attributes?

```
chrome: `Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/70.0.3538.77 Safari/537.36`

smc_git_rev: 97dbda84f57f91100ec8495aeead92efa07f09e4

Date: 2020-09-29 11:22:18.155327

`t.update_gutters is not a function`

```
TypeError: t.update_gutters is not a function
    at t.eval (eval at <anonymous> (https://cocalc.com/app:1:7557), <anonymous>:2408:3547)
    at i (eval at <anonymous> (https://cocalc.com/app:1:7557), <anonymous>:258:13894)
    at eval (eval at <anonymous> (https://cocalc.com/app:1:7557), <anonymous>:258:13413)
    at function.e._wrapper.e._wrapper (eval at <anonymous> (https://cocalc.com/app:1:7557), <anonymous>:439:1980)
    at eval (eval at <anonymous> (https://cocalc.com/app:1:7557), <anonymous>:439:3482)
```

haraldschilly:
also lean:

[...]

PR #4895, opened 2020-09-29
webapp/lean: safeguard debounced calls -- addresses some of #4893

haraldschilly:
# Description

See #4893

I also changed the ordering of close calls: i.e. first set the field in the parent + close there, then the derived child. The other way around some attributes get deleted before the `_state` is changed.

well, maybe changing that ordering isn't a good idea. or not, IDK.  I'm confused. in both cases I get an exception upon closing:

```
Uncaught (in promise) TypeError: Cannot read property 'close' of undefined
    at Actions.close (actions.ts?47df:500)
    at Actions.close (actions.ts?81f3:179)
    at Object.remove (register.ts?99e9:85)
```

The above is strange: tries to close any terminals via `this.terminals.close();` but I don't see how that terminal manager got removed. It shouldn't be removed before it cleaned up any terminals, right?

### DIFF
--- a/src/smc-webapp/frame-editors/lean-editor/actions.ts
+++ b/src/smc-webapp/frame-editors/lean-editor/actions.ts
@@ -55,16 +55,22 @@ export class Actions extends BaseActions<LeanEditorState> {
     this.data_queue = [];
 
     this.debounced_process_data_queue = debounce(() => {
+      if (this._state === "closed") return;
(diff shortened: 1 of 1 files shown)


williamstein:
From the person who saw this bug in production:

> "Thanks very much. I’m not seeing the error now so I think it’s worked!"

PR #4902, opened 2020-09-30
clean button a unicode skull and change Icon to be a FC

haraldschilly:
# Description

This makes the Icon a functional component  and enhances it to show unicode chars as icons as well. Here is chrome and firefox:

Screenshot from 2020-09-30 12-20-57

Screenshot from 2020-09-30 12-20-47

deployment: webapp only


williamstein:
Wow, this is a really cool idea.  We could also have a parameter for making an icon using antd's icons, and have everything use this same "Icon" component...

PR #4908, opened 2020-10-01
npm: update Antd & co

haraldschilly:
# Description

this updates antd, some react defs, icons, etc. However, it breaks the share server. The problem is this:

```
  <rejected> ReferenceError: requestAnimationFrame is not defined
      at Object.<anonymous> (/home/user/cocalc/src/smc-webapp/node_modules/rc-util/lib/raf.js:17:3)
```

basically: `r_misc` → loads `Tip`, which in turn is Antd code... and in turn it uses `rc-util` all over the place. My idea is to globally define `requestAnimationFrame` and probably also `cancelAnimationFrame` when the share server loads.  I tried it, but it didn't work. Maybe you have an idea? A thought I just had is to only load "Tip" and similar components only for the webapp, while we replace them by something without functionality for the share server.

Besides that, just the webapp alone seems to work fine with this.


williamstein:
> Maybe you have an idea? 

Don't load Tip, or if we do, replace Tip by something completely different when running under node.js.  Basically, instead of changing requestAnimationFrame, change Tip.

src/smc-webapp/r_misc/index.tsx
```diff
@@ -14,22 +18,35 @@ export { CloseX2 } from "./close-x2";
...
-export { DateTimePicker } from "./date-time-picker";
-export { PathLink } from "./path-link";
-export { HelpIcon } from "./help-icon";
-export { TimeAgo, TimeAgoElement, is_different_date } from "./time-ago";
+export const DateTimePicker = is_frontend
+  ? require("./date-time-picker").DateTimePicker
+  : Null;
+export const PathLink = is_frontend ? require("./path-link").PathLink : Null;
+export const HelpIcon = is_frontend ? require("./help-icon").HelpIcon : Null;
+const time_ago_module = is_frontend
```
williamstein, line 46: Does this approach break all the Typescript typing?  If so, it's not acceptable.

williamstein: I just did a test and in master (so without your changes), doing `<TimeAgo foo="bar" date={new Date()}>` gives a big type error (yeah!).  After checking out this branch the error disappears.  Thus unfortunately this approach is simply not going to work, since typing is extremely important.

williamstein:
I just did the "npm: update & co" part of this separately, since it makes no sense to block all the other upgrades (which are trivial and require no code change) due to antd issues.

williamstein:
I'm closing this, since the problem goes away completely by just uploading to the latest versions.  Yeah!

williamstein:
7eaa2f5603caa15848c2415f82bd5f0a20dfea7e

PR #4917, opened 2020-10-06
hub/syncstring maintenance: do not order by string_id to speed up the query

haraldschilly:
# Description

this query takes about `query_time_ms=5456`, despite indexing. without sorting, there is no need to look through all found results, right? and there is no pressing need to archive them in order...

deploy: hub syncstring restart

testing: I didn't test this

### DIFF
--- a/src/smc-hub/postgres-blobs.coffee
+++ b/src/smc-hub/postgres-blobs.coffee
@@ -620,7 +620,6 @@ exports.extend_PostgreSQL = (ext) -> class PostgreSQL extends ext
                     query    : 'SELECT string_id FROM syncstrings'
                     where    : [{'last_active <= $::TIMESTAMP' : misc.days_ago(opts.age_days)}, 'archived IS NULL']
                     limit    : opts.limit
-                    order_by : 'string_id'
                     cb       : all_results 'string_id', (err, v) =>
                         syncstrings = v
                         cb(err)


williamstein:
The found results is nothing but a list of a few thousand uuid's, and sorting them presumably takes a few milliseconds.    I can't believe this change can possibly make any difference at all in performance, unless there is some weird postgresql performance bug.

That said, I agree that the order shouldn't matter at all for what we're doing.  A random order might even be better...

haraldschilly:
I don't see how this can be a performance bug? the difference is that in the sorted case it goes through all IDs and filters on them, and then takes the limit. in the unsorted case, it immediately picks the archived index key (well, at least in my query here), and only filters on the last active criteria. the upper bounds for the cost estimations for the index scans, in the sorted case is 535088, while in the other one it is 300626. So, based to the statistics postgres knows about that table and indices, this makes this about 40% faster on average. 

```
smc=# explain select string_id from syncstrings where archived IS NULL and last_active <= NOW() - '60 days'::INTERVAL order by string_id limit 10;
                                             QUERY PLAN
-----------------------------------------------------------------------------------------------------
 Limit  (cost=0.56..14.00 rows=10 width=41)
   ->  Index Scan using syncstrings_pkey on syncstrings  (cost=0.56..535088.60 rows=398147 width=41)
         Filter: ((archived IS NULL) AND (last_active <= (now() - '60 days'::interval)))
(3 rows)

smc=# explain select string_id from syncstrings where archived IS NULL and last_active <= NOW() - '60 days'::INTERVAL limit 10;
                                                 QUERY PLAN
-------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.56..8.11 rows=10 width=41)
   ->  Index Scan using syncstrings_archived_idx on syncstrings  (cost=0.56..300626.35 rows=398147 width=41)
         Index Cond: (archived IS NULL)
         Filter: (last_active <= (now() - '60 days'::interval))
(4 rows)
```

williamstein:
I stand corrected!  Please merge!

PR #4927, opened 2020-10-09
project info: info.processes could be undefined

haraldschilly:
# Description

I'm not sure why at all, but apparently this could be undefined. Updated type info and rejecting an incomplete object.


haraldschilly:
merging this as well. now building this and the previous PR.

PR #4930, opened 2020-10-10
hub/db: nestloop setting specific to a user query, not global

haraldschilly:
# Description

A possible idea how to tune nestloop for a single type of user query only. I tried a few approaches, and well, I think this can work. The background is that this setting shouldn't be global, but only for specific queries where we saw back performance. So far, as I can tell, that's the `file_use` query where there is no entry at all – that's where it is completely off. The risk with this here is that there might be other queries suddenly taking longer in rare cases. 

I'm deploying this in "test" now, just to see if everything still works. In my dev project it was fine.


src/smc-hub/postgres/set-pg-params.ts
```diff
@@ -0,0 +1,42 @@
...
+
+interface Opts {
+  client: Client;
+  query: string;
+  params: string[];
+  pg_params: { [key: string]: string };
+  cb: (err?, result?) => void;
+}
+
+// run the actualy query after a setup quey in a transaction
```
williamstein, line 21: This is full of typos

src/smc-hub/postgres/set-pg-params.ts
```diff
@@ -0,0 +1,42 @@
...
+  try {
+    await client.query("BEGIN");
+    for (const [k, v] of Object.entries(pg_params)) {
+      // SET LOCAL: only for this transaction!
+      // NOTE: interestingly, $1, $2 params do not work … but this isn't user-facing
+      const q = `SET LOCAL ${k} TO ${v}`;
+      await client.query(q);
+    }
+    const res = await client.query(query, params);
+    cb(undefined, res);
```
williamstein, line 35: IMPORTANT: Lines 35 and 36 should be swapped.  The cb(undefined, res) must happen after the await.  Otherwise, if `await client.query("COMMIT")` fails (likely!), then cb will be called _twice_, once with no error and once with an error; not good.

PR #4941, opened 2020-10-13
speed up the "always_running" query

haraldschilly:
# Description

this is a simple optimization and well, maybe there will be other queries asking for this set of projects in the future as well.

### DIFF
--- a/src/smc-hub/postgres/always-running.ts
+++ b/src/smc-hub/postgres/always-running.ts
@@ -15,7 +15,7 @@ export async function projects_that_need_to_be_started(
 ): Promise<string[]> {
   const result = await database.async_query({
     query:
-      "SELECT project_id FROM projects WHERE settings#>>'{always_running}' = '1' AND state#>>'{state}' NOT IN ('running', 'starting')",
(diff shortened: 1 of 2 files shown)


src/smc-hub/postgres/always-running.ts
```diff
@@ -15,7 +15,7 @@ export async function projects_that_need_to_be_started(
 ): Promise<string[]> {
   const result = await database.async_query({
     query:
-      "SELECT project_id FROM projects WHERE settings#>>'{always_running}' = '1' AND state#>>'{state}' NOT IN ('running', 'starting')",
+      "SELECT project_id FROM projects WHERE settings ->> 'always_running' = '1' AND state ->> 'state' NOT IN ('running', 'starting')",
```
williamstein, line 18: Wait does this notation actually impact performance, or you just like it more???

haraldschilly: well, I like it more, but there is also a semantic reason. the `#>>` syntax means to "extract an object from somewhere in the jsonb" and then "convert it to text". so, this is a more general thing. The `->>` syntax means to "get the value of a field" and convert it to text. What happens here is clearly to get a specific value.

PR #4943, opened 2020-10-15
make files listings git aware

haraldschilly:
# Description

I wanted to start with a simple step to make the files listing git aware. So, if the listing is part of a git repo, it should tell the front end where the home of the git directory is. That's minimal overhead and from there we can expose commands, etc.

However, in order to do that, I had to refactor some parts of the file listing code and stumbled over some details. e.g. the type info says it is a string, but it isn't etc. with what's here, I can browse around and it also updates if a file appears or is deleted. Where it fails is reacting to creating/removing a directory and listing files in a new empty directory. I'm unable to figure out what part of the code is getting triggered and why this doesn't work.

### DIFF
--- a/src/smc-project/directory-listing.ts
+++ b/src/smc-project/directory-listing.ts
@@ -17,8 +17,12 @@ Use ?random= or ?time= if you're worried about cacheing.
 Browser client code only uses this through the websocket anyways.
 */
 
+import { join } from "path";
 import { lstat, stat, readdir, readlink, Dirent, Stats } from "fs";
 import { callback } from "awaiting";
+import { spawn as child_process_spawn } from "child_process";
+import { promisify } from "util";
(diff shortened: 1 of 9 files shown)


williamstein:
Looking through this I feel like you made too many changes.  E.g., 
```
-  const { listing } = data;
+  const { files } = data;
```

Why change "listing" to "files"?   I prefer "listing", since that implies order, and I also tend to think of a file and a "directory" as being both in a listing, but files to me suggests just the files (I know -- in UNIX "everything is a file", but still).

williamstein:
Another comment is just that CoCalc directories used to be git aware (in 2013 and 2014) and efficiencies issues were a MASSIVE problem with this.  Basic questions like "when was this  file last changed in the git repo" can be very expensive to answer.  So caching can be really important.  A lot also depends on what you want to accomplish though.

PR #4946, opened 2020-10-16
hub/db: _get_remember_me casts the hash to text, which slows this down

haraldschilly:
# Description

This is a 200x improvement for those cookie queries. Take away: either don't cast – or cast to the correct type.

```
smc=# explain (analyze, verbose, costs) SELECT value, expire FROM remember_me WHERE hash = 'sha512$6cc3e...499'::CHAR(127);

 Index Scan using remember_me_pkey on public.remember_me  (cost=0.54..2.76 rows=1 width=162) (actual time=0.036..0.037 rows=1 loops=1)
   Output: value, expire
   Index Cond: (remember_me.hash = 'sha512$6cc..4499'::character(127))
 Planning time: 0.079 ms
 Execution time: 0.051 ms
(5 rows)

smc=# explain (analyze, verbose, costs) SELECT value, expire FROM remember_me WHERE hash = 'sha512$6cc3e...9'::TEXT;

 Seq Scan on public.remember_me  (cost=0.00..7479.69 rows=591 width=162) (actual time=0.018..108.994 rows=1 loops=1)
   Output: value, expire
   Filter: ((remember_me.hash)::text = 'sha512$...74499'::text)
   Rows Removed by Filter: 118564
 Planning time: 0.060 ms
 Execution time: 109.015 ms
(6 rows)
```

regarding deployment, this is certainly relevant for hub-proxy, but also all others using auth

### DIFF
--- a/src/smc-hub/postgres/remember-me.ts
+++ b/src/smc-hub/postgres/remember-me.ts
@@ -26,7 +26,8 @@ async function _get_remember_me(
       cache,
       query: "SELECT value, expire FROM remember_me",
       where: {
-        "hash = $::TEXT": hash.slice(0, 127),
+        // db-schema/auth defines the hash field as a "bpchar", hence do not cast to TEXT – otherwise this is a 100x slowdown
+        "hash = $::CHAR(127)": hash.slice(0, 127),
       },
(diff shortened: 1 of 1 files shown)


haraldschilly:
merging, building in test

PR #4963, opened 2020-10-23
updating npm packages, including Antd

haraldschilly:
# Description


src/smc-webapp/r_misc/number-input-backend.tsx
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+import * as React from "react";
+
+interface Props {
+  number: number;
+  unit?: string;
+}
+
+export const NumberInput: React.FC<Props> = (props: Props) => {
```
williamstein, line 13: What the heck is this?  It says "Input" but it is clearly not an input.  Confused.

williamstein:
Why is there so much new/modified code here related to r_misc? Are all these changes required by updating antd from 4.7.0 to 4.7.2?  If so, what the hell...?

haraldschilly:
no it's not, the pr isn't done yet

williamstein:
OK, got it.  Just curious.   Maybe a separate PR just upgrading antd?

haraldschilly:
I cleaned this up. newer antd, also another package, and many dependencies (audit fix). I also modernized this "Tip" component, so, for a proper test, hover over a project tab to see that there is still something popping up.

PR #4976, opened 2020-10-29
admin/user_search: sort users by last created or active

haraldschilly:
# Description

the results of an admin search for users shows up in a random order, at least for me. the sort indicator tells it should be, though.

## before, some search in production:

Screenshot from 2020-10-29 12-51-48

## after, with data in my dev project

Screenshot from 2020-10-29 13-33-55


williamstein:
Seems better and less likely to be "randomized" by string conversion like you're hitting (but I don't).

PR #4985, opened 2020-11-02
jupyter: snippets side panel

haraldschilly:
# Description

* based on the idea for the terminal guide, re-implement  that snippets assistant
* webapp only deployment


williamstein:
This is amazing and much better than the snippets modal!  It looks gorgeous!

There are a number of little things that are annoying/confusing to me.    I think I should merge this, make it live, and make a couple of bug issues listing the problems, since this is already more useful than snippets, and the problems aren't that serious.
For example, the +/- zoom buttons don't do anything -- it's a bug, but not a blocker since the old snippets modal didn't even have +/- buttons.  Also, if you close the notebook panel, then re-open it (by splitting and changing) inserting code is broken.  I'll make some followup issue.

PR #5008, opened 2020-11-16
webapp/admin/license: enable "memeber"  by default

haraldschilly:
# Description

Way too often I forgot to enable member hosting for a trial. With that, it's on by default and shown to the admin. No implications for a general user. Only webapp affected...


williamstein:
Good idea; I like this.  I'm also going to make ram/cpu/memory default to 1 (instead of blank=0).

williamstein:
Also, I'll make the run limit default to 1 to avoid accidentally giving away an "infinite" license.

williamstein:
Note, I moved where the default is set.

haraldschilly:
oh ok, I see, makes sense to set it from the start

PR #5018, opened 2020-11-19
npm/hub: rm pdfkit

haraldschilly:
# Description

I don't see any usage of `pdfkit`. Removing this, gets rid of ~70 dependent packages :muscle: 


williamstein:
Yes, I recently removed that due to stripe now having an API that provides a public invoice link.

PR #5052, opened 2020-12-01
nbgrader: introduce multiple choice question/test pair

haraldschilly:
# Description

Well, this ended up a bit different from what I tried first, but it works and makes sense. In particular, I see room to restructure this later. What this does is to introduce 3 new templates for nbgrader:
*  multiple choice question: markdown, no grading, readonly
*  multiple choice answer: essentially assigning to a variable, no grading, editable
*  multiple choice test: a sanity check for the expected result set and a hidden test for the actual point, readonly

I also added a mechanism that sets the cell type if it is inserted. Otherwise, it feels very odd that someone has to change cell type first in anticipation of whatever is inserted – if it is the wrong type, just a general stub doesn't help much.

In any case, I wonder how to check if this is backwards compatible. There is this `state_to_value` function, which maps 32 combinations to 9 … and well, with some parameters I don't really know what they even mean. I think it is okay, because I added a  fallback if the new `multiple_choice` variant isn't set. 

teacher:

Screenshot from 2020-12-02 14-31-06

student

Screenshot from 2020-12-02 14-36-28

good/bad result and an additional stub test as a general sanity check

Screenshot from 2020-12-02 14-36-39

Screenshot from 2020-12-02 14-37-16


williamstein:
Wait, I'm really really worried about this, and also I don't understand why it is necessary.  Doing what you're doing here is philisophically a massive change from everything I've done so far with nbgrader.  In particular, if anybody uses this new cell type that you added, they will be creating nbgrader content that is completely locked into cocalc, which can't be used with upstream nbgrader at all.     I've had as a key design constraint from the beginning of working on this that:

 - upstream nbgrader enhanced notebooks would work in cocalc (i.e., take nbgrader notebooks from elsewhere and you can just use them in cocalc), and
 - conversely, you can use nbgrader in cocalc to make notebook, then actually use upstream nbgrader to work with them.  (Except for things like official nbgrader not supporting Octave for high minded reasons.)

What you've done is I'm sure very useful, but completely breaks the second point above (and maybe even does so silently).    Is that necessary???  We know people are using cocalc + nbgrader and making multiple choice questions without having to extend the nbgrader metdata at all.  Why did you do so?

When I meant "extend" in our discussion on Monday, I just meant adding things like the templates for each language, and actually supporting using languages like Octave and R.

Thoughts?

haraldschilly:
Well, I don't really understand why this was coded like that in the first place. If I just add new question/test types, they are replaced by others somehow. The selector switches to something else. That's why I had to introduce a new category of cell types, to make this distinct. Regarding these two points, I don't understand how it breaks it. Is this explained somewhere? I would hope it falls back to the default categories.

haraldschilly:
I'll try to run nbgrader with this and see what happens. I don't have any other idea.

williamstein:
Maybe I am just wrong? That would be ideal.

williamstein:
The very first line of your patch is
```
  if (cell.getIn(["metadata", "nbgrader", "multiple_choice"]) == true)
```
which tells me you've made up a new field in the nbgrader metadata.  This will either cause upstream nbgrader to immediately terminate (declaring the input unsupported), or at best (!) to completely ignore this cell.  Either way, it's definitely not going to work outside cocalc.

Does **that** make sense?

williamstein:
I see now -- you added some fallback metadata as well..  I tested it and it works well enough.  I'm going to merge this.

PR #5058, opened 2020-12-02
webapp: disable some css that causes \sqrt{x} to not show a sqrt

haraldschilly:
# Description

building this for test now, will make live if it works


haraldschilly:
well, and more general, looking at that file, there is also `img svg` right above. this means "match svg tags inside of img tags". that never happens.

williamstein:
You're now just breaking something else though which I fixed just
yesterday, right?   What we should do is apply this css only to img and not
to svg at all.

PR #5078, opened 2020-12-10
Hub Self Terminate

haraldschilly:
# Description

This is for letting the hub's healthcheck fail after a random time. Initially, I implemented something on the level of kubernetes with a script, but this is much better. I also typscripted some parts, i.e. that hub_register module, because this needs to access the database functionality check.

In any case, by default almost nothing changes, no changes to concurrent or concurrent-warn – just typescripted them.

## `/healthcheck`

 If an environment variable is set like this (for testing):

`export COCALC_HUB_SELF_TERMINATE=0.05,0.1`

then `/healthcheck` says

```
healthchecks:
concurrent 0 < 300 – OK
uptime 8.33s – terminating in 2m10s – OK
```

and after some time, sends a 404 page like

```
healthchecks:
concurrent 0 < 300 – OK
uptime 3m38s – expired, terminating now – FAIL
```

## `/alive`

What also changed is `/alive`: it will tell a proxy to no longer use it, either if the DB doesn't work (like before) or if the healthcheck is about to fail – the minimum of 10 minutes or 10% of expected uptime. This means for us, 10 minutes before the hub will be restarted, it will no longer get new connections. HAProxy should list it as "down" and continues to probe it, until it is removed.


williamstein:
Looks great to me (assuming it fully works of course). Please feel free to merge it and make it live whenever you want.

haraldschilly:
I read more on haproxy. the idea will work, but for haproxy it must be done a little bit different…

haraldschilly:
I used the opportunity of very low usage and deployed this right now

PR #5093, opened 2020-12-12
./tmux-start-all didn't work figured out I had to do dev/project/tmux…

realtimclemans:
I added `dev/project/` in front of `tmux-start-all`


williamstein:
I don't use tmux anymore for cocalc dev...

realtimclemans:
What do you use?

williamstein:
Seems relatively harmless...

Issue #5045, opened 2020-12-01
Add link to public worksheets in header


realtimclemans:

williamstein:
Where would the link go?  There is a link in the "i CoCalc" info page, and it could be moved to a more obvious place.

realtimclemans:
At top left between CoCalc and Help.

haraldschilly:
I think you mean top right. Christian Stump once had a similar wish. My only concern is to use up too much space overall.

PR #5097, opened 2020-12-13
Add public files link to navbar

realtimclemans:
Implements #5045 
Screenshot:
Screen Shot 2020-12-13 at 2 32 11 PM


src/smc-webapp/app/page.tsx
```diff
@@ -88,6 +88,11 @@ const PAGE_STYLE: React.CSSProperties = {
 
 const positionHackHeight = `${NAV_HEIGHT - 4}px`;
 
+// smc-util/theme is set in ./render.coffee
+const BASE_URL = global["BACKEND"]
+  ? require("smc-util/theme").DOMAIN_NAME
+  : require("../misc").BASE_URL;
```
haraldschilly, line 94: well, the hardcoded domain name is a fallback. please don't use it. the backend should use the site settings from the database, which will set this domain name as a fallback – see `get_server_settings_cached` method. I wonder why you set a domain name at all, though. 

and more general, is this really being used by the backend? maybe just don't render this tab, no need to do so.

Maybe it's best to just import `APP_BASE_URL` from misc and that's it. … and please import with `import` at the top, not `require` inline.

realtimclemans: I copied the code used for the info page. I just pushed the change you asked for.

williamstein:
I think this uses too much valuable space, and I'm not as "proud" of the share server at this point.

PR #5108, opened 2020-12-18
idea for a smart editor for unknown files

haraldschilly:
# Description

this fallback "editor" actually checks the content. main point here is to avoid shooting yourself in the foot by just one wrong click.

in general, I see this as a first step. in the future, certain mime types could be registered with an editor automatically, or offer running a binary file (executable ELF) in a terminal … and of course, offer more than just the code editor.

to test this, I added random bytes to a file `foo.baz`, vim-edited a file `bla.bar`, or even copied an ELF binary to `name.bin`. 

deployment: webapp only

Screenshot from 2020-12-18 18-15-18

Screenshot from 2020-12-18 18-17-28


williamstein:
Wow, that was impressively quick!?

williamstein:
This is pretty much genius.  We'll see what impact it has on support, but your design/idea is way better than what I was thinking!  Great job!

PR #5176, opened 2021-01-18
remove docx2txt, instead use pandoc for markdown conversion

haraldschilly:
# Description

 also add capabilities check for pandoc


williamstein:
Looks good to me!   This makes a lot of sense, and also fits nicely with my plan to make markdown files way better soon :-).

williamstein:
@haraldschilly  -- When you open the file it still says "Opening converted plain text file instead..." in an alert.  Maybe it should say "markdown" instead of "plain text"?

PR #5194, opened 2021-01-29
webapp/collabs: modernize react component and wrap remove button in a popover confirm

haraldschilly:
# Description

so, besides the FC, the problem is that I can't see the confirmation here (the  last remove button is clicked) without scrolling. quite suble.

Screenshot from 2021-01-29 17-27-00

with this patch it's more obvious

Screenshot from 2021-01-29 17-25-49

TEST: yes, I checked, works fine, there is no real change in functionality

DEPLOY: webapp only


williamstein:
Thanks!!

Issue #5222, opened 2021-02-25
latex/config: can't change spellcheck language


haraldschilly:
not sure if this is a new bug or just weird – I didn't investigate in detail. but in any case, I can't select a different language here:
 
![Screenshot from 2021-02-25 16-47-52](https://user-images.githubusercontent.com/207405/109178619-5548ee80-7789-11eb-937d-4b400c888bd9.png)

It's the same for chrome and firefox. Maybe this is stuck to the language of the browser? If so, we should get rid of the dropdown, it's confusing.

PR #5232, opened 2021-03-03
webapp: fix spellcheck dropdown

haraldschilly:
# Description

See #5222 ... I have no idea if react bootstrap is fixable, but this replaced it with antd. 


haraldschilly:
I'll make this an FC, then we can merge it

williamstein:
> I have no idea if react bootstrap is fixable

The best fix is to rewrite everything to not use it, then delete it!

PR #5246, opened 2021-03-09
Project Datastore UI

haraldschilly:
# Description

~~this is blocked by #5251~~

deployment: manage and project must be updated in the cluster. they're probably already, though. then, hub first, which will not cause troubles because of #5251, and then static. once this is done, merge/push the [datastore branch](https://github.com/sagemathinc/cocalc-doc/tree/datastore) for updating the documentation.


haraldschilly:
ok, I've finished testing yesterday and well, I don't see how this will cause real issues. Regarding deployment, I suggest I should do it.

williamstein:
> ok, I've finished testing yesterday and well, I don't see how this will cause real issues. Regarding deployment, I suggest I should do it.

+1 to you fully releasing it.  

I did skim your code just now, though of course not closely enough to catch any problems.   I must say, your comments in this particular PR are **excellent**!

PR #5259, opened 2021-03-31
subscription ←→ license sync

haraldschilly:
# Description

well, I thought this through and ran tests. I'm basically confident that this will not cause severe troubles. Also, even if there is an issue, we can switch back to the previous version, which should undo all the mistakes. In the end, nothing gets deleted … only the expiration is set or unset.

I can certainly deploy it. It's currently running (with a different image) as `hub-stripesync-test` shadowing `hub-stripesync`.


williamstein:
Great work.  Please deploy!

williamstein:
(and merge)

PR #5268, opened 2021-04-10
License limit

haraldschilly:
# Description

this is more or less done. I'll deploy for testing.

this also updates jest + expect. in particular, I saw a problem working with a super old version of expect, while the newer one is more or less part of jest.

there is also an info-note in the UI now, which explains why some licenses aren't active.


src/smc-util/upgrades/quota.ts
```diff
@@ -199,233 +238,346 @@ function calc_default_quotas(site_settings?: SiteSettingsQuotas): Quota {
...
+  return selected.reduce((acc, cur) => {
+    acc[cur] = site_licenses[cur];
+    return acc;
+  }, {});
+}
+
+// there is an old schema, inherited from SageMathCloud, etc. and newer iterations.
+// this helps by going from one schema to the newer one
+function upgrade2quota(up: Partial<Upgrades>): RQuota {
+  const dflt_false = (x) =>
```
williamstein, line 318: Sorry but it is pretty funny that you abbreviated "default" to "dflt" (I assume).

williamstein:
This is a ton of work.  I skimmed it but can't tell for sure if it is right, of course.   I think if it works when you test it, then you should deploy it.

PR #5273, opened 2021-04-12
webapp/jupyter: reverting 4dfb08af22389787 because it breaks sliders

haraldschilly:
# Description

this undo's https://github.com/sagemathinc/cocalc/issues/5228 because sliders no longer show up


williamstein:
Probably the right fix is to only hoist up some types of html elements rather than everything.

Issue #5284, opened 2021-04-14
Support formatting IPython magics


mforbes:
Clicking `Format` on a cell with

```
%pylab
```
results in

```
Error: Error: Python formatter "yapf" exited with code 1:

  File "/tmp/tmp-71qjxfBx4Uhki.tmp", line 1
    %pylab inline
    ^
SyntaxError: invalid syntax
```

This is related to [issue 218 with yapf](https://github.com/google/yapf/issues/218), and makes some ense since this is not valid python code, but it is annoying when working with notebooks.  The [Black jupyter-notebook extension](https://github.com/drillan/jupyter-black) deals with this by [explicitly filtering out magics](https://github.com/drillan/jupyter-black/blob/master/jupyter-black.js).

haraldschilly:
I don't think this has anything to do with black.

This extension comments out percent-lines, runs the formatter, and substitutes those lines back. See https://github.com/drillan/jupyter-black/blob/dd01285bf7bca65a939309727a3a4b95743c9439/jupyter-black.js#L26 calling `re.sub( ... )` twice.

I can imagine we could implement this for jupyter as well.

PR #5287, opened 2021-04-15
jupyter/formatter: support %magics 

haraldschilly:
# Description

ticket #5284

the only thing I noticed is that the format button doesn't always work. the menu entry does, though. I don't understand why. 

my test were just simple things like

```
%time 123

def ff(x):
    pass
```

### DIFF
--- a/src/smc-webapp/jupyter/parsing.ts
+++ b/src/smc-webapp/jupyter/parsing.ts
@@ -9,6 +9,7 @@ This can *ONLY* be used from the browser!
 */
 
 const { endswith } = require("smc-util/misc");
+import { Syntax } from "../../smc-util/code-formatter";
 
 declare const CodeMirror: any; // TODO: import?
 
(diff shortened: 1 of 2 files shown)


williamstein:
> the only thing I noticed is that the format button doesn't always work.

Is this a new bug you found or something involving this PR?

src/smc-webapp/jupyter/parsing.ts
```diff
@@ -35,3 +36,25 @@ function last_style(code: string, mode = "python") {
   CodeMirror.runMode(code, mode, (_, s) => (style = s));
   return style;
 }
+
+// calling escape and then unescape should be idempotent
```
williamstein, line 40: In this comment, you're not using the mathematical/English word "idempotent" correctly.  It means an element of a group that is equal to its square, e.g., x*x = x.   What you mean is that the composition of escape followed by unescape is the identity map, which is just a completely different thing.

src/smc-webapp/jupyter/parsing.ts
```diff
@@ -35,3 +36,25 @@ function last_style(code: string, mode = "python") {
...
+// python is inspired by https://github.com/drillan/jupyter-black/blob/master/jupyter-black.js
+export function process_magics(
```
williamstein, line 42: Is it really possible to solve this problem using regexps?  E.g., What happens with this valid Python code that has no magics in it (it's just `3%2` but on two lines)?
```
3\
%2
```

williamstein:
It occurs to me that I have several significant concerns with this, and yet exactly what you've written is massively better than what is live in Cocalc now.  So... I'm merging this, and you can read my comments and make a follow up issue or PR later.

williamstein:
OK, I made followups

https://github.com/sagemathinc/cocalc/issues/5290

and 

https://github.com/sagemathinc/cocalc/issues/5289

Issue #5290, opened 2021-04-16
jupyter: any line with % anywhere in it is potentially formatted incorrectly when using Format


williamstein:
Put this in a cell and do `Cell -> Format selected cells`:
```
a = 3    +2
b = 3    %2
```
and you get the following, where the second line is not correctly formatted.
```
a = 3 + 2
b = 3  %2
```
You should get
```
a = 3 + 2
b = 3 % 2
```

This is because of what I explain here: https://github.com/sagemathinc/cocalc/pull/5287#pullrequestreview-637303434
and is caused by https://github.com/sagemathinc/cocalc/pull/5287.  Of course this is way better than any use of a cell magic completely breaking everything.

williamstein:
I have concerns about PR #5290, but it fixes the examples above, and is probably good enough for our use case.

PR #5291, opened 2021-04-16
jupyter/format magics: process each line individually 

haraldschilly:
# Description

yes, #5290 has a good point. assuming those magics start with the line, this processes the lines individually. then, such % signs are excluded automatically. I also changed the marker to be `#%%%#` which is even more unusual than in the original code. the comment should be a bit better now as well.

### DIFF
--- a/src/smc-webapp/jupyter/parsing.ts
+++ b/src/smc-webapp/jupyter/parsing.ts
@@ -37,7 +37,7 @@ function last_style(code: string, mode = "python") {
   return style;
 }
 
-// calling escape and then unescape should be idempotent
+// (unescape ∘ escape)(code) ≡ code should hold, while escape(code) is valid python
 // python is inspired by https://github.com/drillan/jupyter-black/blob/master/jupyter-black.js
 export function process_magics(
   code,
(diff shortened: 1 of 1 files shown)


williamstein:
This is slightly better but it does *not* work.    My first example in the comments on your PR still fails:
```ts
escape = (code) => code
        .split("\n")
        .map((line) =>
          line.replace(/^\%(.*)$/, (m, g1) =>
            g1 != null ? `#%%%#%${g1}` : /* should not happen */
          )
        )
        .join("\n");
```
Then
```ts
escape('2\n%3')
"2\n#%%%#%3"
```

Another comment is that the trick to avoid bad patterns in a way that will definitely work is to search the input for that pattern first.  This is like how markdown uses n backticks when n-1 backticks appear in a fenced block.

Life is short and I'm not against writing code like this 
```
g1 != null ? `#%%%#%${g1}` : /* should not happen */
```
which anybody who reads it can definitely construct examples that break it.  However, I would write the comment as "This absolutely CAN happen, but it's unlikely and life is short." since if I see "should not happen" in the code, I read that to mean that it really should not happen, "not -- it definitely DOES happen, but usually it doesn't.".

williamstein:
And to add, I would not be opposed to merging this code as is if the comments were correct.  For example, instead of saying the following

> " (unescape ∘ escape)(code) ≡ code should hold, while escape(code) is valid python"

It would be correct to say

> "We would like that the following two conditions hold:
>      (1) `(unescape ∘ escape)(code) ≡ code`
>      (2) `escape(code)` is valid python if code is valid python combined with ipython magics.
> However, the code below does not actually satisfy both of these conditions for all input and
> can probably dangerously mangle carefully crafted input.  Life is short, and somebody should
> fix this properly (is there a parser that Jupyter itself uses for processing magics that we could
> hook into?)  See <github issue>."

Sorry for being a stickler for correctness, but I'm a mathematician.    I'm of course very aware that I've written tens of thousands of lines of code with horrible bugs, some intentional and hopefully commented like the above.   But I think saying "this is wrong -- TODO -- I just don't have time to do this right now" is way better in a comment, then saying "this is right" when it isn't.   In code this is an option.  In math proofs it is too as long as you remain aware that the entire argument is not valid until the heuristic gap is filled in.  I miss doing math research.

williamstein:
Anyway, I'll merge this if either you think of a way to fix it, or change the comments (or just tell me you are OK with me changing the comments).    I could also waste time trying to think of some subtle way that this code could mangle valid user input (i.e., change the meaning of code), but that seems like a waste of time.

PR #5294, opened 2021-04-16
hub/db: check run_quota for always_running projects

haraldschilly:
# Description

I saw that this query is used by the manage backend to check which projects should be started. The problem is, as far as I understand it should check the run quota, not the settings. In particular, this covers licenses as well.

Deployment: `manage-idle`

### DIFF
--- a/src/smc-hub/postgres/always-running.ts
+++ b/src/smc-hub/postgres/always-running.ts
@@ -7,15 +7,14 @@ import { delay } from "awaiting";
 import { PostgreSQL } from "./types";
 import { callback2 } from "../smc-util/async-utils";
 
-// Return an array of project_id's of projects that have the always_running quota
-// set (or a license that would provide it -- TODO!), but are not in the running
-// or starting state.
+// Return an array of project_id's of projects that have the always_running run_quota set,
+// but are not in the running or starting state.
 export async function projects_that_need_to_be_started(
   database: PostgreSQL
 ): Promise<string[]> {
   const result = await database.async_query({
     query:
-      "SELECT project_id FROM projects WHERE settings ->> 'always_running' = '1' AND state ->> 'state' NOT IN ('running', 'starting')",
+      "SELECT project_id FROM projects WHERE run_quota ->> 'always_running' = 'true' AND state ->> 'state' NOT IN ('running', 'starting')",
   });
   const projects: string[] = [];
   for (const row of result.rows) {


williamstein:
>  saw that this query is used by the manage backend to check which projects should be started. The problem is, as far as I understand it should check the run quota, not the settings. In particular, this covers licenses as well.

This seems completely wrong to me.  Isn't run_quota *only* set when there is an attempt to run the project?    Imagine a project with always running true for some reason (e.g., due to project settings via admin), but one has never tried to run it. The run_quota field in the database is empty.  This PR would change the code from working properly to being broken.

Of course you're right that there is a bug, exactly as it says in the comment, namely the license part isn't implemented.

Another issue -- imagine the run_quota says that always_running is on, but the *reason* for that has been removed (e.g., settings or license changed), and there has been no attempt to run the project.   I think again that this change would break things by making the project run even though it shouldn't.  But it would I guess do that only once, since making it run would presumably then fix run_quota back...

haraldschilly:
I assumed all "always_running" project must be started once explicitly. Is this behavior somewhere specified? What's the specification of "always_running" ?

williamstein:
> I assumed all "always_running" project must be started once explicitly. Is this behavior somewhere specified? What's the specification of "always_running" ?

Here's what's in the license purchase page for a specification of "always_running":

image

In particular, 

> "- run long computations and never have to wait for project to start. Without this, project will stop if it is not actively being used. See project init scripts. (Note: this is NOT guaranteed 100% uptime, since projects may sometimes restart for security and maintenance reasons.)"

In particular, what determines the "always running" status is the combination of admin settings and licenses (we do not sell an always running upgrade).   Of course, we can just decide to change this definition if we want, and in retrospect your suggestion sounds pretty good to me.

haraldschilly:
My interpretation was based on member hosting and internet only work after a restart. I implied the same for always running, that's all. My main concern is that the backend's scheduling algorithm definitely assumes a project restart in order to properly schedule the pod. That text doesn't fully explain it as well, because when are licenses activated? My assumption is they need a project start, because that's when its decided if the license gets activated.

Regarding invalidating the upgrade after a license expires:  yes, this run quota filed will be changed, the project continues to run under the usual idle timeout regime, and most likely be de-scheduled very soon. I don't know if removing the admin always running setting is handled as well, but well, that's out of scope of this PR.

The last case I can think of is a license providing always running, but it's activation is in the future. I don't know if there is code that handles that, but it's yet again outside the scope of this PR. If there is no implementation for that, I suggest a similar check like for the license expiration, triggering project startup. This will evaluate the license condition and the project will either run with or without the always running run quota. But I would be careful with starting projects without direct user intervention, because that could lead to confusion.

PR #5298, opened 2021-04-19
jupyter/formatting: avoid messing up code cells starting  with the escape sequence

haraldschilly:
# Description

This reduces the chances of messing up the code. Usual content should work as before, except when a line starts with `#%%%#`. This pops up an error message.


williamstein:
Good test case:
```py
a = 2 + 3
a = r"""
#%%# foo
"""
```

PR #5299, opened 2021-04-19
modernize files search box code

haraldschilly:
# Description

* functional component
* clean up some details
* using useEffect to run the command async. ~~my real problem is how to actually deal with that `id` condition in the original code. the closure architecture updates the reference id, which stops this from working. but in any case, I using `useEffect` and just showing the result when it arrives is good enough for this.~~ ok, got it, useRef is what I needed, which makes total sense. I was mentally stuck with how the code was written before. … and besides that,  I didn't even remember there was a command mode :-) 

IMHO the next step here would be to rewrite the input box to use Antd, but that's for another ticket.


williamstein:
>  I didn't even remember there was a command mode :-)

Now that John is no longer around cocalc to lobby for it... :-)

Issue #5282, opened 2021-04-14
course invite email does not customize the sender's email/name properly


haraldschilly:
1. configure your own domain, name, email, etc.
2. invite a student
3. the "from" field is not customized, but it should

Issue #4881, opened 2020-09-23
hub/email: recreate smtp server instance when config changes


haraldschilly:
In https://github.com/sagemathinc/cocalc/blob/master/src/smc-hub/email.ts `init_smtp_server`, record the settings (as a hashed key?) that were used to instantiate the server and recreate it if the settings change.

williamstein:
Alternatively, add a message to the frontend UI saying "you must restart the server".

Or alternatively only cache the setting for a minute.

PR #5303, opened 2021-04-21
db/always_running: x IN <array>, not x NOT IN <array>!

haraldschilly:
# Description

optimizing a query, without breaking functionality, hopefully ;-)

deployment: manage and hub

## [Checklist](https://github.com/sagemathinc/cocalc/wiki/PR-Checklist):
- [ ] Testing instructions are provided, if not obvious
- [ ] Release instructions are provided, if not obvious

PR #5304, opened 2021-04-21
hub/email: fix invite email sender config

haraldschilly:
# Description

* issue #5282
* issue #4881
* a missing callback when inviting someone (I generalized the code, though)
* fixing replyTo attribute for nodemailer (also, updated it)

### DIFF
--- a/src/smc-hub/client.coffee
+++ b/src/smc-hub/client.coffee
@@ -1329,36 +1329,6 @@ class exports.Client extends EventEmitter
                                 to           : locals.email_address
                                 error        : err
                             cb(err) # call the cb one scope up so that the client is informed that we sent the invite (or not)
-                        
-                        
-                    #try
-                    #    email_body = create_email_body(subject, mesg.email, locals.email_address, mesg.title, mesg.link2proj, await @allow_urls_in_emails(mesg.project_id))
-                    #catch err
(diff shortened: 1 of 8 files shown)


haraldschilly:
so far, I only tested this in my dev project. we definitely have to double check in "test". its affecting course and project invite emails, especially with a custom smtp setup. deployment: hub, the part for "static" can wait.

haraldschilly:
I've merged the #5303 into this one  here as well. Well, in any case, we should update the hub just once to get everything in.

williamstein:
> we definitely have to double check in "test" – building now. 

And... did you test it there successfully?  Also, does this **require** updating the hub before building static, or are the frontend changes backward compatible with the non-updated hub?

haraldschilly:
The webapp part is independent, it's kind of "cosmetic" only.

My tests in "test" were successful.

But, the most important change is for CoCalc instances on different domains! The reason is those "from name/from email" pairs were hardcoded. I either never realized that or just forgot implementing it. So, now, instead of `invites@cocalc.com` the sender will be the organization's email address from the site settings config. This is important, because otherwise emails sent from a different domain easily end up as spam, because the sender's domain doesn't match. However, the "reply to" is the email from the cocalc user, as usual.

PR #5309, opened 2021-04-23
hub: only fallback to non-token collab adding, if there is no token

haraldschilly:
# Description

while testing #5308 I saw a confusing error message when entering a made-up project invite token. this suppresses the fallback test, such that the original error appears. Also, it says more than just `""`.


williamstein:
Merged and I also updated the comment I had written, which was wrong/misleading.

PR #5326, opened 2021-04-30
make cursor classes FCs

haraldschilly:
# Description

I think this works, but in general, I'm not happy with the positioning of it. That's what I actually wanted to look into. 

So, either we merge this quickly and I'll try to fix the CSS for the positioning (but I'm not really sure how it should look) or I'll keep tinkering around.

~~OK, no, I think I messed up something with the Cursor class, when moving it's parts around. Will fix it.~~ nope, just confused.

I'll look into that cursor positioning another time…

### DIFF
--- a/src/smc-webapp/jupyter/cursors.tsx
+++ b/src/smc-webapp/jupyter/cursors.tsx
@@ -16,10 +16,9 @@ import { Map } from "immutable";
 import {
   React,
   Component,
-  rclass,
-  rtypes,
   ReactDOM,
   Rendered,
+  useTypedRedux,
 } from "../app-framework";
 
 import { times_n } from "./util";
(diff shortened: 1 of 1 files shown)


src/smc-webapp/jupyter/cursors.tsx
```diff
@@ -39,137 +32,116 @@ interface CursorProps {
...
-    nextProps: CursorProps,
-    nextState: CursorState
-  ): boolean {
-    if (this.props.time !== nextProps.time) {
-      this.show_name(HIDE_NAME_TIMEOUT_MS);
+export const Cursor: React.FC<CursorProps> = React.memo(
+  (props: CursorProps) => {
+    const { name, color, top, time, paddingText } = props;
+
+    const mounted = React.useRef<boolean>(false); // TODO: don't do this
```
williamstein, line 39: FYI, we have an official "am I mounted?" hook called useIsMountedRef.

Issue #5365, opened 2021-05-31
ipynb→pdf: broken for filenames like foo[AS-3A].ipynb


haraldschilly:
1. create a notebook named `foo[AS-3A].ipynb` or similar
2. file → export pdf via latex
3. nbconvert fails ... either it's an upstream bug or somehow the filename isn't passed down correctly.

workaround: rename your file or duplicate it. those square brackets seem to trigger a search pattern.

if that's truly in nbconvert (i.e. it tries to be smart) we have to switch to a temporary filename to avoid this – or maybe there is a command line option to disable this?

```
Error: Traceback (most recent call last):
  File "/usr/local/bin/jupyter-nbconvert", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.8/dist-packages/jupyter_core/application.py", line 270, in launch_instance
    return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/traitlets/config/application.py", line 663, in launch_instance
    app.initialize(argv)
  File "<decorator-gen-7>", line 2, in initialize
  File "/usr/local/lib/python3.8/dist-packages/traitlets/config/application.py", line 87, in catch_config_error
    return method(app, *args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/nbconvert/nbconvertapp.py", line 285, in initialize
    self.init_notebooks()
  File "/usr/local/lib/python3.8/dist-packages/nbconvert/nbconvertapp.py", line 314, in init_notebooks
    globbed_files = glob.glob(pattern)
  File "/usr/lib/python3.8/glob.py", line 21, in glob
    return list(iglob(pathname, recursive=recursive))
  File "/usr/lib/python3.8/glob.py", line 57, in _iglob
    yield from _glob1(dirname, basename, dironly)
  File "/usr/lib/python3.8/glob.py", line 85, in _glob1
    return fnmatch.filter(names, pattern)
  File "/usr/lib/python3.8/fnmatch.py", line 52, in filter
    match = _compile_pattern(pat)
  File "/usr/lib/python3.8/fnmatch.py", line 46, in _compile_pattern
    return re.compile(res).match
  File "/usr/lib/python3.8/re.py", line 252, in compile
    return _compile(pattern, flags)
  File "/usr/lib/python3.8/re.py", line 304, in _compile
    p = sre_compile.compile(pattern, flags)
[...]

PR #5366, opened 2021-05-31
webapp/nbconvert: sanitizer for filenames

haraldschilly:
# Description

see #5365


williamstein:
As far as I can tell this PR introduces a function `export function sanitize_nbconvert_path(path: string)`.  I have some issues with this:

1. I can't tell what sanitize_nbconvert_path does based on the comment above.  Also, this function is really long and scary, and your comment "this is a bit tricky" gives me no confidence.
2. It seems like you're assuming the only problem with globs is with square brackets.  This isn't true.  The problem is with left square brackets, asterisks, and question marks as I figured out [here](https://github.com/jupyter/nbconvert/issues/911#issuecomment-852230805).

Also, as I mention in the link above, the source code of `glob.escape` in Python is basically one line:
```
>>> import re
>>> a = re.compile('([*?[])')
>>> a.sub(r'[\1]', 'foo')
'foo'
```

Can't you rewrite sanitize_nbconvert_path to use the same code (but in javascript)?  It would be much shorter, easier to understand, and cover all cases....?

williamstein:
(I do see this is work in progress, but hope these comments are helpful.)

haraldschilly:
Ok, yes, that's more elegant. I think I've finished this, at least the two most prominent nbconvert usages work and also slides. My test filename was `12[A-Z]*?.ipynb` (i.e. those square brackets need to be pairs for a pattern, and yes, that filename fails in prod). 

those screenshots show what's going on. for the normal nbconvert, the filename is changed in project's `jupyter.ts` and for the headless chrome variant, it's the util itself (such that the command can be used from the command line as well)

Screenshot from 2021-06-09 17-44-14

Screenshot from 2021-06-09 17-42-33

I also tested that "slideshow" frame in jupyter, which calls nbconvert as well.

PR #5370, opened 2021-06-10
webapp/jupyter: FC-ing components

haraldschilly:
# Description

This makes a lot of changes, but most of them are quite trivial. There were a few complications, though. I'm not 100% what's going on with the codemirror related components and how to test this. It's as close to the previous code as I could, though. What I didn't touch is `codemirror-static` because it is for the share server anyways? In any case, that's one that's left. The other component I didn't touch is `nbviewer`.

What did cause errors are widgets, that's why there is a try/catch in `widgets.tsx` where it cleans up the view. I made a comment in the code about that. It's definitely related to that `useEffect` cleanup, which happens slightly different than the `will unmount` before.

What I did test is everything else I could figure out. e.g. clicking entries in the menu, find/replace, download to pdf, … iframe with a website, widgets, embedding an interactive plot from bokeh, sage 3d plot canvas, long output, folding, uploading an image attachement to a markdown cell, exporting an autograder test cell to a student notebook, etc. all that works fine in dev.

regrading packages, I updated that lru cache (no api change) and removed rooks again. Do we use lru-cache anywhere else? I don't see a warning that versions diverge, though.

Also, there are more hooks in hook.ts now. I think it's better to have such simple ones under our control, especially when some of them have a weird api in other projects.


williamstein:
> Do we use lru-cache anywhere else?

I feel like we've used it all over the place over the years.  I like it in any case.

PR #5406, opened 2021-07-04
package moved to namespace

accursoft:
original version is deprecated


williamstein:
Thanks.  I'm going to instead uninstall this, since it turns out we don't use this jquery plugin anymore.

PR #5432, opened 2021-08-02
hub/analytics: compile analytics-script.ts with a different config.

haraldschilly:
# Description

problem is, firefox crashes right upon start, because it doesn't know how to deal with that implicit `exports` variable in `/analytics.js` ... the code started with `Object.defineProperty(exports,"__esModule",{value:!0}); ` and with this fix here, it's gone

This is an attempt to compile that file differently … in particular to target browsers. I hope I understood this right, i.e. the file ends in `./dist/analytics/` now

made this a blocker, because fixing this holds us back from updating the hub

### DIFF
--- a/src/smc-hub/analytics.ts
+++ b/src/smc-hub/analytics.ts
@@ -31,7 +31,9 @@ import { getLogger } from "./logger";
 // dist/analytics.js and also analytics-script.ts
 // gets compiled to dist/analytics-script.js.
 export const analytics_js = UglifyJS.minify(
-  fs.readFileSync(join(__dirname, "analytics-script.js")).toString()
+  fs
+    .readFileSync(join(__dirname, "analytics", "analytics-script.js"))
+    .toString()
 ).code;
 
 function create_log(name) {
--- a/src/smc-hub/analytics/analytics-script.ts
+++ b/src/smc-hub/analytics/analytics-script.ts
@@ -79,6 +79,3 @@ window
   })
   //.then(response => console.log("Success:", response))
   .catch((error) => console.error("Error:", error));
-
-// so it is a module.
-export {}
--- a/src/smc-hub/analytics/tsconfig.json
+++ b/src/smc-hub/analytics/tsconfig.json
@@ -1,0 +1,9 @@
+{
+  "extends": "../../packages/tsconfig.json",
+  "compilerOptions": {
+    "rootDir": "./",
+    "outDir": "../dist/analytics/",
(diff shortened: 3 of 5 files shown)


williamstein:
Here's the code you wrote that is crashing:
```js
  // 2. CoCalc analytics
  const ctag = w.document.createElement("script");
  ctag.src = join(w.app_base_path, "analytics.js?fqd=false");
  ctag.async = true;
  ctag.defer = true;
  w.document.getElementsByTagName("head")[0].appendChild(ctag);
```

Here's an article about the correct syntax to import es6 modules, https://www.sitepoint.com/using-es-modules/, and it looks like this:
```html
<script type="module">
  import { tag } from './html.js'
  const h1 = tag('h1', '👋 Hello Modules!')
  document.body.appendChild(h1)
</script>
```
So the right fix is to import analytics.js properly as an ES6 module.  

I'll do this later today, rather than this PR.

haraldschilly:
Oh, the link has some pointers. So, in the original code above, just `ctag.type = "module" ` should be enough to fix this.

williamstein:
> just ctag.type = "module" should be enough to fix this.

NICE!

williamstein:
There are two other places that mention analytics.js:

 - a POST request to send account_id -- probably doesn't matter
 - in base-page.tsx, where adding another type="module".

I've made these two changes in my live code at https://cocalc.com/projects/10f0e544-313c-4efe-8718-2142ac97ad11/files/cocalc/src/smc-webapp/share/base-page.tsx?session=default and will try to figure out how to test this later today...  Maybe I'll just push it and build on test.cocalc .com

williamstein:
Sadly using type="module" makes no difference:

Screen Shot 2021-08-02 at 12 56 54 PM

williamstein:
Way to easily reproduce this with various approaches in Firefox console:

image

```js
$("body").append($('<script type="module" src="/analytics.js?fqd=false" async="" defer=""></script>'))
```

haraldschilly:
so, one thing I don't really know is what "common js" vs. "es<number>" really means. the lower es-number are browsers, and well, a low one and not a module should work universally. maybe here in this case, this is more than a "module" and actually this common js setup? My feeling is this is only for node.js … but I never looked into this in any detail.

williamstein:
OK, I ended up just defining exports if it isn't defined (so on Firefox only for some reason...?) as part of the uglifyjs usage higher in analytics.ts.  This actually seems like the least hack-ish simplest approach.

PR #5446, opened 2021-08-13
X11: octave gui and spyder

haraldschilly:
# Description

* if octave gui works, check if there is a simple wrapper script. that way it won't be activated if gui is broken
* fix spyder exe name


williamstein:
I like this; at least I can use it to workaround that weird permissions issues on docker too.

PR #5456, opened 2021-08-18
frontend: support assembler code

haraldschilly:
# Description

my goal here is to support assembler code in `*.asm` files. problem: syntax highlighting doesn't do anything. strange. I must be missing some detail.

ref: https://codemirror.net/mode/gas/index.html


williamstein:
I fixed and improved it in some follow up commits and it will be live soon...

haraldschilly:
👍🏻

PR #5519, opened 2021-10-15
project/dedicated disk: symlink for convenience

haraldschilly:
# Description

this creates a symlink to the directory where dedicated disks are mounted. very similar to "data". it only creates the link if there isn't already a file or link with that name in the home dir and the disk actually exists.


williamstein:
I merged and addressed my trivial comments.

PR #5565, opened 2021-11-17
next/header: change double dash to long dash

haraldschilly:
# Description

such a double dash doesn't look good, long dash suits better


williamstein:
Thanks.  This is in my "muscle memory" from using LaTeX, where "--" is an endash.

PR #5575, opened 2021-11-24
frontend: no need to set version cookie only for https 

haraldschilly:
# Description

* this cookie is not really a secret
* Date.now is more concise


williamstein:
Agreed.

PR #5595, opened 2021-12-07
dedicated VM: dynamically caluclate the per project quotas 

haraldschilly:
# Description

the general motivation here is to leave 2gb of ram on each node for system services. this adds a function to derive the quota limits form the machine spec string. besides removing a source of typo errors, adjusting the numbers later on will be much easier.

~~I have to add I haven't tested this yet, only step left, but it should be fine.~~

I've also written a function to derive the price directly from the specs. This results in almost the same numbers, hence, lots of code, no changes ;-)

new: (the example 3 is different from what's online right now, because it is twice the memory)

Screenshot from 2021-12-08 16-14-21

### DIFF
--- a/src/packages/util/upgrades/dedicated.ts
+++ b/src/packages/util/upgrades/dedicated.ts
@@ -11,6 +11,8 @@ import {
   dedicated_disk_display,
 } from "@cocalc/util/types/dedicated";
 
+type VMFamily = "n2";
+
 // derive price to charge per day from the base monthly price
 function rawPrice2Retail(p: number, discount = false): number {
   // factor of 2 and an average month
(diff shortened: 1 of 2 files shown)


src/packages/util/upgrades/dedicated.ts
```diff
@@ -19,52 +21,86 @@ function rawPrice2Retail(p: number, discount = false): number {
...
+  return {
+    price_day: getDedicatedVMPrice(data),
+    spec: quotas, // the spec for actually setting up the container and communicated publicly
+    quota: { dedicated_vm: spec },
+  };
+}
 
-for (const vmtype of VMS_DATA) {
+// generate all dedicated VM specs we want to offer
+function* getVMData() {
```
williamstein, line 93: Fancy :-)

PR #5598, opened 2021-12-08
database/always_running: does not respect admin settings

haraldschilly:
# Description

pretty minimal, shouldn't impact the DB, since it is already indexed. looks like this case was not ported over? in any case, what's missing is updating the DB version, but I don't want to cause a merge conflict.

### DIFF
--- a/src/packages/database/postgres/always-running.ts
+++ b/src/packages/database/postgres/always-running.ts
@@ -10,13 +10,15 @@ import { PostgreSQL } from "./types";
 // but are not in the running or starting state.
 // We only check for stable states, i.e. there is no state transition going on right now.
 // Ref: @cocalc/util/compute-states.js
-// Performance: only an `x IN <array>` clause uses the index, not a `NOT IN`.
+// Performance:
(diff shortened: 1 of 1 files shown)


williamstein:
If we make this change, won't cocalc start all projects that have the always running admin setting checked, but are currently stopped?  You explicitly argued against doing precisely this, and we changed it so that such projects aren't started (which is why only the run_quota is is the current query).  I seems to me like this PR is exactly half breaking something that you wanted from long ago?

haraldschilly:
oh wait, that's a good point. I certainly got confused by this when working on cocalc cloud, but I see the point. I should check the DB.

haraldschilly:
ok, well, I get your point and this certainly has an effect, but I think it's fine. Unless I fundamentally miss something, for us here in production, there are zero projects that would be started. i.e. I checked via `select project_id, state ->> 'state', title from projects where settings ->> 'always_running' = '1';` (15 projects, all "ours", all are running). So, I'm ok with that change, it makes this setting consistent.

williamstein:
In that case, i'll merge it.

PR #5604, opened 2021-12-10
next/share: tell search engines to "noindex" if unlisted

haraldschilly:
# Description

if something is unlisted, tell search engines to actually not index it

please look carefully at what I wrote, first time I looked at that code

### DIFF
--- a/src/packages/next/components/share/layout.tsx
+++ b/src/packages/next/components/share/layout.tsx
@@ -22,7 +22,7 @@ interface Props {
 }
 
 export function Layout({ title, children }: Props) {
-  const { siteName } = useCustomize();
+  const { siteName, noindex } = useCustomize();
   return (
     <>
       <Head>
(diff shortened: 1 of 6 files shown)


williamstein:
I will rewrite this and add better docs.  The intention with useCustomize is it is a context hook that provides global information about customization of the server configuration (e.g., the site name, is anonymous sign up allowed, etc.).  It thus doesn't make any sense to include information in useCustomize that is specific to a given page.   I doubt I documented this at all.

That said, I ended up later including information about the signed in user optionally in useCustomize, with the argument being that it is used to "customize how you see the current page".  I think that was a mistake and I should put the account stuff in a different hook.   

I'll take over this PR and finish it tomorrow...

haraldschilly:
ah ok, well, maybe rename it to `usePageConfig` ? or another one with such a name?

williamstein:
Are you OK with me defining "customize" to mean "customize how the page is rendered". Then it's fine to include noindex in there and account.  It's very general, but that's not necessarily bad.

PR #5605, opened 2021-12-11
share for authenticated

haraldschilly:
# Description

~~well, I tried to start #5603  … but as you predicted, no idea how to implement the option for the frontend. It just fails with `Cannot coerce: no field 'authenticated' in table public_paths` and I wished I know where I can add that field.~~ ... ok, this was fixed, see below

I should probably just look into the server side part → more or less done, as far as I can tell

---

summary:

* util: adds a column `authenticated` to the public paths table. there is also a high level definition about the meaning of all 3 boolean flags in `util/consts/ui.ts`. 
* frontend: if the kucalc site setting isn't for `cocalc.com`, a 4th option "Authenticated" is displayed in the dialog for the share config. 
* frontend: account/public paths: add option to filter by "authenticated" as well, and for that
* database: return authenticated field for that "all public paths" query.
* next: similar to the above, when editing a shared file there is also such a 4th option to choose "authenticated"
* next: if "authenticated" is true and there is essentially no account_id, you can't see the file (a 404 just as if it doesn't exist). 
* next: similarly, if you're not signed in, you don't see such a shared file in the list of files on those project or account pages.


williamstein:
See QUERIES at the top of https://github.com/sagemathinc/cocalc/blob/53fb19e08b0e827d909f8fea1c3356a22e901d42/src/packages/frontend/project_actions.ts
where you must add "authenticated".

williamstein:
@haraldschilly are you still stuck on this?  I just fixed the conflicts.

haraldschilly:
I'll check. I was tied up with other stuff.

PR #5623, opened 2021-12-29
Server settings readonly

haraldschilly:
# Description

This is an extension to controlling those server settings via environment variables. I.e. there is a new column "readonly" and it's set to true if the hub sets them in "kucalc" mode. Then, the UI blocks and informs the admin about that in the server settings. Screenshot below.

Additionally, I made a change in the db-schema, where something that looks like a check is moved to a `check_hook` and I also added a protection to block readonly fields from being written to. 

Finally, the validity of env variable values is checked as well – and reported back to the admin as an exception.

The packages to update are hub+server+util for the additional field and checks, and obviously frontend to make use of it. In particular, `frontend/admin/site-settings.tsx` queries for a new field which requires updating the hubs first. … but if not, I checked by reverting the hub changes and the UI just says `user get query not allowed for site_settings.readonly` without unfolding the form. 

Screenshot from 2021-12-29 11-28-41

### DIFF
--- a/src/packages/database/postgres-server-queries.coffee
+++ b/src/packages/database/postgres-server-queries.coffee
@@ -245,7 +245,7 @@ exports.extend_PostgreSQL = (ext) -> class PostgreSQL extends ext
                     'name::TEXT'  : opts.name
                     'value::TEXT' : opts.value
                 if opts.readonly?
-                    values.readonly = !!values.readonly
+                    values.readonly = !!opts.readonly
                 @_query
                     query    : 'INSERT INTO server_settings'
                     values   : values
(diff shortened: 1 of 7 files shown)


williamstein:
>  which requires updating the hubs first. …

Thanks. I think that's all only relevant if I'm updating cocalc.com itself, which I'm not going to worry about since this new functionality isn't really relevant (it's just us doing admin after all, and we never change admin settings).  

I will of course update the packages since you use them in building cocalc-cloud.

williamstein:
On second thought, I'll very likely make these new packages live on cocalc.com, just to avoid issues when updating later... and because usage is very low right now due to the holidays.

PR #5637, opened 2022-01-06
record explicitly why a license is not providing upgrades and cleanup UI

### LINKED ISSUE #5635: licenses beyond run limit are removed, not deactivated
The situation I'm wondering about is the following:

1. I have a license with a run limit of 20 (but details won't matter)
2. I apply it to a regular project, it's active, providing upgrades, ok
3. I also apply it to 20 student projects, just to get it used up to the limit, and also start all student projects. The course project is also using that license. I.e. in total we're above 20 usages.
4. Now, I go back to my regular project and stop it, and then start again, the license is removed.
    - the DB really says `site_license = {}` now.
5. Furthermore, even if I add the license again and restart, it is immediately removed.

I don't like that the license is just gone. I would expect it stays with the project, but is eventually deactivated, not providing any upgrades.

This happens in kucalc as well as in a dev project.

Also, after explicitly stopping all student projects of step 3, I can add that license to my other regular project and restart – which activates the license as expected. So, this is 100% caused by hitting the run limit and can be reversed easily.

Also interesting, I read carefully what the box says, and during step 4/5 it claimed to be used 1 more than the limit. So, maybe just some slight inconsistency, but in any case, it was at or above the run limit.

Screenshot from 2022-01-06 18-40-47

haraldschilly:
# Description

goal is to fix #5635 while not breaking at least two other tickets

ok, I untangled the code and well, the core idea is to classify a license by several states: `"expired" | "exhausted" | "valid" | "future" | "ineffective"` – which is documented in the code. Only in the case of expired it is removed, in the other cases it stays around and in one of them it's actually used.

**change** after some discussion, don't remove an expired license. instead, explicitly record the status of a license in a new field (no impact for other services) and expose that status to the user. Then, it's 100% clear why a license doesn't provide upgrades.

this PR also cleans up the UI, i.e. present only the relevant information first (we can tweak this later), and everything else can be expanded (just like before)

also, I noticed some problems, and it turns out there is just a simple heuristic in the **database** to check if a license is being used. I modified those queries to ignore the "status" field. I don't know if there are more queries like that. Of course, once this is fully deployed, those queries should check for the status field + run state of the project. This is just a step in between...

testing: it starts to get interesting if you have several minimal licenses and apply them to projects and courses. i.e. you quickly run into a situation where only one of them is providing upgrades and then it is available for another project. the text in the previous alert box is sometimes wrong, I only made minimal changes to avoid making this PR even larger.

deployment:
**the frontend code must be deployed first** ... the reason is that there is this new field "status" in the license info and when viewing licenses, it tries to parse it as if it is a quota.
there are also changes to the database, they are backwards compatible.  the logger in "backend" is extendable now, like debug itself.

Screenshot from 2022-01-13 12-40-20

Screenshot from 2022-01-13 12-29-08


src/packages/database/postgres/site-license/hook.ts
```diff
@@ -34,218 +38,313 @@ interface License {
...
+    if (upgrades == null) {
+      // This is to make typescript happy since QuotaSetting may be null
+      // (though I don't think upgrades ever could be).
+      throw Error("bug");
+    }
+    const quota = license.get("quota");
+    if (quota) {
+      upgrades["quota"] = quota.toJS() as SiteLicenseQuotaSetting;
+    }
+    // remove any zero values to make frontend client code simpler and avoid waste/clutter.
```
williamstein, line 254: There's a function in @cocalc/util to do this...   `import { removeNulls } from "@cocalc/util/misc";
`

haraldschilly: well, I didn't know about this function, but it doesn't do the same. `removeNulls` checks `if (obj[field] != null) {` ... while here it also removes the entry if it is the number zero. I don't want to change this as part of this PR, just keep it as it was before.

src/packages/database/postgres/site-license/hook.ts
```diff
@@ -34,218 +39,309 @@ interface License {
...
+ * The *value* for each key records what the license provides and whether or
+ * not it is actually being used by the project.
+ *
+ * If the license provides nothing new compared to what is already provided
+ * by already applied **licenses** and upgrades, then the license is *not*
+ * applied.
+ *
+ * related issues about it's heuristic:
+ * - https://github.com/sagemathinc/cocalc/issues/4979 -- do not apply a license if it does not provide upgrades
+ * - https://github.com/sagemathinc/cocalc/pull/5490 -- remove a license if it is expired
```
williamstein, line 58: FYI: I disagree with "remove a license if it is expired"...  We've had users and myself find it confusing.

src/packages/database/postgres/site-license/hook.ts
williamstein: OK

src/packages/frontend/account/licenses/managed-licenses.tsx
```diff
@@ -75,11 +73,15 @@ export const ManagedLicenses: React.FC = () => {
     if (licenses.size == 0) {
       return <div>You are not the manager of any licenses yet.</div>;
     }
+
+    const site_licenses: SiteLicenses = licenses.toJS().reduce((acc, v) => {
```
williamstein, line 77: This code is hard for me to read.  Yes, I know what it does, but... ugh.  Why not just be more straightforward in this case?

Issue #5642, opened 2022-01-11
catch __save_local_view_state throwing "The quota has been exceeded"


haraldschilly:
there is at least 2x a chance of throwing `The quota has been exceeded` (code duplication) in `__save_local_view_state`. if that happens, localstorage should be deleted (or most of it?!)

workaround: delete this manually

- firefox: https://support.mozilla.org/en-US/kb/storage#w_delete-site-storage-for-individual-websites

PR #5643, opened 2022-01-11
frontend/editors: always use wrappers to access localStorage

haraldschilly:
# Description

address #5642

this is just the first step. I plan to do a follow up PR on top of this one which actually clears out parts  of local storage.


slel:
Thanks for this and for the follow-up at #5644!

PR #5644, opened 2022-01-11
Clear local storage heuristic

haraldschilly:
# Description

- follow up of #5644
- the main idea is to delete certain keys if there is an exception upon saving a value. This is mainly deleting the editor entries, if they're not recently used. In particular, most of the configuration is kept as a it is, including the session information, etc.


williamstein:
This is pretty cool.  I spent a while right now searching for an exiting LRU-cache implemented on top of localStorage, and strangely I couldn't find a single one!  That's really weird, since it's exactly what we actually need and it seems like it would be popular, and it's more or less what you just implemented (though yours is more tuned to our use case).   I found this comment by the lru-cache devs (lru-cache being super popular for in memory and something we use): https://github.com/isaacs/node-lru-cache/issues/106
and it's clear they seem to not even get what localStorage is (i.e., they suggest async-cache, not realizing that localStorage is NOT asynchronous).

Merging!

slel:
The description reads "follow up of 'Clear local storage
heuristic' #5644" -- did you mean some other issue or
pull request, such as #5642 maybe?

williamstein:
Yep, but it's too late now.

haraldschilly:
yes, I was searching for this a while ago, and I think the main obstacle is to not have a timestamp (for created and last accessed). that's IMHO a design flaw. All implementations that add a timestamp and search for it are a ways in terms of space and compute power. What I wrote introduces a list of LRU pointers, that's all we need. And also interesting, there are several keys showing up which aren't under "our" control. They must come from jupyter lab or other services … that's why I focus removing entries to the ones we know about. … and yes, extracting this as a standalone "local-storage-lru" package would be great - it's small and self contained.

PR #5647, opened 2022-01-13
files: compact right hand side icons, adding view icon for selected file types

haraldschilly:
# Description

since there was a problem with viewing html files, this could help. also getting rid of that download button, more or less


williamstein:
This is a really good idea.

PR #5650, opened 2022-01-14
frontend: tweak "Pending" message if not running in production

haraldschilly:
# Description

a status message only makes sense for cocalc.com


src/packages/util/compute-states.ts
```diff
@@ -17,7 +17,43 @@
 //
 // The icon names below refer to font-awesome, and are used in the UI.
 
-export const COMPUTE_STATES = {
+export type State =
```
williamstein, line 20: @haraldschilly why do you think these 12 lines of code (explicitly listing all the states here) rather than 1 line of code `export type State = keyof typeof COMPUTE_STATES;` is better????   I think the way it was before with keyof is 100% equivalent typescript, easier to maintain (significantly less redundancy), etc.  I'm not sure why you changed it.

williamstein: And now I understand -- it is because you want to spell out the type of COMPUTE_STATES ahead of time.  That's a good idea.

PR #5665, opened 2022-01-19
frontend/account: show licenses tab in kucalc mode for cocalc and onprem

haraldschilly:
# Description

show account/licenses in `kucalc=yes` and `=onprem`, that's all. there are no restrictions for the "next" pages in place, good!


williamstein:
I don't 100% understand why you did this. How am I supposed to develop licenses in my dev project or other settings, if they only show up for cocalc.com and on-prem?   Also, do licenses not work for cocalc-docker? That might be the case, but if so, the solution is to make them work, rather than get rid of them entirely, right?

haraldschilly:
I develop them by switching to `kucalc=onprem|yes`  In turn, I don't understand why showing licenses has anything to do with "commercial". They're used to modify the run quotas of projects (memory, and in the future GPU etc.). Only the part that's about purchasing something is controlled by "commercial".

Regarding cocalc docker, I don't know if licenses have any effect, I've no idea. I just assume there are no cgroup quotas and hence they don't have a real effect.

haraldschilly:
I added a clause to show them in `commercial = true` mode as well.

PR #5669, opened 2022-01-20
frontend: switch to using @cocalc/local-storage-lru

haraldschilly:
# Description

* util:
  *  there was a local storage wrapper, but it's frontend only code … hence I moved this to frontend to keep things where they are (otherwise, `window` must be declared, etc.)
  * there is still a tiny bit of usage of localStorage (clock skew) … and that's using it directly with a try/catch. 
* frontend: 
   * I found various places where `localStorage` was used. there is also some ancient code in `editor.coffee` … which I converted to typescript and kept everything backwards compatible
* what I tested:
  * jupyter notebook (newer code) and sagews (older editor code). ... open chat, split frames, move them, ... and refresh the page etc. I also debugged some of the local storage functions directly and they do work fine. 
  * I also first went back to the code currently in master, configured an markdown editor with a few frames and a sagews → then with the new code, checked if they still understood the config. all good.


williamstein:
Well this is seriously above and beyond!

williamstein:
I merged this thinking I understood it.  Then I realized it breaks starting projects, building next, etc., and that it's a lot of work for me to fix it, especially because fixing it involves a whole new package that is developed and published separately.  So since I'm the only one working on the master branch I just did a hard reset.  That said, I don't see a way to re-open this pull request. 

In any case, you can't refer to "window" at all and expect this code to work when imported on the backend.  So please fix all that.

PR #5721, opened 2022-02-09
frontend: update jupyter widgets and jquery

haraldschilly:
# Description

```
@interact
def _(r=slider(-3,3,0,25)):
    print(r)
```

is broken in sage 9.4 notebooks (and other version of sage) – but not something equivalent in python 3. the error is in jquery-ui's widget.jss itself.

- my first idea was to update jquery and jquery-ui. well, no difference.
- also updated jupyter widget packages
- my next idea was to go back to older versions of static, and well, even `app-1.7.0.html` from half a year ago shows that errors. that's very odd.
- next, I looked up the test case I use for end to end testing, and even more surprising, that test works – just like I remembered.
```
@interact
def func(k = slider(0, 10, .1)):
    print("k: %s" % k)
    s = (1 - sqrt(k))^2
    print("s: %s" % s)
```
- OK. got it. that 3rd argument is zero. That must throw off everything in some way. In any case, let's just keep this PR, it's fine to update these libs from time to time.


slel:
Related:

- [sage-support, 2022-01, Interact broken](https://groups.google.com/g/sage-support/c/FlLA_7tUb0M/)

haraldschilly:
how is this related? I don't understand.

slel:
Not related. Sorry for the noise.

PR #5741, opened 2022-02-17
next: custom 404 page

haraldschilly:
# Description

Screenshot from 2022-02-17 12-09-46

### DIFF
--- a/src/packages/next/pages/404.tsx
+++ b/src/packages/next/pages/404.tsx
@@ -1,0 +1,66 @@
+/*
+ *  This file is part of CoCalc: Copyright © 2022 Sagemath, Inc.
+ *  License: AGPLv3 s.t. "Commons Clause" – see LICENSE.md for details
+ */
+
+import { join } from "path";
+import Head from "next/head";
+import Footer from "components/landing/footer";
+import LandingHeader from "components/landing/header";
+import { Layout as AntdLayout } from "antd";
+import basePath from "lib/base-path";
+import getCustomize from "@cocalc/server/settings/customize";
+import { Icon } from "@cocalc/frontend/components/icon";
+import { COLORS } from "@cocalc/util/theme";
+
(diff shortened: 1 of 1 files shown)


williamstein:
fyi - @haraldschilly you absolutely can't do getStaticProps for this page, because it depends on information only available at runtime (e.g., to render the header), and the point of getstaticprops is "if you export a function called getStaticProps (Static Site Generation) from a page, Next.js will pre-render this page at build time using the props returned by getStaticProps.".

williamstein:
> yes, I know about data being stale, it's just that showing any recognizable info and a link back to the min page is IMHO still better than the blank page saying nothing useful. 

The way you've code this, it completely breaks building cocalc-docker, since it's not possible **at build time** to run the getStaticProps function at all.  This is a blocker that has to be fixed.

> I didn't want to spend time on some client side fetching and processing of data, that could be a new enhancement ticket of its own!?

No matter what this won't be client side.  it's always definitely going to be server side, like most of our pages (e.g, the landing page). That's what
```export async function getServerSideProps(context) {
  return await withCustomize({ context }, { name: true });
}
```
does.   Just change the code from getStaticProps to getServerSideProps so it isn't completely broken?

haraldschilly:
meh, github just swallowed my reply. There is no `getServerSideProps`. next.js argues this might overwhelm the server if there are too many requests, security, blabla → https://nextjs.org/docs/messages/404-get-initial-props

With that "client side query" enhancement, what I mean is to `useEffect` a query our `customize` endpoint, get the `siteName` directly from there, etc. I didn't want to do this. So, with that, let's just go ahead and hardcode the name or leave it out.

williamstein:
Showing a page with a custom server side headers is something we already do properly and have for a long time. Look at pages/index.tsx. Using useEffectis very much the wrong solution here, as it is much slower.   

I can take this over.

williamstein:
I now see that 404 is a special case and getServerSide props is not allowed.

PR #5759, opened 2022-02-24
frontend: React.FC-ing FrameTree+Leaf+FormatBar

haraldschilly:
# Description

this gets rid of two components used widely, hence there is some chance to introduce a problem, but I wasn't able to find anything during testing. the only change in functionality is sorting of colors – the old code accessed the "`.code`" attribute, which doesn't exist. I stayed true to that intention and well, now the colors are sorted by color.

### DIFF
--- a/src/packages/frontend/frame-editors/frame-tree/editor.tsx
+++ b/src/packages/frontend/frame-editors/frame-tree/editor.tsx
@@ -6,12 +6,11 @@
 import {
   React,
   CSS,
-  rclass,
-  rtypes,
-  Component,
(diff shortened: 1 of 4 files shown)


williamstein:
This will cause me some difficult merge conflict resolution with my branch... since I completely got rid of that css-colors menu and use the general component with all color selectors.  But I'll sort that out.

williamstein:
the commit getting rid of all this color stuff after merging this pr: c492ccd008c8196a68c97d2507a4284fc41af121

Issue #5761, opened 2022-02-25
pdf.js: chinese fonts don't load


haraldschilly:
when trying to render a PDF with chinese fonts in PDF.js, it isn't able to dynamically load additional fonts. error:

> Error during font loading: The CMap "baseUrl" parameter must be specified, ensure that the "cMapUrl" and "cMapPacked" API parameters are provided.

related issue seems to be: https://github.com/FranckFreiburger/vue-pdf/issues/190#issuecomment-788019420

test document:

```
\documentclass{article}
\usepackage{fontspec}
\usepackage{ctex}
\begin{document}
IMPORTANT: select xelatex

中文 这个问题解决了么 , "筲箕灣" and "柴灣".
\end{document}
```

WORKAROUND: native pdf viewer

PR #5762, opened 2022-02-25
Chinese LaTeX

haraldschilly:
# Description

- Ref #5761 
- add pdfjs-dist to our "CDN", but since it installs so many other packages, I added a whilelist mechanism. I also blacklisted typescript. That should avoid bloat. I hope this is ok and doesn't break anything? ... OK, I double checked the "dist" dir: "commander" is gone, "typescript" as well (intentional), and "pdfjs-dist" added. 
- fix the PDF.js config to load those additional files.

test:
- when opening the test document, I see a request to `https://cocalc.com/14eed217-.../port/5000/cdn/pdfjs-dist-2.9.359/cmaps/Adobe-GB1-UCS2.bcmap` and characters show up.

Screenshot from 2022-02-25 15-13-47

### DIFF
--- a/src/packages/cdn/package-lock.json
+++ b/src/packages/cdn/package-lock.json
@@ -14,7 +14,259 @@
         "codemirror": "^5.63.3",
         "katex": "^0.15.0",
         "mathjax": "^2.7.9",
+        "pdfjs-dist": "^2.9.359",
         "typescript": "^4.5.2"
+      }
+    },
+    "node_modules/@types/eslint": {
+      "version": "8.4.1",
+      "resolved": "https://registry.npmjs.org/@types/eslint/-/eslint-8.4.1.tgz",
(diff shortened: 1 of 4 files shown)


williamstein:
In cocalc-docker I had to install the 1GB package `apt install latex-cjk-all` to get this to compile.  Even then, whatever changes you made here didn't fix the problem at all with pdfjs and rendering.  Maybe they will somewhere else, but they don't in cocalc-docker for me.

haraldschilly:
Hmm, weird. I thought this is somewhat self contained inside of pdf.js. Maybe you need the fonts to actually compile the latex code to PDF? In any case, if you open that PDF file in the pdf.js viewer (the standard viewer, no need to do latex), it should issue a network request to load the binary file from a directory in pdfjs-dist. I could image in fails with docker, because of dynamic http requests via an insecure SSL/localhost connection.

PR #5769, opened 2022-03-02
jupyter: maybe fix a "TypeError: Cannot read properties of undefined 

haraldschilly:
# Description

There was a user reporting a `TypeError: Cannot read properties of undefined (reading 'handler')\n at HTMLDocument.eval`  … I don't now which document nor how to reproduce it, but just by going through the code I found this detail, which looks at least like a good idea to check.

### DIFF
--- a/src/packages/frontend/jupyter/output-messages/mime-types/register.ts
+++ b/src/packages/frontend/jupyter/output-messages/mime-types/register.ts
@@ -65,7 +65,8 @@ export function getHandler(type: string): Handler {
   if (h != null) return h.handler;
   for (const typeRegexp in HANDLERS) {
     if (type.match("^" + typeRegexp + "$")) {
-      return HANDLERS[typeRegexp].handler;
+      const hRegex = HANDLERS[typeRegexp];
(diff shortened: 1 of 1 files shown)


src/packages/frontend/jupyter/output-messages/mime-types/register.ts
```diff
@@ -65,7 +65,8 @@ export function getHandler(type: string): Handler {
   if (h != null) return h.handler;
   for (const typeRegexp in HANDLERS) {
     if (type.match("^" + typeRegexp + "$")) {
-      return HANDLERS[typeRegexp].handler;
+      const hRegex = HANDLERS[typeRegexp];
```
williamstein, line 68: This is a long way of writing `return HANDLERS[typeRegexp]?.handler;`

haraldschilly: well, that wouldn't return the `FallbackHandler` or do I miss something? I could rewrite all of that as well, but I didn't want to spend more time.

PR #5775, opened 2022-03-09
expose run quotas in project settings + usage

haraldschilly:
# Description

changes
- database: I don't know if `get_project_quotas` is actually used, but it was wrong. I modified it to use the famous `quota` function.
- frontend:
   - the boxes in the settings, they're now antd based. to make this look homogeneous, also components like the SSH keys needed some small changes. In any case, various `react-bootstrap imports are removed.
   - the main change is the area where quotas can be adjusted. that's now driven mainly by the "run quota". old quota upgrades are less dominant (only a summary if there are any).  The main hurdle is to get everything in one place.
   - for on-prem and docker, less information is shown. in particular `member_hosting` makes no sense for on-prem.
 - project:
    - more status information is required, but it's backwards compatible. 
    - I also added rate limiting, compact data via rounding, etc. so, yes, this will add additional traffic, but only very little.
 - details: the more I test the more small issues I find:
    - course: when user has no upgrades and clicks on the button to apply upgrades, nothing was shown. well, now at least it tells the user there are no upgrades and should get a license → 3cc912b9fc768
    - project control "stop idle projects": this didn't take the run_quota into account (well, it didn't exist before). this would cause projects to be stopped immediately, if the admin quota setting were ever to set "0" as the idle timeout → 813d90defe

**testing**: I deployed "pr-5775' tagged artifacts in our testing environment, where I found some edge cases … but now I'm confident to deploy this. 

**deployment**:
1. ⚠️  hub first  (because the `run_quota` is exposed read only)
2. project and static


haraldschilly:
@DrXyzzy once this is merged and deployed, it has an impact on documentation

williamstein:
Wow, this is a massive improvement.  :shrug:

PR #5777, opened 2022-03-10
Fix some typos in subset of src/packages

stweil:
Most typos were found by codespell.

Signed-off-by: Stefan Weil <sw@weilnetz.de>

### DIFF
--- a/src/packages/assets/misc/talk-20121004/index.html
+++ b/src/packages/assets/misc/talk-20121004/index.html
@@ -132,7 +132,7 @@ Stay tuned...
 
 
 <div class="slide">
-<h1>Longterm Programing Goals (part 1)</h1>
+<h1>Longterm Programming Goals (part 1)</h1>
 <ul>
 <li><b>Simple starter page:</b> where you can type Sage code in a
     box (no login required) with tab completion and interact support.
@@ -153,7 +153,7 @@ Stay tuned...
 </div>
 
 <div class="slide">
-<h1>Longterm Programing Goals (part 2)</h1>
+<h1>Longterm Programming Goals (part 2)</h1>
 <ul>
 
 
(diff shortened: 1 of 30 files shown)


stweil:
@haraldschilly, I tried to exclude third party code in my pull request. If there are still parts in this pull request which should be excluded I can do so, but need your advice.

src/packages/backend/package-lock.json
```diff
@@ -4407,7 +4407,7 @@
       "version": "0.11.0",
       "resolved": "https://registry.npmjs.org/node-pre-gyp/-/node-pre-gyp-0.11.0.tgz",
       "integrity": "sha512-TwWAOZb0j7e9eGaf9esRx3ZcLaE5tQ2lvYy1pb5IAaG1a2e2Kv5Lms1Y4hpj+ciXJRofIxxlt5haeQ/2ANeE0Q==",
-      "deprecated": "Please upgrade to @mapbox/node-pre-gyp: the non-scoped node-pre-gyp package is deprecated and only the @mapbox scoped package will recieve updates in the future",
+      "deprecated": "Please upgrade to @mapbox/node-pre-gyp: the non-scoped node-pre-gyp package is deprecated and only the @mapbox scoped package will receive updates in the future",
```
haraldschilly, line 4410: this is a package lock file, they're automatically generated by `npm`, the package manager behind javascript. changing it directly is either pointless or someone asks for troubles :-)

src/packages/assets/misc/talk-20121004/index.html
```diff
@@ -132,7 +132,7 @@ <h1>sage-support mailing list a few days ago</h1>
 
 
 <div class="slide">
-<h1>Longterm Programing Goals (part 1)</h1>
+<h1>Longterm Programming Goals (part 1)</h1>
```
haraldschilly, line 135: wow, I didn't realize a talk from 2012 is still in the code. this starts to be of historic interest :-)

haraldschilly:
I made two comments about files we should probably keep as they are. All the other changes look good to me, especially those in "next", because they're actually seen by users.

src/packages/assets/misc/talk-20121004/index.html
williamstein: Also "Programing" is a correct spelling of that word, which is more popular in American English.  

> "Long story short, either spelling of “programming” is acceptable and its popularity generally affects its trends in the English language. However, “programming” with two “m’s” is less likely to be docked by a Word processor set to American English." -- https://dictionarykiwi.com/2016/10/17/programming-versus-programing-which-is-preferred/

williamstein:
I'll merge this as is.  I don't care about spelling of programming (I like two m's better, actually), and the package-lock is just a comment, so will probably get reverted next time we update it.

PR #5778, opened 2022-03-11
Add missing semicolons (fixes recommendations from LGTM)

stweil:
Signed-off-by: Stefan Weil <sw@weilnetz.de>

### DIFF
--- a/src/packages/assets/misc/talk-20121004/ui/default/slides.js
+++ b/src/packages/assets/misc/talk-20121004/ui/default/slides.js
@@ -145,7 +145,7 @@ function go(step) {
 		ne = document.getElementById('slide0');
 		snum = 0;
 	}
-	if (step < 0) {incpos = incrementals[snum].length} else {incpos = 0;}
+	if (step < 0) {incpos = incrementals[snum].length;} else {incpos = 0;}
 	if (incrementals[snum].length > 0 && incpos == 0) {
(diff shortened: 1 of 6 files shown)


stweil:
LGTM currently shows more than 400 [alerts for cocalc](https://lgtm.com/projects/g/sagemathinc/cocalc). Some of them are really serious, but require much more work than the trivial changes in the pull request here.

haraldschilly:
well, the last two snippets are in the actual code, but I don't think this is serious. The other changes (assets) are an old talk and well, I wonder why threejs is there. This is probably a directory we should rather get rid of. I have to check …

williamstein:
It seems harmless to merge this.  Add getting rid of the three.js code as a new issue for later?

stweil:
> well, the last two snippets are in the actual code, but I don't think this is serious

None of those missing semicolons is serious. It's more an issue to get uniform code. That's why LGTM classifies it as recommendations.

I think it is good to address recommendations nevertheless, of course without forgetting to address the real errors as well.

williamstein:
> I think it is good to address recommendations nevertheless

It's all merged.

williamstein:
Incidentally, in our own codebase we frequently run prettier with default options, which should always add in semicolons.  Our official style policy for our code is "prettier with the default settings".

PR #5790, opened 2022-03-18
Next/SSO

haraldschilly:
# Description

refactors everything around authentication, improves what the info/conf row in the database should contain, improve passport configuration, React.FC-ing landing/sign-in/passport components, ... the main new addition are dedicated  `/sso/*` pages: an index page and one more detailed page for each mechanism. non-public SSO mechanisms are also hidden by default from the sign-up/in pages, unless they're configured to show up. that avoids unnecessary clutter. However, if you try to sign up with an exclusive domain, the pages point you to the SSO endpoint. Exclusive domains also block you from linking with other methods and changing your email address (both could be used to circumvent the control of the external sso mechanism). Also, you can't change your email address to have a domain, which is currently governed by an exclusive SSO

**tests**
- create "normal" account, change password, etc.
- if the SSO mechanism provides the name fields and set to update on sign in (by default that's off), it "fixes" your name fields
- sign in using exclusive SSO, grabs existing account with the same email address, forces you to go through its sign in process, etc.
- you can't unlink from an exclusive SSO, nor can you change your email address
- tested in particular is OAuth2 and SAML, local servers, and SAML via Google Workspaces. OIDC does work, but passport.js does not support some aspects (PKCE?) … so, not 100% OIDC. 
- as public SSO mechanisms, only google is tested – but all 3 other ones are the same, no change in behavior
- password reset

**deployment**:
1. hub (since there is a new DB field). the one I tested has tag `2022-08-24-153746-next-sso`
2. then static: tested tag is `2022-08-24-142303-next-sso`
3. force "new version" (version is already incremented)

rollback: it's possible to revert the hub (static not so much due to the version change, and it's changes are less deep, so, should be fine)

some screenshots

Screenshot from 2022-03-25 15-53-58

Screenshot from 2022-03-25 15-53-26

Screenshot from 2022-03-25 16-02-48

Screenshot from 2022-03-25 16-46-36


williamstein:
Wow!

haraldschilly:
far from ready, though 🙈

PR #5792, opened 2022-03-21
Join method from path module should not be used for joining URIs part…

belonesox:
Join method from path module should not be used for joining URIs part, because it eats double slashes and make incorrect URI schema in URLs.

The bug that fixed here looks as 

image

### DIFF
--- a/src/packages/frontend/collaborators/project-invite-tokens.tsx
+++ b/src/packages/frontend/collaborators/project-invite-tokens.tsx
@@ -219,7 +219,7 @@ export const ProjectInviteTokens: React.FC<Props> = React.memo(
           <br />
           <br />
           <CopyToClipBoard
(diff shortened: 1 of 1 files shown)


williamstein:
I think this change will definitely break CoCalc in many settings, e.g., when the app base path is nontrivial.  Our definition for the base path is given here:

https://github.com/sagemathinc/cocalc/blob/master/src/packages/backend/base-path.ts#L12

```
[...] It must start with a leading slash /, but does not
end with one unless it is '/'.  It also should not include ://.
Some examples of valid basePaths:
- /
- /10f0e544-313c-4efe-8718-1111ac97ad11/port/5000
...
```

This code in this PR
```
${base}app
```
will work when the basepath is `/`, but NOT in any other cases.

I will merge this as is and then fix it, since of course I care the most about the actual bug being fixed ASAP.

Thanks again for your contribution!!

PR #5795, opened 2022-03-23
Boost licenses backend

haraldschilly:
# Description

this implements all the special handling to process `{quota: {boost: true}}` licenses. I should deploy this in test to double check if it works – apart from that its done.

- built packages tagged `pr-5795` and deploying hub as `2022-04-05-142827-boost-pr-5795` 
- and manage: `2022-04-05-125142-boost-pr-5795`
- I tested this and found one final tricky bug. It runs fine in "test".

deployment:
- manage must be first, because the effective quota is calculated based on the licenses set in the hub (backwards compatible, of course)
- then at least the websocket hub, because that's where the "hook", for deciding which licenses to use, runs.

TODO:
- [ ] there is an inconsistency if there is no regular license at all
- [ ] figure out why hub crashes

## [Checklist](https://github.com/sagemathinc/cocalc/wiki/PR-Checklist):
- [ ] Testing instructions are provided, if not obvious
- [ ] Release instructions are provided, if not obvious

PR #5796, opened 2022-03-23
Boost and Dedicated Resources Licenses

haraldschilly:
# Description

- step 2 after #5795 is done
- basically, this is done. the major cornerstones are:
   - 2 additional sections in the "store"
   - more diverse and stricter typing of everything that goes on behind the scenes – otherwise I would have lost my sanity
   - boosts: those licenses are basically "regular" licenses, except that their minimum quotas start at 0 and in turn, their upper limit is capped by the default minimum you can get. their activation depends on an active regular license of the same type. same type? well, idle timeout and member hosting must match.
   - dedicated disks: subscription with a fixed monthly fee, which derives from GCPs fee. there is no room for discounts, since this is not a shared resource or any other effect is in place.
   - dedicated VMs: for a fixed interval a VM is provisioned. no room for discounts as well.
- additionally
   -  bugfixes for the license activation (ironing out edge cases which are relevant for regular licenses as well). also, there is now a warning in the admin license editor form about editing dedicated resources (it's not possible). 
   - starting to change the memory using to "G" everywhere
   - typescripting "utils/theme.js→ts` because I found a wrong use and well, after that I also fixed the fallout

what's left: I revisited the management service for all this and well, looks good. I'll deploy this in "test" next.

regarding packages: there are **no** changes to project (only a refactoring detail), cdn, assets and sync. hub has direct code change, only indirectly via hook.ts in the database package.

hence for deployment:
- static: the new interface elements (showing dedicated resources), admin … must come first
- hub
   - websocket because that's the one doing the evaluation for the site license config
   - next: the store, to be able to purchase this


src/packages/frontend/project/explorer/path-navigator.tsx
```diff
@@ -23,44 +23,46 @@ export const PathNavigator: React.FC<Props> = React.memo((props: Props) => {
...
-
-    const is_root = current_path[0] === "/";
 
     const current_path_depth =
       (current_path == "" ? 0 : current_path.split("/").length) - 1;
     const history_segments = history_path.split("/");
+    const is_root = current_path[0] === "/";
+
+    v.push(
+      // yes, must be called as a normal function
```
williamstein, line 33: Ooops  -- don't leave a comment in "yes, must be called as a normal function" when the code *doesn't* do that anymore.

williamstein:
> what's left: I revisited the management service for all this and well, looks good. I'll deploy this in "test" next.

Your actually deployment plan doesn't mention manage at all, so I'm worried that it must be incomplete.  Could you read over your deployment plan again and fill in any missing details?   Alternatively, do the deployment yourself?

haraldschilly:
It's all good … but yes, I think it's best if I do this on my own. I can target Wednesday, because then I have enough time to deal with arising issues.

williamstein:
OK, that's a good plan.  I did skim the code, but it's really too big to review, and only testing it can tell, which I assume you've done already.  Obviously, at least the style of the code looks very good.

Issue #5824, opened 2022-04-01
extract and postgresql 14


williamstein:
As explained here: https://github.com/sagemathinc/cocalc/pull/5818#issuecomment-1086039789

we need to make some changes to our code to support PostgreSQL 14, due to changes in how the `EXTRACT` command works in PostgreSQL.

williamstein:
There are I think 5 uses of this in our codebase...

haraldschilly:
yes, I'll look into this

PR #5825, opened 2022-04-01
postgres 14: EXTRACT returns a numeric, which is converted to a strin…

haraldschilly:
# Description

- see issue #5824
- the conversion in `postgres-blobs.coffee` is not necessary, I don't see why. Must be related to rethinkdb.
- the other client/server data serializations make sense. I've thought about this, and to me it makes sense to convert to a float (pg calls it "double precision")  like before. 

tests:
- [x] un-archiving a series of patches works fine
- [x] listings on the share server show timestamps when last edited
- [x] stopping idle projects … I just manually set the last edited field to something in the past and project stopped as expected


williamstein:
Very nice code!

PR #5842, opened 2022-04-07
Stripe fix rounding error

haraldschilly:
# Description

follow up of #5837


williamstein:
@haraldschilly -- heads up -- when I update cocalc to use ESM modules, I think it's going to be a big "patch bomb" that messes up any outstanding PR's.  This is the only outstanding PR right now.

PR #5902, opened 2022-05-03
k3d

williamstein:


williamstein:
I made significant progress:

Screen Shot 2022-05-04 at 10 16 21 PM

The main issues really deeply under the hood are problems with how I implemented buffers, both easy and subtle, and also hooks for custom serialization.   I ended up putting in some incorrect code to work around issues, but I now understand pretty much 100% exactly what to do to fully do this properly, and also be able to support other custom widgets.  I'll definitely also want to similarly fully support ipyvolume.

PR #5905, opened 2022-05-04
project/jupyter-lab: add the --collaborative flag

haraldschilly:
# Description

well, all this does is to add this flag. there is also an extension for sharing links, but I think this is pointless.


williamstein:
I know I suggested this, but I have some concern about just turning this on by default for everybody, since the current design is known to cause significant data loss and other major problems, with no "TimeTravel" to save you.

Basically, I remain concerned about: https://github.com/jupyterlab/jupyterlab/issues/12154

williamstein:
On the other hand, we do have filesystem snapshots, so maybe I shouldn't be so concerned.

PR #5911, opened 2022-05-05
frontend/latex: add "no output dir" options and some fixes to sanitization functions

haraldschilly:
# Description

- see #5910
- this also fixes nonsense sanitizations when someone selects "<disabled>" and similar issues

### DIFF
--- a/src/packages/frontend/frame-editors/latex-editor/actions.ts
+++ b/src/packages/frontend/frame-editors/latex-editor/actions.ts
@@ -9,11 +9,7 @@ LaTeX Editor Actions.
 
 const HELP_URL = "https://doc.cocalc.com/latex.html";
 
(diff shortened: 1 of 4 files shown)


williamstein:
Reminder to check for typescript errors:
```
../frontend/frame-editors/latex-editor/actions.ts:383:9 - error TS2532: Object is possibly 'undefined'.

383     if (cmd.get(0).startsWith("false")) {
            ~~~~~~~~~~

../frontend/frame-editors/latex-editor/build-command.tsx:10:16 - error TS6133: 'fromJS' is declared but its value is never read.

10 import { List, fromJS } from "immutable";
                  ~~~~~~
```

If you're using VSCode it often misses things, as compared to `npm run tsc`.

PR #5936, opened 2022-05-19
db/always-running: do not start deleted projects

haraldschilly:
# Description

1. deleted projects should not be started if they're set to always running


williamstein:
I've merged this and updated @cocalc/database (patch update)

PR #5954, opened 2022-05-27
a little refactoring to try an experiment with slate + latex

williamstein:


williamstein:
I tried this.  It was interesting, and I could definitely see how to do a reasonable mapping from the AST produced by unified-js from a latex document to a slate editable document.  Going the other way seems quite difficult without significant potential loss of information.  On the other hand, with some work, and significant limitations in what we parse (leaving big chunks unparsed), this may be of interest.  Closing, for now.

Issue #4896, opened 2020-09-29
Port the latex_envs jupyter extension to cocalc


williamstein:
See https://pypi.org/project/jupyter_latex_envs/

We could probably make this generic and work for more general markdown in cocalc - not just jupyter!  There is really nothing jupyter specific about this.  It also seems reasonably backwards compat.

williamstein:
I'm looking into this a little more carefully and unfortunately I'm less optimistic about how difficult this is.   The reasons are that:

- Looking at the page on latex_envs, it doesn't really say exactly what it does.  It justs lists some ambiguous bullet points.
- Looking at actual source code in latex_envs, I see things like this:
```
...
            data = data.replace('\\\\', '!sl!!sl!')
            data = re.sub(r'%([\S\s ]*?)\n', r'!cc!\1!nl!',data)
            #data = data.replace('%', '!cc!')
            data = data.replace("/begin", "\\begin")
            data = data.replace("/end", "\\end")
....
```
and I have no clue why or what is going on.  It certainly seems replacing "/begin" by "\\begin" is going to result in accidentally breaking a lot of normal things, i.e., this is a hacky ill-defined extension that you would only want to use when you explicitly need it.  When you need it, then it is probably very useful.

williamstein:
I did some more research on latex_envs. 
I read through some of the testing/example docs, and it seems like most of what latex_envs
does is basically take latex that current markdown parsers in Jupyter view as "broken latex formulas"
and interpret it using latex.  For example:

\begin{itemize}
\item Hello\item World
\end{itemize}

gets considered as latex via jupyter due to the \begin/\end tokens.  Then, since of course mathjax knows nothing 
about this non-math construction, it's an error.     This addresses one of my biggest concerns about
latex_envs, namely that it would involve adding a lot of incompatibilities to parsing markdown.  However,
that's not the case at all -- instead, it's mostly just making it so the current parsing mechanism works
in more generality.

Interestingly, this latex.js project I mentioned recently

https://latex.js.org/playground.html
[...]

PR #5982, opened 2022-06-03
support latex environments in markdown that go beyond basic math formulas (issue 4896)

williamstein:
#4896


williamstein:
Screen Shot 2022-06-03 at 12 43 15 PM

williamstein:
This shows mixing markdown and latex...

Screen Shot 2022-06-03 at 7 59 40 PM

williamstein:
I will close this for now, and reopen it if/when come back to this approach.

PR #5999, opened 2022-06-16
…

haraldschilly:
# Description

the main problem is that I saw a problem with collecting a student's assignment. it's not working. the UI just says the error is "true" and I don't know which part is really throwing the error. I somewhat think it has something to do with the code for writing a file to a project, because that's what collecting does as well. Hence that's why I've labeled this a blocker.

So far, I only modernized the writeTextFileToProject and ensureContainingDirectoryExists code, but that didn't help. I'm mayb also confused by how to test a change to the project code (I'm building the backend and project package, and then restart, but it doesn't seem to have an effect). … or maybe I'm just looking at something completely wrong.


williamstein:
Your changes fix the problem with this function (and uniformly improve it), so very likely the only issue is probably " how to test a change to the project code ".

williamstein:
I think your PR fixes the problem (after testing with make-dev).

PR #6003, opened 2022-06-17
proof of concept that it is possible to use our "static" app inside our next.js app

williamstein:
Critical foundations for things to come.


williamstein:
There's no reason to make this a PR, since I don't at all necessarily plan to merge this. This is just a proof of concept to see if it is even an option.  I don't at all necessarily think it is the best approach to this problem.

williamstein:
OK, this commit ce3edd8b71d1466789e6fd63f40872d0de994b65 is the alternative approach of using iframe embedding, and it is clearly a superior approach.

PR #6011, opened 2022-06-21
ipywidgets -- working on implementing custom messages

### LINKED ISSUE #1934: support interactive plots in CoCalc Jupyter notebooks
Also, ensure "%matplotlib notebook" works, which maybe uses widgets?

WORKAROUND: Use classical Jupyter mode, where widgets are fully supported.

williamstein:
This is to fix #1934.  It's probably 80% there.


williamstein:
I'll re-open this when I work on it again.

PR #6013, opened 2022-06-22
hub/db: database connectivity healthcheck + some overall fixes

haraldschilly:
# Description

well, figuring this out is confusing, but I think I get the general direction. The main problem is, if the DB is down, the healthckeck endpoint quickly deteriorates, because it depends on the database. so, the general solution is to make them work regardless of the DB. Main obstacle to get this right is the express server itself: it checks for a virtual host, with caching, and once that caching is over, the query fails. ~~What I don't understand right now is where exactly things go wrong. I tried to try/cache a couple of calls related to the database, but it's still some other call somewhere that breaks…~~

ok, I think I figured out what's going on. same for the metrics endpoint, which stops working as well. I also added try/catch blocks to at least log when errors happen.

I also figured out why sometimes the hub gives up on connecting to the DB – in one case there is a timeout for retrying. I removed this. Instead of timing out, I think it's better to continue trying … and if the health check fails at some point, the hub will be terminated. Also, there is now a limit to how many queries can pile up before the DB connects again.

also, found out why debug spits out all timestamps twice.


williamstein:
Awesome!!

PR #6065, opened 2022-08-03
frontend: let users add their license easily when they create a new project

haraldschilly:
# Description

This adds a bit of "antd" to the new project dialog and a dropdown to add or select a license. I also saw traces of giving projects a default name, which didn't work. It's fixed and I'm also adding a timestamp to the default name.

how it looks now:

Screenshot from 2022-08-04 10-55-38

and after expanding the add license part:

Screenshot from 2022-08-04 10-55-23


williamstein:
Note: we should add a "Buy a license" link somewhere in the vicinity...

PR #6133, opened 2022-09-19
onprem: license quota to make /ext read/writeable

haraldschilly:
# Description

- this is for onprem only. admins get a way to define `{ext_rw:true}` which mounts the readonly `/ext` directory as read/write. this is only for the license.
- it hides member hosting, that makes no sense for onprem
- this tweaks a bit the UI of the license admin panel
- it does not touch the embedded payment form, and for non-onprem there is no change in actual functionality


williamstein:
I wish this "sizewatcher" actually made sense, as this makes no sense:

image

PR #6135, opened 2022-09-23
onprem: load SSO configs via a file

haraldschilly:
# Description

This introduces a simple for administrators to configure the SSO settings, without having to fiddle around with the DB. Later, we can always implement something more sophisticated via a UI.

In any case, the idea is the env var `$COCALC_SSO_CONFIG` points to a file, containing a JSON config. This could easily be a k8s secret mounted as a file as well. If this file exists, upsert/delete operations are done accordingly, before the actual sso endpoint initialization happens.


haraldschilly:
and I looked into that size watcher, that's a bug in 1.2.1: https://github.com/adobe/sizewatcher/issues/52

PR #6145, opened 2022-09-29
store: url encode form settings

haraldschilly:
# Description

This encodes the form's state in URL query parameters, which makes it possible to send a link to a specific setting.  This is only for site-license and boost. 

One interesting test to check the functionality is to open a URL like `/store/site-license?uptime=medium&run_limit=100&ram=4`. The form is populated with default values, except for the 3  in the query.


williamstein:
Wow, soooooo good :-)

Issue #6156, opened 2022-10-05
package versions out of sync: mocha and parse-domain


williamstein:
According to
```sh
~/cocalc/src$ npm run version-check
```

mocha and parse-domain are out of sync version wise, caused by changes by @haraldschilly in the last month:

```sh
~/cocalc/src$ git blame packages/frontend/package.json |grep mocha
906f77ec119 src/smc-webapp/package.json        (William A. Stein  2021-05-16 20:37:05 +0000 175)     "@types/mocha": "^5.2.7",
83c0bb9f958 src/packages/frontend/package.json (Harald Schilly    2022-09-07 18:25:27 +0200 184)     "mocha": "^10.0.0",
~/cocalc/src$ git blame packages/server/package.json |grep parse-domain
8ca779a248c (Harald Schilly 2022-08-05 18:55:44 +0200 58)     "parse-domain": "^5.0.0",
~/cocalc/src$ git blame packages/hub/package.json |grep parse-domain
c62103f6f42 src/packages/hub/package.json (William Stein    2021-10-16 00:57:20 +0000  68)     "parse-domain": "^4.1.0",
```

@haraldschilly can you look into these?

haraldschilly:
yes, I'll look into this

PR #6162, opened 2022-10-07
npm: update mocha+@types/mocha+ts-mocha and parse-domain

haraldschilly:
# Description

see #6156 

this looks simple, but there is a little problem with newer parse-domain versions. 

> will now be released as native ECMAScript module. There's no CommonJS build anymore.

breaking change: https://github.com/peerigon/parse-domain/releases/tag/v6.0.0

ok, no idea how to fix this. My guess is, some upgrade must be done with the `tsconfig` settings.

What I did is to update only to version 5.


williamstein:
> ok, no idea how to fix this. 

This is the same nightmare rabbit whole I went down with JupyterWidgets, because they did the same thing.

See https://stackoverflow.com/questions/70396400/how-to-use-es6-modules-in-commonjs for what looks like a pretty good answer, namely "Simple answer: you don't. Either transpile them to commonjs, or use es modules all the way."

We are definitely not going to "use es modules all the way" anytime soon, since we have a massive amount of code, and es modules are not sufficiently ready (there are a lot of subtle problems that come up and some really ugly and unfixable issues, e.g., you have to write `import ... from "a/b.js"` instead of `import ... from "a/b"` everywhere, which is extremely ugly.  There's no end in site to that import problem, due to a fundamental disagreement between typescript and node devs.  I did run scripts to autoconvert tons of code, but it's just horrible.

So the option is to transpile from es modules back to commonjs.  This is doable in most cases, but there can be some subtle issues with this.  For anything involving webpack the solution is to use swc-loader (faster version of esbuild or babel).  For plain typescript code targetting node, I'm not sure at the moment if typescript can fully do this, or if you also have to involve swc or esbuild or babel.

Issue #6218, opened 2022-11-25
fix the many broken tests in src/packages/server/prices.test.ts


williamstein:
Maybe half the tests in src/packages/server/prices.test.ts are broken, with the numbers and times not matching up.  

I think @haraldschilly spent a lot of time writing these tests, then fixing related issues that @DrXyzzy  found related to pricing.  Thus it would be good to follow up and fix the tests.  Maybe the new output is all correct and the tests need to be updated?  I have no idea.

PR #6222, opened 2022-11-28
test server/prices: fixing tests 

haraldschilly:
# Description

see #6218

this fixes the price calc tests (they changed with the price increase) and also added tests for the midnight rounding


williamstein:
Stoked that  I already know this passes tests due to https://github.com/sagemathinc/cocalc/actions/runs/3564161273

PR #6224, opened 2022-11-28
project: optionally launch the sshd server

haraldschilly:
# Description

this moves the sshd server launch to the project, i.e. instead of doing this in the init.sh file. later, we can figure out if we ever want to customize this, get rid of the start script, etc.


williamstein:
The CI caught a typescript mistake!   https://github.com/sagemathinc/cocalc/actions/runs/3565955280

haraldschilly:
> The CI caught a typescript mistake! 

that's outdated. I wonder if there is a way to automatically abort a run, if newer stuff is in the PR.

PR #6227, opened 2022-11-29
project: optionally run an init script

haraldschilly:
# Description

This moves the logic of running `project_init.sh` to the project hub itself. Same behavior. However, there is also an important change: I added this parameter to the `launchProjectDaemon` function, which has implications for single and multi user project modes. So, that means we get feature parity with "kucalc" projects.

### DIFF
--- a/src/packages/project/init-kucalc.ts
+++ b/src/packages/project/init-kucalc.ts
@@ -9,6 +9,7 @@ import * as projectSetup from "./project-setup";
 import { activate as initAutorenice } from "./autorenice";
 import * as dedicatedDisks from "./dedicated-disks";
 import * as sshd from "./sshd";
+import * as initScript from "./init-script";
 import { getLogger } from "./logger";
 
 export default async function init() {
@@ -38,4 +39,6 @@ export default async function init() {
   }
 
   await dedicatedDisks.init();
+
+  initScript.run();
 }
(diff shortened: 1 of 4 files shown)


williamstein:
I wonder if this will make debugging "project boot up time" easier?  Right now projects start very quickly in kubernetes, but it takes just as long or longer until you can actual connect to them and use them.  I found our old architecture difficult to benchmark and debug, so I'm excited about this!

haraldschilly:
Well, previously the spawned process for the init script also runs in the background, so, I don't think this has any impact on the startup time. Getting rid of all this makes it certainly easier to test and debug, that's for sure.

williamstein:
> Well, previously the spawned process for the init script also runs in the background, so, I don't think this has any impact on the startup time. 

The "startup time" is a combination of several things (2 and 3 are in parallel):

1. The Kubernetes pod actually starting
2. The cocalc project itself starting a network server listening
3. CoCalc knowing the correct network address of the pod and connecting to the server

All these steps take time. The init stuff is happening in parallel with 2 and 3 and slows them down.

I think step 2 itself takes much more time than I would like.  If the project did absolutely nothing to listen for a network connection, then I suspect 1-3 would be significantly quicker than it is now.    But I don't know for sure.  It might just be that our internal private networking stack is a bit slow to add new hosts.  I don't know for sure.

PR #6232, opened 2022-11-30
database/pool test: fix off-by-one

haraldschilly:
# Description

- occasionally, there is a +/- 1 difference and "close to" is only for fractions of one after the decimal point
- `valueOf` is somehow discouraged since [mdn docs say](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/valueOf#description) "This method is usually called internally by JavaScript and not explicitly in code."

### DIFF
--- a/src/packages/database/pool/util.test.ts
+++ b/src/packages/database/pool/util.test.ts
@@ -10,7 +10,8 @@ test("use the timeInSeconds code gen function", () => {
 import { expireTime } from "./util";
 
 test("using expireTime to compute a time in the future", () => {
-  const now = new Date().valueOf();
-  const now10 = expireTime(10).valueOf();
-  expect(now10 - now).toBeCloseTo(10000);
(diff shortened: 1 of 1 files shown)


williamstein:
Maybe the right fix regarding toBeCloseTo is to look up the docs and see it has a second parameter which is the number of digits to compare?   https://jest-bot.github.io/jest/docs/expect.html#tobeclosetonumber-numdigits

haraldschilly:
Well, I was hesitant, because it says "control how many digits **after** the decimal point to check" and I don't want to add a division or a negative param for the digits. The ideal fix would be to make a change upstream to introduce a check for closeness with a delta > 1. But who knows what tricky edge cases there could be :face_with_spiral_eyes:

williamstein:
good point, no worries, and of course note that I did merge this as is.

PR #6237, opened 2022-12-01
admin config for "datastore" and FC-ing settings body

haraldschilly:
# Description

- for onprem mode, a toggle to show the "datastore" config panel in the settings
- there was a problem when the default value is true, i.e. the frontend didn't check properly if it should use a default fallback for an undefined value. also made the conversion to a boolean idempotent.
- I also noticed that the body for the settings isn't a React FC yet, so I fixed that as well.

### DIFF
--- a/src/packages/frontend/account/types.ts
+++ b/src/packages/frontend/account/types.ts
@@ -15,7 +15,7 @@ export interface AccountState {
   active_page: string;
   user_type: string;
   account_id: string;
-  groups?: string[];
+  groups?: List<string>;
   terminal: Map<string, any>;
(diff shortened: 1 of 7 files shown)


williamstein:
👍

haraldschilly:
Ok, I tested this as far as I could, and it looks ok. Converting this "Body" raised most of my concerns, though.
1. The type is now `groups?: List<string>;`. That's at least consistent now.
2. I don't have any of these old upgrades, so I can't test them. But like before, it's just passing down these summary functions from the store, so that should still work.

Issue #6241, opened 2022-12-05
store/site-license undefined user crashes computing the cost of the license


haraldschilly:
Somehow, the state of "user" is sometimes `undefined`. I don't understand why, but my best guess is it has something to do with a `useProfile` hook setting the initial value in `store/usage-and-duration` for the form item "user".

The error then happens in `util/licenses/purchase/compute-cost.ts` where `user` is `undefined` and hence throws an error:

> `Error: unknown user undefined`

williamstein:
This was the exact same problem as #6240.

Issue #6240, opened 2022-12-05
Application error when viewing license store


DrXyzzy:
Seen with recent Firefox and Chrome, on MacOS 12.5.1.

1. Open an incognito window in Chrome.
2. Browse to https://cocalc.com/
3. Click "Store" in upper horizontal navbar.
4. Click the big link for "Site License Upgrade".
5. Get a white browser window with the message: `Application error: a client-side exception has occurred (see the browser console for more information).` Console shows 2 instances of `Error: unknown user undefined...`

williamstein:
Oh, this is caused by trying to determine the user type (commercial or academic).

PR #6243, opened 2022-12-05
next/store: avoiding #6241 but not a proper fix

haraldschilly:
# Description

this is a quick fix for #6241 but does not address the root cause (I don't know what's going on, I'm not able to cleanly reproduce #6241 but I am sure this fix at least avoids that error)

### DIFF
--- a/src/packages/next/components/store/compute-cost.ts
+++ b/src/packages/next/components/store/compute-cost.ts
@@ -91,7 +91,7 @@ export function computeCost(
         throw Error("must be a quota upgrade license");
       }
       const {
-        user,
+        user = "business", // workaround for https://github.com/sagemathinc/cocalc/issues/6241 (does not fix root problem)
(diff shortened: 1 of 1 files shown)


williamstein:
> this is a quick fix for https://github.com/sagemathinc/cocalc/issues/6241 but does not address the root cause (I don't know what's going on, I'm not able to cleanly reproduce https://github.com/sagemathinc/cocalc/issues/6241 but I am sure this fix at least avoids that error)

But I already fixed it and pushed the fix to master before you started working on this.  It was the same as #6240.  You couldn't reproduce it probably because I already fixed it here :shrug:  : https://github.com/sagemathinc/cocalc/commit/561db8559e9de7dbe969fa36e8fd8b0f93b9550d

williamstein:
It can't hurt to merge this as an extra layer of safety.

PR #6246, opened 2022-12-05
nocodb admin visibility

williamstein:
Goal is to provide a way to get better visibility into an existing cocalc server using nocodb.


williamstein:
We're not going to use nocodb after all, due to it not really addressing our core issues, and also the AGPL license is kind of a show stopper for us.

PR #6262, opened 2022-12-07
hub/proxy: sketch about changing websocket caching params

haraldschilly:
# Description


haraldschilly:
closing this. The main take away for me is that `node-http-proxy` is no longer maintained and I wasn't able to find a good replacement. The goal was to disable caching the response for the websocket upgrade, because my suspicion is that it is cached, which causes a delay establishing the connection to a project. What didn't work about this implementation is a suitable callback for the response, because these "proxyResWs" is not implemented (or however it should be called) – or maybe they actually make no sense to be implemented and I don't know how to handle this. The best for now is to just keep this as it is. In production, cloudflare helps us with disabling any caching.

PR #6263, opened 2022-12-07
frontend: fix DateTimePicker displaying format 

### LINKED ISSUE #6261: all (?) date selectors are broken
for me, they just show "LLL" instead of a date

courses:

Screenshot from 2022-12-07 10-55-43

admin/licenses:

Screenshot from 2022-12-07 10-56-11

haraldschilly:
# Description

- **this fixex #6261**
- I think this is the equivalent to "LLL", but I can't remember exactly what we had. I'm always in favor to format dates in ISO, that's also the default by antd.
- However, since this came up from time to time, I've added `Z` to the format to show the time zone offset. see https://day.js.org/docs/en/display/format for options.
- obligatory screenshot for a due date in a course:
   Screenshot from 2022-12-07 13-32-38

### DIFF
--- a/src/packages/frontend/components/date-time-picker.tsx
+++ b/src/packages/frontend/components/date-time-picker.tsx
@@ -11,12 +11,12 @@
        https://ant.design/components/date-picker/
 */
 
-import { React } from "../app-framework";
-
 import { DatePicker } from "antd";
 import dayjs from "dayjs";
 
-export function DateTimePicker(props: {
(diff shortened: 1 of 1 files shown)


williamstein:
> I think this is the equivalent to "LLL", but I can't remember exactly what we had.

I think we had something automatic using moment.js.  I think it relied on having every single locale in the world wasting space in some shockingly large js bundle (moment was a shockingly large part of the webpack download).

PR #6313, opened 2023-01-04
frontend/jupyter: add shortcut to delete a cell with the delete keys

haraldschilly:
# Description

I instinctively use the delete key all the time, but it never works. This adds a shortcut for it.


williamstein:
I would be really concerned about this due to divergence from Jupyter and danger of people loosing work.  However, we have excellent undo and Timetravel, so it's easy to recover from any mistake.

PR #6330, opened 2023-01-09
Remove use of react-sortable-hoc

williamstein:


williamstein:
Top project nav done.  Notice that antd tabs are nicely rounded and also have a little "..." to show any tabs not fully in view.  So instead of the tabs all just getting really small and hard to read if there are a lot, they get "smaller", and then at some point any not on screen appear in the "..." menu fully visible.  This is much nicer on mobile, obviously, and generally **much** more accessible:

image

williamstein:
image

PR #6339, opened 2023-01-13
frontend/admin: improve editing json data fields, support for patches, and show these quotas to the user

haraldschilly:
# Description

this got a bit longer than just the json fields, since I also worked on their representation to make it more transparent about what's going on

- the admin json editor: it's just a better textarea. I briefly looked into this json editor you mentioned, but I don't want to work on this on top of all that. better for a PR later down the road …
- site settings: if disconnected, it's possible to click on the header to try loading again
- replacing max upgrades and default quotas with that simple json editor
- adding help-circles for more information, because it is hard to remember what the fields mean.
- The above implies there is a new optional field `help` for the server settings (the site default spec file)
- various enhancements to show the `ext_rw` and `patch` quota to the user, if it is set
- admin license editor: the save/cancel buttons appears after edits at the bottom as well, since that form is gaining in height
- patches: there is a validation function for checking if the json patches are roughly in the correct form. For ease of use in the UI, there is also a `loadPatch` wrapper, which doesn't throw (just returns an empty array if there is a problem).
- the "describe quota" function now also knows about patches and ext_rw, and hides some details for onprem setups.
- quota function: very important, the ext_rw and patch quotas are extracted before selecting the licenses. That's currently a bug, because e.g. ext_rw might go away. There are now additional tests to cover this! i.e. `const ext_rw = extract_ext_rw(site_licenses);` and `const patch = extract_patches(site_licenses);`
- quota: also more precise typing about mandatory and optional fields in the computed total quota
- (unrelated) add note about throttling auto save in IDEs

Screenshot from 2023-01-17 17-06-19

Screenshot from 2023-01-17 17-06-03


williamstein:
Possibly consider https://www.npmjs.com/package/@microlink/react-json-view  as its already in the cocalc frontend (e.g., used in Jupyter) and provides a reasonably nice easy to use JSON editor component.

Issue #6190, opened 2022-10-25
upgrade to nextjs version 13


williamstein:
Motivation: https://nextjs.org/blog/next-13

williamstein:
Blocked on:

- [ ] https://github.com/sagemathinc/cocalc/issues/6196 -- react-sortable-hoc

williamstein:
Some notes:

- I'm having problems even with next 12.3.x so downgrading for now to the 12.2 series.  The main issue is the CSS weirdly doesn't fully load on first page load.
- I did add dnd-kit to the frontend and used it a bunch in the new admin crm system.  So that's a step in the right direction to getting rid of react-sortable-hoc.

Issue #5887, opened 2022-04-27
upgrade to react 18


williamstein:
We are on react-17, and I wanted to open this issue to keep track of upgrading to react 18.  My expectation is that this won't be easy because it breaks a lot of backward compat.  Also, definitely when react 18 came out a few weeks ago, many of the dependencies of cocalc definitely didn't yet support it.  Let's track that progress here.  Also, we use specific ancient broken things, e.g., some react-bootstrap beta, and it would be a miracle if it just happens to work fine with react 18; much more likely, we'll have to bite the bullet and [finish our rewrite](https://github.com/sagemathinc/cocalc/issues/3782) and get rid of it!

- this looks like an open issue still: https://github.com/ant-design/ant-design/issues/35282
- virtuoso had some subtle issues: https://github.com/petyosi/react-virtuoso/issues/647

PR #6340, opened 2023-01-13
upgrade to React18 & next 13, thus fixing #5887 & #6190

williamstein:
See #5887


williamstein:
Only current known issue: Loading /support the first time results in a warning about server side rendering problems.

williamstein:
I think I'll also do this next 13 upgrade -- https://github.com/sagemathinc/cocalc/issues/6190

It's unblocked by react 18 and might resolve the issue with SSR?

PR #6358, opened 2023-01-20
util/quota: avoid modifying passed in objects

haraldschilly:
# Description

- this is a very subtle problem I have never seen before, but makes 100% sense: somewhere the quota function modifies licenses, which have unintended consequences when used again. all tests just assume a single call!
- however, the postgres/site-license/hook progressively applies licenses and in some cases, subsequent calls with modified licenses start to be incompatible.
- there is now a test applying partially and then all licenses, which trigger this problem.
- but it doesn't fail, because quota now does a deep copy of all arguments


williamstein:
Ready to merge? Seems safe.  I'm making a new release at the moment...

haraldschilly:
yes, looks good. I just wanted to see if it build + all tests pass

PR #6362, opened 2023-01-23
Playwright automated testing support

williamstein:


williamstein:
I did a lot here and it's pretty cool... but it also feels surprisingly flaky (given the marketing hype around playwright), and like a waste of time relative to other options, to be honest.  So I'm closing this for now.

PR #6368, opened 2023-01-24
frontend/jupyter: make notebook show up without a kernel

haraldschilly:
# Description

Ref: https://github.com/sagemathinc/cocalc/issues/6360

Well, there are a lot of changes, since I started to sort imports as well. One of the changes that had a large impact was changing `public store: any;` to `: JupyterStore;`. The kernel status indicator als had issues and I removed the borders to make it look better, same gray tone as the menu bar, and now there is also the mode, which was not showing anything (and also made a line break and was no longer visible)

during testing, I found many corner cases. this is mainly what I checked for:

1. account setting: "ask for kernel"
   1. create new notebook → show menu
   2. kernel → remove kernel → save notebook and close it … then when opening again, there is no kernel.
2. account setting: "don't ask for kernel, python3 as default"
   1. create new notebook → selects python 3
   2. remove kernel like above → save/close/open → still no kernel (i.e. it doesn't set my default kernel)
3. In general, running a cell without a kernel spits out an error message
   - such an error message (or any others) are removed when the kernel changes
5. I also checked that whiteboard cells work
6. Another interesting edge case: a new notebook with an unknown kernel, continuing without a kernel, did also cause an exception (in `output-messages/mime-types/data.tsx`) → hence changing the typing of that hook to be optional
7. I also checked that some ipynb file renders in the share server
---

This is some progress to make the content of a notebook show up without a kernel. I had to fix many TS errors, because somewhere I think I made an object property more than "any" and hence all this noise.

~~However, I need help. What's not working at all is saving a notebook. I can create an empty one (using `touch x.ipynb`) and add cells + markdown, but the "Save" button stays active. On disk, the file stays empty as well.~~

~~The root problem is that the project want's a kernel, because of this line:~~

```
const ipynb = this.store.get_ipynb(this.jupyter_kernel.get_blob_store());
```

~~I tried to come up with an idea to just get the blob store directly, but since this is in a file that explicitly states to not import from the project, I don't know what to do.~~

~~Once notebooks can be saved, I can finish this PR.~~


williamstein:
When I check this out do `pnpm build-dev` to build everything, projects don't start.  The problem is illustrated by
```sh
~/cocalc/src/packages/project$ pnpm exec cocalc-project
/Users/wstein/build/cocalc/src/data/projects/812abe34-a382-4bd1-9071-29b6f4334f03/cocalc/src/packages/node_modules/.pnpm/codemirror@5.65.11/node_modules/codemirror/lib/codemirror.js:18
  var userAgent = navigator.userAgent;
                  ^

ReferenceError: navigator is not defined
    at /Users/wstein/build/cocalc/src/data/projects/812abe34-a382-4bd1-9071-29b6f4334f03/cocalc/src/packages/node_modules/.pnpm/codemirror@5.65.11/node_modules/codemirror/lib/codemirror.js:18:19
```

This would be caused by importing some things from the frontend and part of importing the project library, which can't be loaded in node.js.  

Since I think `pnpm test` passed with this branch, I'm going to see if I can add a jest test that imports what ` pnpm exec cocalc-project` imports (to some extent), so we can catch this sort of thing more quickly.

williamstein:
I refactored the code to move a bunch of stuff only needed in browser-actions... to browser-actions. Then the project startup issue mentioned above went away...  I also added a test to prevent this surprise.

williamstein:
**With my last commit you can now save to disk!**

The trick was just to allow creating the JupyterKernel class even if the kernel isn't set. That's fine.  We just don't allow *spawning* such kernels.

PR #6369, opened 2023-01-24
next/config: fixing style, adding email verification, ...

haraldschilly:
# Description

- next/config: name, etc.
- next/config: account
- I saw email verification is still missing, and I added it. screenshot from  config/account/email:
     Screenshot from 2023-01-25 15-47-12
-  This depends on the hub, so, I added this as a reference. I believe that's against the overall plan, but well, there are already warnings about circular dependencies, and this only adds a bit more of these. In any case, this works.
- While implementing the above, I struggled with some postgres typing. Being more specific with some types opened up a can of worms. e.g. now it is possible to specify the expected type of a query/async_query. So, many additional touches to files, but none of them should have any impact (famous last words :shrug:)

### DIFF
--- a/src/packages/database/postgres/account-queries.ts
+++ b/src/packages/database/postgres/account-queries.ts
@@ -136,7 +136,7 @@ export async function set_email_address_verified(
   const { db, account_id, email_address } = opts;
   assert_valid_account_id(account_id);
   assert_valid_email_address(email_address);
-  return await db.async_query({
+  await db.async_query({
     query: "UPDATE accounts",
     jsonb_set: { email_address_verified: { [email_address]: new Date() } },
     where: { "account_id = $::UUID": account_id },
--- a/src/packages/database/postgres/passport.ts
+++ b/src/packages/database/postgres/passport.ts
@@ -16,7 +16,6 @@ import {
   set_email_address_verified,
 } from "./account-queries";
 import {
-  CB,
   CreatePassportOpts,
   DeletePassportOpts,
   PassportExistsOpts,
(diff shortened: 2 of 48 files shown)


williamstein:
> I believe that's against the overall plan, but well, there are already warnings about circular dependencies, and this only adds a bit more of these. In any case, this works.

Those warnings are from pnpm, since in some cases pnpm install with workspaces can't handle circular deps.  They don't seem to be a problem for us, due to the explicit rules I put in workspaces.py.   

Regarding the actual Javascript/Typescript dependencies, since we're mostly using import now (instead of require), circular dependencies aren't a problem.

That said, I often view circular dependencies as a good opportunity to refactor.  In particular, this code for email verification should maybe be moved to packages/server, and then hub and next both use that.

williamstein:
> This depends on the hub.

I now see looking at your code that you made @cocalc/server depend on @cocalc/hub by adding this line:

```ts
import { verify_email_send_token } from "@cocalc/hub/auth";
```

I don't like this at all.  This breaks building our kucalc manage-* image, since the Dockerfile to do that contains:

```sh
RUN cd /home/manage/cocalc/src && \
  export PACKAGES="util,database,backend,server" && \
  ./workspaces.py install --packages=$PACKAGES && \
  ./workspaces.py build --packages=$PACKAGES
```

Notice that we'll have to add more packages to the above, and that'll significantly increase build time and size.

I realize it's a lot of work to refactor the code to move `verify_email_send_token` to @cocalc/server, where it belongs.

I guess we can temporarily just make building manage-* worse...

Any thoughts?

williamstein:
I think i'll merge this and if the dep on hub causes too much trouble, then I'll just comment that part of the PR out and we can fix it later.

haraldschilly:
I full agree, this is not ideal and should be refactored. It's not that easy, because that  code I saw depends on a coffee file, and who knows what else. My more general direction is to redo the entire "emailing" aspect in a "server" subdirectory, remove any dependency on external templates – and instead use something to handle email templates by the application itself and hence make them configurable for 3rd parties.

PR #6375, opened 2023-01-27
frontend/files: revisit file rename/duplicate warnings

haraldschilly:
# Description

- ref https://github.com/sagemathinc/cocalc/issues/6370 … I added the case where there is a space around the extension as a blocker for renaming
- I also added a red warning, if there are other reasons why you can't rename. e.g. right now if the file is too long, the button is disabled and not telling you why.

Screenshot from 2023-01-27 13-52-26

Screenshot from 2023-01-27 13-51-28

Screenshot from 2023-01-27 13-51-44

### DIFF
--- a/src/packages/frontend/project/explorer/action-box.tsx
+++ b/src/packages/frontend/project/explorer/action-box.tsx
@@ -297,38 +297,73 @@ export const ActionBox = rclass<ReactProps>(
       this.props.actions.set_all_files_unchecked();
     }
 
+    private filename_length_test(name: string): boolean {
+      return name.length > 250;
+    }
+
+    private filename_illegal_chars(name: string): string | false {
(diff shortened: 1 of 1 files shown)


williamstein:
I merged this since it is a massive improvement (and also much better than what Windows/MacOS do, imho). 

One little issue:

1. Change the filename to have a space at the beginning
2. You can't click submit but there is also no error message.

It should either allow the behavior or show an error.   Technically in Linux files can start with a space, so I think showing a warning and allowing the behavior is possibly the correct thing to do (though one could make the same argument about filename extensions).   It seems like maybe in MS Windows having a space at the beginning of a filename is very similar to a dot at the beginning of a filename in UNIX, so maybe this is a reasonable thing to do.

PR #6383, opened 2023-01-31
next/subnav: push down the logic when to show it 

haraldschilly:
# Description

only show the subnav menu, when there is actually data for it


haraldschilly:
@williamstein I'm not waiting for the tests to pass (off for the day) … this should do it, because now it's the sub-nav (which holds the actual data) who decides. and not some weird test a level higher.

PR #6413, opened 2023-02-07
fix antd deprecations

haraldschilly:
# Description

- project's "modal" visible→open and make it a banner
- menu dropdown: instead of overlay={menu} use items={items}, where items is a datastructure.
- project/new: I never liked this and well, here is my maybe controversial change. the buttons are larger, easier to read, and it's 2 columns on narrower screens. Improving this, adding templates, etc. … that's for later

  Screenshot from 2023-02-08 14-54-05

- format title bar: font type, size, ... and "title"
- that "file actions" dropdown
- slate: that "insert" dropdown
- share: the is a dropdown for selecting a license
- frame editors: spellcheck in configuration frame
- jupyter: main menu rows, kernel type selection, etc.


williamstein:
I'm so glad you're doing this!  I keep meaning to and never getting around to it. THANKS!!

PR #6414, opened 2023-02-07
[needs review] Provides working OAuth1 SSO flow (should not break OAuth2 SSO)

belonesox:
# Description

* Provides working OAuth1 SSO flow (should not break OAuth2 SSO, but still not check). 
* Improve creation Profile from SSO data, if provided «realname» attr.
* Small diffs for easy verifications.

- [* ] Testing instructions are provided, if not obvious
- [*] Release instructions are provided, if not obvious

### DIFF
--- a/src/packages/server/auth/sso/oauth2-user-profile-callback.ts
+++ b/src/packages/server/auth/sso/oauth2-user-profile-callback.ts
@@ -10,21 +10,21 @@
 const safeJsonStringify = require("safe-json-stringify");
 import { parseOpenIdProfile } from "@cocalc/server/auth/sso/openid-parser";
 import { UserProfileCallbackOpts } from "./types";
-
+import jwt_decode from "jwt-decode";
 export function addUserProfileCallback(opts: UserProfileCallbackOpts) {
   const { strategy_instance, userinfoURL, L2, type } = opts;
   if (userinfoURL == null) throw new Error(`the userinfoURL is required`);
-  strategy_instance.userProfile = function userProfile(accessToken, done) {
-    L2(`userinfoURL=${userinfoURL}, accessToken=${accessToken}`);
+  strategy_instance.userProfile = function userProfile(accessToken, tokenSecret, params, done) {
+    L2(`userinfoURL=${userinfoURL}, accessToken=${accessToken}, tokenSecret=${tokenSecret} `); 
+    L2(params); 
 
     let oauth = this._oauth;
     if (this._oauth2) {
       oauth = this._oauth2;
       oauth.useAuthorizationHeaderforGET(true);
-    }
-    oauth.get(userinfoURL, accessToken, (err, body) => {
-      L2(`get->body = ${safeJsonStringify(body)}`);
-
+    }    
+    oauth.get(userinfoURL, accessToken, tokenSecret, (err, body) => {
+      // L2(`get->body = ${safeJsonStringify(body)}`);
       let json;
 
       if (err) {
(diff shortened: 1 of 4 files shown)


williamstein:
Thanks!  

Can you use the labels PR-works in progress or PR-needs review?  It might be you're not allowed to use those labels yet (due to silly github permissions), in which case don't worry about it.

belonesox:
Is there some policy-workflow about these labels?  

Anyway, looks I am not allowed to edit labels (and other issue fields) at all.

image

williamstein:
> Anyway, looks I am not allowed to edit labels (and other issue fields) at all.

Evidently this is a "not implemented yet bug" in Github:  https://github.com/neuroanatomy/BrainBox/issues/304

Github did implement PR workflows, which are probably better than using labels.  They implemented that years after we got used to using labels for PR's though, so we haven't switched.  We should switch.  In the meantime:

> Is there some policy-workflow 

The policy workflow is to label a PR "work in progress" when you're working on it, then "needs review", when it is ready to be reviewed, and finally "needs work" if the review revealed something wrong.  Since github doesn't allow non-contribs to set labels, can you edit the title of your PR to start with [work in progress] or [needs review]?    That will work just as well.

src/packages/server/auth/sso/oauth2-user-profile-callback.ts
```diff
@@ -10,21 +10,21 @@
...
 import { UserProfileCallbackOpts } from "./types";
-
+import jwt_decode from "jwt-decode";
 export function addUserProfileCallback(opts: UserProfileCallbackOpts) {
   const { strategy_instance, userinfoURL, L2, type } = opts;
   if (userinfoURL == null) throw new Error(`the userinfoURL is required`);
-  strategy_instance.userProfile = function userProfile(accessToken, done) {
-    L2(`userinfoURL=${userinfoURL}, accessToken=${accessToken}`);
+  strategy_instance.userProfile = function userProfile(accessToken, tokenSecret, params, done) {
+    L2(`userinfoURL=${userinfoURL}, accessToken=${accessToken}, tokenSecret=${tokenSecret} `);
```
williamstein, line 18: I'm concerned this just introduced a potential security vulnerability here.  Before the code here (and below that is now commented out) took some care to not print a secret in the logs, but now it does seem to print a secret in the logs.   Generally, we always try to avoid writing secrets to our logfiles, to reduce our attack surface.

src/packages/server/auth/sso/oauth2-user-profile-callback.ts
```diff
@@ -59,7 +59,12 @@ export function addUserProfileCallback(opts: UserProfileCallbackOpts) {
       try {
         json = JSON.parse(body);
       } catch (ex) {
-        return done(new Error(`Failed to parse user profile -- ${body}`));
+        try { // OAuth1 body, need base64-like decoding
+            json = jwt_decode(body);
+        }
+        catch(ex){
```
williamstein, line 65: Here you catch the exception but seem to not refer to it, which is a typescript error (given our settings). E.g., something like this is better:

```ts
return done(new Error(`Failed to parse user profile -- ${body} -- error: ${ex}`));
```

williamstein:
Ping -- see my comments.  @belonesox   thanks!

belonesox:
Sorry for delaying my reactions, was very busy (partially because I was starting new course on fresh installation of cocalc, with OAuth1 authorizations, I am using cocalc+codeserver even in my teaching videos, [like that](https://vimeo.com/showcase/10185384)). 

Absolutely agree with review, I little surprised why I did not catch «unused exception» error. 

Also, I planned to test all this with OAuth2 (to check that OAuth1 support  did not break OAuth2, but still cannot find time to set up test setup for it).

williamstein:
Thanks!

williamstein:
> Also, I planned to test all this with OAuth2 (to check that OAuth1 support did not break OAuth2, but still cannot find time to set up test setup for it).

Merged.  I'll test it and fix/revert if there is an issue, but I don't see where there would be one.

PR #6434, opened 2023-02-14
csv viewer using ag grid.

williamstein:
I'm going to create this PR, then close it, since ag grid looks and works nicely, but it's a pretty heavy dependency with a massive API surface, and probably overkill.  That said, the extra MB would only get downloaded on first opening a csv file.

This uses the free open source version of https://www.ag-grid.com/


williamstein:
Here's what it looks like.  It does fully work, with column resizing.   It's fast to scroll, of course.

image

williamstein:
I went with something custom built on our react-virtuoso instead.

PR #6435, opened 2023-02-14
csv viewer using reactgrid.

williamstein:
- this fully works
- grid view is read only
- I don't at all know if I want to merge this though, since there are many other approaches to grids, and I'm unsure of https://reactgrid.com/   It seems unsupported and sort of dead.


williamstein:
I went with something custom built on our react-virtuoso instead.

PR #6440, opened 2023-02-15
frontend/chat+slate: antd modernize @-dropdown

haraldschilly:
# Description

This touches the "@" menu, to get rid of the deprecation, and well, I also made a font slightly smaller and made sure to popup to the top in chat (since for me, it sometimes opens up to the bottom and then I can't see or select some). I also tested this in slate markdown, works ok as well.


williamstein:
I'm going to wait on this, since I just made the font larger 2 days ago, so I could better see the avatars, and also emojis, which use these same components.  There may be a better solution to the font issue (e.g., making only the emoji/avatar) larger, but it's possibly not trivial.

PR #6468, opened 2023-02-22
frontend/antd: make dialog modals be on top of popovers -- #6467

### LINKED ISSUE #6467: frontend/frame-editor: additional buttons popup is in front of modals
jupyter: click on "..." and "halt". The modal is below the menu. That's a matter of fixing the z-indices.

Screenshot from 2023-02-22 11-48-09

haraldschilly:
# Description

this fixes #6467

it also removes two divs, which are replaced by left and right margins :shrug: 

### DIFF
--- a/src/packages/frontend/_antd_fix.sass
+++ b/src/packages/frontend/_antd_fix.sass
@@ -40,3 +40,10 @@ html
   flex-direction: column
   > .ant-alert-banner
     display: block
+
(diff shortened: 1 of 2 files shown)


williamstein:
> it also removes two divs, which are replaced by left and right margins 🤷

FWIW the most popular web dev podcast (syntax.fm) which I listen to a lot, recently argued for using space divs instead of margins in some cases.   I'm fine with the change though.  And of course there were a bunch of weird versions of that html before the one you saw...

Issue #6454, opened 2023-02-17
side-chat: slate editor buttons hidden by scrollbar


haraldschilly:
there is somewhere an "overflow-x" that hides the buttons. instead, I guess, we need a "..." hidden button solution like there is with the frame editor buttons.

![Screenshot from 2023-02-17 15-46-04](https://user-images.githubusercontent.com/207405/219689119-4d8b3fcc-1f01-44b8-a86b-762116ca76c1.png)

williamstein:
harald fixed this!

PR #6479, opened 2023-02-24
frontend/side-chat: fixing hidden slate icons

haraldschilly:
# Description

Ref #6454

With this, there is a toggleable Popover with these EditBar-buttons

Screenshot from 2023-02-24 16-33-25


williamstein:
heads up -- it looks like this PR breaks editing past chat messages; when double clicked they come up as blank.

williamstein:
I fixed the bug introduced here.

PR #6525, opened 2023-03-08
next/pricing/institutions: introduce "overcommitment"

haraldschilly:
# Description

I get the point of the disclaimer. In this version, however, I made this more explicit by adding an overcommitment ratio and explained this in a separate paragraph. Also, this reduces the prices by a lot and should be easier to follow.

Screenshot from 2023-03-08 13-24-53


williamstein:
Wait -- my banner was 

image

which I think pretty clearly explains a fairly concrete thing.  You've completely replaced that entire explanation by this:

image

I'm very concerned that what you've replaced it by is significantly less clear to a reader...

haraldschilly:
I added an entire paragraph explaining what "number of projects" means, in particular starting with the statement, that this not equal to the number of individuals.

williamstein:
OK, that make sense.

PR #6531, opened 2023-03-09
next/billing/subscriptions: fix the "cost" column

haraldschilly:
# Description

Right now, the "Cost" column in the overview of subscriptions only lists the cost for a single unit. I.e. run limit/quantity is not taken into account. This was a bit tricky, because the info is actually in the license data, but that isn't always available (like, it is for the description column as well). In any case, here are two rows from my dev project, shows the actual cost in the cost column.

(I'm also adding some typing to have less surprises about which field is where, etc.)

Screenshot from 2023-03-09 11-19-00

Screenshot from 2023-03-09 11-18-39


williamstein:
Thanks!  I've noticed this and realized it would be hard to fix.

PR #6536, opened 2023-03-10
next/npm: do not call build-dev twice from build target

haraldschilly:
# Description

This is just tiny, and we discussed this and forgot about it. I think that build-dev step (the first one of two) in next's build target is pointless. I was wondering if this should be "build-deps" instead, but well, I don't think so. We always call make-dev (and that workspaces script). So, with that change it's more straight to the point.

Why the change? Running build-dev twice wastes a lot of time.

### DIFF
--- a/src/packages/next/package.json
+++ b/src/packages/next/package.json
@@ -25,7 +25,7 @@
   ],
   "scripts": {
     "preinstall": "npx only-allow pnpm",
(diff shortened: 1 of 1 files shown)


williamstein:
> . I was wondering if this should be "build-deps" instead,

We used to use build-deps, but that got too difficult to maintain.  Any reference to that should be deleted.  You're right -- It's probably why the "build-dev" accidentally got put in twice.

PR #6541, opened 2023-03-11
chatgpt integration

williamstein:
See https://github.com/sagemathinc/cocalc/issues/6392, though this might turn out to be nothing like that initially.


williamstein:
Progress

image

williamstein:
Next commit adds ability to call chatgpt from the browser frontend client (with no limits yet, of course):

image

PR #6566, opened 2023-03-22
server/sso: additional heuristic if no first and last name in profile

haraldschilly:
# Description

There is a stupid edge case,where querying the OAuth2 endpoint for the userinfo does not work (no idea why) and still no name info comes back. This adds the same heuristic that would clean up the profile to the general `sanitizeProfile` function, i.e. that `if (!opts.first_name && !opts.last_name) {` case.


haraldschilly:
@williamstein I'm merging this, because this is a bit urgent to check if this small change helps. I should have communicated this internally.

PR #6572, opened 2023-03-23
server/sso: replace the deprecated saml module by the new one

haraldschilly:
# Description

The current SAML module is deprecated – [see here at the top](https://www.npmjs.com/package/passport-saml)

This replaced it with the new one. However:

**IMPORTANT**: do not deploy this immediately. we use SAML and from my local testing I have noticed, it does testing of something called "audience" – that's certainly one of the bugfixes for this method. so, if that assertion is set in the response, and we don't have that set, it will fail. The simplest way to go is to just disable it, i.e. in the `conf` settings, we have to add `{… , "audience": false, … }`.
 
a long break later, now it breaks (the "new" variant, the default "old" is fine)

```
Error: Invalid document signature                                                                                   
    at SAML.validatePostResponseAsync (/cocalc/src/packages/node_modules/.pnpm/@node-saml+node-saml@4.0.5/node_modules/@node-saml/node-saml/src/saml.ts:696:15)                                                                         
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)                                   
```

that's a config issue. adding `{…, "wantAuthnResponseSigned": false }` to `conf` fixes this. the point is, only the one and only assertion in the response has a signature. 


williamstein:
I don't know what to do with this PR.  I'm going to leave it unmerged until you finish it.

williamstein:
Harald: "i will merge this"

haraldschilly:
I fixed this branch today and it's ok to merge it. By default, it keeps using the old version 3 of that module. Only if the environment variable `COCALC_SSO_SAML=new` it uses the version 4. The main complication is it is necessary to change the configuration when updating. They're just not 100% compatible. The two additionally required fields are in the top comment.

haraldschilly:
I added an additional explicit type parameter, this should help with any kind of issue that could come up.

PR #6575, opened 2023-03-23
hub/analytics: fix cookie, cleanup old code

haraldschilly:
# Description

- the correct fix is to have two cookies, since they have different purposes and well, the analytics cookie does work differently
- I noticed uglify-js is outdated and not used in other packages. Hence updating it once and removing it twice. I added a quick test to make sure it is working in the hub.
- 


williamstein:
Closing because not touched in 8 months. Reopen if this is still a priority.

PR #6596, opened 2023-03-30
frontend/file notifications less intense

haraldschilly:
# Description

 instead of alert-red, this is a subtle shade of gray change


williamstein:
I think we need to think this through more, and have something more fine grained (similar to, but better than, github).

williamstein:
OK, we'll do it for a few days and see what happens.

williamstein:
During the weekend!

williamstein:
I've reverted this since it was entirely due to not wanting to see chatgpt threads result in notifications, and now they don't.

PR #6641, opened 2023-04-18
frontend/admin/settings: confirm save-all, and disable autocomplete, ...

haraldschilly:
# Description

- refactor: breaking apart, subdirectory … no major changes, though
- input fields get an `autocomplete="off"` [see this ref](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#values)
- modal confirmation box listing exactly which value will be changed before saving


haraldschilly:
> What is left to do?

Those "json string values", just a special case. Now they behave like the other ones (i.e. save button lights up) and no weird extra "you have to save" message.

PR #6661, opened 2023-04-26
project blobstore on disk

haraldschilly:
# Description

instead of sqlite, this is supposed to use a simple file based store. ~~however, this is more tricky than I thought, because everything with files is async and this propagates deeply into jupyter.~~

Ok, this works! I got confused so many times, at least I think I learned a bit about what the jupyter backend is doing. My far-reaching goal to make the api fully async didn't work out. So, this is still in the world of sync calls.

### DIFF
--- a/src/packages/backend/data.ts
+++ b/src/packages/backend/data.ts
@@ -44,6 +44,13 @@ export const projects: string =
   process.env.PROJECTS ?? join(data, "projects", "[project_id]");
 export const secrets: string = process.env.SECRETS ?? join(data, "secrets");
 export const logs: string = process.env.LOGS ?? join(data, "logs");
+export const BLOBSTORE = getBlobstore();
+
+function getBlobstore(): "sqlite" | "disk" {
+  const env = process.env["COCALC_JUPYTER_BLOBSTORE_IMPL"];
+  if (env === "disk") return "disk";
+  return "sqlite";
+}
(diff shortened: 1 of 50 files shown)


williamstein:
Also, how is it that you're writing something from scratch that beats sqlite at its own game in a few hours?  I'm just curious, since that seems like something I used to do and realize it was a bad idea.  See, e.g., the first few years of sagenb's source code...

haraldschilly:
>  how is it that you're writing something from scratch that beats sqlite 

well, my unemotional answer is, this depends on the metric. if the metric is "don't stall starting the node process for minutes and then throw a lock message, even if there isn't even an existing database file", then this beats this. also, this isn't a SQL database. another difference is that instead of one large file, there are many smaller ones. network file-systems certainly do cool tricks to avoid sending a large file over the wire, but with that it's not necessary to do rely on that.

williamstein:
I'm even more dubious than before with storing one file for every item, so there's potentially a ton of disk activity.  I'm also not at all convinced that just changing an option for how to connect with sqlite or using the async driver instead might be a better approach.  And sqlite3 is a great key:value store.  All that said, this appears to only be used when explicitly enabled, with a fallback to the existing working sqlite3, and your code also looks nice, so I don't see any harm in merging it.  Merging!

PR #6667, opened 2023-04-28
project/blobstore-disk: fix crash uncaught unlink

haraldschilly:
# Description

delete files fix and be more precise about how many are deleted


haraldschilly:
this is just a quick fix, doesn't touch other parts so I merge it. seems like one last change didn't make it to the original PR...

PR #6754, opened 2023-06-01
Project context

haraldschilly:
# Description

~~This is just an experiment.~~ I think this is something we should have, for some aspects, but there are maybe drawbacks I don't see. One interesting detail I have in my mind is to only instantiate some react hooks once, and share their effects, instead of instantiating many times.

I think this works fine now. My main worry is, once this is used in more places, we  accidentally use a component outside the realm of that context provider. the default state is a misleading fallback. That's why I added a check for the project_id, [which raises an exception if this happens](https://github.com/sagemathinc/cocalc/pull/6754/files#diff-2a5113353ecdb6107e8de7676ffd42b7e09e6f6e4fe3ef43c4b7cdcb6e6c70eeR52). 

I also mulled over the idea to optionally set some context values, but in the end that's pointless, because this just means we can't use the context.

```
export function useProjectContext(
  overwrite: Partial<ProjectContextState> = {}
) {
  const context = useContext(ProjectContext);
  return useMemo(() => ({ ...context, ...overwrite }), [context, overwrite]);
}
```

One important detail is the `useProjectStatus` hook, which is for metrics and alerts. It moved inside the context hook, hence [this change](https://github.com/sagemathinc/cocalc/pull/6754/files#diff-edfc36f0ebb855e42ce544351240b68df359eba6fe064e796638975b4bb99e1aL74) and the [corresponding line](https://github.com/sagemathinc/cocalc/pull/6754/files#diff-2a5113353ecdb6107e8de7676ffd42b7e09e6f6e4fe3ef43c4b7cdcb6e6c70eeR66) in the context.

PS: there is also a padding css fix for the trial banner included


williamstein:
>  One interesting detail I have in my mind is to only instantiate some react hooks once, and share their effects, instead of instantiating many times.

Be careful.  E.g., if you put a bunch of things in one hook, then it will cause components that don't need all of those things to be updated a lot even though it is completely unnecessary.  It's a double edged sword.

I see your concern about:
```ts
  if (context.project_id === "") {
    throw new Error(
      "useProjectContext() must be used inside a <ProjectContext.Provider>"
    );
  }
```

rather than make the project_id of type 
```
   project_id?:string
```
I think your design choice means that all of the checks are done at runtime rather than at compile time, thus making bugs more likely. That said, it would be such a headache to have to check that project_id is defined in every component, so screw that...

That said, again, I think putting an exception right in the middle of the context provider is probably never ever going to work except in the case when we want it to blow up in our face.  The reason is because hooks have to all get called every time the component is rendered.  Hooks can't raise exceptions as that breaks react.  So if we ever mess this up, expect scary react tracebacks.  Maybe that is perfectly fine!

Obviously, so much of the code is sooo much nicer with this context hook!

PR #6770, opened 2023-06-06
frontend/flyout/log: also load the complete log

haraldschilly:
# Description

- fixes https://github.com/sagemathinc/cocalc/issues/6765 by not deleting the table.
- there is also the recent file list for the "home" part, which doesn't use the "_all" variant. So, I'm not sure if this is a good fix or not. In general, I wasn't aware that there are two separate data sources for the activity log. I thought there is one and with that button, it just gets longer.


williamstein:
> I thought there is one and with that button, it just gets longer.

I wish, but I didn't know how to implement that at the time.    We have the same issue with the projects and projects_all virtual table.   It might be possible to just increase the size of the same sync'd table using the limit option.  I improved a lot of the sync table functionality recently when writing our internal .cocalc-crm thing...

haraldschilly:
I see

PR #6779, opened 2023-06-13
flyout: 5th round of improvements

haraldschilly:
# Description

-  make the entire file item area clickable, but prevent the dropdown menu from triggering a click
- instead of caret sign, make background like hover. border's don't work well.
- load "full log" button
- keyboard scroll support for files in log entry + opening via return key
- get rid of popovers + remove associated account "other setting"
- fix opening file from log entries (re-using existing function)
- search: 
   - compact settings, use help icon for additional info
   - nested scrollbars if necessary
   - do not copy if text is selected, also add a small copy button
- flyout vs. tasks: remove key handler when focused
   - however, the focus never flipped back to tasks. how does chat do this? I had to change some code in `task-editor/list.tsx` **this is the only change outside flyouts, and it puzzles me, see comment**

### DIFF
--- a/src/packages/frontend/editors/task-editor/list.tsx
+++ b/src/packages/frontend/editors/task-editor/list.tsx
@@ -10,16 +10,17 @@ List of Tasks -- we use windowing via Virtuoso, so that even task lists with 500
 import { List, Set as immutableSet } from "immutable";
 import { useEffect, useMemo, useRef } from "react";
 import { Virtuoso, VirtuosoHandle } from "react-virtuoso";
-import useVirtuosoScrollHook from "@cocalc/frontend/components/virtuoso-scroll-hook";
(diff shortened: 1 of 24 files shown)


src/packages/frontend/editors/task-editor/list.tsx
```diff
@@ -147,7 +148,8 @@ export default function TaskList({
   }
 
   function on_click(e) {
-    if (e.target === mainDivRef.current) {
+    // test, if e.target is a child of mainDivRef.current
+    if (mainDivRef.current.contains(e.target)) {
```
haraldschilly, line 152: this is my only change outside "flyouts" with an impact. in particular, I'm puzzled how regaining the key handler ever worked, because the click target is as good as never the mainDivRef.

PR #6811, opened 2023-06-30
frontend/flyout/files: modal file selection

haraldschilly:
# Description

this makes file selection a modal mode, which can be triggered in various ways (click on icon, explicit button, shift/ctrl key clicks). otherwise it opens the file. also, adding details like closing an active editor with a middle mouse click.


williamstein:
This is pretty clever!

PR #6837, opened 2023-07-19
frontend: support *.ini files

haraldschilly:
# Description

since I just came across that file type ...

Screenshot from 2023-07-19 10-11-13


williamstein:
> since I just came across that file type ..

Idea: add a link to create an issue or support ticket whenever somebody gets the "no such editor" editor...

PR #6847, opened 2023-08-03
hub: if paygo is active -> make all licenses ineffective

haraldschilly:
# Description

WIP

unclear:

- I haven't found any use of `database/postgres/project-action-hooks.ts`. any idea what this is? This PR deletes it.
- in kucalc, I'm guessing site licenses are processed first, and then paygo is eventually activated. That can't work, paygo must come first.
   - ok, think I have a plan. the only way I see how this can work is to not call the site license hook in the kucalc case, in the hub. I'm not sure about the implications, though. Instead, manage has to do the paygo-handling, then do the site license processing, and then compute and process the run_quota.

---

well, despite the implications for kucalc, this seems to work. here for my test project: either the license is active, or with paygo, it is inactive.

### paygo is active

Screenshot from 2023-08-03 16-50-35

### or the license is active

Screenshot from 2023-08-03 16-47-32


williamstein:
> in kucalc, I'm guessing site licenses are processed first, and then paygo is eventually activated

No that is not what happens.  It first checks for pay as you go. Read the code you find in the manage addon by searching for "pay as you go"...

src/packages/server/projects/control/kucalc.ts
```diff
@@ -55,7 +55,10 @@ class Project extends BaseProject {
     }
     try {
       this.stateChanging = { state: "starting" };
-      await this.siteLicenseHook();
+
+      // TODO: once "manage" processes site licenses, we can remove this line!
+      // Manage has do the equivalent of this.computeQuota()
+      await this.siteLicenseHook(false);
```
haraldschilly, line 61: I should add, `siteLicenseHook(false)` should stay there until the equivalent of https://github.com/sagemathinc/kucalc/pull/114 has been patched in cocalc-cloud

PR #6863, opened 2023-08-10
frontend/avatar: this should fix #6862

### LINKED ISSUE #6862: frontend crash when searching for non-existent user
Easy to reproduce: users → add collab → I tried this with "foo@user.com" and it crashes here:

**Cannot read properties of undefined (reading 'startsWith')**

```
        function w(e) {
            return e.account_id.startsWith("chatgpt") || e.account_id.startsWith("openai-") ? o.React.createElement(f.Z, {
                size: e.size,
                style: e.style
            }) : o.React.createElement(x, g({}, e))
        }
```

haraldschilly:
# Description

- ref: #6862

### DIFF
--- a/src/packages/frontend/account/avatar/avatar.tsx
+++ b/src/packages/frontend/account/avatar/avatar.tsx
@@ -48,10 +48,11 @@ interface Props {
   style?: CSSProperties;
 }
 
-export function Avatar(props) {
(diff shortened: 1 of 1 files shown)


williamstein:
This is too much "bandaid instead of fixing the real problem", so I'm closing it and will push my own fix.

PR #6876, opened 2023-08-18
PoC gpu support

haraldschilly:
# Description

WIP


williamstein:
Please re-open this if you are going to work on it.  Not sure if we really need this given our strategy and the difficulty involved.  It might be better to support this only in cocalc-cloud?

PR #6882, opened 2023-08-23
Switch from Privacy Shield to Data Privacy Framework

novoselt:
# Description

I *think* that we can publish it now, because we are already listed as participating in DPF:
https://www.dataprivacyframework.gov/s/participant-search/participant-detail?id=a2zt0000000TUtjAAG&status=Active

Paragraphs that had explicitly suggested text from DPF website are replaced with those, others are edited to replace Privacy Shield with DPF.


williamstein:
Now live at https://cocalc.com/policies/privacy

PR #6905, opened 2023-09-08
Account creation time based tweaks

haraldschilly:
# Description

this contains two tweaks based on the account creation time:

1. after 2023-09-22: the "flyout only" vertical bar behavior is the default, no change for older users or those who've set this explicitly
2. the news item counter does not count news items that are older than the account creation time (no change for actually showing the news items, this is only for that blue notification number)

additionally, I started to be worried about not seeing the full-page button. I added a one-step mini tour, that triggers the tooltip popup once and backs it with the tours array. During testing that, I stumbled over a weird behavior: https://github.com/sagemathinc/cocalc/issues/6929 and made an ugly fix. I don't know how those tables work, so, that's the best I was able to come up with.

The tooltip itself is this:

Screenshot from 2023-09-21 10-57-47

and after either clicking fullpage once or the dismiss button, it reverts back to the usual mouse-over behavior with less text:

Screenshot from 2023-09-21 10-58-08

---

I also made some changes to how the mobile view of flyouts looks like. At some point, maybe flyouts as a vertical fullpage should be the default on mobile. they work clearly better on narrow screens. That's for later in another ticket, though.


williamstein:
> the "flyout only" vertical bar behavior is the default

I'm very nervous about making this change.  I still periodically need to access the old bar, and I'm also worried that the docs might be confusing.

If there were a maximize button next to the x to close a flyout panel, that showed the full size page then hid the flyout panel, I would merge this.

I.e., at the top next to that close x, put our standard fullscreen icon:

image

like we have for chat and all the file editors:

image


When you click that icon:

- flyout side panel closes
- navigate to corresponding full page

This is a million times more discoverable for a new user than finding that button near the bottom of the actions bar, reading and understanding the menu items, etc.

williamstein:
An example of how this would be a weird change to make:

1. New user creates a new project or opens their first project.  They will see the full file explorer page.
2. They open a file and edit it.
3. There is absolutely no obvious/discoverable way to get back to the full file explorer page, which they naturally may want to do.

haraldschilly:
yes, adding a button to expand it in that header is a good idea

williamstein:
Another issue with flyout only as the default is that file actions are are very nondiscoverable, to put it mildly.  E.g., to rename a file you have to:

1.  Click the file type, which is already a sort of easter egg/weird thing to do which you might never figure out.
2. Then look all over and find near the bottom some little icons with no words and hover over each until you see the "Rename" tooltip. 

It seems to me the odds somebody could figure that out is much lower than with the full size file explorer, where there is an always visible checkbox, and a big word "Rename" at the top.  It's not great, but it's much more discoverable, and possibly familiar since it's how chromebooks were at some point...

haraldschilly:
with the latest comment, the flyout button mode "both" is kept as the default for everyone

Issue #6920, opened 2023-09-17
Display the full name of the time zone used


novoselt:
It may not always be necessary, but is very helpful at least when setting up the license start/end time and when viewing invoices with time stamps.

I think it is also preferable to use something like "Mountain Daylight Time" rather than "MDT" to make it both more visible and less ambiguous, especially if the time zone is detected incorrectly for some reason.

williamstein:
Solved by PR #6938 !

PR #6938, opened 2023-09-25
#6920: Added time zone label to license start/stop times.

schrodingersket:
This pull request only partially addresses #6920; I'm quite sure how to spoof invoices to test changes, but I did get the timezone name sorted out for license start/end dates:

image

### DIFF
--- a/src/packages/next/components/misc/date-range.tsx
+++ b/src/packages/next/components/misc/date-range.tsx
@@ -15,6 +15,7 @@ interface Props {
   maxDaysInFuture?: number; // don't allow dates this far in the future from now
   disabled?: boolean;
   initialValues?: DateRangeType;
+  suffix?: string;
 }
 
 export default function DateRange(props: Props) {
@@ -25,6 +26,7 @@ export default function DateRange(props: Props) {
     maxDaysInFuture,
     disabled = false,
     initialValues = [undefined, undefined],
+    suffix
   } = props;
 
   const [dateRange, setDateRange] = useState<DateRangeType>(initialValues);
@@ -114,6 +116,9 @@ export default function DateRange(props: Props) {
             : undefined
         }
       />
+      {suffix && (
+        <span style={{marginLeft: '5px'}}>{suffix}</span>
+      )}
     </div>
   );
 }
--- a/src/packages/next/components/store/usage-and-duration.tsx
+++ b/src/packages/next/components/store/usage-and-duration.tsx
@@ -21,6 +21,15 @@ interface Props {
   extraDuration?: ReactNode;
 }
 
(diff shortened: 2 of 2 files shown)


williamstein:
Thanks.  This is basically an old dialog for admins to edit licenses for heavy touch customers, and not something users would ever see.  It's not sure it's a good idea to show the local timezone here, since it's the one place it could be misleading, since as an admin looking at it, you would guess CoCalc is trying to show the timezone for the user, not you. 

Most of the code for the store that you see here https://cocalc.com/store is in

https://github.com/sagemathinc/cocalc/tree/master/src/packages/next/components/store

It's part of the nextjs app in cocalc, which is separate from the single page app that you're editing in this PR.  
Note that hot module reloading is currently broken for our nextjs app for reasons we haven't had time to figure
out (but can, if only there were more time), so after making edits to the code in 

https://github.com/sagemathinc/cocalc/tree/master/src/packages/next/components/store 

you have to do a page refresh, which is annoying.

schrodingersket:
> It's not sure it's a good idea to show the local timezone here, since it's the one place it could be misleading, since as an admin looking at it, you would guess CoCalc is trying to show the timezone for the user, not you.

Understood. I'll rescind that change :+1: 

From #6920:

> It may not always be necessary, but is very helpful at least when setting up the license start/end time and when viewing invoices with time stamps.

Are the invoices here referring to those found in the store at `/billing/receipts`? Or is this referring to the statements that show up in `/settings/statements`? If it's the former, then I'm not sure there's much that can be done since those are generated by Stripe.

image

williamstein:
The invoices are generated using stripe.  But they don't have the license info on them anyways anymore -- they just say that you added credit to your account. 

I think a very useful and helpful initial goal would be to show the local timezone here to users:

image


https://cocalc.com/store/site-license?user=business&period=range&run_limit=1&member=true&uptime=short&cpu=1&ram=2&disk=3&range=2023-09-25T16%3A32%3A51.058Z_2023-10-02T16%3A32%3A51.058Z

schrodingersket:
> I think a very useful and helpful initial goal would be to show the local timezone here to users:

Done! The styling and form structure in that particular area was a bit challenging, so I went ahead and refactored things a little bit so that the custom license term follows the same look and feel of the rest of the form:

(with valid input)
image

(with invalid input)
image

williamstein:
Looking good!  Let me know when it's ready to merge.

schrodingersket:
> Looking good! Let me know when it's ready to merge.

We are good to go. Just had a quick rebase to do to clean up the commit history on my end.

williamstein:
Thanks!

PR #6941, opened 2023-09-27
store: disk speed wording

haraldschilly:
# Description

I think it's best to just state the values and point to their website.


williamstein:
>  point to their website.

I strongly agree.  People can do their own benchmarks to be sure, and the numbers are wildly different (better usually), than the guaranteed mins that Google suggests...

PR #6949, opened 2023-09-29
frontend/explorer: fix behavior when project is stopped

haraldschilly:
# Description

- when project is stopped and we don't have a cache of the listing, do not show the loading indicators
   - instead, tell the user the project must be started
- also, when there is no cached listing, do not start the project implicitly 

how I tested this (and where the limit to this is):

- stop the project
- delete the database entry for the listing in a path: `delete from listings where path = ...`
- refresh the browser client, because directories are cached, or close/open the project
- maybe stop the project again (because a refresh sometimes triggers a project start)
- with that, I can click through known directories while the project isn't running, while I see a warning text for unknown dirs.

with that, I get this:

[screencast-localhost_5000-2023.09.29-17_45_38.webm](https://github.com/sagemathinc/cocalc/assets/207405/8f974029-2d4c-4dda-b1c6-3b48db0c205e)

### DIFF
--- a/src/packages/frontend/account/public-paths/public-paths.tsx
+++ b/src/packages/frontend/account/public-paths/public-paths.tsx
@@ -3,7 +3,8 @@
  *  License: AGPLv3 s.t. "Commons Clause" – see LICENSE.md for details
  */
 
-import { Button, Checkbox, Table } from "antd";
+import ShowError from "@cocalc/frontend/components/error";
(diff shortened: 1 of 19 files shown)


williamstein:
OK, this actually has a bunch of really difficult (for me) conflicts to resolve, but probably easy for you.

haraldschilly:
In retrospect, I shouldn't have coded this in parallel with the other PR. Conflicts are resolved now.

Issue #6928, opened 2023-09-20
Current and running balance when viewing transactions


novoselt:
Related to #6919

In the admin view of user purchases it would be handy to see the current balance.

Moreover, it would be nice to see the balance after each transaction in the admin and user interface, similar to how banks show it for accounts.

novoselt:
I think this is fully implemented now!

Issue #6919, opened 2023-09-17
Make the credit balance easily visible


novoselt:
Perhaps it should not be on the screen all the time, but whenever the user is in the store, especially while configuring licenses, it would be very helpful to see the current balance. Maybe in some box in the corner.

williamstein:
@schrodingersket this could be ideal for you -- not a bug fix, but fun to add.

williamstein:
Scope -- this issue is just adding it in the https://cocalc.com/store

PR #6969, opened 2023-10-07
#6928, #6919: Added current and running balance to `PurchasesTable` component

schrodingersket:
# Description

The running balance now shows as a new column in the `PurchasesTable` component, and a new field is shown at the bottom with the user's current balance:

image

Addresses #6928 and #6919.


williamstein:
Nice!

PR #6979, opened 2023-10-13
mention: add note to email about how to reply

haraldschilly:
# Description

This is just a small note in the sent email, to tell users to not just reply.


williamstein:
@haraldschilly can you also add to the todo list that we *should* support email replies?  GitHub does, and you're improving our email infrastructure, and it could be an interesting challenge for someday.

PR #7003, opened 2023-10-25
obviously wrong shebang for command.sh

belonesox:
Obviously wrong shebang for command.sh
(just found debugging why «failed to load VS Code» on my setup)


williamstein:
Thanks!

Issue #7014, opened 2023-10-29
If DEBUG + DEBUG_FILE   is set for Hub in multi-user mode with sudo, Hub cannot start projects


belonesox:
When I try to debug cocalc, I set something like

```
  DEBUG=cocalc:*,cocalc:silly:*
  DEBUG_FILE=/var/log/colalc/cocalc-debugging.log
```
running Hub as sudo, with `--mode=multi-user`.

In such case it cannot start projects!

With 
* only DEBUG → OK
* DEBUG + DEBUG_FILE → Fail

So, then I should start Project from debugging Hub, (`server/projects/control/util.ts`, `launchProjectDaemon`), it failed with exit code 0. 
```
2023-10-29T11:04:39.508Z: cocalc:debug:project:BUG (uncaughtException) initializing uncaughtException handler
```

![image](https://github.com/sagemathinc/cocalc/assets/1609739/f9d548e8-faf0-4f1d-a3a2-fb61d9fd05ab)

Probably (I am not sure…), this is because it inherites from Hub environment `DEBUG_FILE`, and failed starting project, because Project, runned by unpriviliged user, tried to write to same DEBUG_FILE.

if we disable `DEBUG_FILE` setting for Hub, it creates project daemon without problems.

So may be better strip all `DEBUG_FILE` from env, running Project?
Because in `packages/project/client.ts` I see separate logic to enable debugging in projects

```js
const DEBUG_FILE = join(HOME, ".smc-DEBUG");
if (fs.existsSync(DEBUG_FILE)) {
  DEBUG = true;
```

williamstein:
> So may be better strip all DEBUG_FILE from env, running Project?

Yes.  Do it!  Thanks.

> Because in packages/project/client.ts I see separate logic to enable debugging in projects

That's just a coincidence that the variable is named the same thing.

PR #7018, opened 2023-10-30
Solving #7014, see https://github.com/sagemathinc/cocalc/issues/7014#…

belonesox:
# Description

One-line patch for solving #7014 + lets log stdout/stderr during launching project daemon runner.

### DIFF
--- a/src/packages/server/projects/control/util.ts
+++ b/src/packages/server/projects/control/util.ts
@@ -113,14 +113,6 @@ export async function launchProjectDaemon(env, uid?: number): Promise<void> {
       uid,
       gid: uid,
     });
-    child.stdout.on('data', (data) => {
-      // Lets's log stdout of process that running daemon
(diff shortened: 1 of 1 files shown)


src/packages/server/projects/control/util.ts
```diff
@@ -113,6 +113,14 @@ export async function launchProjectDaemon(env, uid?: number): Promise<void> {
       uid,
       gid: uid,
     });
+    child.stdout.on('data', (data) => {
```
williamstein, line 116: Please only do this if the logging is actually enabled.  E.g., `stdout ${data.toString("utf8")}` could take a lot of time to do nothing if logging isn't even enabled.

belonesox: I removed additional logging. Actually, regular stdout here consists of only a couple of lines before running the project daemon, but it's okay. I don't want to check the logging level here because all the other debug calls around it don't do it.

Issue #7019, opened 2023-10-30
Creating customized tokens that will have tunable expiration time


belonesox:
I implemented accepted idea from https://github.com/sagemathinc/cocalc/discussions/6483

Additionally to standard "two weeks token" (for «UI backward compatibility»),
we can create token with customizeable expiration time.

![image](https://github.com/sagemathinc/cocalc/assets/1609739/bd2eaf29-e8e4-4890-ae05-b1013190a5e2)

![image](https://github.com/sagemathinc/cocalc/assets/1609739/d396843b-efac-42bb-9b77-dc2d95941926)

Probably, there are possible to UI improvement (set reasonable expiration time by default, day? month?… ), 
may be something else. But it worked.

PR #7020, opened 2023-10-30
Implemented #7019 https://github.com/sagemathinc/cocalc/issues/7019

belonesox:
# Description
* First discussed here → https://github.com/sagemathinc/cocalc/discussions/6483
* Connected issue to implement #7019


williamstein:
Yes, this is way better from a security point of view!

PR #7031, opened 2023-11-06
Updated `src/README.md` to include instructions for using NVM and Python...

schrodingersket:
…virtual environments for local development. Added `requirements.txt` to install required Python utils. Added `pnpm` as a dev dependency.


williamstein:
Nice!  So much better :-)

Issue #6998, opened 2023-10-23
python 3.10: structural pattern matching syntax highlighting


haraldschilly:
Essentially, highlight `case` and `match` keywords in a code like the following. currently, those words are treated like a variable (a default syntax styling).

```{python}
match x:
   case "foo":
       return 1
```

This is probably easy by adjusting `frontend/codemirror/mode/python.js`

williamstein:
This is supported in the current version from codemirror 5: https://codemirror.net/5/mode/python/index.html

williamstein:
fixed by https://github.com/sagemathinc/cocalc/pull/7032

PR #7032, opened 2023-11-06
#6580: Updated `python.js` CodeMirror mode.

### LINKED ISSUE #6580: jupyter: misleading state when opening readonly files
1. create file `ro.ipynb`
2. enter something like 1+1, whatever, and save/quit/close it
4. terminal: `chmod ugo=r ro.ipynb`
5. check: `ls -l ro.ipynb` says `-r--r--r-- 1 user user ...`
6. when you open it, it looks fine. … but at the latest when I save, I see something is wrong:
   
   > Error: Error: error writing file: Error: EACCES: permission denied, open /home/user/ro.ipynb'

What I expect: I am not sure, I certainly want to read what's there, but I don't think it should run. Or we allow starting a kernel, with a warning (?), but all input cells should be marked as "readonly & delete protected" and we disallow switching the kernel. Certainly, we can't allow saving the file.

Note: the same works fine for markdown/code files.

schrodingersket:
# Description

Fixes #6580.

Before:

image

After:

image

(file copied from https://github.com/codemirror/codemirror5/blob/659df46b1f53cd94952058e23687f9e58e1b997e/mode/python/python.js)

### DIFF
--- a/src/packages/frontend/codemirror/mode/python.js
+++ b/src/packages/frontend/codemirror/mode/python.js
@@ -1,10 +1,5 @@
-/*
- *  This file is part of CoCalc: Copyright © 2020 Sagemath, Inc.
- *  License: AGPLv3 s.t. "Commons Clause" – see LICENSE.md for details
- */
-
 // CodeMirror, copyright (c) by Marijn Haverbeke and others
-// Distributed under an MIT license: http://codemirror.net/LICENSE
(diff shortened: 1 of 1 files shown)


schrodingersket:
Just for reference: I accidentally tagged this with the wrong issue number. This PR fixes #6998, and *not* #6580 .

PR #7046, opened 2023-11-15
expire more table data

haraldschilly:
# Description

## :warning: do not merge

* db/file_access_log: introduce "expire" column, with either the PII retention or 1 year as fallback
* add test for all possible pii expiration values, since we parse them
* db/central_log: expire after 1 year, unless it's a special case
* db/copy_paths: add an expiration for that kucalc related table
   * with the additional detail, that scheduled copy events are postponed 1 month after the scheduled time, at least
   * also added a new test for `control/kucalc.ts`
* db/jupyter_api_cache and jupyter_api_log: expire after 1 month (log is used for abuse, and only looks at 1 hour ago … 1 month is enough data to still have enough insight)

---

Test: I went through the tables in our kucalc test environment, including checking the copy event and the API call for scheduled copies

### DIFF
--- a/src/packages/database/postgres-base.coffee
+++ b/src/packages/database/postgres-base.coffee
@@ -1006,6 +1006,7 @@ class exports.PostgreSQL extends EventEmitter    # emits a 'connect' event whene
 
     # Go through every table in the schema with a column called "expire", and
     # delete every entry where expire is <= right now.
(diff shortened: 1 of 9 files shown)


haraldschilly:
this is fine to merge, just adds new columns. if there is no value (i.e. NULL) in the expire col, it won't fall on our head, because those rows will not be deleted.

Issue #7012, opened 2023-10-27
Allow users to NOT use their credit balance when making a purchase


novoselt:
Someone may have a lot or a little on their balance, but prefers to pay for a license in full using a credit card to get reimbursed later. At checkout there should be an option to use the credit balance or not.

PR #7065, opened 2023-11-25
Feature/7011

schrodingersket:
# Description

Added initial support for line items on Stripe invoices.

Some extra eyes on this (especially testing with various purchase configurations) would be greatly appreciated, since I'm still new to the invoicing/payment subsystem.


williamstein:
The Github actions checks fail.  I'm sure they pass on your computer.  You might want to try debugging directly on Github Actions using that thing I added?

williamstein:
I did look over the code itself and it looks pretty good, and it's nice that there's so many tests.  It also looks relatively unlikely to actually break something.

schrodingersket:
They do, but only via `ts-jest` (which makes debugging tests in my IDE
easier). I'm working on sorting out the difference between running that way
and compiling/running the tests as separate commands to un-bork the build.

williamstein:
> They do, but only via `ts-jest` 

I would use ts-jest instead of jest on the compiled files, if I could have figured out how to get it to work automatically on the CLI.

williamstein:
I built this and tried it and see lots of "undefined" in the invoice:

image

schrodingersket:
Yep - working on that now. It seems that there's a pretty big discrepancy
between the types asserted by function arguments (which is what the code
changes were based on) in the files that were touched and the actual
objects returned from the database, so I'm looking into tracking that down
to see where the differences are coming from.

williamstein:
> I'm looking into tracking that down to see where the differences are coming from.

Thanks!  That could really help improve development later!!

schrodingersket:
@williamstein I am able to reproduce the first part of the issue you outlined in the `master` branch, so it looks like this refactor uncovered existing behavior. The steps to reproduce are:

1) Start a project (ideally with high resource usage)
2) Add `x` amount of user credit.
3) Add a license to the shopping cart which costs more than `x + 2.5`.
4) Click the `Add $$ credit to your account...` button
5) Wait a minute or 10 for the running project to consume some of the user's existing account balance
6) Attempt to complete the Stripe payment.

I suspect that the rounding that occurs in the `getChargeAmount` function in `src/packages/util/purchases/charge-amount.ts` lessens the frequency of this happening in practice, since rounding up would add just enough buffer room in the user's balance to allow the transaction to go through despite the small-but-not-negligible project charge that happens while the user is entering card details to complete the transaction.

There's not really a great way around this, besides forcing the user to charge the entire transaction balance even when a partial credit is available. In that case, #7012 would provide the fix while also clearing up the second error you ran into.

williamstein:
> There's not really a great way around this, besides forcing the user to charge the entire transaction balance even when a partial credit is available. 

Alternatively, look at the current pay as you go rate, and add 30 minutes to that. 

For users using compute servers or pay as you go project upgrades, it could be extremely common that they have a pay as you go rate of say $0.25/hour.   In that case, we could just add $0.125 to give
them up to a half hour to complete the purchase.

williamstein:
In any case, I view this issue I noticed as NOT a part of this PR, and something for a followup issue! This is now https://github.com/sagemathinc/cocalc/issues/7099

williamstein:
I'm going to take over and do the "run limit" description request that Andrey tacked on at the end.  **I'm not doing this because I don't think you couldn't easily do it, but because it's much easier to understand code you're reviewing by adding on little feature to it, then just passively reading it.**

PR #7087, opened 2023-11-30
frontend/jupyter: add "computer servers" tab to the kernel selector

haraldschilly:
# Description

This adds a tab to the kernel selector (instead of the default list of all kernels), pitching compute servers and listing available images. 

(the "no kernel" box just happens because this is me on my dev machine)

Screenshot from 2023-11-30 16-25-36

### DIFF
--- a/src/packages/frontend/jupyter/select-kernel.tsx
+++ b/src/packages/frontend/jupyter/select-kernel.tsx
@@ -516,9 +516,9 @@ function ComputeServerInfo() {
   const { project_id } = useProjectContext();
   const actions = useActions({ project_id });
 
-  // get the keys of the IMAGES map and sort them by IMAGES[key].label
+  // sort all images with a jupyter kernel by IMAGES[key].label
   const sortedImageKeys = sortBy(
-    Object.keys(IMAGES),
+    Object.keys(IMAGES).filter((key) => IMAGES[key].jupyterKernels !== false),
     (key) => IMAGES[key].label,
   );
 
--- a/src/packages/util/db-schema/compute-servers.ts
+++ b/src/packages/util/db-schema/compute-servers.ts
@@ -28,6 +28,7 @@ interface ImageBase {
   source: string;
   versions: VERSIONS;
   authToken?: boolean; // if true, image has web interface that supports configurable auth token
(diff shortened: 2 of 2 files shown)


williamstein:
Just a comment about that screenshot -- neither ollama or CUDA dev have a Jupyter kernel.

williamstein:
What's the status of this?  You removed all the labels.

haraldschilly:
There was a test failure, now fixed. Ok. I see. I'll add a field to the server definitions, indicating if there is a jupyter kernel and filter on that.

williamstein:
> I'll add a field to the server definitions, indicating if there is a jupyter kernel and filter on that.

Yep, and I think everything else has a kernel.

I very much plan to move those definitions out of db-schema/compute-servers.ts, but it doesn't matter until there is a lot of usage.    Also, I'm not sure where to move it to.  It's a very similar problem to the one we have with software environments in kucalc.

haraldschilly:
The software env for cocalc cloud can be configured via a config map, exposed in json file. Doing this for the kucalc environment as well is an open task. It's just a bit more complex situation, since it's more than a single list, that's why it isn't 1:1 compatible. Certainly doable, though. In any case, IMHO compute servers could treated in a similar way.

I pushed a small commit, now the full list looks like all of them should have a kernel – otherwise it is missing.

Screenshot from 2023-11-30 17-30-13

williamstein:
> In any case, IMHO compute servers could treated in a similar way.

The actual full range of possibilities are defined by:

- the code here https://github.com/sagemathinc/cocalc-compute-docker, and
- what is actually built
- what has been tested and marked as such.  This involves private images hosted in google cloud for the GCP platform integration, but will be different for other integrations, e.g., for on prem, it's defined by what's on Dockerhub.  Tested is defined by google cloud labels.

So I think the solution involves documenting all possibilities via something in the code in https://github.com/sagemathinc/cocalc-compute-docker, then having some system for marking things as "tested", and finally having cocalc itself assemble information...  Also, It would be nice if there were a way for admins via the UI to manually adjust things, e.g., to turn off an image or revert a version.

haraldschilly:
yes, tests are happy as well :-)

PR #7094, opened 2023-12-01
support google vertex ai 

haraldschilly:
# Description

## deployment

first hub, then static

## tl;dr of changes

* site settings
  * "Google's VertexAI UI" enabled boolean (analog to openAI UI)
  * "Google GenerativeAI API key" secret (with a link to how to get the API key)
* the ChatGPT logo in the toolbar and similar is now a "head with AI" icon. We can change that later, mainly a placeholder for a really good generally understood AI assistance icon, that's vendor neutral.
* charging: gpt4 and the 16k model do still issue purchases, while 3 and palm2 are free
* chat/regenerate button: yes, this just sends the info again and processes the result, no changes
* markdown file type: I added a special case for "md" files, because they were prompted as "yaml frontmatter code files". Now, they're just called "Markdown documents". This improved the prompt, because there is no longer some random yaml frontmatter ending up in the reply
* course config for student projects: the actual "key" to disable it hasn't changed, but I've slightly generalized the name and phrase in the UI

## testing

below is what I actually tried in the end to do, and it's orange where it seems to work on a technical side, but PaLM2 isn't happy with the prompt. Below are examples and more details.

| aspect                | gpt3   | gpt4 | palm2 |
|-----------------------|--------|-----|---------|
| chat handle  | :green_circle: | :green_circle: | :green_circle: (`@palm2`) | 
| generating notebook   | :green_circle: | :green_circle: | :green_circle: | 
| notebook cell AI assistant | :green_circle: | :green_circle: | :green_circle: | 
| generating cell | :green_circle: | :green_circle: | :green_circle: | 
| explaining error in notebook | :green_circle: | :green_circle: | :orange_circle: [1] |
| help with latex error | :green_circle: | :green_circle: | :orange_circle: [2] |
| terminal assist | :green_circle: | :green_circle: | :orange_circle: [3] |
| markdown summarize [4] | :green_circle: | :green_circle: | :green_circle: | 
| markdown custom prompt [5] | :green_circle: | :green_circle: | :green_circle: | 

[1] often bails out, does not return anything. instead, user sees error that this prompt did not generate a result

[2] similar to the above, and interestingly, if you reply to the error about not having returned anything, it will continue to tell you more – although in this particular case here, the answer is wrong (but at least, there is an answer)

Screenshot from 2023-12-12 14-48-37

[3] similar to the above, this happens for a terminal chat. very interesting. maybe the questions need to be just simple enough?

Screenshot from 2023-12-12 14-52-57

---

[4] summarize/complete in markdown. Yet again, PaLM2 is funny. It can't make sense of git output, but when I start writing a few lines about learning terminal/bash, it throws a huge essay into the chat.

4.1: explain what's in a terminal:

Screenshot from 2023-12-12 15-00-18

4.2: write about learning bash

Screenshot from 2023-12-12 15-00-32

and just for fun, I copied all that output into the markdown file and asked PaLM2 to summarize it. It did!

Screenshot from 2023-12-12 15-04-11

---

[5] custom prompt, without any context (equivalent to an empty document): *"Write a table about pros and cons of allowing children to use a smartphone under the age of 10:"*

Screenshot from 2023-12-12 15-06-35

and then, asking it in a reply to clean this up and making it shorter

Screenshot from 2023-12-12 15-07-04

---

Far from done, but here is a first sign of life:

Screenshot from 2023-12-07 19-51-05

Screenshot from 2023-12-07 19-48-01


williamstein:
Is it intentional that you can't use @ to create a chat with palm?  That was the first thing I would have implemented but it seems to be missing.

image

williamstein:
When I try to use the notebook generator I see a big error at the top:

image

My standard test case for the notebook generator is "Draw a spiral" or "Plot a spiral".  This palm 2 model totally fails at doing that for either python or octave.  Chris on the chat said yesterday that maybe we should be using a different model ("codey"?) for code, and that maybe palm2 is really bad for code.

I tried "Plot sin(x)" for "Generate..." in an octave notebook and got this:

image

I.e., there is a final block of text included with code.

williamstein:
I feel like these issues are minor, that the LLM functionality is overall improved by this PR a lot, so it's probably best for the issues mentioned above to be in a followup PR.

PR #7112, opened 2023-12-09
Compute servers: naturally work with filesystem of compute server directly.

williamstein:
This is a work in progress and it will take a while, since it's a big "all or nothing" sort of project.

---

NOTES:

- must restart hub's when doing the release to update database permissions for the listings table
- rebuild the cocalc npm package that the compute servers run


williamstein:
After **two months** of this being in progress, I just pushed fixes for the last blocker issues on my big todo list for this!  Now it's all about integration testing and finally releasing this :-).

PR #7163, opened 2024-01-04
next/sign-up: role tags and optional contact

haraldschilly:
# Description

- Compared to before, only 4 roles
- tags/contact only shows up on the full sign-up page – not the minimal in-place ones.
- By default, at least one role must be selected. This can be made optional real quick by setting the env var `COCALC_SIGNUP_REQUIRE_TAGS=false`.
- professionals (and only them) are asked to be contacted + they can add a "goal". That "goal" is optional, at least for now it feels better that way.
- in the DB, the accounts get `tags: {personal,contact}` and eventually also the sign up intent.
- There are no implications for SSO – kind of not implemented/ignored. Something for later if we need it.
- No implications for on-prem setups

apart from that
- fix the UI indication that something is missing, e.g. the 6 letter requirement for the form submission
- reorg the bottom button statements. email comes earlier, hence move it up.
- spell out that the email field is for emails, just like first and last name

## UI after our discussions

initially

Screenshot from 2024-04-10 16-28-33

after selecting professional and adding a goal

Screenshot from 2024-04-10 16-28-58

that selection then ends up in CRM like

Screenshot from 2024-04-10 16-32-59

## Rollout

no changes to the DB, although I suggest to update hub-next first and then the frontend.


haraldschilly:
@novoselt this is the adjustment you were talking about yesterday. I'll post a screenshot once I'm done for the day. Then we can iron out the exact wordings and merge it.

haraldschilly:
ok @novoselt … above are screenshots and the only really relevant part to review is [this list](https://github.com/sagemathinc/cocalc/pull/7163/files#diff-c163c256ebb92c45da5e4e07d30b5e87dae5d370f97c9c2c857a5a485152097cR668)

There is also some issue with an npm lockfile, must be a result of a recent merge. I'll fix that as well.

novoselt:
I thought about these for a bit:
- Can we please replace "1" with "one"? For some reason it really bugged me before and now ;-)
- The colors don't seem to mean anything, but they make it difficult to see that some buttons are highlighted. I would stick to the same neutral color for all and then something more standing out for the selected ones
- "May we contact you" is a standard question to opt-out/in of marketing emails, that's not what we were discussing, I think. Can it be "Would you like to schedule a video call to ...?" Not sure about ... part, "help us understand your needs better and make suggestions"?
- If the box is checked, there may be a **required** text box "Please give us some idea of what you would like to discuss to help us prepare and have the right people on the call:" My hope is that those people who really want to talk and we can actually gain something from it won't be deterred by such a perfectly legitimate question, while we will not get requests from "random people"
- In a similar way, perhaps that box should only appear if something other than "Personal" or "Student" is picked. Perhaps only for "Commercial", "Government", "Instructor", "Researcher".
- The selection of roles seems fine (do we really have artists?! and collaborator is somewhat strange), but the order is a bit random, although I am not sure what would be better
- I am very much in favour of even number of buttons and having them split equally between two rows!

What will actually happen if someone does want a video call? Can there be a Zendesk ticket created automatically? We also do need to be able to contact that person, so we have to have a valid email.

williamstein:
> There is also some issue with an npm lockfile, must be a result of a recent merge. I'll fix that as well.

It was due to your other two PR's conflicting.  I think I fixed it in master.

haraldschilly:
> * Can we please replace "1" with "one"? For some reason it really bugged me before and now ;-)

Well, has been like that before. We would need something that converts numbers to english words, since the number is configurable. I bet there is a library that does that. Right now, my goal is to just get this done and see what happens. Such a detail could also be an unrelated follow up ticket.

> * The colors don't seem to mean anything, but they make it difficult to see that some buttons are highlighted. I would stick to the same neutral color for all and then something more standing out for the selected ones

Well, there were colors before, they're just random. Since the user just clicked, they know they clicked. It's not like you have to recognize an existing set of selections without an action first. But sure, I can tweak the space from where the colors are sampled. Adding a bit of color at least draws attention and makes it look more friendly.

> * "May we contact you" is a standard question to opt-out/in of marketing emails, that's not what we were discussing, I think. Can it be "Would you like to schedule a video call to ...?" Not sure about ... part, "help us understand your needs better and make suggestions"?

I can change that text to say anything. For me, as someone signing up, it would be very weird if it says something about a "video call". Maybe that's not even what we end up doing in that particular case. Is your point that this sounds like a newsletter? The main intention should be that this is a one time only thing, if at all.


> * If the box is checked, there may be a **required** text box "Please give us some idea of what you would like to discuss to help us prepare and have the right people on the call:" My hope is that those people who really want to talk and we can actually gain something from it won't be deterred by such a perfectly legitimate question, while we will not get requests from "random people"

The implementation only allows for tags and I don't want to make this overly complicated. My counterpoint is, we should just start with this based purply on tags and contacting directly, and then – if this is really necessary – make this more complex.


> * In a similar way, perhaps that box should only appear if something other than "Personal" or "Student" is picked. Perhaps only for "Commercial", "Government", "Instructor", "Researcher".

Yes, that's possible, but again, I don't really want to spend too much time on this. Which tags should it be, only those 4? Why not "artist" and "other"?

> * The selection of roles seems fine (do we really have artists?! and collaborator is somewhat strange), but the order is a bit random, although I am not sure what would be better

Collaborator: there are certainly cases where someone signs up, because they got an invitation to collaborate. That's probably their main motivation at this point. This tries to capture this along the other roles a user can have.

> * I am very much in favour of even number of buttons and having them split equally between two rows!

Well, they just wrap around where the line breaks.

> 
> What will actually happen if someone does want a video call? Can there be a Zendesk ticket created automatically? We also do need to be able to contact that person, so we have to have a valid email.

What happens is the tag,that's all. Then we can figure out what to do. If this is selected frequently, I wouldn't create tickets automatically – just overwhelms us. I would rather wait first and start manually, then iterate on that. The email address is in the account information and at that point we don't know if it is valid or not. 

Any comments on the hover texts?

haraldschilly:
Ok. there are a couple more changes

* The colors are duller
* There is now a simple conversion function for integers to english words – there are libraries out there, pretty sophisticated, but they are large and full of features we don't really need for this. Hence I just wrote something simple for now …
* I noticed an unused field in the database, which can be used for such a contact information text. That's now shown in all cases. I don't think we should restrict this right now, without knowing more how this works out.

Screenshot from 2024-01-05 10-56-55


Screenshot from 2024-01-05 10-57-15

haraldschilly:
more changes:

* renaming the text field description for being contacting to "intended use:"
* making that text field for being contacted mandatory
* slight rewording of the contact checkbox

Screenshot from 2024-01-05 16-56-51

novoselt:
Thank you for detailed comments and taking care of one, Harald! Looks way better in my opinion ;-)

I am still in favour of a more direct question. If we don't mention a video call, it still can be something like "Do you want us to contact you?" instead of "May we reach out to you?". The latter to me in this context feels kind of like begging - we know that nobody likes to be pestered, but could you pretty please let us do it anyway? And every polite person is supposed to say "Well, sure!" while thinking "No, but I guess I should let you..." With "Do you want us to contact you?" it is more about actual facts and desires with no implied pressure.

Of course I am not a native speaker and I may be overthinking it ;-)

williamstein:
I think it could also be helpful to get some feedback from Blaec about this PR before it goes live.  Unlike us, Blaec has *manually* tried to get this information about hundreds (or thousands) of users over the last few months, and has also done a lot of actionable stuff with it.

haraldschilly:
I figured out how to expose the `sign_up_usage_intent` field in the CRM view as well. Also, only showing these tags for the main site as choices.

@novoselt I can change it to whatever you want, no problem. Sounds better to me as well. I wonder about the list of tags and tooltip texts: [this list](https://github.com/sagemathinc/cocalc/pull/7163/files#diff-c163c256ebb92c45da5e4e07d30b5e87dae5d370f97c9c2c857a5a485152097cR670).

williamstein:
Look at new accounts created in the last hour:

image

Just deleting the existing tags seems like a bad move, given how much info they clearly convey right now.

novoselt:
> Just deleting the existing tags seems like a bad move, given how much info they clearly convey right now.

What do they actually convey, however, and how are we going to use these data?..

williamstein:
> What do they actually convey, however, and how are we going to use these data?..

Spend a few hours looking at what new users actually do, and compare that to what the tags are that they selected, then come back and ask the question.

PR #7174, opened 2024-01-10
7150: Added balance refresh to `PurchasesTable` refresh handler.

schrodingersket:
Also added `express` to `package.json` to allow `pnpm run c` out of the box, along with support for attaching v8 debugger to `hub` when running in development mode.

# Screenshot

image

# Testing

Click the "Refresh" button in the `PurchasesTable` component at http://localhost:5000/settings/purchases. Verify that the balance is refreshed.

### DIFF
--- a/src/package.json
+++ b/src/package.json
@@ -40,7 +40,7 @@
   },
   "homepage": "https://github.com/sagemathinc/cocalc",
   "devDependencies": {
-    "express": "^4.18.2",
+    "express": "^4.17.1",
     "pnpm": "^8.10.2"
   }
 }
--- a/src/package.json
+++ b/src/package.json
@@ -40,6 +40,7 @@
   },
   "homepage": "https://github.com/sagemathinc/cocalc",
   "devDependencies": {
+    "express": "^4.18.2",
     "pnpm": "^8.10.2"
   }
 }
--- a/src/packages/frontend/purchases/purchases.tsx
+++ b/src/packages/frontend/purchases/purchases.tsx
@@ -240,6 +240,11 @@ export function PurchasesTable({
     }
   };
 
+  const refreshRecords = async () => {
+    await getPurchaseRecords();
+    await getBalance();
+  }
+
   useEffect(() => {
     getBalance();
   }, []);
@@ -296,7 +301,7 @@ export function PurchasesTable({
           />
           {showRefresh && (
             <Refresh
-              handleRefresh={getPurchaseRecords}
+              handleRefresh={refreshRecords}
               style={{ marginRight: "8px" }}
             />
           )}
--- a/src/packages/hub/package.json
+++ b/src/packages/hub/package.json
@@ -89,7 +89,7 @@
     "sinon": "^4.5.0"
   },
   "scripts": {
-    "preinstall": "only-allow pnpm",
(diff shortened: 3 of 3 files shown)


williamstein:
According to the CI your removal of "pnpm" in the package.json was a little overly aggressive.  Can you fix that?

williamstein:
Maybe only-allow is slightly different than any other package, since it's critical to make sure somebody doesn't do `npm install`, and you end up with a catch 22 if you need only-allow to check whether you can install only-all.  Does that make sense?

I copied this from a pnpm tutorial, for what it is worth.

schrodingersket:
> Maybe only-allow is slightly different than any other package, since it's critical to make sure somebody doesn't do `npm install`, and you end up with a catch 22 if you need only-allow to check whether you can install only-all. Does that make sense?

Yep - now that you mention it, I remember coming to a similar conclusion a little while back. I think the way out of that is to explicitly specify the version of `only-allow` in the run command (e.g., `npx --package only-allow@x.x.x only-allow pnpm`

> I copied this from a pnpm tutorial, for what it is worth.

For sure. I'm not super opposed to always using the latest version, especially for packages with a very small API surface that's unlikely to change (like `only-allow`); implicitly installed dependencies are just a code smell for me, so I usually bring those up when I see them. That said, there are definitely occasions to ignore code smells, and it sounds like this might be one of them.

haraldschilly:
@schrodingersket I stumbled over this today. I want to run `npm run hub-project-dev` but due to removing "pnpm" [exactly here](https://github.com/sagemathinc/cocalc/commit/12d572b4173#diff-83ad6a907d4a785a321f7ea9a9c7a86c5d2bb56c22854c50820d2c8f8621f749R96) it is no longer able to find `"build"`. Maybe those `pnpm´s are necessary, when they're chained together with `&&`s ? Or maybe I'm doing something wrong and I have to call this differently? In any case, the `*-nobuild` variant works, so I can work around that. Just some feedback from my side...

schrodingersket:
> @schrodingersket I stumbled over this today. I want to run `npm run hub-project-dev` but due to removing "pnpm" [exactly here](https://github.com/sagemathinc/cocalc/commit/12d572b4173#diff-83ad6a907d4a785a321f7ea9a9c7a86c5d2bb56c22854c50820d2c8f8621f749R96) it is no longer able to find `"build"`. Maybe those `pnpm´s are necessary, when they're chained together with `&&`s ? Or maybe I'm doing something wrong and I have to call this differently? In any case, the `*-nobuild` variant works, so I can work around that. Just some feedback from my side...
> 
> **edit** and there are also other targets with `&&`s where `pnpm` was removed. So, they probably fail in a similar way.
> 
> **edit2** but what's working for me now is the server side nodejs debugging. that's very cool to have!

@haraldschilly Thanks for letting me know; looks like I just got the `bin` and `scripts` fields mixed up. I hadn't tested those yet, which is why I'd left the PR in progress. Here's the new one with the appropriate changes (and everything tested :) - https://github.com/sagemathinc/cocalc/pull/7176

Glad to hear debugging is working for you! Are you debugging via Chrome debugger or with your IDE? The latter was a bit painful for me to get set up with IntelliJ - curious what your experience is like.

Issue #7171, opened 2024-01-08
Cancelling a subscription might be off by a factor of 100x


novoselt:
Transaction 16877 cancels an erroneous subscription for a refund. It was $0.3391 less than purchasing price, so this was the cost for about 27 minutes. But with $5.43/month the cost of 27 minutes is $0.00339375, which is very close except for the two leading zeroes. Looks like a confusion in code between cents and dollars.

williamstein:
> Looks like a confusion in code between cents and dollars.

Interesting.  All the cocalc code uses dollars as the unit, but stripe uses cents.   

I have unassigned @schrodingersket since this is high priority, and anything genuinely high priority can't be assigned to him as he is working very much part time, right?

williamstein:
Also, changing the title to "might", since a factor of 100x would mean hundreds of dollars off for other cases that are more than a few minutes, and we haven't seen that, right?

novoselt:
Makes perfect sense!

williamstein:
There is absolutely no way that the refund is off by a factor of 100x.  E.g., I have a test $9.05/month subscription that was 4 days into it's period.   I just canceled it now and got a refund for $7.81, which seems almost exactly right.    There's clearly no way that things are just off by a factor of 100x in general.

I realize now that your title is very ambiguous and vague.  it is "Cancelling a subscription seems to be off by a factor of 100x".  I interpreted to mean "Cancelling a subscription refunds an amount that is off by a factor of 100x".     I think now you actually mean "Cancelling a subscription isn't exactly right, and I noticed one number is roughly 100x another number."

williamstein:
I've moved this back to assigning to @schrodingersket since I don't really know what this ticket really means.  If it really were a clear case that we were charging 100x too much, then it would be a blocker to be fixed ASAP, but that's not the situation. Sorry for any noise.

novoselt:
It is indeed an observation on a single transaction which I've mentioned. Perhaps with the exact details of that license-edit it is possible to track down where exactly the discrepancy is coming from by plugging in its parameters into some tests.

schrodingersket:
FWIW: This difference between resume/cancel is exactly what I'm working on resolving since that needs to be fixed in order to resolve a couple of the other issues I'm working on.

Issue #7161, opened 2024-01-03
Cancelling a subscription should NOT depend on available balance!


novoselt:
Yet I got this message for this transaction:
![image](https://github.com/sagemathinc/cocalc/assets/1993227/e75bbbb2-57d6-440e-b7e8-63e5c000c961)

schrodingersket:
One (likely related) observation I made while investigating #7141: when a license is cancelled, its new end date is set 10 minutes into the future (see https://github.com/sagemathinc/cocalc/blob/master/src/packages/server/purchases/cancel-subscription.ts#L35) and the remaining subscription amount is refunded from that point up through the natural termination of the subscription. This results in a non-zero charge to the account (i.e., the `$0.01` amount seen above) and the rest is refunded after this change. If the user cannot charge anything to their account (due to, e.g., a negative account balance), that would almost certainly trigger the above error.

Issue #7141, opened 2023-12-16
Resuming a subscription is still broken and confusing


novoselt:
I am trying to resume a subscription and see
![image](https://github.com/sagemathinc/cocalc/assets/1993227/102e1ef3-e929-4a5e-8a1e-17b8f3b3623d)
**Why can't I see here that exact amount for which I will be charged?!** It HAS to be computed to initiate the transaction, can't we first compute it, then display the message, then process?.. But it gets worse:
![image](https://github.com/sagemathinc/cocalc/assets/1993227/a04cf4b1-17a3-4052-b047-46ac430b9c3c)
- first of all, it was a lie, that I will be charged at most $5.43
- it is unclear how this amount was calculated, although it is clear in this case that it is wrong
- if I want to resume the subscription and I don't have enough credit, my intention is most likely to pay for it, so the next step should be dealing with the payment system, NOT getting an error message and going to some other page to do some unrelated balance top up, trying to keep in mind how much do I need...

What has to be done:
1. Compute the resume charge correctly, the formula is "monthly cost" * ("current period end" - "now") / ("current period end" - "current period start") and there are no other variants. "Now" is this formula has to be computed only once, to avoid discrepancies in displayed values in different places including the actual charge.
2. Display the actual compute charge like: "You will be charged $2.34, which is the prorated cost for the rest of the current period, till Januber 35."
3. If there is not enough credit, initiate the payment transaction! This is the step to deal with and explain the minimal transaction amount and that the difference will be credited to the account, not lost!

This is a very high priority - subscriptions are not only for paying users, but for users who have the intention to continue paying in the future, the experience should be smooth, quick, and transparent!

PR #7178, opened 2024-01-15
#7161: Fixed subscription cancel/resume logic when computing cost to resume or cancel an existing subscription.

schrodingersket:
# Description

This pull request reconciles the cost to cancel or resume an existing subscription. In particular, this pull request now computes the cost to resume a subscription based on the stored hourly rate of the subscription. There is still some follow-up work to do, so this is _not_ ready for merge, but these changes are ready for feedback. The items yet to be addressed are:

1) There is a discrepancy of precisely $0.02, which is introduced [here](https://github.com/schrodingersket/cocalc/blob/e78b23b3e07a5f0703d9c184266addbc1b7ab437/src/packages/util/purchases/cost-to-edit-license.ts#L177); the `modifiedPrice.discounted_cost` field is set to $0.02 when a refund is issued for an existing subscription license. It's not yet clear whether this should be removed when issuing a subscription.

2) Upon subscription cancellation, the cancelled subscription should be updated to current rates so that when resumed, the user is charged according to the latest subscription cost.

Note that this pull request only changes the backend code used to compute these costs, and is only a partial fix for #7141. This PR _should_ address #7171 and #7161, but will require further testing before closing those out.

### DIFF
--- a/src/packages/next/pages/api/v2/purchases/cancel-subscription.ts
+++ b/src/packages/next/pages/api/v2/purchases/cancel-subscription.ts
@@ -23,6 +23,6 @@ async function get(req) {
     throw Error("must be signed in");
   }
   const { subscription_id, now } = getParams(req);
-  await cancelSubscription({ account_id, subscription_id, now });
+  await cancelSubscription({ account_id, subscription_id, cancelImmediately: now });
   return { status: "ok" };
 }
--- a/src/packages/server/purchases/cancel-subscription.ts
+++ b/src/packages/server/purchases/cancel-subscription.ts
@@ -1,39 +1,42 @@
 import getPool, { PoolClient } from "@cocalc/database/pool";
+import getLicense from "@cocalc/server/licenses/get-license";
+
 import editLicense from "./edit-license";
 import { getSubscription } from "./renew-subscription";
-import dayjs from "dayjs";
-import getLicense from "@cocalc/server/licenses/get-license";
 
 interface Options {
   account_id: string;
   subscription_id: number;
-  now?: boolean;
+  cancelImmediately?: boolean;
   client?: PoolClient;
 }
 
 export default async function cancelSubscription({
   account_id,
   subscription_id,
-  now,
+  cancelImmediately,
   client,
 }: Options) {
   const pool = client ?? getPool();
+  const now = new Date();
+
   await pool.query(
-    "UPDATE subscriptions SET status='canceled', canceled_at=NOW() WHERE id=$1",
-    [subscription_id]
+    "UPDATE subscriptions SET status='canceled', canceled_at=$1 WHERE id=$2",
+    [now, subscription_id]
   );
-  if (now) {
+  if (cancelImmediately) {
     const subscription = await getSubscription(subscription_id);
     const { metadata, current_period_end } = subscription;
     const license = await getLicense(metadata.license_id);
     let end;
-    if (license.activates != null && new Date(license.activates) > new Date()) {
+
+    if (license.activates != null && new Date(license.activates) > now) {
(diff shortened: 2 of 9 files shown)


williamstein:
Despite reading this and it seeming right, I merged this and then did just one testing with an existing business subscription on our dev server (canceling/resuming/canceling/resuming), and it's very wrong (i.e., every cancel/resume reduces my balance significantly) -- probably you forgot about business v academic?   I'm 100% certain this is with your code.  Also, I tested with an academic license and it works fine.

image

I'm going to just try to fix this myself, since this is already merged and in general fixing this functionality is obviously very important.     I should probably just fix absolutely all known bugs in the billing code this week, since it's taking too long (I was holding off before due to compute servers).

schrodingersket:
Sounds good. I tested with a business license with the default configuration - I'm going to give this a test with the configuration you have above so I can reproduce and investigate for my own edification.

williamstein:
It's not a problem with "business" but with the *run limit*:

image

I guess the correct invariant to aim for is: 

- create a license with all parameters pretty random
- immediately cancel it
- see that the amount refunded is VERY close to the charged amount
- start it again: see that the amount charged is VERY close to the initial charged amount

schrodingersket:
Looks like the refund computation for a partial period is only computing the refund for a single project.

schrodingersket:
Changing this line:

https://github.com/sagemathinc/cocalc/blob/master/src/packages/util/licenses/purchase/compute-cost.ts#L167

to this:

```
    base_cost = hours * info.cost_per_hour / (quantity || 1);
```

seems to do the trick (modulo the `0.02 * n` charge we previously discussed).

williamstein:
Thanks!  You perfectly saved me time since I ended up fixing another bug that I hit first, which was that the amount for the initial purchase didn't include the discount (for the initial purchase), and hadn't got to this issue.

I rewrote the code like this to be more robust:
```ts
  let base_cost;

  if (subscription === "no") {
    // Compute license cost for a partial period which has no subscription.
    // Or, the license is a subscription, but we're only computing the cost for a
    // portion of the subscription period (which occurs when computing refunds
    // for a cancelled subscription).
    //
    if (start == null) {
      throw new Error("start must be set if subscription=no");
    }
    if (end == null) {
      throw Error("end must be set if subscription=no");
    }
    // scale by factor of a month
    const months = (end.valueOf() - start.valueOf()) / ONE_MONTH_MS;
    base_cost = months * cost_per_project_per_month;
  } else if (partialPeriod) {
    // Compute license cost for a partial period. This happens when the license is a subscription,
    // but we're only computing the cost for a portion of the subscription period (which occurs when
    // computing refunds for a cancelled subscription)
    //
    if (start == null) {
      throw new Error("start must be set if computing for a partial period");
    }
    if (end == null) {
      throw Error("end must be set if computing for a partial period");
    }
    if (!info.cost_per_hour) {
      throw Error(
        "cost_per_hour is required when computing license cost for a partial period",
      );
    }
    const hours = (end.valueOf() - start.valueOf()) / ONE_HOUR_MS;
    const cost_per_project_per_hour = info.cost_per_hour / (quantity ?? 1);
    base_cost = hours * cost_per_project_per_hour;
  } else if (subscription === "yearly") {
    // If we're computing the cost for an annual subscription, multiply the monthly subscription
    // cost by 12.
    //
    base_cost = 12 * cost_per_project_per_month;
  } else if (subscription == "monthly") {
    base_cost = cost_per_project_per_month;
  } else {
    throw Error(
      "BUG -- a subscription must be yearly or monthly or a partial period",
    );
  }
```

williamstein:
And finally things are consistent:

image

PR #7180, opened 2024-01-15
simple -- make the editor title bar much simpler and more organized using menus

williamstein:
The highest possible impact change regarding reducing the initial complexity of CoCalc is to move most of the title bar buttons to organized menus.   Keyboard shortcuts could then be documented (and added when missing) to make it easy to get quick access to each command, quicker than clicking a button.   This is what it looks like so far:

image

image

image

image


williamstein:
image

PR #7182, opened 2024-01-16
fix bugs in purchasing

williamstein:


src/packages/util/licenses/purchase/compute-cost.ts
```diff
@@ -132,13 +134,13 @@ export function compute_cost(info: PurchaseInfo, partialPeriod: boolean=false):
   cost_per_project_per_month *=
     COSTS.user_discount[user] * COSTS.sub_discount[subscription];
 
-  let base_cost = cost_per_project_per_month;
+  let base_cost;
 
   if (subscription === "no") {
     // Compute license cost for a partial period which has no subscription.
-    // b) The license is a subscription, but we're only computing the cost for a
+    // Or, the license is a subscription, but we're only computing the cost for a
```
schrodingersket, line 141: One quick note since you have a PR open anyway - this comment really should just be

```
 // Compute license cost for a partial period which has no subscription.
```

since the rest is extraneous from an overeager copy-paste on my side.

PR #7184, opened 2024-01-18
frontend/flyout/files: duplicate controls at the top

haraldschilly:
# Description

for better or worse, this adds an "action bar" row at the top. other online file explorers alternate between the navigation/filter bar and the action buttons to safe space. We could try that in a follow up modification.

Screenshot from 2024-01-18 14-04-19

Screenshot from 2024-01-18 14-03-55


williamstein:
NOTE: I've crashed my browser a few times with this, due to the controls at the top appearing even though nothing is checked.  I stupidly didn't get a stack trace and can't reproduce it easily.  I'll look at the code and see if that helps.

Issue #7173, opened 2024-01-10
Polish further license info presentation


novoselt:
At the moment I see in the shopping cart:
![image](https://github.com/sagemathinc/cocalc/assets/1993227/e1da8936-1a75-4f4b-ae48-dba2aaf76d4b)

1) While it does look pretty, it is not immediately obvious why dates are displayed "separately" in bold. This may be a good idea, even though it slightly complicates my copy-pasting ;-)
2) More confusing is seeing "1 simultaneously running project" in the "main text" and on the button. Perhaps this is the reason why this number was not shown in the license description before? I would much rather keep the text and remove the button. I didn't even realize at first that this is a button. Clicking it allows one to edit the number of projects, but there is also "Edit" button and I am not convinced that we need a shortcut for a particular field.
3) Clicking the Edit button actually leads to "Application error: a client-side exception has occurred (see the browser console for more information)."

williamstein:
> Clicking the Edit button actually leads to "Application error: a client-side exception has occurred (see the browser console for more information)."

Edit works fine for me and has in all cases I've ever tested.  This error could happen for a specific shopping cart that has corrupted data, so I tried impersonating your account and indeed editing the first item shows the error:

<img width="571" alt="image" src="https://github.com/sagemathinc/cocalc/assets/1276278/6452ce1f-4ba1-4cd9-8460-cd84c08c4ce8">

and drilling down the actual error happens parsing a timestamp in the highlighted line here:

<img width="1248" alt="image" src="https://github.com/sagemathinc/cocalc/assets/1276278/53e0029e-5821-4b53-9054-d0f2e9aa2823">

Drilling down further, there is a function in src/packages/next/components/store/usage-and-duration.tsx called `getTimezoneFromDate` and the error is from the "formatToParts" call there

<img width="545" alt="image" src="https://github.com/sagemathinc/cocalc/assets/1276278/7780cfb0-508e-4766-b151-bac2f8ee7ead">

williamstein:
@schrodingersket hopefully the previous comment makes it easy for you to fix the crash?

novoselt:
And obviously there are use cases for impersonating admins ;-)

williamstein:
Since I was just looking at this, I'm going to fix just the date parsing right now (but nothing else).

williamstein:
> And obviously there are use cases for impersonating admins ;-)

Just used it to confirm this crash is fixed for you.

williamstein:
Can just delete that button.

PR #7199, opened 2024-01-22
#7173: Removed license `run_limit` edit button in `ShoppingCart` component

schrodingersket:
...and moved license period into description box.

# Description

See screenshot for new view:

image


williamstein:
Ready for review?

PR #7231, opened 2024-02-01
next/compute-infographic: change to "neural networks"

haraldschilly:
# Description

@schrodingersket this changes that string to "neural networks"


schrodingersket:
@haraldschilly it looks like the change in text also change the color of that particular label to black, which is at odds with the rest of the color scheme. Was that intentional for emphasis?

haraldschilly:
@schrodingersket oh no, now I see what you mean. No intention, but I don't think this matters. I see this as a quick for now, to have something visual to grab attention, but beyond that it's just a placeholder to be filled in later.

PR #7232, opened 2024-02-01
frontend/x11: fix fallout of using our own reuseInFlight

haraldschilly:
# Description

see title, also binding all other wrapped method calls, just to avoid surprises … in any case, with that I can run X11 apps in my dev project, format code, etc.

### DIFF
--- a/src/packages/frontend/frame-editors/x11-editor/xpra-client.ts
+++ b/src/packages/frontend/frame-editors/x11-editor/xpra-client.ts
@@ -6,21 +6,21 @@
 // Use Xpra to provide X11 server.
 
 import { join } from "path";
+import { throttle } from "underscore";
+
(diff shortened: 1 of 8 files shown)


src/packages/frontend/frame-editors/x11-editor/xpra-client.ts
```diff
@@ -63,7 +63,7 @@ export class XpraClient extends EventEmitter {
   constructor(options: Options) {
     super();
     this.record_active = throttle(this.record_active.bind(this), 30000);
-    this.connect = reuseInFlight(this.connect);
+    this.connect = reuseInFlight(this.connect.bind(this));
```
haraldschilly, line 66: essential is this line, a missing `bind(this)` ... just like with the throttle in the line above.

other changes are just sorting and formatting

PR #7238, opened 2024-02-04
Chore/7213: Inline login component

### LINKED ISSUE #7213: change the "Explorer Published documents..." button on the landing page with a link to the shared files
This button on the landing page:

image

currently reveals its own weird non-searchable listing of published files, which also includes github things people visit.  Instead, completely replace this by a link to 

https://cocalc.com/share

which is much better -- it hides the github links, it has search, etc.

schrodingersket:
# Description

Fixes #7213. Definitely had some scope creep here (cf. https://www.youtube.com/watch?v=8fnfeuoh4s8). In general, this PR implements a standardized view for auth-like flows (e-mail verification, sign up, password reset, etc.) as a result of updating the inline sign up/sign in component to be embeddable into the main landing page. As those changes propagated, I ran into some additional places where verbiage could be updated and typos fixed.

More specifically, the following changes are implemented in this PR:

- Initial landing page clean-up. Added in-place login/sign up form to landing page. Added brief site description to header.
- Sign in, sign up, and password reset components are now much more responsive to parent containers; all auth components now share a common design language via `packages/next/components/auth/fragments/auth-page-container.tsx` component.
- Simplified password reset UI.
- Fixed (or at least ameliorated) small-screen layout for site licenses.
- Added org logo when user is logged in.

The majority of the changes here are in the landing page, so that's where I think feedback would be most valuable. Pending a clean build, this PR is ready for merge from a functionality perspective, but should be treated like an initial draft of new landing page verbiage since that page is probably the single most important page for new users.

Some screenshots:

Unauthenticated landing page (before):
image

Unauthenticated landing page (after):

image

Authenticated landing page (before):

image

Authenticated landing page (after):

image


williamstein:
Some automated tests fail due to react compilation errors.  These are the sort of things you might not see if using vscode, but are clear if you do: `pnpm clean; pnpm build-dev` at the top level once.  Doing `pnpm build` is even more strict.

williamstein:
The screenshots and design look great!!

novoselt:
I like how it looks overall and that it became more compact, so that info on compute servers  became visible!

I would remove the second line at the top, however: "Open CoCalc is a ..." since it does not add much to the first line, but does take up space and reading time.

Instead of "Don't want to provide any information?" I would have everything on the same line: "New to Open CoCalc? Sign Up or Try Open CoCalc without creating an account" - on the one hand it saves vertical space, on the other hand let's not put into users head that they don't want to provide us any information. They do, they just don't realize it yet ;-)

williamstein:
> I would remove the second line at the top, however: "Open CoCalc is a ..." since it does not add much to the first line, but does take up space and reading time.

This might be something that is at least very important for on prem sites, and Harald specifically added config info to admin settings for what goes there.  So removing it needs to be aware of this possibility.

schrodingersket:
> I like how it looks overall and that it became more compact, so that info on compute servers became visible!

Making compute servers (specifically GPU support) was one of the primary design goals of the refactor :)

> I would remove the second line at the top, however: "Open CoCalc is a ..." since it does not add much to the first line, but does take up space and reading time.

> Instead of "Don't want to provide any information?" I would have everything on the same line: "New to Open CoCalc? Sign Up or Try Open CoCalc without creating an account" - on the one hand it saves vertical space, on the other hand let's not put into users head that they don't want to provide us any information. They do, they just don't realize it yet ;-)

Agreed! I do like that much better. How's this? 

image

schrodingersket:
> > I would remove the second line at the top, however: "Open CoCalc is a ..." since it does not add much to the first line, but does take up space and reading time.
> 
> This might be something that is at least very important for on prem sites, and Harald specifically added config info to admin settings for what goes there. So removing it needs to be aware of this possibility.

My thoughts exactly. I'll check out what's currently in `master` to make sure that's somehow preserved in the description change.

schrodingersket:
With all of the above changes, here's where we're at now:

image

I added the `siteDescription` tagline below the embedded promo video since there was some empty space there and it didn't feel like a totally unreasonable layout association to make.

With the registration view as the default, there's a bit of a compromise with respect to the compute server blurb: the descriptive text and image disappears, but the headline is still clearly visible.

schrodingersket:
All comments should now be addressed :+1:

PR #7274, opened 2024-02-12
Added support for self-signed SSL configuration with Postgres.

schrodingersket:
# Description

This is an _in-progress_ PR; the functionality and tests to add support for environment-driven SSL configuration for Postgres is complete, but documentation is still in progress.

## New Configuration

This PR introduces five new (optional) environment variables which are named in accordance with existing PSQL environment variables. These are:

- `SMC_DB_SSL`: If set to "true" (case-insensitive), this setting enables SSL connection using the default system certificate store. This means that the certificate presented by the external RDBMS must be in the host OS' trust store. If any of the next four variables are set, it is assumed that SSL is to be enforced, and the value of this variable is ignored.
- `SMC_DB_SSL_CA_FILE`: Specifies a path to the certificate authority file.
- `SMC_DB_SSL_CLIENT_CERT_FILE`: Specifies a client certificate file to be used for client authentication.
- `SMC_DB_SSL_CLIENT_KEY_FILE`: Specifies a client certificate key file to be used for client authentication.
- `SMC_DB_SSL_CLIENT_KEY_PASSPHRASE`: The passphrase used to encrypt the file specified in `SMC_DB_SSL_CLIENT_KEY_FILE`.

When none of these environment variables are provided, database connections are executed as before.

## Status

- ~~Add support for validating against a custom certificate authority.~~
- ~~Add support for client authentication.~~
- ~~Unit tests.~~
- ~~E2E test certificate validation with self-signed certificates/certificate authority.~~
- Update [documentation](https://github.com/sagemathinc/cocalc-docker/blob/master/docs/external-postgresql.md)
- ~~E2E test client certificate authentication.~~
- ~~Add to [data.ts](https://github.com/sagemathinc/cocalc/blob/master/src/packages/backend/data.ts)~~


haraldschilly:
I'm happy to merge this. Without knowing all the details, the only little detail I can find is a discrepancy between  the "node postgres" package doc and this: https://node-postgres.com/features/ssl has a parameter `rejectUnauthorized: false` and here is no such switch. Default is true, according to [node.js/tls](https://nodejs.org/api/tls.html#tls_tls_connect_options_callback). My impression is you want this to be 100% secure or break. I wonder if in the real world, that `rejectUnauthorized` should be configurable. In any case, that's an easy change for later.

PR #7277, opened 2024-02-13
Reintroduce auth component changes

schrodingersket:
# Description

This PR brings back the authentication changes from https://github.com/sagemathinc/cocalc/pull/7238 (along with https://github.com/sagemathinc/cocalc/pull/7231 and https://github.com/sagemathinc/cocalc/pull/7251, which were also reverted in https://github.com/sagemathinc/cocalc/commit/2b88d398271ef856ad6e04a5f5e9642dba6cce89), modulo the changes to the landing page.

## Testing Instructions:

- [ ] The most important thing to test here (other than logins) is using an anonymous account to run a notebook with an external license as described here: https://github.com/sagemathinc/cocalc/issues/7252.
- [ ] Verify that the `Store` sub-menu header in the CoCalc Store directs to the Store Overview page.
- [ ] Verify that the new link to the Voucher Center in the Store Overview page does in fact redirect to the Voucher Center.
- [ ] Verify that the license computation header shows correctly in mobile on the `/store/site-license` page.


williamstein:
I think the top navbar looks out of balance or lopsided with this change:

image

The wrap behavior when you make the browser thinner is also not as good:

image

And it gets weirder:

image

Though on iphone 12 maybe this PR looks better:

image

Is it necessary to change the layout of the top navbar at all in this PR?

schrodingersket:
> Is it necessary to change the layout of the top navbar at all in this PR?

Not at all! Those changes were made in order to float the Sign In/Sign Up/Try buttons over to the right in light of some prior discussion we'd had, which had a side-effect of changing the wrapping behavior. Happy to backtrack that change and re-examine later when we have time to outline the desired behavior in more detail. Otherwise, I can look into spending some time trying to reproduce the old wrapping behavior precisely.

Just let me know which direction you'd like to go (:

williamstein:
> Happy to backtrack that change

Yes, please backtrack that change.

schrodingersket:
> Yes, please backtrack that change.

Done!

williamstein:
This works fine!

PR #7309, opened 2024-02-28
frontend/jupyter: show placeholder text in empty code cell

haraldschilly:
# Description

This is surprisingly hard and I spent way more time on this than I anticipated.

* The first and solved obvious detail is to show the cell generation dialog and to start inserting the output into the current cell.
* When I select an empty cell, it shows codemirror with a cursor, other cells aren't active. So, showing this only for a selected (?) cell but not in edit mode is kind of pointless, because this almost never happens to me.
* instead, this shows it for all empty code cells, similar to what I saw with colab in our video calls (it doesn't show any of that for me over here in the EU, though)
* related to focus/selected, the remaining problem is that clicking on a cell the usual way should change it to edit mode. I can't figure out how this works. The `click_on_cell` function is getting called, but doesn't switch the cell to edit mode.
* it also adds a checkbox if the previous cell should be included or not

### DIFF
--- a/src/packages/frontend/client/openai.ts
+++ b/src/packages/frontend/client/openai.ts
@@ -46,6 +46,7 @@ export class LLMClient {
     return await this.queryLanguageModel(opts);
   }
 
+  // ATTN/TODO: startExplicitly seems to be broken
   public languageModelStream(opts, startExplicitly = false): ChatStream {
     const chatStream = new ChatStream();
     (async () => {
--- a/src/packages/frontend/jupyter/cell-input.tsx
+++ b/src/packages/frontend/jupyter/cell-input.tsx
@@ -6,13 +6,18 @@
 /*
 React component that describes the input of a cell
 */
+
+import { Button, Tooltip } from "antd";
+import { delay } from "awaiting";
 import { Map } from "immutable";
 import { useCallback, useEffect, useRef, useState } from "react";
-import { Button, Tooltip } from "antd";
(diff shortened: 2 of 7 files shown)


williamstein:
> When I select an empty cell, it shows codemirror with a cursor, other cells aren't active. So, showing this only for a selected (?) cell but not in edit mode is kind of pointless, because this almost never happens to me.

- [ ] I consider this a blocker, but I could do the work to fix it.

> related to focus/selected, the remaining problem is that clicking on a cell the usual way should change it to edit mode. I can't figure out how this works

- [ ] That's of course a blocker.

> it also adds a checkbox if the previous cell should be included or not

Great idea to clarify exactly what is sent!  We could also add a checkbox to include all content, or all content above the current cell.  (This is of course starting to look a lot like the Assistant dialog at the top...)

It sounds like I should take over this PR and fix the remaining two issues above?  Plus also fix that the markdown comment is left in edit mode (which bugs me a lot).

haraldschilly:
> It sounds like I should take over this PR and fix the remaining two issues above? 

Yes, that would be great!

williamstein:
ok, I'll plan to finish this.

williamstein:
I redid it from scratch via a very different approach that really does involve keeping the placeholder text in the SAME codemirror component as actual editing, rather than not rendering that at all.  This will avoid problems down the line.
I fixed some other issues, e.g., the markdown being stuck in edit mode.

It would be good if you could extract your code toward https://github.com/sagemathinc/cocalc/issues/7311 at some point!

Issue #6463, opened 2023-02-20
change all instances of "new Date().valueOf()" to "Date.now()"


williamstein:
We use "new Date().valueOf()" in over 100 places in our codebase, since that works everywhere, and used to be the right thing to do, etc.:
```sh
~/cocalc/src$ git grep "new Date().valueOf()" |wc -l
     110
```

However, @haraldschilly has been randomly changing instances of "new Date().valueOf()" to "Date.now()", since it's "better".    

Some notes:
- "Date.now()" is NOT supported in old browser (e.g., IE), but that doesn't matter, since CoCalc isn't supported at all in older browsers, either.
- "Date.now()" is barely faster than "new Date().valueOf()" these days.  It's easy to find this sort of statement in Stackoverflow - -"They are effectively equivalent, but you should use Date.now(). It's clearer and about twice as fast." from a few years ago.  I just tried https://jsbench.me/f5k9ckm6lh/1 and "new Date().valueOf()" was maybe 10% slower for me on Chrome. 

The point of this ticket is simple: do a once and for all change of all instances of "new Date().valueOf()" to "Date.now()" in our codebase.

haraldschilly:
also:
* [that MDN page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/valueOf) says: 
  > This method is usually called internally by JavaScript and not explicitly in code.
* and by calling the static `now` method we avoid instantiating the date object. One object less to garbage collect :man_shrugging:

schrodingersket:
This should be good to close out :+1:

williamstein:
thanks!

PR #7328, opened 2024-03-04
#6463: Changed all instances of `new Date().valueOf()` to `Date.now()`.

schrodingersket:
# Description

...since presumably, we don't need to support browsers earlier than IE9.


williamstein:
> ...since presumably, we don't need to support browsers earlier than IE9.

Agreed.  In fact, we only officially support: **recent Chrome, Firefox and Safari with no extensions**

PR #7431, opened 2024-04-08
Supporting Rstudio Server

haraldschilly:
# Description

- [x] one-click rstudio server

~~Unfortunately I'm stuck. For whatever reason, such a named port does not work.~~

Interestingly, it is impossible to run this in a dev environment.

Deployment: hub first (for the proxy magic), then static, and then project (only when the configuration detects rserver, it will show it in the frontend)

### DIFF
--- a/src/packages/comm/project-configuration.ts
+++ b/src/packages/comm/project-configuration.ts
@@ -28,6 +28,7 @@ export interface MainCapabilities {
   jupyter: boolean | Capabilities;
   formatting: Capabilities; // yapf & co.
   hashsums: Capabilities;
+  rserver: boolean;
   latex: boolean;
   sage: boolean;
   sage_version?: number[];
@@ -49,6 +50,7 @@ export interface Available {
   jupyter_lab: boolean;
   jupyter_notebook: boolean;
   jupyter: boolean;
+  rserver: boolean; // RStudio Server
   x11: boolean;
   latex: boolean;
   sage: boolean;
@@ -69,6 +71,7 @@ export const NO_AVAIL: Readonly<Available> = {
   jupyter_lab: false,
   jupyter_notebook: false,
   jupyter: false,
+  rserver: false,
   sage: false,
   latex: false,
   rmd: false,
@@ -89,6 +92,7 @@ export const ALL_AVAIL: Readonly<Available> = {
   jupyter_lab: true,
   jupyter_notebook: true,
   jupyter: true,
+  rserver: true,
   sage: true,
   latex: true,
   rmd: true,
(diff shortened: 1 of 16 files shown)


williamstein:
I also spent a lot of time over on compute-server (e.g., https://github.com/sagemathinc/cocalc-compute-docker/issues/9) on an issue with basically this same title.  A big surprise for me is that the version of R that got run by rstudio was some ancient version that maybe was included in the package, rather than the one I expected.    

In case you hit this, you can give the path to the version of R that you actually want to run when starting R studio as follows:

```sh
# RSTUDIO_WHICH_R
#  - https://github.com/sagemathinc/cocalc-compute-docker/issues/9
#  - https://forum.posit.co/t/use-a-different-r-version-temporarily-in-rstudio/20848/7
export RSTUDIO_WHICH_R=/usr/local/bin/R
```

williamstein:
I also hit issues with graphics support, where the solution was creating this file Rprofile.site, whose location depends on the version of R:
```sh
user@prod-2709:~$ more  /usr/local/lib/R/etc/Rprofile.site
options(bitmapType='cairo')
```

and I also had to compile R from source with 
```
./configure \
    --prefix=/usr/local \
    --enable-R-shlib \
    --enable-R-profiling \
    --enable-memory-profiling \
    --with-blas \
    --with-lapack \
    --with-x \
    --with-cairo \
    --with-jpeglib \
    --with-readline \
    --with-tcltk

```

Anyway, just FYI...

In all cases I could never get websocket terminals to work, which are I think just slightly faster. This is unrelated to my proxy server, since the problem also happens with ssh port forwarding.

williamstein:
Regarding this:
```
user@prod-2709:~$ more  /usr/local/lib/R/etc/Rprofile.site
options(bitmapType='cairo')
```
I just wasted an hour due to not having a frickin' newline at the end of the file, which causes R to just ignore it!!  https://github.com/easybuilders/easybuild-easyconfigs/issues/4918#issuecomment-417486880

haraldschilly:
Well, in any case, this is now deployed in our "test" and I can do various things. e.g. plotting in Rmd, running R files, and even "knit" an Rmd file to HTML and then show it in a popup. That's below (which means RServer → generate file → properly point to file and open it does work as well)

Screenshot from 2024-04-15 16-53-47

haraldschilly:
An interesting setting is how the RSession is constructed. The first setting here lets someone point to the "R" instance to use – if it isn't in the PATH

https://docs.posit.co/ide/server-pro/reference/rserver_conf.html#rsession-which-r

PR #7439, opened 2024-04-12
support querying elements in an array

haraldschilly:
# Description

this is necessary for the CRM data, to filter the accounts/tags field

my main worry is, I hope this does not break anything else.

Screenshot from 2024-04-12 18-27-40

### DIFF
--- a/src/packages/frontend/frame-editors/crm-editor/views/view-menu/search.tsx
+++ b/src/packages/frontend/frame-editors/crm-editor/views/view-menu/search.tsx
@@ -1,12 +1,12 @@
-import { useMemo } from "react";
 import { FilterOutlined } from "@ant-design/icons";
-import { Button, Input, InputNumber, Popover, Select, Space } from "antd";
-import type { ColumnsType } from "../../fields";
-import { getFieldSpec } from "../../fields";
 import { Icon } from "@cocalc/frontend/components";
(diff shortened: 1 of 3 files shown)


williamstein:
Yes, this will be nice!

williamstein:
(The other nearby missing big feature is AND queries involving the same field twice.  I never got to that...)

haraldschilly:
Yes, I see, but that's for another day. In any case, FWIW all tests pass. You can check for yourself, I tried to keep this patch minimal.

src/packages/frontend/frame-editors/crm-editor/views/view-menu/search.tsx
```diff
@@ -151,12 +151,15 @@ function SearchBy({
...
+    if (fieldSpec.type === "boolean") {
       return [
         { value: "IS" as Operator, label: "IS" },
         { value: "IS NOT" as Operator, label: "IS NOT" },
       ];
     }
+    if (fieldSpec.type === "array") {
+      return [{ value: "ANY" as Operator, label: "ANY" }];
+    }
     return OPERATORS.filter((op) => op != "==").map((op: Operator) => {
```
haraldschilly, line 163: oh I see, here it appears again. So, maybe what's left is to filter "ANY" like "==". … but that shouldn't be a real problem...

PR #7446, opened 2024-04-15
Added Events News Type

schrodingersket:
# Description

This PR adds the "Events" functionality to CoCalc as a subset of the full "News" feature. Some important things to note:

News items with the `event` channel are rendered and managed at `/about/events` instead of the typical `/news` page. This means that those events will _not_ show up in the typical News view, nor are do they show up in the news banner or user notification view.

The "Save" button has also been updated to redirect the user to the created and/or edited news item, and this holds for both canonical news items and the new event items; in the former case, the admin is simply re-routed to the `/about/events` page.

Here's what the new view looks like for admins:

image


haraldschilly:
I made a couple of small tweaks. I'll test this once more when I have more time – but apart from that it should be good to go.

williamstein:
This looks pretty safe.  @haraldschilly if you see problems, they can be follow up issues.

williamstein:
And thanks!!!

haraldschilly:
Oh no! I'm currently working on it.

williamstein:
>  I'm currently working on it.

I got tricked by "it should be good to go."   In any case, I didn't make anything live in production, so no worries.

PR #7468, opened 2024-04-22
Feature/7411

schrodingersket:
# Description

This completes the initial port of team biographies from https://about.cocalc.com/the-people-behind-sagemath-inc-and-cocalc/

Here's an example bio:

image


haraldschilly:
I think it is about time to change the "About" link the footer to point to this page.

In any case, the only detail I stumble over is that it does not show any items for me locally. I bet this is due to me having set some site settings differently – I saw there is a `filteredDataSource = useMemo(...)` in the `index-list.tsx`.

haraldschilly:
Ok, got it, there are `landingPages: true,` in the items and I happen to have them disabled. These fields shouldn't be set, because those are no landing pages – either we show all the about/* pages or none of them. 

So, in any case, this is out of scope, clearly, because I don't know why these switches are there. There is a `cocalc.com` mode and `on-prem` modes. IMHO they should decide if these pages show up or not – I don't see why we have to fine tune this and then you wouldn't run into such intricate traps.

PR #7486, opened 2024-04-25
npm: update langchain google for gemini pro 1.5

haraldschilly:
# Description

Trivial package update, should make it work.

### DIFF
--- a/src/packages/pnpm-lock.yaml
+++ b/src/packages/pnpm-lock.yaml
@@ -33,10 +33,10 @@ importers:
         version: 29.5.12
       jest:
         specifier: ^29.7.0
-        version: 29.7.0(@types/node@18.16.14)
+        version: 29.7.0(@types/node@18.19.31)
       ts-jest:
         specifier: ^29.1.2
-        version: 29.1.2(@babel/core@7.24.4)(jest@29.7.0)(typescript@5.2.2)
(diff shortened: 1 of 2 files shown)


williamstein:
This obviously doesn't change the list of available models:

image

and I don't see anything in admin settings related to Gemini versions specifically.  How does version 1.5 get enabled?

haraldschilly:
Hmm, not sure what's going on. Is this in your dev project? 1.0 and 1.5 show up by default, unless models to be shown are selected explicitly. Here is a screenshot of chat in my dev project:

Screenshot from 2024-04-25 17-49-45

williamstein:
On https://cocalc.com Gemini 1.5 is not user selectable or visible.  Why?

haraldschilly:
OK. I figured it out. The websocket service sends the `/customize` endpoint and right now runs on f50a7f9d3a794124e0 which only marks gemini pro as a working option: https://github.com/sagemathinc/cocalc/blob/f50a7f9d3a794124e0b57afcc7bd9162ff64dd07/src/packages/util/db-schema/llm-utils.ts#L134 … therefore, the web client only gets a list of LLMs without it.

williamstein:
OK, that clears everything up.  I'm glad I didn't make the latest master code live before this PR for hub-websocket :-)
l will later today though. 

Thanks!

Issue #7513, opened 2024-05-02
project crash upon upload in snapshot dir


haraldschilly:
I'm not sure if the upload POST and the crash are actually related, but they're at least right next to each other.

```
2024-05-02T13:17:20.900Z: cocalc:debug:project:handle-message-from-hub received a message {"event":"query","id":"90721163-fc14-499e-8786-1647f9854ca0","...":"..."}

2024-05-02T13:17:27.693Z: cocalc:http:project:upload upload POST 

2024-05-02T13:17:27.693Z: cocalc:http:project:upload upload POST  ensure target path exists...  /home/user/.snapshots/2024-05-02-122434

2024-05-02T13:17:27.694Z: cocalc:http:project:upload upload POST  parsing form data...

2024-05-02T13:17:27.697Z: cocalc:error:project:BUG (uncaughtException) 
BUG ****************************************************************************
 {"errno":-13,"code":"EACCES","syscall":"open","path":"/home/user/.snapshots/...ipynb"} uncaught
Exception Error: EACCES: permission denied, open '/home/user/.snapshots/...ipynb'

Error: EACCES: permission denied, open '/home/user/.snapshots/...ipynb'
2024-05-02T13:17:27.698Z: cocalc:debug:jupyter:pool killKernel pid= 227

```

PR #7515, opened 2024-05-02
project/upload: check explicitly for write perms to avoid #7513

haraldschilly:
# Description

Well, #7513 is a bit mysterious, since there is a try/catch block. However, this patch adds an extra check and throws a more human readable error message.


williamstein:
This definitely seems like a good idea!

PR #7531, opened 2024-05-08
next/store: warnings for low/high cpu/mem and tweaking profiles

haraldschilly:
# Description

This is a quick idea I had. The license configurator will explicitly warn about maybe a bad experience with a low memory choice, and also warn about picking a compute server or VM instead of a high memory or cpu cores choice. This gives immediate feedback very early, right where the choice happens. It's also conditional on the action of the user, hence due to being dynamic it is more visible.

… and of course, params and the text in detail could be adjusted.

## high cpu/ram

Screenshot from 2024-05-08 18-18-38

## low mem

Screenshot from 2024-05-08 18-18-01

## otherwise, no warnings, this is the default config (3gb ram)

Screenshot from 2024-05-08 18-27-48

### DIFF
--- a/src/packages/next/components/store/quota-config-presets.tsx
+++ b/src/packages/next/components/store/quota-config-presets.tsx
@@ -47,7 +47,8 @@ type PresetEntries = {
 
 // some constants to keep text and preset in sync
 const STANDARD_CPU = 1;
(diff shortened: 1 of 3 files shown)


williamstein:
Definitely an improvement!

PR #7547, opened 2024-05-13
7474: Migrated `src/packages/{database,jupyter,project,static,sync,terminal,util}` test scripts to use TypeScript.

schrodingersket:
# Description

These tests all pass on my machine (from the `src` directory, I ran `pnpm clean && pnpm make-dev && pnpm test`), but I'd recommend waiting for a successful GitHub build before merging just in case there are breaking differences between dev/build servers.

Note that `src/packages/hub` cannot be migrated until that module no longer mixes CoffeeScript and TypeScript.


williamstein:
Thanks!!

PR #7548, opened 2024-05-13
ux/7532: Extracted landing page link rendering logic to a single source of truth.

### LINKED ISSUE #7532: logic for what goes in the footer is different than the header
Sorry for this being vague, but here's the header and footer on my dev server:

Header:
image

Footer:
image

Notice that "Store" and "Features" are NOT in the header, but are in the footer.  

It's actually really, really bad having links to features and store on random installs of cocalc, since that confuses users, messes up our seo, etc.

schrodingersket:
# Description

I'm opening this as a draft for now so that we can go over the business logic that dictates which page links should appear when. Once the business logic is satisfactory, this PR is ready to merge.

The link rendering logic has been consolidated to a single source of truth so that business logic is all in one place; see https://github.com/sagemathinc/cocalc/compare/master...schrodingersket:ux/7429?expand=1#diff-53c6b656252dc3b6591f8121477a8c83492b50363c59fc139ed2619d04c0d135R78 for the current configuration, which mirrors the existing logic as much as possible.

Fixes #7532.

### DIFF
--- a/src/packages/next/lib/with-customize.ts
+++ b/src/packages/next/lib/with-customize.ts
@@ -11,6 +11,8 @@ import { KUCALC_COCALC_COM } from "@cocalc/util/db-schema/site-defaults";
 import { getSoftwareEnvironments } from "@cocalc/server/software-envs";
 import { DEFAULT_COMPUTE_IMAGE } from "@cocalc/util/db-schema";
 
+import { CustomizeType } from "./customize";
+
 const revalidate = 30;
 
 interface Options {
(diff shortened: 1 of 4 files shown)


src/packages/next/lib/with-customize.ts
```diff
@@ -73,6 +75,38 @@ export default async function withCustomize(
...
+  customize.enabledPages = {
+    auth: {
+      try: !customize.account && customize.anonymousSignup,
+    },
+    about: {
+      index: customize.landingPages,
+      events: customize.isCommercial,
+      team: customize.landingPages,
+    },
+    compute: customize.isCommercial,
```
williamstein, line 87: hsy: maybe use the explicit enable or disable of compute servers for the site...

schrodingersket:
@williamstein This should be good to merge whenever you're ready for it.

PR #7552, opened 2024-05-15
frontend/flyout/files: fix #7550

### LINKED ISSUE #7550: flyouts/files: crash in filesStore.toJS
I saw this error

```
TypeError: e.toJS is not a function
    at https://cocalc.com/static/app-a8f436d26aa45c7b90d2.js?f72bb682e99e8565f71e:2:5750944
    at Object.Xo [as useMemo] (https://cocalc.com/static/load-f686fcd7daedf046f9eb.js?f72bb682e99e8565f71e:2:74239)
    at t.useMemo (https://cocalc.com/static/load-f686fcd7daedf046f9eb.js?f72bb682e99e8565f71e:2:157829)
    at flyout (https://cocalc.com/static/app-a8f436d26aa45c7b90d2.js?f72bb682e99e8565f71e:2:5750834)
[...]
```

which must be happen in `flyouts/files.tsx`, in the useMemo, in that line:

```
const files: DirectoryListing = filesStore.toJS();
```

apparently, `fileStore` can be null. … and `directoryListings` has type `any`. There should be a type info for it, signalling it can be null.

ok, on a closer look, `filesStore` must be something else, i.e. there needs to be a check for `toJS`.

haraldschilly:
# Description

I tried to come up with a type in the store, but no luck. So, just something local here, and fixing #7550.


williamstein:
It says work in progress, but this seems like an obvious and safe improvement over master, right?

PR #7599, opened 2024-05-31
frontend/jupyter: run all cells above/below dropdown button in cell toolbar

haraldschilly:
# Description

- this adds a small dropdown to get quick access for running all cells above or below
- I'm also changing the icons, because they are not telling much, or even point in the wrong direction
- this also adds support for rotated menu icons

Screenshot from 2024-05-31 17-06-29


williamstein:
When does it include the current cell?  The naming strongly suggests it never includes the current cell.

williamstein:
image

williamstein:
image

PR #7618, opened 2024-06-08
OpenAPI doc: Completed first draft of Compute Server docs.

schrodingersket:
# Description

Added first draft of OpenAPI docs for the Compute Server API.

This should be considered an incremental update, as some endpoints have not been fully E2E tested in accordance with the API docs. In addition, the following endpoints are still in progress (but are minimally documented now) due to their deeper object structure:

- `/api/v2/compute/get-images`
- `/api/v2/compute/get-images-google`
- `/api/v2/compute/check-in` (pending the current shared storage work for compute servers)
- `/api/v2/compute/get-template[s?]`
- `/api/v2/compute/set-server-configuration`

### DIFF
--- a/src/packages/next/lib/api/schema/compute/create-server.ts
+++ b/src/packages/next/lib/api/schema/compute/create-server.ts
@@ -1,0 +1,56 @@
+import { z } from "../../framework";
+
+import { FailedAPIOperationSchema } from "../common";
+
+import { ProjectIdSchema } from "../projects/common";
+
+import {
+  ComputeServerCloudSchema,
+  ComputeServerColorSchema,
+  ComputeServerIdBodySchema,
+  ComputeServerTitleSchema,
+  ServerConfigurationSchema
+} from "./common";
+
+// OpenAPI spec
+//
+export const CreateServerInputSchema = z
+  .object({
+    configuration: ServerConfigurationSchema,
+    title: ComputeServerTitleSchema,
+    color: ComputeServerColorSchema,
+    cloud: ComputeServerCloudSchema,
+    project_id: ProjectIdSchema
+      .describe(
+        "The project id that this compute server provides compute for."
+      ),
+    idle_timeout: z
+      .number()
+      .describe(
+        `The idle timeout in seconds of this compute server. If set to 0, the server will
+        never turn off automatically. The compute server idle timeouts if none of the tabs 
+        it is providing are actively touched through the web UI. _Not yet implemented._`
+      )
+      .optional(),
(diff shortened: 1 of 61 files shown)


src/packages/next/lib/api/schema/compute/create-server.ts
```diff
@@ -0,0 +1,56 @@
...
+        `The idle timeout in seconds of this compute server. If set to 0, the server will
+        never turn off automatically. The compute server idle timeouts if none of the tabs 
+        it is providing are actively touched through the web UI. _Not yet implemented._`
+      )
+      .optional(),
+    autorestart: z.boolean()
+      .describe(
+        `If true and the compute server stops for any reason, then it 
+        will be automatically started again. This is primarily useful for 
+        stopping instances.`
```
williamstein, line 39: This should say "This is primarily useful for spot instances.`"  You were copying from packages/util/db-schema/compute-servers.ts which had a different typo "This is primarily useful for stop instances." but you fixed it in the wrong way.

williamstein: Incidentally, you could through Javascript directly import and use the text from the files in util/db-schema.  It's all in one big javascript object "SCHEMA".

williamstein:
I made two inline comments, but nothing is blocking releasing this, so merged as is.

src/packages/next/lib/api/schema/compute/create-server.ts
schrodingersket: Whoops :sweat_smile: Thanks for the feedback!

> Incidentally, you could through Javascript directly import and use the text from the files in util/db-schema. It's all in one big javascript object "SCHEMA".

At one point, I'd thought that it might be better if the schemas were more or less isolated from the rest of the codebase (since they effectively define an interface) to keep concerns separate, but this could be a pretty good way to maintain the whole "single source of truth" paradigm. I'm honestly not sure which is the better way to go here, so if you have a preference I'd be happy to run with that.

williamstein: Since in both cases we're 100% documenting a field (with no impact except documentation), and this very example illustrates how such documentation of the exact same thing can diverge and become inconsistent, I would say the code should clearly be the same.    I do this sort of refactoring constantly - that's what a lot of the current util package contains, is code  or data that should be used in multiple places.

I don't mean to replicate too much.  But documentation of a field is certainly fair game.  The typing schema is different/dangerous/tricky though (e.g., one is oriented toward postgres) so I wouldn't touch that.

PR #7641, opened 2024-06-27
update antd

haraldschilly:
# Description

Antd 5.18.3

* I thought this is a smooth update, but it's not. I'll fix the fallout…
   * mostly it's around the children of a button. effectively, many little "top: -Npx" corrections can be removed now. Also, if there is more than just text on a button, wrapping this in `<Space>...</Space>` is a good idea – or at least a plain `<div>`.
* unrelated: I saw a set of LLMs in in chat/llm/regenerate was missing
* my https://github.com/sagemathinc/cocalc/pull/7640 had a mistake, dayjs wasn't updated in the frontend. Why didn't this the package version checker see? In any case, back then antd didn't care about types for the `TimePicker`, but now it does. Hence I noticed this problem.
* another unrelated little fix: when filtering by #tag in the projects list, it says `'0'` instead of `#[tag]`: 1bc317f4702
* regarding testing, it's kind of impossible to go through every dialog, but I checked a lot of them. There was never a functional issue, only spacing and alignments, with things on a button.

### DIFF
--- a/src/packages/frontend/account/other-settings.tsx
+++ b/src/packages/frontend/account/other-settings.tsx
@@ -409,7 +409,7 @@ export function OtherSettings(props: Readonly<Props>): JSX.Element {
       <Panel
         header={
           <>
-            <AIAvatar size={22} /> AI Settings
+            <AIAvatar size={18} /> AI Settings
           </>
         }
       >
--- a/src/packages/frontend/admin/page.tsx
+++ b/src/packages/frontend/admin/page.tsx
@@ -89,8 +89,8 @@ export function AdminPage() {
       key: "llm-testing",
       label: (
         <div style={headerStyle}>
-          <AIAvatar size={16} style={{ marginRight: "8px", top: "-5px" }} />{" "}
-          Test LLM Integration
+          <AIAvatar size={16} style={{ marginRight: "8px" }} /> Test LLM
+          Integration
         </div>
       ),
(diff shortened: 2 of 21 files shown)


haraldschilly:
The only thing I can't test is the next/store. It fails grabbing the "profile". I.e. this is the error I get in the network info. I tried everything I could think of to build/rebuild it. It works in prod, though. … if this is related to what we already discussed, is this another case where there is a problem? /cc @schrodingersket 

Screenshot from 2024-06-28 15-09-41

schrodingersket:
@haraldschilly thanks for pinging me on this - I'll check it out here in just a little bit. Incidentally, there's an environment variable that can disable JSON schema validation in dev (I don't have it off the top of my head, but will track that down for you once I'm out of the meeting I'm about to jump into) if you're currently blocked.

haraldschilly:
@schrodingersket well, don't worry too much about me – it's just one detail to double-check after building this in prod, that's all. most likely, this has no impact, since everything else did work fine.

williamstein:
>  it's just one detail to double-check after building this in prod, that's all. most likely, this has no impact, since everything else did work fine.

The store is completely broken in production mode.   I'll look into fixing it right now.

williamstein:
Never mind -- the store just didn't come up initially, probably because of the hubs restarting (on test.cocalc.com).

PR #7667, opened 2024-07-05
hub: cleanup data of deleted projects and accounts

haraldschilly:
# Description

Periodic hub job that cleans up data of projects and accounts marked deleted,
designed to run safely against large backlogs (hundreds of thousands of
deleted projects, tens of thousands of deleted accounts).

### Projects (`cleanup_old_projects_data`)
Runs for projects with `deleted = true` and `users IS NULL`:

1. Delete all `patches` for each syncstring of the project (chunked by ctid),
   then drop the syncstring rows themselves.
2. Clear `site_license`, `status`, `last_active`, `run_quota` on the project row.
3. On-prem only: delete the project's files on disk and any shared files,
   with a safety assertion that the resolved path contains the project_id UUID.
4. Chunked-delete rows referencing the project_id in ~10 associated tables
   (`file_access_log`, `file_use`, `jupyter_api_log`, `mentions`,
   `openai_chatgpt_log`, `project_log`, `public_paths`, `shopping_cart_items`,
   `copy_paths`, `listings`, `project_invite_tokens`).
5. Set `state.state = 'deleted'` so the project is skipped on subsequent runs.

### Unlink (`unlink_old_deleted_projects`)
Sets `users = NULL` on projects flagged deleted for more than 30 days.
Chunked so first runs on large backlogs are many small UPDATEs rather than
one huge transaction.

### Accounts (`cleanup_deleted_account_pii`)
Scrubs PII from accounts flagged `deleted = true` once the site's
`pii_retention` grace period has elapsed. At deletion time
`server/accounts/delete.ts` already clears `email_address` and `passports`;
this job additionally clears `first_name`, `last_name`,
`email_address_before_delete`, `email_address_verified`,
`email_address_challenge`, `email_address_problem`. The `accounts` row is
kept so references to `account_id` elsewhere (projects, central_log, etc.)
still resolve.

Retention is driven by the existing **`pii_retention` site setting**
(accepts `"never"`, `"30 days"`, `"3 month"`, `"1 year"`, ...):
- When `"never"` (the default) the scrub is a no-op.
- Otherwise gating uses a new `accounts.deleted_at` timestamp column set in
  `markAccountDeleted`; accounts are scrubbed once
  `NOW() - deleted_at >= pii_retention`.
- On first run, legacy deleted accounts with `deleted_at IS NULL` are
  backfilled with `NOW()`, so the existing backlog gets a full retention
  window from deploy time rather than being scrubbed immediately.

## Scale-safety primitives (`bulk-delete.ts`)

- Shared `throttledRunner` loop: adaptive chunk size targeting a ~100ms
  round-trip, backing off via `MAX_WAIT_S` (30s cap, configurable) so total
  DB utilization stays around `COCALC_DB_BULK_DELETE_MAX_UTIL_PCT` (10% by
  default).
- `bulkDelete` uses this runner for DELETEs; `unlink_old_deleted_projects`
  and `cleanup_deleted_account_pii` use it for chunked UPDATEs.

## Schema changes

- `accounts.deleted_at TIMESTAMP` — set to `NOW()` whenever `deleted`
  flips to true; used by the PII scrub grace period.
- `accounts` index on `deleted` — for efficient scans of the deleted subset.
- `syncstrings` index on `project_id` — so the cleanup can find a deleted
  project's syncstrings without a seq scan.

All three are auto-created by `syncSchema` at hub startup.

## Gating

A new site setting `delete_project_data` (default `no`) gates the actual
destructive operations on projects. On-prem / single-user dev mode
additionally deletes files from disk.

Account PII scrub is gated by `pii_retention` (default `"never"`).

## UI

\"Modernize\" visibility and delete controls (popconfirm, switch). Text is
unchanged.

Screenshot from 2024-07-17 16-38-45

## Known follow-ups (not in this PR)

- `blobs` table referencing deleted projects is still intentionally skipped —
  data could be shared across projects (no refcount), and blobs are only
  accessible if you know the content hash, so leaving them is acceptable.

## Deploy notes

- **On a large database, pre-create the two new indexes manually before
  deploying** — `syncSchema` uses plain `CREATE INDEX` (not `CONCURRENTLY`,
  see the comment in `postgres/schema/indexes.ts`), which would block at
  hub startup on a huge table. Do this first, then deploy:

  ```sql
  CREATE INDEX CONCURRENTLY IF NOT EXISTS accounts_deleted_idx
    ON accounts (deleted);
  CREATE INDEX CONCURRENTLY IF NOT EXISTS syncstrings_project_id_idx
    ON syncstrings (project_id);
  ```

  The cleanup functions refuse to run (loud log, early return) if either
  index is missing, so a forgotten precreate fails safe rather than
  doing sequential scans.
- Keep `delete_project_data = no` until first run has been observed.
- Keep `pii_retention = never` unless/until the account PII scrub is
  desired. When enabling, existing deleted accounts will first be backfilled
  and will then respect the full retention window.
- For the first real run against a large backlog, set `INTERVAL_H` high and
  watch metrics (`cocalc_database_delete_projects_total`, including the new
  `fs_delete_failed` label which indicates on-disk deletes that skipped the
  tombstone and will be retried on the next run).
- Unlink (`unlink_old_deleted_projects`) is intentionally ungated — it runs
  whenever the cleanup hub runs. Once a project's `users` is nulled, the
  restore endpoint refuses (see below), so this phase is effectively
  permanent. Deploy only once the on-prem/site operator is comfortable
  with that.
- After unlink or full purge, the `/api/v2/projects/restore` endpoint
  refuses with "this project has been permanently deleted and cannot be
  restored", rather than silently producing an operationally dead row.


williamstein:
huge +1!   It'll be a good foundation for doing this better in general.

src/packages/database/postgres/delete-projects.ts
```diff
@@ -52,20 +76,244 @@ async function get_account_id(
...
+  // collecting tables, where the primary key is the default (i.e. "id") and
+  // the field to check is always called "project_id"
+  const tables = [
+    "public_paths",
+    "project_log",
+    "file_use",
+    "file_access_log",
+    "jupyter_api_log",
+    "openai_chatgpt_log",
+    "blobs",
```
williamstein, line 250: Think twice about blobs.  I would not delete the blobs.

williamstein:
I'm working on implementing an exam mode for courses.  This must involve creating a new project per student for each exam, because it seems it is VERY important to instructors that students don't have access to the exam when the exam is over (and also temporarily don't have access to their own work).  This means that the student should be removed from the exam project and their work collected when the exam ends.    It would be natural at that point to delete the exam project.
However... it would also be VERY good for the instructor to have access to the TimeTravel data from the exam, which means *not* deleting the exam project... or at least the edit history in the database.

I understand a big point of this PR is to make it so when you delete a project, all TimeTravel edit history also gets permanently purged.  I'm just trying to figure out what to do here...

No comment is needed, probably -- I just wanted to write this down.

src/packages/server/nats/tiered-storage/clean.ts
```diff
@@ -10,7 +10,7 @@ import { archiveProject, archiveAccount } from "./archive";
 
 const logger = getLogger("tiered-storage:clean");
 const log = (...args) => {
-  logger.debug(...args);
+  logger.debug("", ...args);
```
williamstein, line 13: strange -- I didn't need to do that with my typescript setup...

PR #7679, opened 2024-07-10
implement new usage profiles for users

williamstein:
- this works as is
- not in master yet, since I want to rethink it, e.g., maybe change the time to be a date, etc.


williamstein:
I'm going to pause this pending having a discussion with the company, since it's so unclear what to include...

williamstein:
I'm closing this for now.  I'm just not sure this is the right way to go...

Issue #7666, opened 2024-07-05
API for async project exec


haraldschilly:
There is an `project_exec` in the v1 api … but, it became apparent there is a need to make it possible to run asynchronously. I.e. a job (command + args) is launched, a unique ID is returned, and later the callee can query by job ID to check the status and get the returned stdout+stderr. (status codes I can think of: "running" | "error" | "missing" | "success").

My idea how to implement this is by letting the project spawn a sub-process (child_process), but to not block on it. Instead, it stores the spawned process internally in a map, where the key is just an UUID or a simple number. In any case, with that, it should be possible to check up on a spawned sub-process by ID. Results & status are stored in memory.

The more complex detail is how to properly route the request from the hub to the project.
My feeling is, it isn't too hard to extend whatever `project_exec` does right now, but I haven't looked at the code. So, maybe the hard part is already in place? Just needs a bit of "high level" treatment to fit into the v2 framework?

Concerns about memory usage:
* Another detail is how to deal with too much output. Probably the best idea is to cap stdout and stderr, e.g. at 1 MB, and only keep the remainder of the output (aka "tail").
* There should also be an overall limit to how many such outputs are stored, and the stored output should be deleted from memory, once it has been returned by an API call.

As an extra, I think it would also be cool to add some information about the process, similar to `/usr/bin/time` in linux. I'm thinking of e.g. the output of [process.resourceUsage](https://nodejs.org/docs/latest-v18.x/api/process.html#process_process_resourceusage) … such that callers know a bit more about what was going on inside the project.

williamstein:
FYI, I think this will be relatively easy to implement, and the hard part is fortunately mostly already done. Deciding on all those parameters is very, very helpful.   

What's your motivation for this?

PR #7686, opened 2024-07-12
async project exec

haraldschilly:
# Description

* implementing #7666 for the project itself
* adding more details to the v2 API (some were missing)
* more tests
* typescript aggregate.js
* my main confusion was this new openapi validation, which is not working as I assumed, when there are more options for the input object. In any case, I think the code I wrote goes into the correct direction.
* the only real change for existing calls is adding a new `{type: "blocking"}` indicator to the response. I can't think of any reason why this is could become a problem.

### DIFF
--- a/src/packages/backend/execute-code.ts
+++ b/src/packages/backend/execute-code.ts
@@ -17,6 +17,7 @@ import { tmpdir } from "node:os";
 import { join } from "node:path";
 import shellEscape from "shell-escape";
 
+import { envToInt } from "@cocalc/backend/misc/env-to-number";
 import getLogger from "@cocalc/backend/logger";
 import { aggregate } from "@cocalc/util/aggregate";
 import { callback_opts } from "@cocalc/util/async-utils";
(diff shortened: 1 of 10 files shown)


src/packages/util/types/execute-code.ts
```diff
@@ -20,12 +40,24 @@ export interface ExecuteCodeOptions {
   env?: object; // if given, added to exec environment
   aggregate?: string | number; // if given, aggregates multiple calls with same sequence number into one -- see @cocalc/util/aggregate; typically make this a timestamp for compiling code (e.g., latex).
   verbose?: boolean; // default true -- impacts amount of logging
+  async_call?: boolean; // default false -- if true, return an ID and execute it asynchroneously
```
williamstein, line 43: " asynchroneously" is not spelled right.

src/packages/backend/execute-code.ts
```diff
@@ -115,16 +151,90 @@ async function executeCodeNoAggregate(
...
+      const started: ExecuteCodeOutputAsync = {
+        type: "async",
+        stdout: `Process started running at ${start.toISOString()}`,
+        stderr: "",
+        exit_code: 0,
+        start: start.getTime(),
+        job_id,
+        status: "running",
+      };
+      asyncCache.set(job_id, started);
```
williamstein, line 170: jonathan: you could use a third arg to set the timeout here to be bigger than opts.timeout

PR #7711, opened 2024-07-27
Update Refunds section in ToS

novoselt:
- drop some details about upgrading/downgrading
- explicitly state that CoCalc Credit is nonrefundable
- PAYG services may be terminated as soon as zero balance is hit

# Description


williamstein:
You should update the " <h2>Last Updated: May 5, 2023</h2>" section at the top of the file you're editing.

PR #7713, opened 2024-07-28
Account Management API: Added new `/api/v2/accounts/set-name` API route…

schrodingersket:
…to edit the first, last, and username for an account.

### DIFF
--- a/src/packages/next/pages/api/v2/accounts/set-name.ts
+++ b/src/packages/next/pages/api/v2/accounts/set-name.ts
@@ -1,0 +1,74 @@
+/*
+Set account {user/first/last} name.
+*/
+
+import userQuery from "@cocalc/database/user-query";
+import userIsInGroup from "@cocalc/server/accounts/is-in-group";
+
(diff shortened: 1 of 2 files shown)


src/packages/next/pages/api/v2/accounts/set-name.ts
```diff
@@ -0,0 +1,74 @@
...
+  // This user MUST be an admin:
+  if (account_id && !(await userIsInGroup(client_account_id, "admin"))) {
+    throw Error("Only admins are authorized to specify an account id.");
+  }
+
+  return userQuery({
+    account_id: account_id || client_account_id,
+    query: {
+      accounts: {
+        // Any provided values must be non-empty in order for userQuery to SET values
```
williamstein, line 44: That's true, and a good reason why we should also support an explicit set/get option instead of this implicity "non-empty" userQuery thing that determines that it is set instead of get.  That was a mistake on my part in designing user_query long ago.   I'm glad you figured that out though!

PR #7722, opened 2024-07-30
api/exec await + latex

haraldschilly:
# Description

- coming up with the right semantics to start, get and await an async job
- using this to enhance latex document compilation. this is way harder than I hoped for :-( ... the more I debug, the more problems come to the surface.
   - the hardest bug was a case, where a process that exits in a forced way (e.g. kill) has no exit code. Then, the backend never returns, leaving the callback chain hanging. So, this is now fixed, because the exit code falls back to 0 in that case.

- this fixes https://github.com/sagemathinc/cocalc/issues/3783 by letting users click a stop button if there is no progress
   
   Here is how this looks after >15 seconds. Of course, there is no indication why. The log stops and cpu is high. More informative than before.

   Screenshot from 2024-08-09 18-20-27

- after compilation is done, this adds peak known memory usage and duration at the top of the logs.
   
   Screenshot from 2024-08-09 18-24-16


haraldschilly:
This is by far not perfect, but it is definitely a step forward.

PR #7761, opened 2024-08-16
I18N

haraldschilly:
# Description

preliminary support and tooling. language is in the account settings, can be changed via a dropdown – if enabled – or query parameter `?lang=[locale]`. A bullet proof way to reset the language to english is https://cocalc.com/app?lang=en 

## ops

For admins, there is a setting to only enabled selected languages. If none are selected, all are available. Everything can be disabled  by only setting english as an available language – language could still be changed via the query param.

Screenshot from 2024-08-26 14-26-20

## deployment

due to that new `i18n` admin setting, hub must come first – then frontend. (it won't break otherwise, just won't show any translations)

## misc fixes

* the menu tooltip title for closing a frame claims: "Close all frames to restore the default layout.", which is not true. I'm changing this to explicitly select restoring the default layout.
* meaning of "delete variables" in the context of restarting a jupyter kernel → "deleting variable values", because the variables are still there, it's their state or value that's gone.

## frontend testing

this also changes the test task, and some aspects of testing, in the frontend. I think it isn't setup to run tests properly at all, right now. So, further down the road, "enzyme" and similar packages have to go.


williamstein:
Setting the language was just discussed in the jupyterlab meeting: https://github.com/jupyterlab/jupyterlab/issues/13990

PR #7765, opened 2024-08-19
Fix Project Collaborator Permission Check

schrodingersket:
# Description

Added `await` call to enforce collaborator permissions check for {copy-path,copy-url,start,stop,touch} project operations.


haraldschilly:
This reminds me of https://typescript-eslint.io/rules/no-misused-promises/ ... but does this linting rule work with the `!` negation?

haraldschilly:
> This reminds me of https://typescript-eslint.io/rules/no-misused-promises/ ... but does this linting rule work with the `!` negation?

or, to frame this more general, I'm sure it makes sense to have a `pnpm lint` target, for each package, which does some of these typescript checks for us.

PR #7788, opened 2024-08-27
frontend/frame-editors: nail down type of each frame editor spec + translation fix

### LINKED ISSUE #7789: Repeated explain in AI tools
**Describe the bug**
image

**Expected behavior**
The lower item is actually for fixing code, probably a result of i18n

haraldschilly:
# Description

So, basically, instead of `{object} as EditorDescription`, this types each variable and adds a `type` to each such spec object. Otherwise, type errors might not show up. Also, there is an attribute used, which isn't defined in `EditorDescription`: `renderer?: "canvas"`. I've added it, with a note to check up on this.

The core detail is that this is-a-TimeTravel check now looks like this:

```typescript
      if (
        spec.type === "timetravel" &&
        !(actions instanceof TimeTravelActions)
      ) { ... }
```

---

This also fixes #7789 … since there was a test failure.


williamstein:
Very nice!

PR #6804, opened 2023-06-29
frontend/project: start refactoring top right tab area

haraldschilly:
# Description

wip

## [Checklist](https://github.com/sagemathinc/cocalc/wiki/PR-Checklist):
- [ ] Testing instructions are provided, if not obvious
- [ ] Release instructions are provided, if not obvious

PR #7793, opened 2024-08-28
frontend/flyout: porting over the project homepage from #6804 

haraldschilly:
# Description

This ports over the project homepage from #6804. There are no other changes, hence this is only relevant for the `vertical_fixed_bar==flyout` mode.


williamstein:
What does this do?  Can you include a screenshot?  Is it bringing back the original homepage, which I think nobody uses or wants (?) or something else?

haraldschilly:
In that vertical-tab mode with flyout only, the homepage looks like that right now. I don't like this, because it is quite chaotic.

2024 08 29-old-homepage

 This PR changes that homepage to look like the following. I like this much more, because it just mentions the project, shortcuts to common tasks like opening files or creating new ones, and the recent files. This was part of that PR #6804 back then. There is absolutely no change with anything else (I would like to, but this is not part of that PR).

2024 08 29-new-homepage

williamstein:
@mforbes (and others) made a strong argument based on user testing that the "Home" button should just show the file explorer in the home directory.  We listened and made that change about 6 months ago and people like it.   I don't understand what this PR does.

williamstein:
I misunderstood this. Now that I see it ONLY impacts the situation when:

image

and nothing else, and there is a good logical reason for this.

PR #7829, opened 2024-09-11
npm: update express, express-sessions, and express-rate-limit

haraldschilly:
# Description

motivated by body-parser, which is a direct dependency and should be updated

test: I can run next, frontend and api pages in my dev setup, also checked if I can run a jupyter notebook, etc. So, this seems to be safe.


haraldschilly:
There was an error about consistency. Maybe all of them should have one?

https://github.com/sagemathinc/cocalc/actions/runs/10812015547/job/29992698919

haraldschilly:
Not so long ago (weeks?) I tried to install everything fresh without a lock file or existing node_modules directories. This was completely broken. Therefore I know the versioning information is all the package files is not sufficient to end up with a working build – i.e. this doesn't work as it is. Effectively, we have to rely on the lockfile and hence all caret-signs are either exactly what we have or the lockfile defines it. But sure, I can change it however you like.

PR #7835, opened 2024-09-12
Restrict student search to use only emails

### LINKED ISSUE #7832: course: make it so student search ONLY works via email address by default (growing pains)
Searching by student name given how many millions of accounts we have is potentially confusing, as many people in the world have the same name, and we don't divulge any info about found people (e.g., their email, institute, etc.), so there is no way to judge if you found the right person.

Of course, exactly this same concern applies to adding collaborators to projects!

I don't think the API has a way to specify that search is only by email address, so this will be a medium-difficulty ticket.

This also impacts on prem that might use single sign on only with no emails, and a small group of people... where searching by name is very reasonable.

enjeck:
# Description
Fixes #7832 
See it working:

https://github.com/user-attachments/assets/b36a8e89-eb42-4d1f-956c-c719ab3ca356

### DIFF
--- a/src/packages/frontend/course/students/students-panel.tsx
+++ b/src/packages/frontend/course/students/students-panel.tsx
@@ -41,6 +41,7 @@ import {
 import * as util from "../util";
 import { Student, StudentNameDescription } from "./students-panel-student";
 import { HelpBox } from "../configuration/help-box";
+import { is_valid_email_address as isValidEmailAddress } from "@cocalc/util/misc";
 
 interface StudentsPanelReactProps {
   frame_id?: string; // used for state caching
@@ -253,6 +254,7 @@ export const StudentsPanel: React.FC<StudentsPanelReactProps> = React.memo(
         select = await webapp_client.users_client.user_search({
           query: add_search,
           limit: 150,
+          only_email: true,
         });
       } catch (err) {
         if (!isMounted) return;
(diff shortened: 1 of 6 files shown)


src/packages/frontend/course/students/students-panel.tsx
```diff
@@ -313,7 +315,7 @@ export const StudentsPanel: React.FC<StudentsPanelReactProps> = React.memo(
     }
 
     function student_add_button() {
-      const disabled = add_search?.trim().length === 0;
+      const disabled = add_search?.trim().length === 0 || !isValidEmailAddress(add_search?.trim() ?? "");
```
williamstein, line 318: I think the input is one or more email address, separated by whitespaces and/or commas, whereas above you are I think searching for only *ONE* email address.    Often an instructor pastes in a bunch of email addresses that they are copying out of a spreadsheet or somewhere else.

williamstein:
> @williamstein I imagine there'll be more work to be done here, but hopefully this is a good start?

I made a line comment about checking whether the input is an email address -- it should really be a test for one or more email addresses.

Also, there needs to be a user-configurable fallback to the current behavior.  That's what I meant by "This also impacts on prem that might use single sign on only with no emails, and a small group of people... where searching by name is very reasonable."   I think it would be sufficient to add a checkbox right below the search button (and have the checkbox only appear when the input is nonempty), which restores the current behavior.

williamstein:
This looks good and I'll merge as is.  A little polish can happen in another PR.

williamstein:
I just changed the UI slightly to hopefully make it clearer (given minimal effort).  I then opened another easy issue that you might want to solve:

https://github.com/sagemathinc/cocalc/issues/7843

PR #7839, opened 2024-09-13
use filename when creating file via extension dropdown

### LINKED ISSUE #7385: Using extension dropdown ignores the entered filename
1. Click on "+New"
2. Enter some file name
3. Use the extension dropdown to pick something, say .course
4. A new file is created with the chosen extension and the file name that was in the text box BEFORE I entered what I wanted.

If I use a big button instead of the dropdown, the entered name is preserved.

enjeck:
# Description
Fixes #7385

### Before

https://github.com/user-attachments/assets/4d3ebaa7-fefa-4f76-8735-473bb69ebbe7

### After

https://github.com/user-attachments/assets/d682b446-1d13-48b4-93e2-1022b8b7c342

### DIFF
--- a/src/packages/frontend/project/new/new-file-page.tsx
+++ b/src/packages/frontend/project/new/new-file-page.tsx
@@ -99,6 +99,7 @@ export default function NewFilePage(props: Props) {
   const [creatingFile, setCreatingFile] = useState<string>("");
 
   async function createFile(ext?: string) {
+    const filename = inputRef.current?.input.value;
     if (!filename) {
       return;
     }
@@ -122,6 +123,7 @@ export default function NewFilePage(props: Props) {
   }
 
   function submit(ext?: string) {
+    const filename = inputRef.current?.input.value;
     if (!filename) {
       // empty filename
       return;


haraldschilly:
hmm, I see, I wonder as well why the state isn't updated properly? My second thought is, is that dropdown-create activity used elsewhere as well, and hence there might be a problem overwriting the filename.

williamstein:
My guess is that this bug is because filename is in a closure (i.e., a function).  filename is updated in the component, but the copy of the function createFile  (etc.) that is actually called on submit is an *older* copy that has an old version of filename.

All this code used to be coffeescript with classes, and we migrated it over the years to typescript with functional components.  This is one of those subtle traps that comes up in such migration.  Often we leave the class methods as nested functions in the functional component (instead of further refactoring the code), which can cause bugs.

A fix is to get the value using some sort of reference, which is basically what you did.

haraldschilly:
@enjeck thank you for the PR, but we need a CLA (contributor license agreement) for all code contributions on file. It's just a standard document to clarify it can be used by CoCalc and the company behind. You do not have an email address on your github profile and I don't know how to contact you. So, please reach out to me ( hsy@cocalc.com ) and I'll send you a form.

PR #7846, opened 2024-09-14
show email address when adding students by email

### LINKED ISSUE #7843: course: when adding students by email address, show the email address next to the result
Consider this:

image

You don't know which user you're selecting, since the name is arbitrary.  If you search for a student by email address, then the email address should be shown in parens after their name.

enjeck:
# Description
Fixes #7843 
Shows email address in parenthesis alongside the name when adding students by email. If there's no name, in which case the email address is already shown, we don't show it. 
Screenshot from 2024-09-14 13-31-04


williamstein:
Thanks.  There's is a bit of tricky logic in doing this, which I think you got right.

PR #7877, opened 2024-09-25
frontend/latex: fix "help me fix" in gutter

haraldschilly:
Down the tree, help me fix contains internationalization, but it is somehow rendered outside the root, where the context provider is defined.

So, this adds another context, just with the english defaults, to avoid that bug. It would be great if this is rendered in the same overall context ...

Error: 

```
Could not find required `intl` object. <IntlProvider> needs to exist in the component ancestry. 
```


williamstein:
> TODO: make this part of the application react root.

I don't know if that is desirable or even possible, since this is a bit of DOM that is managed by the CodeMirror editor.

williamstein:
> I set this to review before waiting the test to finish … but this should be fine.

I manually tested it in my dev project and it didn't crash.

PR #7888, opened 2024-09-28
courses: add exams and group assignments

williamstein:
See

- https://github.com/sagemathinc/cocalc/issues/7886
- https://github.com/sagemathinc/cocalc/issues/7110
- https://github.com/sagemathinc/cocalc/issues/196

This will simply make a way to configure the "location" where student(s) work on an assignment:

image

That will also the workflow, e.g., there's a way to start the exam, which configures the exam projects and adds students to them, end the exam, which removes students and collects the work.  

There will also be something very similar for group assignments.


williamstein:
Also, lockdown browser support could be useful for exam mode.  A free option: https://safeexambrowser.org/about_overview_en.html

PR #7920, opened 2024-10-09
npm: update to typescript 5.6 and update cookie

haraldschilly:
* just a minor update to stay on top of thing ... https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/ ... e.g. this includes checks for always-true if clauses
* update cookie to 1.0.0 + install @types/cookie in sync-client for dev


williamstein:
Sorry it has a massive pnpm lock merge conflict.  Can you redo it (since it's just maybe two pnpm commands)?

williamstein:
I updated typescript -- I didn't touch cookie.

haraldschilly:
ok, let's see if this builds, but I think the conflict is fixed

PR #7946, opened 2024-10-15
npm: express/+session and microlink/react-json-view

haraldschilly:
- this does not attempt to update to [express 5](https://expressjs.com/en/guide/migrating-5.html) … we should probably wait a bit, but this is something to keep an eye on. This release already includes [backports](https://github.com/expressjs/express/releases/tag/4.21.1), though.
- small update to express-session to update its "cookie" dependency


williamstein:
> this does not attempt to update to [express 5](https://expressjs.com/en/guide/migrating-5.html) … we should probably wait a bit,

I ❤️ express :-)

PR #7953, opened 2024-10-16
static/webpack/npm: get rid of html-minify-loader

### LINKED ISSUE #7945: update outdated html-minify-loader@1.4.0
In `static/package.json` there is a `loader: "html-minify-loader"`. It's not maintained for many years. 

When I look into the webpack docs, there are 3 others: https://webpack.js.org/plugins/html-minimizer-webpack-plugin/

So, whatever is really going on, to me it looks like we have switch this html minimize loader.

haraldschilly:
Goal is to get rid of html-minify-loader to fix #7945

In the end, this PR just gets rid of the minify loader. I don't know enough about webpack and all of this any more, though. It looks like there is a lot of opportunity to enhance this, but just regarding that loader, it only deals with existing html files – not generating them via plugin or whatever. 

So, all possibly affected assets are:

```
...$ git ls-files HEAD '**/*.html'
frontend/account.html
frontend/console.html
frontend/editor.html
frontend/jupyter.html
frontend/sagews/3d.html
frontend/sagews/d3.html
frontend/sagews/interact.html
```

and surely, they are either small already or will go away anyways.

My understanding is that with this change, the html file is just copied over. There are ways to add a plugin for minifying html and set an optimization option – but that just causes a weird typescript error. In any case, I don't see the point in optimizing these few files.

### DIFF
--- a/src/packages/pnpm-lock.yaml
+++ b/src/packages/pnpm-lock.yaml
@@ -26,10 +26,10 @@ importers:
         version: 29.5.13
       jest:
         specifier: ^29.7.0
-        version: 29.7.0(@types/node@18.19.50)
+        version: 29.7.0(@types/node@18.19.55)
       ts-jest:
         specifier: ^29.2.3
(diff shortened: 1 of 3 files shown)


williamstein:
> surely, they are either small already or will go away anyways.

+1 and I agree with this.

PR #7964, opened 2024-10-18
bookmarks: starting with starred files

haraldschilly:
Implementing a rather simple "backup" of starred files in a new bookmarks table. Every time a project is opened or a star is set/removed, it reconciles local storage and what's stored in the database.

A follow up idea is to do the same for notebook cells.

also included:
- stricter typing of `trunc*` functions in that misc.ts file.
- do not log credentials


src/packages/frontend/project/page/flyouts/store.ts
```diff
@@ -0,0 +1,119 @@
...
+import {
+  FlyoutActiveStarred,
+  getFlyoutActiveStarred,
+  storeFlyoutState,
+} from "./state";
+
+// Additionally to local storage, we back the state of the starred files in the database.
+// Errors with the API are ignored, because we primarily rely on local storage.
+// The only really important situation to think of are when there is nothing in local storage but in the database,
+// or when there is
```
williamstein, line 26: This comment ends midsentence: "The only really important situation to think of[...] or when there is [NOTHING]"

williamstein:
Question: why not just make a new user_query spec synctable?  You have pages of code with api endpoints specifically for setting, getting, etc., some bookmarks.   Isn't this all automatic and much, much more uniform by just adding a few declarative lines in db-schema?   

I strongly agree with using the api for some types of set/get/update database operations when there are a lot of complicated things that might also happen, e.g., creating a purchase or creating a compute server.  However, for bookmarks, I feel like this could just be a few lines of code in a file in util/db-schema/bookmarks.ts say.     If it is just that it's confusing how to do that, I can give it a try.

williamstein:
The argument for using db-schema when it is applicable is that it's much more likely to be secure.  It's just one declarative approach that applies automatically and it's very hard to do something wrong.   Also, you get changefeeds automatically, so you could have bookmarks that are synchronized between all browser clients (with different localStorage) if we ever wanted, and it's just trivial and very robust (since it has to be).  

When I was implementing all the coordination to define which compute server is editing a file, etc., I wrote it like this with a bunch of api endpoints.  I then realized I was reinventing the wheel and making something that wasn't as robust, and just deleted all that code and rewrote it (much shorter) using a single synctable...

williamstein:
In short, I think there are like 500 lines of worrisome code to define an api for bookmarks, when one could I hope do something arguably better in a few lines of code right here:

https://github.com/sagemathinc/cocalc/pull/7964/files#diff-f981340abf57256db6da8205d33f917b71be152e3c188cbe118331e987eaa66bR44

williamstein:
@haraldschilly did you see my questions above?

williamstein:
I'm going to merge this as is, but I think it would be better to use a declarative user_query spec to read/write the data.  That can be done in a followup PR.

PR #7975, opened 2024-10-22
i18n 2024-10-22

haraldschilly:
This is a small start of translating the course manager.

---

* fixing frame title: do not limit title width if not necessary. e.g. here the german text is cut off, although not necessary. the fix just let it expand as it fits. does work with terminals and more narrow windows as well.

before:
Screenshot from 2024-10-22 12-34-32

after:
Screenshot from 2024-10-22 12-34-09

---

Another issue this fixes is horizontal scrolling of project panel pages. I noticed this yet again when looking for missing translations, because headers/body were cut off and there was no scroll bar. Now, with that change, if there is a necessity to scroll, the scroll bar is actually shown. This also harmonizes all the panels to have the same behavior.

---


williamstein:
Thanks!  This bugs me all the time; I wanted to fix this two days ago, but had to stay focused on some other issues...!

haraldschilly:
I just noticed something similar with the project panel pages, where I wasn't even able to scroll to the side (neither mouse nor scrollbars). This now shows the scrollbar if necessary. My feeling is, this has also an implication for what I had to fight with on that +new page rearrangement. In any case, now I want to actually do some translations ^^

williamstein:
What is "project panel pages"?

Issue #7976, opened 2024-10-22
frontend/next: findDOMNode is deprecated


haraldschilly:
This is basically to track our progress. Will be [removed in the next version of react](https://react.dev/reference/react-dom/findDOMNode).

Upstream tickets:

- https://github.com/react-component/resize-observer/issues/208 (which comes in via antd tooltip, and probably others)

PR #7977, opened 2024-10-22
frontend/flyout/terminal: get rid of deprecated ReactDOM.findDOMNode

haraldschilly:
I keep seeing findDOMNode deprecated warnings. This PR is a baby step towards removing it. Test: the flyout mini terminal shows up just like before, so, this small change is fine.

Ticket #7976 is more general, in particular it references the upstream ticket of the dependency, which is actually triggering that warning I keep seeing.

### DIFF
--- a/src/packages/frontend/project/page/flyouts/files-terminal.tsx
+++ b/src/packages/frontend/project/page/flyouts/files-terminal.tsx
@@ -8,7 +8,6 @@ import { debounce } from "lodash";
 
 import {
   CSS,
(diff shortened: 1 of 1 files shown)


williamstein:
> Ticket https://github.com/sagemathinc/cocalc/issues/7976 is more general, in particular it references the upstream ticket of the dependency, which is actually triggering that warning I keep seeing.

Same, and it's blocked on upstream, annoyingly...

PR #7992, opened 2024-10-30
I18n 20241029

haraldschilly:
* course: lot to do, initial steps...
* translating more of those account level settings
* account/ssh keys: fixing a bug parsing the key (avoiding a crash)
* removing next/settings/editor/keyboard, because of i18n conflicts and well, it is only having an impact for the frontend


williamstein:
It looks to me like this PR also deletes a random subset of https://cocalc.com/config, e.g., https://cocalc.com/config/editor/keyboard

I think it's bad enough that we have two different configuration systems, and that they have slowly diverged, with some things configurable in only one of the two places (e.g., AI model is only in the App).  It would be worse to silently delete random stuff from one of them.   Instead we should -- all at once -- remove the entire https://cocalc.com/config page, and have a redirect to the in-app configuration.  

Right now, with this PR, somebody could easily browse https://cocalc.com/config/ for "the range of configuration options" and come to the conclusion that there is nothing to change the keyboard shortcuts.   It would be better if https://cocalc.com/config/ didn't exist at all (i.e., redirected to the app). 

Does this make sense? Can you do one of the following:

- change this PR to not delete pages from https://cocalc.com/config/*, or 
- entirely delete everything from https://cocalc.com/config/* and put in redirects to https://cocalc.com/settings ?
- or tell me I'm wrong (it's sometimes hard to know what is going on when reading code).

williamstein:
I'm going to merge this anyways, but I think it raises the priority of putting in redirects to  https://cocalc.com/settings 

You'll probably want to anyways, since it means less to translate!

PR #7995, opened 2024-10-31
bookmarks: setup user_query support

haraldschilly:
WIP


williamstein:
In your view of how this will be used, will account_id and project_id always be set?

PR #7998, opened 2024-11-01
API: create account tweaks for onprem

haraldschilly:
* follow up of 433f4b3bcb56688c3edd9271c12f9cbb4c11c753: for onprem, require admin group:
  * When a regular user wants to create a user: `{'issues': {'api': "Error: Only users in the group 'admin' are allowed to create new users."}}`
  * When an admin wants to create a user, works fine (account_id returned)
* I also typed that db database object, and had to deal with the fallout. No change in functionality, just many small typing details.
* OpenAPI spec, also documenting that issues object upon failure.

Screenshot from 2024-11-01 14-16-57


williamstein:
THANKS!!!

PR #8022, opened 2024-11-13
next: add a missing key={} to an <A>

haraldschilly:
I saw a warning in the logs … and this adds a  `key={index}`, that's all


williamstein:
nice -- I also forget about that second argument to map.

PR #8043, opened 2024-11-27
frontend: add "verify email" banner

haraldschilly:
- show a banner if your email address is not verified
- buttons: one to open the account settings to fix it, or the same green one to send a verification email
- once sent: tells you to check email and click on the verification link
- if dismissed (gray X) then this is stored in local storage and the banner is back in a week.
- extra: improving the typing of icon names

---

Screenshot from 2024-11-28 14-54-06

---

Screenshot from 2024-11-28 14-54-24


src/packages/frontend/components/icon.tsx
```diff
@@ -681,14 +681,15 @@ try {
 }
 
 // This was nice but unfortunately it exceeds typescript limits.
-//export type IconName = keyof typeof IconSpec;
```
williamstein, line 684: The limits are no longer exceeded...?  Or using "as const" fixes it?

haraldschilly: I think either were are problem, but apparently it is fine now. There was a bit of fallout due to typing "icon" everywhere, but that's easy to fix.

williamstein: Nice!!

PR #8045, opened 2024-11-28
next/i18n: tweak intro text and translate screenshots alt text

haraldschilly:
Screenshot from 2024-11-28 11-31-19


novoselt (approved):
I've tried to look at all the languages where I can at least attempt to see how this version was handled and "synchronized" does seem to get across. Except for Russian. This is weird, but I think that the English original sounds great and clear, Russian version is at least grammatically correct, and overall the point does carry over.

PR #8047, opened 2024-12-02
frontend/cocalc.com: add "Ubuntu 24.04 (Testing)" option

haraldschilly:
with that, someone can choose the new dev image for 24.04. it's at least in a state where it is safe to play around

- ~~this also tests two postgres versions, instead of just one~~ already merged


williamstein:
Merging, but also see https://github.com/sagemathinc/cocalc/issues/8066

PR #8071, opened 2024-12-19
add commenting functionality to document editing

williamstein:
This turns out to be difficult since we are going to fully support storing the complete TimeTravel edit history of where all comments are located, and this requires a somewhat more sophisticated approach to our sync framework that fully supports branching and merging.


williamstein:
I'm closing this, since it's so bitrotten.  I'll start over on it sometime in the future.   The main hard part is actually already done in the NATS branch, which is live now.

PR #8072, opened 2024-12-19
npm updates

haraldschilly:
* update primus (for newer nanoid dependency)
* update express (newer path-to-regexp dependency)


haraldschilly:
I haven't tested this, but those are just minor version updates.

PR #8084, opened 2024-12-30
npm: fix nanoid dependency

haraldschilly:
primus, and some other less important packages, install nanoid<3.3.8. This forces to install a newer nanoid. The [actual changes are minimal, just that pool pollution problem](https://github.com/ai/nanoid/compare/3.3.6...3.3.8)


williamstein:
Now that I read about https://www.npmjs.com/package/nanoid  (= nano id) I want to use it.  I actually need something exactly like this for the commenting code: https://github.com/sagemathinc/cocalc/pull/8071

Anyway glad to be aware of this!

PR #8096, opened 2025-01-06
npm: update next to 14.2.22

haraldschilly:
testing: I tried opening a couple of next pages and everything was fine


williamstein:
I merged this, but then realized your totally forgot to update next in packages/hub.

williamstein:
ok, pushed fix.

PR #8105, opened 2025-01-09
npm: updating to typescript 5.7.3

haraldschilly:
This involves some headaches with Buffer vs. ArrayBuffer vs. Uint8Array. My vague understanding is, that `ArrayBuffer` is the low level object, the data itself, while `Uint8Array` is a view on top of it, which not only adds typing, but also methods to interact with the data. There is also a `SharedArrayBuffer` for web workers, which we don't have.

https://devblogs.microsoft.com/typescript/announcing-typescript-5-7/

Well, this was not easy, but so far I got this to work in my dev setup. (project works, I can eval a notebook cell, call the API, move an ipywidgets slider (and a second instance changes accordingly), etc.). Not sure what is affected in total by this, though.

Also, the only logic change I made is in `enable-messaging-protocol.ts`: there were two blocks starting with `if (!socket.writable) {` which made no sense to me – with this change here these blocks are married together.


williamstein:
This generally looks good to me.

haraldschilly:
Before merging this, I want to add checks for computing these sha1 sums, especially for data.

PR #8110, opened 2025-01-10
project: do not delete all COCALC_ variables

haraldschilly:
... only those for configuring the settings

### DIFF
--- a/src/packages/server/projects/control/util.ts
+++ b/src/packages/server/projects/control/util.ts
@@ -1,19 +1,22 @@
-import { promisify } from "util";
-import { dirname, join, resolve } from "path";
-import { exec as exec0, spawn } from "child_process";
 import spawnAsync from "await-spawn";
+import { exec as exec0, spawn } from "child_process";
 import * as fs from "fs";
 import { writeFile } from "fs/promises";
-import { projects, root, blobstore } from "@cocalc/backend/data";
-import { is_valid_uuid_string } from "@cocalc/util/misc";
-import { callback2 } from "@cocalc/util/async-utils";
-import getLogger from "@cocalc/backend/logger";
-import { CopyOptions, ProjectState, ProjectStatus } from "./base";
-import { getUid } from "@cocalc/backend/misc";
+import { dirname, join, resolve } from "path";
+import { promisify } from "util";
+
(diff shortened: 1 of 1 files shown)


williamstein:
I disagree with this.  This code definitely does not run in the hub. 

Because: (1) that code doesn't actually delete anything, it just defines an object (env) that gets passed in when starting a project explicitly, which ONLY happens in cocalc-docker and dev -- those code paths aren't used in Kubernetes mode, and (2) we only got one message about a problem right at startup, and no more.

I'm stumped about what is going on though. My only idea is to make the place that uses COCALC_BLOB_STORE just fallback to "/blobs" if it isn't set. /blobs is fine as a default -- we mount the gcs bucket via gcsfuse there no matter what other configuration is going on.

haraldschilly:
yes, I am also dubious that this has an effect. /blobs as a fallback is certainly fine.

williamstein:
I'll do a quick 2-liner to add that now.

williamstein:
```diff
~/cocalc/src/packages/database$ git diff
diff --git a/src/packages/database/postgres-blobs.coffee b/src/packages/database/postgres-blobs.coffee
index bb4463b0d9..1ca6aa96f4 100644
--- a/src/packages/database/postgres-blobs.coffee
+++ b/src/packages/database/postgres-blobs.coffee
@@ -211,8 +211,8 @@ exports.extend_PostgreSQL = (ext) -> class PostgreSQL extends ext
                     cb()
                 else if x.gcloud
                     if not COCALC_BLOB_STORE?
-                        cb("no blob store configured -- set the COCALC_BLOB_STORE env variable")
-                        return
+                        # see comment https://github.com/sagemathinc/cocalc/pull/8110
+                        COCALC_BLOB_STORE = '/blobs'
                     # blob not available locally, but should be in a Google cloud storage bucket -- try to get it
                     # NOTE: we now ignore the actual content of x.gcloud -- we don't support spreading blobs
                     # across multiple buckets... as it isn't needed because buckets are infinite, and it
```

williamstein:
https://github.com/sagemathinc/cocalc/commit/264e24bad4b1c9a41fdd3a94ac4497c8b6259729

PR #8117, opened 2025-01-13
hub/startup: set support account to first admin account

haraldschilly:
if `support_account_id` isn't set for onprem, set it to the first admin account. This makes it clear for admins who will be the sender of these messages.

### DIFF
--- a/src/packages/database/postgres/admins.ts
+++ b/src/packages/database/postgres/admins.ts
@@ -3,7 +3,7 @@ import getPool from "@cocalc/database/pool";
 // get account_id's of all admins -- "long" cache
 // *Always sorted by account_id lexicographically.*
 
-export default async function admins(): Promise<string[]> {
+export async function getAdmins(): Promise<string[]> {
   const pool = getPool("long");
   const { rows } = await pool.query(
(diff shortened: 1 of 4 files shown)


williamstein:
1. My intention was the the cocalc support account not be an admin account,
2. I wrote things so that if the support account doesn't exist a brand new account with name "CoCalc Support" is automatically created and set in server settings.

haraldschilly:
Hmm, I had no idea. In my dev environment the field is empty `support_account_id` and I found some fallback code to pick an admin account. So, that just makes this explicit. I'll try to find the code that creates that support account – it's clearly not running in my case.

williamstein:
> In my dev environment the field is empty support_account_id 

It's empty until it's first **actually used**.   There is no need to create it before it is needed.  This also provides a chance for it to be created explicitly during some early configuration step if desired...

> and I found some fallback code to pick an admin account. 

That's for something else, namely admin alerts.

> So, that just makes this explicit. I'll try to find the code that creates that support account – it's clearly not running in my case.

It does run fine.  It just hasn't been run already.  See

https://github.com/sagemathinc/cocalc/blob/f695624960eb720f7ae6a8640d78b65f04afb7ac/src/packages/server/messages/support-account.ts#L17

PR #8119, opened 2025-01-13
send admin messages from the support account

haraldschilly:
By removing "from_id" for sending admin messages, it will fall back to the support user account.


haraldschilly:
Besides sorting imports, this is just removing this line:

```
    return await send({
      subject: `Admin Alert - ${subject}`,
      body: toString(body) + stack,
-      from_id,         <<<<<<<<<<<<<<<<
      to_ids,
      dedupMinutes,
    });
```

PR #8174, opened 2025-02-04
frontend/quota: change presets to a dropdown and special-case for LLM models

haraldschilly:
* lower preset values for LLMs
* instead of showing the preset tags everywhere, this is now a dropdown selector. editing the value works just like before.
* I also attempted to change this for the modal, but I couldn't figure out how to show the modal. I'm somehow assuming that modal is deprecated.

Screenshot from 2025-02-04 11-43-43

Screenshot from 2025-02-04 11-42-15

### DIFF
--- a/src/packages/frontend/client/purchases.ts
+++ b/src/packages/frontend/client/purchases.ts
@@ -1,7 +1,7 @@
 /*
 Functions for interfacing with the purchases functionality.
 
-TODO/DEPRECATE: this module is mostly pointess since I moved essentially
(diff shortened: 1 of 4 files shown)


williamstein:
Quick dumb question.  In the current version you can easily type in an arbitrary amount. In your screenshots, that functionality *seems* completely gone?

image

versus

image


Am I missing something?

haraldschilly:
> Quick dumb question. In the current version you can easily type in an arbitrary amount. 

It's the same component, you can type in a number like before.

PR #8276, opened 2025-04-18
npm package updates

haraldschilly:
* nyc: I think we don't use it, but in any case, now it is ~~updated~~ deleted
* there are coffee-coverage-loader file, which we do not need (at least not use)
* antd: this needs double checking
  * there are two interesting features:
     * https://github.com/ant-design/ant-design/pull/51550
     * https://github.com/ant-design/ant-design/pull/51653
* antd icons: was holding back some dependency as well

This needs testing… I checked next pages and the frontend, various editors, ... everything looks and works as usual


williamstein:
> nyc: I think we don't use it, but in any case, now it is updated

I agree that we don't use it. It seems to be related to determining unit test coverage?  Let's remove it.

PR #8299, opened 2025-04-23
explorer: merge search and terminal

haraldschilly:
So, this isn't perfect, there is really a lot going on, but I think this hits the general idea. All the features I'm aware of are still the same.

What I changed is
* where the "terminal mode" banner appears (further up) 
* checking for the return key for submitting the mini terminal command  in the keydown handler (and not a `<form ...>` tag wrapped around the input box
* The icons for showing other users' presence is further to the right, because in the middle it feels awkward (chat is everywhere to the right)
* The intended flex layout is 2:1, but it's not really visible in the screencast. It also wraps as before, if really narrow.

additionally: this fixes a few typos and Antd deprecations

[screen-capture (2).webm](https://github.com/user-attachments/assets/d76cde9a-9528-491a-a058-a24a8bc640ef)


williamstein:
wow, that looks awesome!!

PR #8356, opened 2025-05-29
next/api: limit password length

haraldschilly:
pretty simple fix to limit the password length. ~~what's left is testing.~~


williamstein:
Andrey: "What about sign in? Same problem."

haraldschilly:
ok. now the sign-in form and the (central to all this) `getAccount` function of the api endpoint, used by other parts, has a password length check as well.

PR #8399, opened 2025-07-01
github: split up actions for each package

haraldschilly:
WIP

### DIFF
--- a/.github/workflows/make-and-test.yml
+++ b/.github/workflows/make-and-test.yml
@@ -46,6 +46,97 @@ jobs:
   build:
     runs-on: ubuntu-latest
 
+    outputs:
+      packages: ${{ steps.packages.outputs.dirs }}
+
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/make-and-test.yml
```diff
@@ -106,7 +194,34 @@ jobs:
...
+      - name: Run tests
+        run: |
+          source venv/bin/activate
+          cd src/packages/${{ matrix.dir }}
+          if [ ! -f package.json ]; then
+            echo "::warning::No package.json found in ${{ matrix.dir }}"
+          elif ! jq -e '.scripts.test' package.json > /dev/null; then
+            echo "::notice::No test script in ${{ matrix.dir }}"
+          fi
+          pnpm run test
```
williamstein, line 226: Just use `pnpm run --if-present test` instead of jq shell commands.  

https://pnpm.io/cli/run#--if-present

Also, do we have packages with no package.json?

Issue #8397, opened 2025-07-01
flyout panel "Show labels" menu item and icon


williamstein:
When labels are visible this is the menu:

<img width="272" alt="Image" src="https://github.com/user-attachments/assets/200d0786-ca0c-44d9-9d95-e9c5020f3275" />

- The icon to the left should not be a checkbox, it should be something that indicates a label.
- The menu text should side "Hide labels" since that's what you are actually doing.

Conversely, when labels are not visible,

<img width="264" alt="Image" src="https://github.com/user-attachments/assets/3db60e2b-4a56-4ee7-bf00-806970c97d30" />

- The icon should be the same -- something that indicates a labels.
- The text "Show labels" is fine.

The above would be consistent with how the menu for hiding/showing the menu toolbar works:

<img width="303" alt="Image" src="https://github.com/user-attachments/assets/bcc8c197-a870-46cd-83f6-04d8c78532ea" />

or e.g., in an operating system how hiding a window works:

<img width="317" alt="Image" src="https://github.com/user-attachments/assets/02a30a02-8fd3-4c3e-884f-5b48169e5c6e" />

PR #8415, opened 2025-07-07
frontend/action bar: show/hide labels menu fix and more padding

haraldschilly:
fixing #8397

Screenshot from 2025-07-07 16-38-42

Screenshot from 2025-07-07 16-42-08

## the label

Screenshot from 2025-07-07 16-42-56


williamstein:
It looks great!

PR #8420, opened 2025-07-08
symbols-bar: optionally show labels and slightly tweak margins

haraldschilly:
below is a comparison of before and after (no intentional change, but the height of the symbol is now set and with that, the margins/padding play out a little bit better)

Screenshot from 2025-07-08 16-48-51

This is the setting:

Screenshot from 2025-07-08 16-49-59

And this is similar to before 1c54e48109f9b

Screenshot from 2025-07-08 16-49-28


williamstein:
I merged your other PR and now this one has conflicts for all the translations.

williamstein:
I merged this since it's clearly way better than not having it.  However, there are still 3 issues, which need to get addressed:

https://github.com/sagemathinc/cocalc/issues/8438

PR #8422, opened 2025-07-09
cleaner top bar

williamstein:
Another idea to clean up the top -- big icons for each project, click "+" to get the projects page, show the title in the middle (probably too big in this screenshot), use the same color for the entire top bar as for the activity bar to clearly distinguish it from the file tabs.

image

image

This is basically just a minor modification of our existing code (so easy)...


williamstein:
not planned

PR #8431, opened 2025-07-10
hub/conat: discover via k8s api

haraldschilly:


haraldschilly:
I tested this in onprem, I get

`cocalc:debug:conat:core:server scan: remote http://10.124.0.25:5000 knows about  Set(2) { 'http://10.124.0.25:5000', 'http://10.124.2.31:5000' }`

which looks promising, and there are two routers – but the health check still fails.

williamstein:
Thanks!

haraldschilly:
ok! the variant where it calls the API directly works in onprem!

PR #8473, opened 2025-07-29
frontend/jupyter: add cell-type switches to the button menu of a cell

haraldschilly:
This is how it looks (including some reorg), but needs more work.

The only usability issue is, if you switch away from a code type, that cell menu is gone as well. Then you're stuck in markdown and the menu where you did the change can't be reached. So, the extra work here is to make that menu available for all cell types, and only show "run" and so on if it makes sense.

What I did is to remove the "Edit" button from rendering it in the "cell input", down to render it in the "cell-buttonbar". With that, everything is in one place. Bonus: it is trivial to enable the "Format" button for markdown as well, although it is a bit strange if the cell isn't in edit mode and edited as markdown. So, I left it disabled as before.

Screenshot from 2025-07-29 12-25-20

---

Ok, and here is with the additional changes for markdown and raw cells. The "Edit" button is actually a toggle now. The menu now allows undo/redo, moving, cutting, etc. for markdown as well.

Screenshot from 2025-07-29 16-12-16

### DIFF
--- a/src/packages/frontend/frame-editors/jupyter-editor/cell-notebook/actions.ts
+++ b/src/packages/frontend/frame-editors/jupyter-editor/cell-notebook/actions.ts
@@ -422,8 +422,7 @@ export class NotebookFrameActions {
       undefined,
       false,
     );
-    let md_edit_ids = this.store.get("md_edit_ids");
-    if (md_edit_ids == null) md_edit_ids = Set();
+    const md_edit_ids = this.store.get("md_edit_ids", Set());
     if (md_edit_ids.contains(id)) {
       return;
     }
@@ -497,9 +496,8 @@ export class NotebookFrameActions {
     this.set_mode("edit");
   }
 
-  cell_md_is_editing(id): boolean {
-    let md_edit_ids = this.store.get("md_edit_ids");
-    if (md_edit_ids == null) md_edit_ids = Set();
+  public cell_md_is_editing(id): boolean {
+    const md_edit_ids = this.store.get("md_edit_ids", Set());
     return md_edit_ids.contains(id);
   }
 
@@ -511,7 +509,6 @@ export class NotebookFrameActions {
       return;
     }
 
-    this.set_cur_id(id);
     if (this.cell_md_is_editing(id)) {
       this.set_md_cell_not_editing(id);
       this.set_mode("escape");
@@ -519,6 +516,7 @@ export class NotebookFrameActions {
       this.switch_md_cell_to_edit(id);
       this.set_mode("edit");
     }
+    this.set_cur_id(id);
   }
 
   public switch_code_cell_to_edit(id: string): void {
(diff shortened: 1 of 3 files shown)


williamstein:
> only show "run" and so on if it makes sense.

Running a markdown cell makes sense -- it means changing it from edit mode to non-edit mode.

williamstein:
> But what's being run? 

The definition of "run" is "compile the markdown source into a visible rendered view".  That said, this would be super annoying if it happened for all viewers at once, so only the particular frame that you're viewing has its mode changed from edit to view.  That's why it's ` frameActions.current?.run_cell(id)` since `actions?.run_cell(id)` has no knowledge of the particular frame.

src/packages/frontend/frame-editors/jupyter-editor/cell-notebook/actions.ts
```diff
@@ -421,8 +422,7 @@ export class NotebookFrameActions {
       undefined,
       false,
     );
-    let md_edit_ids = this.store.get("md_edit_ids");
-    if (md_edit_ids == null) md_edit_ids = Set();
+    const md_edit_ids = this.store.get("md_edit_ids", Set());
```
williamstein, line 425: This change you just made is NOT equivalent and it pretty dangerous.   If 'md_edit_ids' is actually set to undefined or to null in the store, then ` this.store.get("md_edit_ids", Set());` would be undefined or null, NOT Set().  It's much, much safer to do:

```
const md_edit_ids = this.store.get("md_edit_ids") ?? Set()
```

It's also more efficient since above doesn't evaluate Set() unless needed, whereas ` this.store.get("md_edit_ids", Set());` always does.

src/packages/frontend/frame-editors/jupyter-editor/cell-notebook/actions.ts
```diff
@@ -487,7 +487,7 @@ export class NotebookFrameActions {
     const cell = this.get_cell_by_id(id);
     if (cell == null) return;
 
-    if (cell.getIn(["metadata", "editable"]) === false) {
+    if (!this.jupyter_actions.store.is_cell_editable(id)) {
       // TODO: NEVER ever silently fail!
```
williamstein, line 491: feel free to delete my comment "// TODO: NEVER ever silently fail!" -- I've learned people greatly prefer silent failures to errors they don't understand.  At most, log something to the console...

src/packages/frontend/frame-editors/jupyter-editor/cell-notebook/actions.ts
williamstein: That said, the function that implement this.store.get(...) could also be *changed* to basically do the ?? thing.   That said, I generally think ?? is just better, and we shouldn't write new code using that second input to get.  I probably did so also in the last week, but I'm going to stop.

williamstein:
I had a nitpick comment or two, but no actual bug, so I merged this.

PR #8477, opened 2025-07-30
frontend/explorer: enhance breadcrumb nav

haraldschilly:
* this is mode == files only
* by spelling out "Home", next to the icon
* and adding an "up" arrow button (more could be done, but I wanted to start simple)

Screenshot from 2025-07-30 14-17-43

Screenshot from 2025-07-30 14-17-23


williamstein:
Coincidentally in my new filesystem branch I added support for shift+up to go the parent directory.

Issue #8474, opened 2025-07-29
deprecate the  progressive web app functionality for cocalc


williamstein:
1. It adds complexity
2. I've never heard of anybody actually using it.
3. Progressive web app support got much more aggressive in my latest version of Chrome, making this even more annoying and confusing. 
4. CoCalc really just isn't designed to be a progressive webapp, so the resulting thing is not particularly useful at present anyways.

I think we should just either delete the code that enables this or at least put it behind a const that disables it.

PR #8482, opened 2025-08-01
frontend: disable PWA install app 

haraldschilly:
WIP

#8474


williamstein:
Looks good and will be easy to revert if there is an outcry, or interest from an onprem customer or something else...

PR #8492, opened 2025-08-07
next/api/v2: add set-admin-quotas endpoint

haraldschilly:
This helps admins for onprem setups to adjust specific project quotas.

What I tested:

* only admin API keys work, otherwise you get `{'error': 'only admins can set project quotas'}`

* adjusting the quotas
 
without settings, the effective `run_quota` is 

{`"network": false, "cpu_limit": 1, "cpu_request": 0.02, "member_host": false, "idle_timeout": 1800, "memory_limit": 1000, "always_running": false, "memory_request": 200, ...}`

Then, after this call and stopping/starting the project explicitly ...

```
{  "project_id": "...",
    "always_running": True,
    "internet": True,
    "cpu_limit": 2,
    "cpu_request": 1.5,
    "memory_limit": 10000,
    "memory_request": 5555,
    "idle_timeout": 60000,
}
```

... the `run_quota` changed:

`{"network": true, "cpu_limit": 2, "cpu_request": 1.5, "idle_timeout": 60000, "memory_limit": 10000, "always_running": true, "memory_request": 5555, ...}`


williamstein:
This is exactly the sort of functionality where our current api and zod schema makes sense.

PR #8501, opened 2025-08-11
npm: update dependency to get update for form-data@2.5.3 and @4.0.3

haraldschilly:
* @types/request and @types/node-fetch: This is a fix for an (in our case) rather pointless security related dependency issue.
* axios: minor version update, same underlying problem, next dependency


haraldschilly:
the test error is unrelated to this. I think this PR is fine, though

PR #8518, opened 2025-08-19
frontend/llm: record llm prompt history

haraldschilly:


williamstein:
The UI extra looks good. However, it looks like you made a key:value store with exactly two keys and the *entire history* as the values.  So if Blaec uses this for a few weeks (say), and has a thousand entries in the history, then every single time he does one single chat, he'll be sending the entire 1000-entry history to the server and his other browser tabs then get a new copy of this history.     Basically objects in a key:value store are viewed as atomic and every time they are changed, the new object is sent.  So this approach is very, very bad in terms of scalability, server load, etc.      I would be less worried about this, but there's no easy migration later either.

Alternatives: 

- use a dkv, but instead have a different key for each history entry (e.g., a timestamp or counter+random value to avoid conflicts).   With a dkv when the user connects you still (very efficiently) get ALL of the data in a single message, so this is no slower than what you've already done.  The advantage is that whenever there is a new history item only that one item gets sent to the server and other clients -- instead of the entire history.
- use a dstream, which is a stream of values (basically an array).  Again you get the full list of all values on connect, but only new values get sent out.  You can delete old entries.  There's a timestamp automatically.  Basically, you connect, scan through the array and process it somehow into a client-side data structure, and then process updates as they come in.  Note that dkv is basically just built on dstream.


And I'm very happy you were able to implement this so easily, which shows Conat is usable by somebody $\neq$ me!

williamstein:
One other thing -- the history needs to be searchable!    Do this:

https://ant.design/components/select#select-demo-search-box

williamstein:
Please delete all of the claude auto-generated comments that got added all over in the conat code.  They add absolutely no useful information whatever and are making even reading this to review it really hard.

williamstein:
> Ok, but for me this is quite useful. Otherwise I have to open the source code or search for usages to see how the API works. This would be much more convenient for me.

I tried searching this branch so I could list several and show how they are almost entirely noise with absolutely no useful information, but I can't find them now (maybe your deleted the history).  I was going to do this before (I obviously did so in my head), but didn't think it would be necessary.

PR #8561, opened 2025-09-09
frontend/admin/license: fix expire date field 

### LINKED ISSUE #8559: admin/license: can't set expiration field
When I create a license as admin and set an expiration, I get this error:

> Error: postgresql error: invalid input syntax for type timestamp: "{"$L":"en","$d":"2025-09-29T22:00:00.000Z","$y":2025,"$M":8,"$D":30,"$W":2,"$H":0,"$m":0,"$s":0,"$ms":0,"$x":{},"$isDayjsObject":true}"

haraldschilly:
This fixes #8559 by sending the string of the date to postgres. the day.js object does not work.

### DIFF
--- a/src/packages/frontend/site-licenses/admin/license.tsx
+++ b/src/packages/frontend/site-licenses/admin/license.tsx
@@ -7,7 +7,6 @@ import jsonic from "jsonic";
 import { DebounceInput } from "react-debounce-input";
 import { useIntl } from "react-intl";
 
-import { labels } from "@cocalc/frontend/i18n";
 import { Projects } from "@cocalc/frontend/admin/users/projects";
 import { Button, Checkbox } from "@cocalc/frontend/antd-bootstrap";
 import { CSS, Rendered, TypedMap } from "@cocalc/frontend/app-framework";
@@ -15,10 +14,11 @@ import {
   A,
   CopyToClipBoard,
   DateTimePicker,
-  Icon,
   Gap,
+  Icon,
   TimeAgo,
 } from "@cocalc/frontend/components";
+import { labels } from "@cocalc/frontend/i18n";
 import {
   capitalize,
   days_ago,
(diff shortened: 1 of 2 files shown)


src/packages/frontend/site-licenses/admin/license.tsx
```diff
@@ -35,7 +35,7 @@ import { actions } from "./actions";
 import { Managers } from "./managers";
 import Owner from "./owner";
 import { DisplayQuota, EditQuota } from "./quota";
-import { license_fields, license_field_type, ManagerInfo } from "./types";
+import { license_field_type, license_fields, ManagerInfo } from "./types";
```
williamstein, line 38: Why is this switched?  Is it some sort of linting?  It's true that the new order is better since it is alphabetical, but it's a surprising thing for a non-linter to actual change.

src/packages/frontend/site-licenses/admin/license.tsx
```diff
@@ -142,6 +142,12 @@ export const License: React.FC<Props> = (props: Props) => {
     actions.set_edit(license.get("id"), field, new_val);
   }
 
+  function on_date_change(field, _date, dateString): void {
+    // For date fields, use the dateString instead of the Day.js object
+    // to ensure proper serialization for PostgreSQL
+    actions.set_edit(license.get("id"), field, dateString || null);
```
williamstein, line 148: Why use `dateString || null` instead of `dateString ?? null` which seems more idiomatic in modern times.  Who uses || anymore?

williamstein:
**This code is weird.**  Every single change just seems weird.

src/packages/frontend/site-licenses/admin/license.tsx
haraldschilly: all imports are sorted by vs code.

src/packages/frontend/site-licenses/admin/license.tsx
haraldschilly: So, there is an issue, where unsetting the expire date field does not set it to null in the database. The idea is to send a "null" value instead of an empty string. Doesn't work, though.

PR #8565, opened 2025-09-11
latex/rmd/qmd: add streaming exec support

haraldschilly:
let project<->frontend talk via conat and use that stream to show the compilation progress. this basically replicates what llm streaming chat does.

same for qmd and rmd

also enhance the qmd/rmd build panels to show cpu/mem, ability to stop, and force build. also a bit of styling, font size

---

there were also additional issues in aggregate.ts and I think I also figured out why build calls for an ongoing build didn't work. This is now much more robust.

---

latex: I opted to split up stdout and stderr, because there is "informative" stderr output, which is added at the bottom of the combined output, and then the progress in stdout can't be seen any more. 

Screenshot from 2025-09-16 12-17-29

---

## deployment

I'm also wondering about deployment. Probably hub first (conat communication). I assume it needs an update if there is a new service. At the same time we also update the project code. Once this is in place (and after many new projects are running) we deploy static for the actual changes. I added a modal to request or a project restart, if the service is not available. I.e. this is what pops up, if I change the name of the service in the project to not match what the frontend asks for:

Screenshot from 2025-09-17 17-23-37

### DIFF
--- a/src/packages/frontend/frame-editors/latex-editor/build.tsx
+++ b/src/packages/frontend/frame-editors/latex-editor/build.tsx
@@ -89,6 +89,10 @@ export const Build: React.FC<Props> = React.memo((props) => {
     margin: "0",
   } as const;
 
+  function renderErrorHeader(errorIsInformational: boolean): string {
+    return errorIsInformational ? "Output messages" : "Error output";
+  }
+
   function render_tab_item(
     title: string,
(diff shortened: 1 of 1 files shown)


williamstein:
https://github.com/sagemathinc/cocalc/blob/fs2/src/packages/util/throttle.ts

PR #8598, opened 2025-09-26
frontend/projects: star projects, similar to starring files

haraldschilly:
a star for projects on the list and in the about settings. backed similarly to file stars. ... and of course, an additional filter button.

### two projects have a star
Screenshot from 2025-09-26 16-58-29

### with filter enabled

Screenshot from 2025-09-26 16-58-43

### project settings → about 

Screenshot from 2025-09-26 16-59-03


src/packages/frontend/projects/use-bookmarked-projects.ts
```diff
@@ -0,0 +1,154 @@
...
+    | ((changeEvent: {
+        key: string;
+        value?: Record<string, any>;
+        prev?: Record<string, any>;
+      }) => void)
+    | null
+  >(null);
+
+  // Initialize conat bookmarks and set up listeners
+  useEffect(() => {
```
williamstein, line 50: Maybe useAsyncEffect would be slightly better for this?   Search our code for it...

haraldschilly: I tried it for the "LLM History", but it was still leaking (too many event emitters warnings). Seemed to me, something didn't work with the returned function in the useAsyncEffect -- or I was using it in a silly way. With this pattern, however, I had more luck avoiding this.

PR #8599, opened 2025-09-27
npm: update tar-fs, zstd-napi and better-sqlite3

haraldschilly:
3 updates, in particular tar-fs, and the other two appear as dependencies (they still depend on the older tar-fs, but it's good to keep them updated, though)


williamstein:
For what it is worth, tar-fs isn't used anymore in cocalc2.  It was part of websocketfs.

williamstein:
More precisely cocalc doesn't directly use tar-fs anymore. It's potentially a dependency of dependencies in various ways still.

PR #8650, opened 2025-11-19
wcag/contrast: split off some of antd-theme and adjust links color

novoselt:
Somewhat following codex suggestions the base of antd-theme is split off to a separate file so that it can be imported by next pages as well for uniform styling. As a test of this approach, the link color on the main page is changed to a darker brand color to improve contrast:
image


haraldschilly:
I'm ok with merging. There aren't many changes and we can tweak this in follow up PRs. The only detail I wonder is, changing the ".ext" of filenames. I mean, the whole idea is to fade them away because they're not of so much interest. Changing this to high contrast defeats the purpose. So, maybe we should really start adding an accessibility account configuration setting, and only if this is true, such details are adjusted.

If you think this is a good idea, I can introduce such a setting. For the first iteration is is just a simple yes/no setting, but down the road this could be more complex. 

In any case, I've clicked on the merge button, we'll then see how this looks and we can write tickets about fallout details.

PR #8654, opened 2025-11-25
Dim extensions setting

novoselt:
Clarify the meaning of graying out some files, add an option to gray out extensions, make these grays consistent (and a bit darker than the current super light gray), and propagate the behaviour to file explorer, log, flyout panels, and the file names or tabs (or rather their extensions):
image
image
and if we turn them on:
image

### DIFF
--- a/src/packages/frontend/project/page/file-tab.tsx
+++ b/src/packages/frontend/project/page/file-tab.tsx
@@ -536,6 +536,8 @@ const LABEL_STYLE: CSS = {
   whiteSpace: "nowrap",
 } as const;
 
+const DIMMED_STYLE = { color: COLORS.FILE_DIMMED } as const;
+
 const FULLPATH_LABEL_STYLE: CSS = {
   // using a full path for the label instead of just a filename
   textOverflow: "ellipsis",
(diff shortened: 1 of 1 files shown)


williamstein:
This also gets rid of asked toggle, which I think is good.   It overall looks good to me, but I'll let Harald look as well, and run an ai review, since I'm not sure what to do with translations.

haraldschilly:
> @haraldschilly Was it right to remove the translation key? Or should it be done in a separate workflow?

So, from what I can see you haven't removed the `labels.masked_files` key. You modified artifact files, that's all I can see. In i18n/README is a description of the workflow.

src/packages/frontend/account/other-settings.tsx
```diff
@@ -211,6 +210,18 @@ export function OtherSettings(props: Readonly<Props>): React.JSX.Element {
...
   }
 
+  function render_dim_file_extensions(): Rendered {
+    return (
+      <Switch
+        checked={!!props.other_settings.get("dim_file_extensions")}
+        onChange={(e) => on_change("dim_file_extensions", e.target.checked)}
+      >
+        <strong>Dim file extensions:</strong> gray out file extensions so their
+        names stand out.
```
haraldschilly, line 220: @novoselt  If you want to get your feet wet with translations, this here is an ideal place to start. Above in line 132 is the blueprint how to do this. e.g. 

```
 <FormattedMessage
   id="account.other-settings.dim_file_extensions"
   defaultMessage={`<strong>Dim file extensions:</strong> gray out file extensions so their names stand out.`}
        />
```

and then you just have to call the i18n:update function (a chain of extract→upload→translate→download→compile)

I can also do this in a follow up as well, no problem.

haraldschilly: https://github.com/sagemathinc/cocalc/pull/8656

PR #8665, opened 2025-11-28
Billing preferences

novoselt:
- Changed the settings menu so that there is a top level Billing submenu similar to Preferences.
- Submenus are closed by default and only one is open at a time, selecting an item outside of them closes them.
- Added missing tiles for Upgrades and Payment Methods.
- Restyled the menu panel in collapsed/hidden mode: before icons were strangely off-center and submenus had arrows obscuring icons.
- The open/close behaviour is the same as with full size panels, only one is opened at a time.
- Simplified some code to avoid duplication and confusing different behaviour for similar elements when something is changed.
- There are new keys for translations, but translation was NOT done, I need some further advice on using SimpleLocalize.
image 
image


src/packages/frontend/account/account-page.tsx
```diff
@@ -633,6 +667,7 @@ export const AccountPage: React.FC = () => {
 
   return (
     <div className="smc-vfill">
+      <style>{ACCOUNT_MENU_INLINE_CSS}</style>
```
haraldschilly: @novoselt we don't do this by style tags, I'll fix this

src/packages/frontend/account/settings-index.tsx
```diff
@@ -91,6 +96,10 @@ const MESSAGES = defineMessages({
     id: "account.settings.overview.payments",
     defaultMessage: "Manage payment methods and transaction history.",
   },
+  paymentMethods: {
+    id: "account.settings.overview.payment_methods",
+    defaultMessage: "Manage your saved payment methods or add new ones.",
+  },
```
haraldschilly, line 102: I ran `pnpm i18n:update`, then git-added the new json files and the other changes I made as a commit.

```
hsy@x2:~/p/cocalc/src/packages/frontend$ pnpm i18n:update
[...]
✓ All auto-translation jobs completed!
[...]
compile completed in 3 seconds.
```

PR #8670, opened 2025-12-04
accessibility: restructure the policy/statement page, link fresh VPAT

novoselt:
Rendered view:
image
VPAT is now served as a PDF from the official template - it is likely to be more convenient for customers bookkeeping and discussions. The link is somewhat "lost" on the new page, but those who are interested in it really should read the page first.


src/packages/next/pages/policies/accessibility.tsx
```diff
@@ -28,12 +27,37 @@ export default function AccessibilityPage({ customize }) {
...
+                industry standard Jupyter Classic (and Jupyter Lab). These
+                projects have put substantial deliberate efforts into making
+                their products accessible, although they still do not claim to
+                have AA compliance with WCAG.
+              </p>
+              <p>
+                For more specific details, please consult our{" "}
+                <a href="/documents/SageMathInc_VPAT2.5Rev_WCAG_February2025_December2025.pdf">
+                  Voluntary Product Accessibility Template, VPAT®
+                </a>{" "}
```
haraldschilly, line 58: my gut feeling is this should be an uppercase `<A ...>` tag, because they have a prefix logic. The path should be relative as well (just remove the leading "/"). I'm not sure, though, maybe with PDF something isn't ok with that approach.

haraldschilly: but I see, you probably need the leading /

novoselt: If I remove / it resolves to policies/documents instead of just documents, so I would need to move the file then - is this preferable?

As for a vs A codex suggests it should be lower case for the build-in HTML element rather than capital A for the component. I don't find it very convincing, but if I add the necessary import for A and use it, the page renders fine, but the PDF does not open, giving an error instead:
image

PR #8693, opened 2026-01-12
frontend/latex: fix time travel frame path for knitr files

### LINKED ISSUE #8341: timetravel frame editor type change broken for rnw and rtex
Opening time travel by changing the PDF preview frame type to TimeTravel of an rnw or rtex file hangs. Doing the same with other files works for me.

I was able to track the problem down to an exception after 10 seconds happening in `time-travel-editor/actions.ts` → `_init2` → `this.init_syncdoc()`

Image

Image

## Example

1. Create `rnw.rnw` file with that content:

```
\documentclass{article}
\usepackage[utf8]{inputenc}
\usepackage[T1]{fontenc}
\usepackage{url}
\begin{document}

<<>>=
summary(rnorm(50))
@

\end{document}
```

2. Then click on the PDF preview type change of the frame editor → TimeTravel
3. The analog is broken for rtex.

My guess, there is a mismatch of the file extension vs. was TimeTravel tracks.

## Workaround

Click on the blue TimeTravel button.

haraldschilly:
Fix time travel frame type on knitr files by using the source .rnw/.rtex path when switching frames.

Fixes #8341.


haraldschilly:
@codex review

PR #8698, opened 2026-01-14
npm: remove get-random-values to get rid of min-document

haraldschilly:
instead of a thin backwards compatibility wrapper, we assume modern node 20+ and browser, and just use the crypto module's random number generator directly.


src/packages/static/src/rspack.config.ts
```diff
@@ -226,6 +226,8 @@ export default function getConfig({ middleware }: Options = {}): Configuration {
         path: require.resolve("path-browserify"),
         util: require.resolve("util/"),
         assert: require.resolve("assert/"),
+        crypto: false, // get-random-values uses window.crypto in browser
```
williamstein: what about code that doesn't use window.crypto but imports crypto?  E.g., I wrote this in some new code in util:
```
      import { createPrivateKey, createPublicKey, sign, verify } from "crypto";
```
and it will be broken if I merge this.

This PR makes me uncomfortable, especially with patching the code of get-random-value.   

I just looked at this get-random-value package -- it's 40 lines of code; we should delete it and use crypto directly.   Instead of breaking importing from crypto, ensure importing from crypto works and use crypto.getRandomValues directly.   

E.g., this PR adds twice as many lines in unit tests as the code of the get-random-values library...

williamstein (changes requested):

src/packages/static/src/rspack.config.ts
haraldschilly: I was curious what this patching even does. I also think not using it is the best way forward.

haraldschilly: patching is good to know, though. it's specific to the version and hash of a package, so, this could be a useful escape hatch to know about

PR #8705, opened 2026-01-16
perf(frontend): avoid RegExp recompilation in synctex parser

haraldschilly:
Moves the RegExp definition outside the `parse_synctex_output` function scope to prevent recompilation on every call.

Measured improvement: ~8-9% faster execution time (from ~958ms to ~870ms for 100k iterations).


haraldschilly:
@codex review

PR #8706, opened 2026-01-16
fix: use asyncio.sleep in async _retry_with_backoff

haraldschilly:
The `_retry_with_backoff` function in `jupyter.py` was using `time.sleep`, which blocks the event loop when called from the async `jupyter_execute` function.

This commit replaces the synchronous `_retry_with_backoff` with an asynchronous version that uses `asyncio.sleep` to prevent blocking the event loop. The `jupyter_execute` function has been updated to be async and use the new retry logic.

I also fixed a bug where newline characters were being escaped in f-strings.


haraldschilly:
@codex review

PR #8713, opened 2026-01-22
ci postgres in mem + fix doc url links

haraldschilly:
- **ci: speed up postgres in tests**
- **ci: remove broekn retry login for url link testing**

### DIFF
--- a/src/scripts/check_doc_urls.py
+++ b/src/scripts/check_doc_urls.py
@@ -5,6 +5,7 @@ Check that all links to the documentation are valid.
 import os
 import re
 import requests as r
+import time
 from multiprocessing.pool import ThreadPool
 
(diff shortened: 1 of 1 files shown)


williamstein:
> ci: speed up postgres in tests

It's good to do that, but it likely won't make it any faster.   Currently the postgres database is NOT reset between running tests in each file, because resetting it meant having to recreate the schema a lot, which was slow.   Maybe by using a tempfs for postgres, you can now change the tests to reset the database at the start of each file, and that won't make things way slower.    Even better, use a separate database for each file, so we can run multiple files in parallel.

williamstein:
but regarding review -- this looks like a good idea to me; but  of course test it.

PR #8736, opened 2026-02-10
Define precise type for DBPatch in sync/db/doc.ts

haraldschilly:
This PR defines a precise type for `DBPatch` in `packages/sync/editor/db/doc.ts`. Previously defined as `any`, `DBPatch` is now typed as `(number | jsmap[])[]`, which matches the runtime structure of a flat array containing alternating opcodes (-1 for delete, 1 for set) and data arrays (arrays of record objects).

The implementation updates `apply_patch` and `make_patch` methods in `DBDocument` to use this new type, with necessary type assertions where TypeScript cannot infer the alternating structure or `Immutable` conversions.

Additionally, this PR updates the `Document` and `Patch` interfaces in `packages/sync/editor/generic/types.ts`. Since `DBDocument` implements `Document` but uses `DBPatch` instead of the text-based `CompressedPatch`, the interface methods `apply_patch` and `make_patch` (and the `Patch.patch` property) were updated to accept `any`. This resolves type incompatibility errors while allowing different document types to implement their own patch formats.

Transplanted from https://github.com/haraldschilly/cocalc/pull/24.


haraldschilly:
@codex review

PR #8747, opened 2026-02-18
hub/blobs: harden blob serving against stored XSS

haraldschilly:
## Summary
- Force-download non-image blob types (HTML, SVG, JS, etc.) to prevent stored XSS via malicious files served on the hub origin
- Add `X-Content-Type-Options: nosniff` header to all blob responses
- Use proper HTTP status codes (401 for unauthenticated, 403 for unauthorized) instead of 500
- Sanitize error messages so internal details are logged server-side only, not leaked to clients
- Add `cleanupUploadKey()` to prevent upload state accumulation on error paths

## Test plan
- [ ] Upload a `.png` image as a blob — verify it still renders inline in the browser
- [ ] Upload a `.html` file as a blob — verify the browser downloads it instead of rendering
- [ ] Upload a `.svg` file as a blob — verify forced download
- [ ] Check response headers include `X-Content-Type-Options: nosniff`
- [ ] Attempt blob upload without auth — verify 401 response
- [ ] Attempt blob upload to non-collaborator project — verify 403 response
- [ ] Trigger an upload error — verify the response says "upload failed" without internal details

### DIFF
--- a/src/packages/hub/servers/app/blob-upload.ts
+++ b/src/packages/hub/servers/app/blob-upload.ts
@@ -15,6 +15,7 @@ import {
   MAX_BLOB_SIZE,
   //MAX_BLOB_SIZE_PER_PROJECT_PER_DAY,
 } from "@cocalc/util/db-schema/blobs";
+import { is_valid_uuid_string } from "@cocalc/util/misc";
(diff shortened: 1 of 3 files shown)


williamstein:
# :+1:

PR #8754, opened 2026-02-24
backend/process-stats: offload proc scan to worker and harden timing

haraldschilly:
deploy: project only

---

## Summary
- move process table scanning out of the main thread into a dedicated worker thread
- replace promise-heavy per-pid traversal with a synchronous linear `/proc` scan in the worker
- add an inline scan fallback for source/test runs where `process-stats.worker.js` is not present
- harden project-info startup so concurrent `start()` calls cannot create overlapping loops
- cache expensive project-info lookups (lazy jupyter resolver, cgroup path/system fallback reads)
- make project-info scan interval configurable via `COCALC_PROJECT_INFO_SCAN_INTERVAL_S` (float, default `5`)

## Validation
- `pnpm -C packages/backend build`
- `pnpm -C packages/project build`
- `pnpm -C packages/backend test -- execute-code.test.ts`
- runtime smoke check: `ProcessStats.getInstance().processes(...)` from built backend returns process data

### DIFF
--- a/src/packages/backend/execute-code.test.ts
+++ b/src/packages/backend/execute-code.test.ts
@@ -349,6 +349,15 @@ describe("await", () => {
     check(s);
     if (s.type !== "async") return;
     expect(Array.isArray(s.stats)).toBeTruthy();
+    if (!Array.isArray(s.stats)) return;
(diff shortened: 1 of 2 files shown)


haraldschilly:
@codex review

Issue #7997, opened 2024-10-31
add back to the api the ability to create accounts


williamstein:
We had to remove this api endpoint due to fraud / abuse; in fact now, creating an account requires either SSO or going through recaptcha (which proves you're human).    We can bring back a variant of creating an account via the api, but will need to put it behind an initial approval step (where the account using the api has to marked to have special access to this functionality).

williamstein:
- Here's the old endpoint docs: https://doc.cocalc.com/api/create_account.html 
- Probably the way to implement this is just to make the code in src/packages/next/pages/api/v2/auth/sign-up.ts be usable by certain accounts that have been manually approved to have elevated API access.

williamstein:
This is not yet documented at https://cocalc.com/api/v2, so in the meantime here's how to use the new endpoint via the api:

```
API Usage:

curl -u sk_abcdefQWERTY090900000000: \
  -d firstName=John00 \
  -d lastName=Doe00 \
  -d email=jd@example.com \
  -d password=xyzabc09090 \
  -d terms=true https://cocalc.com/api/v2/auth/sign-up

TIP: If you want to pass in an email like jd+1@example.com, use '%2B' in place of '+'.
```

You MUST have a minimum balance under -$100 to use this endpoint.

PR #8755, opened 2026-02-24
fix: anonymous signup for licensed shared files

haraldschilly:
## Summary

Fix anonymous signup for licensed shared files (public paths with `site_license_id` + `unlisted=true`) when general anonymous signup is disabled but "Allow anonymous signup for licensed shared files" is enabled.

### Signed-in user visiting a licensed share
1. Visits the shared file page → sees "Choose Project"
2. Creates a new project or selects an existing one
3. The file is copied to their project
4. The site license from the public path is automatically applied to the project, providing compute resources

### Anonymous user visiting a licensed share
1. Visits the shared file page → sees "Use CoCalc Anonymously" button
2. Clicks the button → anonymous account is created and signed in
3. The file is copied to a new anonymous project with the license applied

**Previously broken:** When a user had any authentication (e.g., a `remember_me` cookie from a previous session, from signing in via the in-page form, or from being logged in in another tab), clicking "Use CoCalc Anonymously" failed with `{"api":"Creation of anonymous accounts via the API is not allowed."}`.

## Root Cause

Commit 433f4b3bcb (Oct 2024, #7997) added API-based account creation, introducing an `owner_id` check in the sign-up endpoint. When the request carries any authentication — even a browser cookie — `getAccountId()` returns a value, and the code rejected anonymous account creation **before** reaching the licensed-shares exception that was already in place further down.

## Changes

1. **Licensed-share exception in `owner_id` block**: Add the same licensed-share check to the early `owner_id` gate, extracted into a shared `isAllowedLicensedShareSignup()` helper (DRY — used in both code paths)

2. **Clear `owner_id` for licensed-share path**: When the licensed-share exception fires, clear `owner_id` so `signUserIn()` runs for the new anonymous account (otherwise the account is created but never signed into — orphaned)

3. **Restrict to browser auth only**: Only allow the licensed-share anonymous exception for cookie-based sessions (`!req.header("Authorization")`), not API-key authenticated requests

4. **Skip registration token for licensed shares**: Registration tokens are an independent policy — anonymous users accessing a licensed share shouldn't need one. The site license is the authorization

5. **Zod schema: support anonymous signup in dev mode**: The OpenAPI schema (`SignUpInputSchema`) had `email`, `password`, `firstName`, `lastName`, and `terms` as required fields. This blocked anonymous signup in dev mode where Zod validation is active (production has it disabled). The schema now uses a `z.union()` of two shapes: regular signup (all fields required) and anonymous signup (`anonymous: true`, other fields optional). The frontend now sends `anonymous: true` for anonymous account creation

## Test plan

- [x] Anonymous user (incognito) on licensed share → "Use CoCalc Anonymously" succeeds
- [x] Signed-in user on licensed share → "Choose Project" copies file and applies license
- [x] Anonymous signup on share WITHOUT `site_license_id` → correctly denied
- [x] API-key anonymous account creation (without valid licensed share) → still rejected
- [x] Regular email signup with registration tokens → still enforced

### DIFF
--- a/src/packages/next/pages/api/v2/auth/sign-up.ts
+++ b/src/packages/next/pages/api/v2/auth/sign-up.ts
@@ -111,6 +111,10 @@ export async function signUp(req, res) {
   const { email_signup, anonymous_signup, anonymous_signup_licensed_shares } =
     await getServerSettings();
 
+  // Track whether this is a licensed-share anonymous signup, so we can
+  // skip checks that don't apply (e.g., registration tokens).
+  let licensedShareSignup = false;
+
   let owner_id = await getAccountId(req);
   if (owner_id) {
     if (isAnonymous) {
@@ -129,6 +133,7 @@ export async function signUp(req, res) {
         // anonymous account gets signed in (signUserIn is skipped when
         // owner_id is set, which would leave an orphaned account).
         owner_id = undefined;
+        licensedShareSignup = true;
       } else {
         res.json({
           issues: {
@@ -175,6 +180,7 @@ export async function signUp(req, res) {
         })
       ) {
         // an unlisted public path with a license when anonymous_signup_licensed_shares is set is allowed
+        licensedShareSignup = true;
       } else {
         res.json({
           issues: {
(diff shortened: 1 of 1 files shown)


haraldschilly:
@codex review

PR #8770, opened 2026-03-08
Unify Assignments and Handouts UI

novoselt:
- Unify Assignments and Handouts UI.
- Keep backend behavior intact; this PR is frontend-focused with naming/refactor cleanup.
- Allow assigning/distributing empty assignment/handout folders (warning remains, but no hard block).
- No-content warning dynamically refreshes.
- Improve i18n coverage and message clarity across course workflow controls/popovers.
- Add/update many message IDs/descriptions for better automatic translation context.

There is still room for improvement both to English wording and translations, but it would be better done on top of this.

### DIFF
--- a/src/packages/frontend/course/common/course-unit-strings.tsx
+++ b/src/packages/frontend/course/common/course-unit-strings.tsx
@@ -68,18 +68,14 @@ export function fileActivityMessages(
     }),
     tip: intl.formatMessage(
       {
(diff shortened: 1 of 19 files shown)


novoselt:
Option C implemented for 7: have uniform behaviour for both types of units, keep the new editor, store current drafts based on id in a similar way to old code.

PR #8814, opened 2026-04-01
Add HubSpot to 3rd parties

novoselt:


novoselt:
@haraldschilly Is this one supposed to be live by now? I don't see it on https://cocalc.com/policies/thirdparties

PR #8826, opened 2026-04-05
stripe: prevent duplicate customers and student-pay transactions

novoselt:
Stripe customer creation is now serialized under a database row lock, so concurrent requests are much less likely to create two Stripe customers for the same CoCalc account.

The checkout-session path for student-pay now participates in the same “outstanding payment” guard that previously only covered payment-intent/invoice flows. Before creating a new student-pay checkout session, CoCalc now checks whether that course project already has an active payment in progress. It records the checkout session id on the project so later retries can detect that state. This closes the gap where a checkout-based course payment could be charged twice in Stripe while CoCalc only recorded a single course purchase.

The user-facing behavior is split by payment type. If there is an outstanding payment intent, users still get the old actionable message directing them to the payments page. If the outstanding object is an open checkout session, users now get a message telling them to try again later or contact support, since the payments page is not a reliable place to cancel or complete a Checkout-backed payment.

### DIFF
--- a/src/packages/database/postgres/stripe/customer-id.ts
+++ b/src/packages/database/postgres/stripe/customer-id.ts
@@ -1,5 +1,6 @@
 import getPool from "@cocalc/database/pool";
 import getLogger from "@cocalc/backend/logger";
+import type { PoolClient } from "@cocalc/database/pool";
 
 const log = getLogger("database:stripe:customer-id");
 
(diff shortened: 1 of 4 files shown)


novoselt:
Looking more into this and potential fixing, it is probably not worth doing it before the upcoming payment model switch. To avoid duplication in the future https://docs.stripe.com/api/idempotent_requests may be useful.

PR #8869, opened 2026-04-28
conat: backport fast-rpc from cocalc-ai

haraldschilly:
## Summary

Backports the **fast-rpc** path from cocalc-ai, plus the supporting
ordering/wakeup/handshake fixes that make it stable on this base.

Typed Conat service calls (the common control-plane hot path) now skip
inbox request/reply, msgpack chunking, and most of the `Message`
abstraction. Selectable via `COCALC_CONAT_SERVICE_TRANSPORT=request`
(default: `fast-rpc`).

### Fast-rpc (1–4)

1. `75631d26af` **conat: add direct RPC fast path** — additive
   (cocalc-ai `3ee2c2f982`). Server `rpcServices` registry +
   `rpc-service`/`rpc-service-close`/`rpc` socket handlers; client
   `rpcService()`, `rpcRequest()`, `rpcCall()`. Nothing routes through
   it yet.
2. `f48e253c86` **conat: add fast RPC benchmark path** — additive
   (cocalc-ai `2c08b0baee`). The `fast-rpc` wire protocol that bypasses
   `Message` construction and msgpack wrapping for raw payloads.
3. `0552fd8091` **conat: route typed services through fast RPC** —
   production switchover (cocalc-ai `b5f1efeb68`). New `core/codec.ts`
   (extracted to break a service/client import cycle); `service/typed.ts`
   rewritten to use `fastRpcRequest` for typed calls under 4 MiB,
   falling back to the legacy chunked path otherwise.
4. `7b246a82a7` **conat: preserve legacy service compatibility** —
   compat layer (cocalc-ai `eddda0b2b0`). Falls back to legacy when
   client mocks lack `fastRpcRequest`/`waitForInterest`, when the
   router responds "no services matching", and on jsdom-style runtimes
   lacking `setImmediate`.

### Socket handshake + ordering (5–7)

These are the cocalc-ai socket changes that the fast-rpc cutover ended
up needing on this base. Originally targeted as out-of-scope; without
them the cluster connect path and ordered-data assumptions drifted
under fast-rpc load.

5. `459f5656b2` **conat/socket: dedicated connect-control handshake** —
   manual re-implementation of cocalc-ai `8f968e4164` minus all
   instrumentation. Replaces the request/reply-based connect step with
   a one-way `publishSync` of a `connect` control message; the server
   replies via `publish` of `connected` on the client subject already
   subscribed to, instead of responding to the request inbox. Each
   attempt gets a unique `SOCKET_HEADER_CONNECT_ATTEMPT` id so stale
   replies for aborted attempts cannot promote a torn-down session
   back to ready. Decouples connect from the generic
   inbox/reply-subscription/request-timeout machinery, which is fragile
   across cluster nodes.
6. `cc4997095b` **conat/socket: setImmediate-batched data delivery +
   flushDataQueue** — re-implementation of the dataQueue slice from
   cocalc-ai `62f005adb6`, with the rest of that commit
   (`publishQueue`, sequential `await handleMesg`, etc.) intentionally
   left out. Buffers inbound `data` events and emits one per
   `setImmediate` turn so back-to-back data events don't get coalesced
   from the perspective of synchronous `EventEmitter` consumers.
   `flushDataQueue()` drains synchronously before emitting `request`,
   so handlers see preceding data first. Falls back to
   `setTimeout(...,0)` on jsdom.
7. `f8e1a5c344` **conat/socket: tighten connect handshake state
   gating + retry publish** — codex-review follow-ups to (5):
   server-side `connect` reply uses async
   `publish(..., { waitForInterest: true, noThrow: true })` instead of
   `publishSync` (so retry connects survive cluster routing races);
   client-side `handleConnected` and `waitForConnected` now both gate
   on `state == "connecting"` so a delayed reply or in-flight loop
   can't operate on a torn-down session.

### Wakeup correctness + sync convergence (8–9)

Independent slices from cocalc-ai `62f005adb6` + `518f737b10` that
don't depend on the larger ordering rework but are needed once typed
services route through fast-rpc.

8. `70ddb91eeb` **conat: backport Group-3 ordering / wakeup correctness
   fixes** —
   - `core/server.ts waitForInterestOnThisNode`: always recheck every
     1s so a "change" event landing between `hasMatch()` and listener
     registration cannot stall a waiter until an unrelated future
     interest update.
   - `core/client.ts Message.respond`: drop `async` and return the
     publish promise directly with a `.catch(()=>undefined)` to
     swallow unhandled-rejection noise from fire-and-forget callers
     without changing observable error semantics for awaited callers.
   - `persist/server.ts` init-error path: derive `errorCode` from
     `err.code` with a permission-denied fallback to 403; emit a
     second `stream-initialized` event so any pending
     `await once(socket, "stream-initialized")` waiters wake up,
     re-check `error`, and rethrow with the proper code instead of
     stalling to the request timeout.
   - `service/service.ts callConatService`: split `waitForInterest`
     from the request itself when retries are allowed. Pre-wait for
     interest using the full timeout, then issue the request with the
     remaining budget and `waitForInterest:false`. Matches what
     fast-rpc's `service/typed.ts` already does and avoids
     cluster-side wait machinery getting tangled up inside the
     publish itself.
9. `d4cea0619c` **conat/sync: backport Group-3B CoreStream + DKV
   delete fixes** — three interlocked slices from cocalc-ai
   `62f005adb6` + `518f737b10`. Builds on the `lastValueByKey` + DKV
   `remote` cache from #8703.
   - `CoreStream.processPersistentSet`: re-emit on duplicate seq so
     downstream consumers (DKV save loops, changefeed waiters) holding
     local state can clear it. Bump `lastSeq`, look up `prev` from
     `kv[key]?.mesg ?? lastValueByKey.get(key)`, emit, return.
   - `CoreStream.waitForLocalPublish` + `setKv`/`setKvMany`/`deleteKv`
     wiring. New helper waits via `until(...)` until the local mirror
     has caught up to a just-published seq. `deleteKv` gains `force`
     and `waitForLocal` options.
   - `DKV.delete`: replace `_delete + save()` with a direct
     `kv.deleteKv(key, { force: true, waitForLocal: false })` plus
     `.then()` cleanup and `.catch()` fallback to `save()`. Some
     callers inspect the underlying stream inventory right after
     `delete`; with the coalesced save loop, that read could overtake
     the write and see stale data.

### Transport / compression (10–11)

10. `d28c374e77` **conat/core/server: document Group-1 deferral
    rationale** — preserves the tried-and-failed list in the
    `socketioOptions` comment so the next attempt knows what was
    already in place when `perMessageDeflate:false` was tried.
11. `209afd1c9d` **conat/core/server: ship Group-1 minus
    perMessageDeflate:false** — manual port of the safe portion of
    cocalc-ai `e7ae84b433` + `365af2a39f`.
    - **Ship**: `transports: ["websocket"]` (clients already force
      websocket; this just refuses long-poll fallback);
      `httpCompression: false` (HTTP-level gzip on the long-poll
      handshake, unused with websocket-only);
      `COCALC_CONAT_SOCKET_IO_COMPRESSION` env var as an operational
      escape hatch (default off; set `1`/`true`/`yes` to flip
      `perMessageDeflate` to full compression for benchmarking or
      production rollback without a redeploy).
    - **Defer**: `perMessageDeflate: false`. Even on top of (5)–(9),
      this still regresses `basic.test.ts "client first, then server"`
      — queued pre-connect writes are lost. cocalc-ai/main carries
      ~1265 commits beyond `e7ae84b433`; there is plausibly more
      plumbing (transport-scoped liveness, idle stats suppression, …)
      that completes the picture. The threshold-`1<<30` workaround
      stays for now and already eliminates per-message zlib cost; what
      we lose vs `false` is just ws-extension send-pipeline overhead.

`cluster.test.ts` gets `jest.setTimeout(15000)` since cross-cluster
connect legitimately needs multiple round-trips of cluster sync and
jest's default 5s budget is too tight under CI parallelism. The bump
gives real connects room to succeed; it does **not** cause failures
to surface faster (it's strictly a longer ceiling).

## Behavior changes worth flagging

- Default service transport flips to `fast-rpc`. Kill switch:
  `COCALC_CONAT_SERVICE_TRANSPORT=request`.
- Long-poll fallback removed (`transports: ["websocket"]`). Any
  environment that proxies WS poorly will need
  `COCALC_CONAT_SOCKET_IO_COMPRESSION` semantics extended (or revert).
- HTTP-level gzip removed on the router (`httpCompression: false`).
  No effect with websocket-only.

## Skipped from the upstream sequence

- cocalc-ai `62f005adb6` minus the slices in (6) and (8): `publishQueue`,
  sequential `await handleMesg`, `waitForServerId` polling. Each tried
  individually or paired with (6); regressed `cluster.test.ts`
  cross-cluster connect.
- cocalc-ai `518f737b10` minus the slices in (9): the DKV save-loop
  switch from `!hasUnsavedChanges()` to `isStable()`. Master's existing
  loop is functionally close and the cocalc-ai variant has subtle
  corners around stale local entries with no diff vs saved.
- cocalc-ai `e7ae84b433` minus the bits in (11): `perMessageDeflate: false`
  (see deferral rationale above).
- cocalc-ai `31c9c1ca22` (rpc-raw benchmark transport). Pure benchmark
  harness, not needed.

## Deployment / rollout

This branch is designed to support an **incremental rollout** without
forcing project restarts or browser refreshes. The dedicated
connect-control handshake (commit 5) is a wire-protocol change, but
the bidirectional compat shims in commit 13 keep both directions of
mixed-version traffic working.

### Compatibility matrix

| Direction | Path | Compat? |
|---|---|---|
| Old client → New server (un-restarted project / unrefreshed browser tab → upgraded hub) | persist, changefeeds | ✓ via server-side inbox-reply shim |
| New client → Old server (refreshed browser tab → un-restarted project) | primus / browser-websocket-api | ✓ via client-side parallel-request shim |
| Old client → Old server | — | ✓ unchanged |
| New client → New server | — | ✓ primary publish/publish path |

### Suggested rollout order

1. **Hub / router / persist / static** — rolling deploy across
   replicas. Brief intra-hub mixed-version windows during the rollout
   converge as soon as all replicas finish; cross-cluster join links
   between mixed-version router nodes will fail to connect during
   that window but recover automatically.
2. **Project servers** — no coordinated restart needed. Old projects
   continue to talk to the upgraded hub via the server-side compat
   shim (their old client uses request/reply for connect; the new
   server replies on the inbox in addition to publishing to
   `clientSubject`).
3. **Browser tabs** — no forced refresh needed. Old tabs talk to the
   upgraded hub via the same server-side shim. As tabs refresh
   organically, they pick up new client code; the new client's
   parallel-request shim keeps them compatible with un-restarted
   project servers.
4. **Project servers upgrade naturally** as users restart projects
   over the following days/weeks.

### What is NOT cross-version compatible

- Cross-cluster conat router join links between mixed-version nodes
  (brief, only during the multi-pod hub rolling window).
- The hub <-> persist server channel uses ConatSocket; if hub and
  persist are deployed as separate processes that cycle independently,
  same shims apply (both directions covered).

### Operational kill switches

- `COCALC_CONAT_SERVICE_TRANSPORT=request` — forces every typed-service
  call to use the legacy `request` path (skip fast-rpc entirely). Set
  on a single hub replica for canary; if call latency or correctness
  regresses, set everywhere and the system reverts to pre-PR-8869
  service-transport behavior.
- `COCALC_CONAT_SOCKET_IO_COMPRESSION=1` (or `true`/`yes`) — re-enables
  full per-message + HTTP compression. Use for benchmarking or to roll
  back the transport changes (commit 11) without a redeploy.

## Test plan

- [x] All 654 backend conat tests pass locally (full-suite, 10
      consecutive runs over the original 11 commits + 4 fix-up
      commits)
- [x] Socket cluster test 5/5 under `maxWorkers=4` (matches CI
      parallelism)
- [x] New `fast-rpc.test.ts` (18 cases): IIFE registration failure,
      fallback semantics (transport-timeout vs response-408 vs
      "no services matching"), `syncRpcServices` reconnect re-registration
      via direct `fastRpcRequest`, close-removes-routing,
      8 MiB oversize fallback
- [x] New `socket/handshake-compat.test.ts` (10 cases): legacy
      request/reply against new server, new client against
      legacy-only-inbox emulator, dual-shim deduplication
- [ ] CI green (build, type-check, full test matrix)
- [ ] Manual smoke test in dev hub before merge
- [ ] Manual mixed-version smoke: old project ↔ new hub, new browser ↔
      old project

### Commit 12 — robustness from review

- **No double-execution on response-level 408.** `fastRpcRequest` now
  tags errors that originate from a socket.io ack timeout (the router
  itself never acked — the request provably did not reach a service
  handler) with `transportTimeout: true`. `callTypedConatService` only
  falls back to legacy on `transportTimeout` + the existing
  `disconnected` / `no services matching` / 413 cases. A 408 in a
  server response (where the inner emit to the service socket timed
  out and the handler may have run) propagates instead of being
  silently retried — which is what mattered for any side-effecting
  service method.
- **`syncRpcServices` retry loop.** Wrapped in `until(...)` mirroring
  `syncSubscriptions` so a transient ack timeout on (re)register no
  longer becomes an unhandled rejection AND silently leaves fast-rpc
  services unregistered.
- **`createServiceHandler` IIFE try/catch.** The fire-and-forget async
  registration now logs (via `getLogger("conat:service:typed")`) and
  swallows; the legacy service is unaffected, and the call-side
  fallback handles requests through the legacy path.
- **`fast-rpc.test.ts` added** with the coverage listed above.

### Commit 13 — bidirectional connect-control compat shim

The dedicated connect-control handshake (commit 5) is a wire-protocol
change. To support incremental rollout without forced restarts:

- **Server-side**: in `socket/server.ts` the `cmd == "connect"` branch
  picks the reply path based on `mesg.isRequest()`. Modern clients
  (publishSync, no inbox) get a publish to `clientSubject`. Old
  clients OR the new client's compat probe (request/reply, has an
  inbox) get an inbox reply. The two paths are mutually exclusive,
  so an old client never receives an unsolicited `cmd:"connected"`
  publish that it would route into TCP recv as garbage.
- **Client-side**: in `socket/client.ts` `sendConnectCommand` fires
  the modern `publishSync` immediately AND schedules a legacy compat
  probe (a real `request`) for 1.5s later. If the modern path
  succeeds before then, the probe is cancelled and never creates an
  inbox subscription — keeping cluster overhead negligible. If the
  probe fires and times out (transient failure or modern server that
  doesn't reply on inbox), the `.catch` reschedules another probe so
  a single transient timeout cannot wedge the socket against a
  legacy server. Successful probes feed through `handleConnected`
  via the same `connectAttempts` dedup as the modern path.
- **`socket/handshake-compat.test.ts` added** verifying both
  directions of the mixed-version matrix and the dedup.

Connect latency: modern-modern is unchanged (publishSync replies in
~10ms locally; probe never fires). New-client → old-server is
~1.5s + one round-trip. The 1.5s deferral was tuned empirically:
0/200ms causes `persist/cluster.test.ts` and `sync/cluster.test.ts`
to time out on cross-cluster failover because per-socket probe
traffic congests interest propagation; 1.5s gives modern handshakes
plenty of room (typical <100ms, worst-case cross-cluster a couple
of seconds) while keeping the new-browser → old-project upgrade
path bounded.

## Still-deferred follow-ups (post-merge, scoped separately)

- `core/server.ts` register-rpc-service path doesn't enforce
  `MAX_SUBSCRIPTIONS_PER_CLIENT/HUB`; an authenticated client can
  register many subjects. Acceptable for now since RPC services come
  from internal/trusted code paths, but worth a guard.
- `CoreStream.waitForLocalPublish` requires exact seq equality for
  keyed writes — a later same-key update can make a successful
  write look like a timeout. Edge case; needs a separate small fix.
- The three parallel maps for RPC service state in `core/client.ts`
  (`rpcServiceQueues` / `rpcServiceImpls` / `fastRpcServiceHandlers`)
  are clunky and same-subject collision-prone. Quality nit, not
  correctness.
- `core/server.ts` `rpc` / `fast-rpc` socket handlers are 80-line
  near-duplicates differing only in payload field names; factor.
- `error_attrs` propagation in `core/client.ts handleRpcRequest`
  forwards `path` / `syscall` to remote callers — minor info-leak
  surface that needs a broader sanitization policy
  (`headers.error` carries `err.message` which often includes the
  filesystem path itself).


williamstein:
Obviously let me know if it finds any bugs...

PR #8883, opened 2026-05-12
next/cookies: improve consent banner behavior

novoselt:
Polish the wording of the cookie banner to reflect all categories, clarify how to change consent later, and for non-authenticated users add a link in the footer that removes all cookies (and saved consent) and reloads the page to bring back the original banner. Also catch the last non-wrapped youtube video.

### DIFF
--- a/src/packages/frontend/cookie-consent/index.ts
+++ b/src/packages/frontend/cookie-consent/index.ts
@@ -20,7 +20,7 @@ export type { CookieCategoryKey };
 
 // Bump this if the cookie categories or banner text change in a way that
 // invalidates prior consent. vanilla-cookieconsent will re-prompt the user.
-export const COOKIE_CONSENT_REVISION = 1;
+export const COOKIE_CONSENT_REVISION = 2;
 
 // Snapshot of the user's consent we persist in account.other_settings. Kept
 // minimal on purpose: the browser cookie is authoritative for the session;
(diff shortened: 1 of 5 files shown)


haraldschilly:
Pair-reviewed locally with Claude Code. Pushed a follow-up commit that addresses four review notes:

1. **Single source of truth for cookie names.** The "Clear cookies" link used to hardcode `[cc_cookie, cocalc_youtube_consent, CC_ANA, _ga*, _gid*]` in `footer.tsx`, duplicating what's already declared in `COOKIE_CATEGORIES.autoClearCookies` (categories.ts). Moved the clearing logic into a new `clearAllConsentCookies()` export in `frontend/cookie-consent/index.ts` that derives the list from the categories module — adding a new category with `autoClearCookies` now propagates automatically, matching the "Adding a new category … everything else falls into place" promise in `categories.ts`.

2. **YouTube cookie handled by its own module.** `clearAllConsentCookies()` delegates to the existing `revokeYouTubeConsent()` from `youtube.ts` instead of hardcoding `cocalc_youtube_consent`. Keeps `YT_COOKIE` encapsulated.

3. **More robust parent-domain enumeration.** The old `getCookieClearDomains` only handled two-segment TLDs (`.cocalc.com`); for hosts like `.cocalc.co.uk` it produced `.co.uk` which browsers reject. New `parentDomainCandidates()` walks every parent suffix in both `host` and `.host` form — works for arbitrary multi-level TLDs without needing the Public Suffix List (invalid candidates get silently rejected by the browser).

4. **Accessible link, matching the surrounding ones.** The "Clear cookies" link used `Typography.Link` with `onClick` and no `href`, which renders an `<a>` that isn't keyboard-focusable. Reverted to the same `<A>` component used by every other footer link, with `href="#"` and an `onClick` that calls `e.preventDefault()`. Removed the `FooterColumnLink` indirection — back to one inline render path.

5. **Update privacy policy.** Adding cookie description, updating timestamps and changes table.

Net diff: `footer.tsx` lost ~50 lines (helpers gone), `cookie-consent/index.ts` gained ~60 lines for the centralized clearer.

